PR #17, opened 2015-02-20
Pre-fetch external document contents so we don't have to request them synchronously

psndcsrv:
This fixes PT story [#88253392](https://www.pivotaltracker.com/story/show/88253392): Firefox not supporting synchronous loading of resource forks (external documents).

It adds pre-fetching of resource fork contents before handing off the main document to the document initialization process. This way we avoid needing to do an XHR fetch of the contents in the middle of the initialization process.


jsandoe:
It would be helpful to add logging of document load failures in authorization.js. That would facilitate debugging of those events.

psndcsrv:
@jsandoe The logging statement is now included.

jsandoe:
Great!

PR #19, opened 2015-02-24
Bug fixes and small updates

psndcsrv:
These commits actually build off the ones in the wip-async-externals branch (see https://github.com/concord-consortium/codap/pull/17).


psndcsrv:
I closed this and opened a new one (#20) so that the branch references would be correct (merge into master instead of wip-async-externals)

PR #20, opened 2015-02-25
Bug fixes and small features

psndcsrv:
These commits actually build off the ones in the wip-async-externals branch (see #17), so the first commit is a duplicate of that.


psndcsrv:
@jsandoe This pull request can now also get merged, since #17 is merged.

PR #21, opened 2015-02-26
Metacog integration

jsandoe:


jsandoe:
This branch was an experimental integration. If we were to go forward on this line we would take a different tack.

PR #32, opened 2015-10-21
Overwriting

psndcsrv:
These are the changes required to support warning the user about a name change that would overwrite an already-existing document.

There are also some fixes for:
- correctly deferring a `save` if saving is already in progress
- forcing multiple saves in order to make sure ID's and parent ID's are accurate
- correctly handling a failed save and marking the current save as done (it used to get stuck)


psndcsrv:
@jsandoe This is ready to go at this point. You can play around with it here: https://codap.concord.org/releases/overwriting/static/dg/en/cert/index.html?documentServer=https%3A%2F%2Fdocument-store-dev.herokuapp.com%2F&undo=true

Let me know if you have any questions!

psndcsrv:
Also, this should be deployed at the same time as https://github.com/concord-consortium/document-store/pull/1

PR #33, opened 2015-10-29
Flexible grouping

jsandoe:
This commit adds to CODAP the ability to rearrange the table heirarchy of a particular set of related data. One can drag attribute to organize the attributes in a table. On can drag between tables of the same context. One can drag to create an new table.
Known bugs: Interactions among contexts. Over enthusiastica drop targets. Other contexts light their drop targets when they should not. Drop targets poach other drop zones. ...


chaddorsey:
Does this mean this feature is on production now?? :-)

--Chad

chaddorsey:
No, not in release until code reviewed, merged, and QA-ed.
Bill

chaddorsey:
No problem – just excited to see it even to this stage!

--Chad

bfinzer:
In data_context.js, the last line is
        iChange.operation = 'resetCollections';
This seems to be a side effect on the input parameter. Intentional?
The comment to the method, doMoveAttribute, says there is a return value '.success' presumably of iChange, but there is no actual setting of this property.

apps/dg/controllers/data_context.js
```diff
@@ -968,12 +1086,12 @@ DG.DataContext = SC.Object.extend((function() // closure
...
-        collectionRecord, collectionID,
+        collection, collectionID,
         prevCollectionID = null;
     for( var i = 0; i < collectionCount; ++i) {
-      collectionRecord = collections.objectAt( i);
-      collectionID = collectionRecord.get('id');
-      if( collectionRecord && (collectionID === childCollectionID))
+      collection = collections.objectAt( i);
+      collectionID = collection.get('id');
+      if( collection && (collectionID === childCollectionID))
```
bfinzer: This seems like the place where order dependency is introduced into getting the parent collection or child collection. True? If so, would now be a good time to get rid of it and do something not order dependent?

apps/dg/models/collection_model.js
```diff
@@ -16,223 +16,624 @@
...
+    caseIDToIndexMap: null,
+
+    /**
+     * Indicates whether parent/child links are configured correctly.
+     * Games using the revised API (Aug 2011) should set this to true.
+     * Note that this property is not a record attribute and is not
+     * expected to be archived with the record. It's a temporary flag
+     * for use until games using the old API are no longer supported.
+     *
+     * ToDo: Believe this property is removable.
```
bfinzer: I also think it is removable. Good if this were the time to do it!

apps/dg/models/data_set.js
```diff
@@ -0,0 +1,276 @@
...
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+// ==========================================================================
+
+
+sc_require('models/DataItem');
+
+/** @class DataSet
+ *
+ * An indexed collection of DataItems, which are in turn .
```
bfinzer: Please complete.

bfinzer:
Just a couple of suggestions for the next commit. Merging now.

PR #34, opened 2015-11-02
Remove duplicate logging events

sfentress:
By adding some instrumentation to the log events, plus a bunch of manual tracing through event paths, I believe I've found all the cases where we were duplicating the logging of undoable events, and removed those.

Each of these log events was logging an event that was already being logged by the undo-redo controller (via the `log` property of an undoable command).

[#104009308]


jsandoe:
Looks good.

PR #42, opened 2016-02-05
Fixed bug in which a newly-opened document would be dirty immediately

kswenson:
Fixed bug in which a newly-opened document would be dirty immediately, triggering auto-save ([PT#113048513](https://www.pivotaltracker.com/story/show/113048513)).

The comment on the line being removed states that the line can be removed once all components are created with undoable actions, which is now true. The effect of this in current production CODAP is limited, but it will be greater once CFM is integrated with its auto-save capability.


jsandoe:
Looks fine to me.

PR #45, opened 2016-02-17
Specify https scheme for help and CODAP project page URLs

kswenson:
Specify https scheme for help and CODAP project page URLs to bypass https/http mismatch warnings.

Note: I considered protocol-relative URLs, but given that protocol-relative URLs are no longer recommended (e.g. http://www.paulirish.com/2010/the-protocol-relative-url/) we just specify https explicitly.


jsandoe:
Looks good.

PR #49, opened 2016-05-12
Invalidate/redraw child table cases when parent case changes

kswenson:
Refresh/redraw child table when changes to parent collection occur [#117946925]


kswenson:
After conversation with Jonathan we decided it would be easier to merge this in if it were based on the "working" branch rather than "master". Since there doesn't seem to be a way to change the base branch of a PR, we close this one and open a new one with the same change based on the "working" branch.

PR #51, opened 2016-05-12
Invalidate/redraw child table cases when parent case changes

kswenson:
Invalidate/redraw child table cases when parent case changes [#117946925]


jsandoe:
Nice!

PR #53, opened 2016-05-17
Support aggregate functions in plotted values and functions

kswenson:
Support aggregate functions in plotted values and functions
Support 'x' or axis attribute name in plotted values and plotted functions
Addresses [[#118567825]](https://www.pivotaltracker.com/story/show/118567825)

With this PR aggregate functions are supported in plotted values and plotted functions, but in a split dot plot only a single value is currently displayed across all sub-plots. Split evaluation is supported by the engine and can be enabled by setting the 'splitEval' property of the DG.PlottedFunctionModel, but the proper rendering of such split values is captured in a separate PT story [[#119751249]](https://www.pivotaltracker.com/story/show/119751249).

To demonstrate that split evaluation is working, set the 'splitEval' property of DG.PlottedFunctionModel to true, after which plotted values will be calculated for the first cell and plotted accordingly. To test computation of other cells, set the 'currentIndex' property of DG.PlottedValueAdornment to the index of the desired cell.

### DIFF
--- a/apps/dg/components/graph/adornments/plot_adornment_model.js
+++ b/apps/dg/components/graph/adornments/plot_adornment_model.js
@@ -24,14 +24,20 @@ DG.PlotAdornmentModel = SC.Object.extend(
 /** @scope DG.PlotAdornmentModel.prototype */
 {
   /**
-    The plot model to which the adornment is attached.
-    @property   {DG.PlotModel}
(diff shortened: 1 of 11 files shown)


jsandoe:
Looks good.
I rebased these changes on top of master and pushed.

PR #62, opened 2016-06-10
Fix plotted value regression

kswenson:
Plotted value code accommodates recent changes -- must specify 'ownerSpec'.
Invalidate the plotted value on global value changes.
Note: I have an in-progress branch which moves the invalidation code from the DependencyMgr to the FormulaContext and will allow the plotted values/functions to be invalidated only when necessary. For now, we simply invalidate the plotted value on any global value change for simplicity.


jsandoe:
Looks good.

PR #64, opened 2016-07-14
Move responsibility for invalidation from DependencyMgr to FormulaContext

kswenson:
Move responsibility for invalidation from DependencyMgr to FormulaContext [#124911061]
Full support for plotted value involvement in dependency tracking/invalidation
- aggregate function stack is captured during compilation for use in run-time bindings
- dependencies tracked for plotted value default argument (e.g. mean())


jsandoe:
Looks good. (Walked through with Kirk on Zoom.)

PR #65, opened 2016-07-16
Add lookupByIndex() and lookupByKey() functions

kswenson:
Add lookupByIndex() and lookupByKey() functions [#126247549]
- Integrate with DependencyMgr so cross-context dependencies are tracked/invalidated like within-context dependencies.


jsandoe:
Looks good.

PR #66, opened 2016-07-18
Handle namespace changes using the DG.DependencyMgr

kswenson:
- DG.DataContext listens for global value and attribute name changes
- DG.DependencyMgr is used to invalidate only those formulas that could be affected
- DG.DependencyMgr tracks undefined identifier references
- DG.DataContext sends out targeted 'dependentCases' notification
- DG.FormulaContext tracks undefined identifier references
- Eliminate DG.CollectionClient's 'attrFormulaDependentChanges' notification [#126259111]
- Eliminate DG.GlobalFormulaContext need to listen to DG.GlobalsController notifications [#126259111]


jsandoe:
Looks good. In person code walkthrough.

PR #68, opened 2016-07-20
Better error messages for lookup functions

kswenson:
lookupByIndex() and lookupByKey() now give more intelligible error messages when data set or attribute lookups fail.

Jonathan encountered an issue while trying to use the lookup functions which resulted in an inexplicable error message rather than something potentially useful.


jsandoe:
Looks good. Reviewed in person.

PR #71, opened 2016-07-23
Fix slider name editing bugs

kswenson:
- Changing the name of a slider with a negative value now works correctly [#126913099]
- Slider names can be edited to have a leading underscore
- Log message for slider name changes includes the correct prior name


jsandoe:
Looks good.

PR #74, opened 2016-07-28
invalidateNodesAndNotify() --> invalidateDependentsAndNotify()

kswenson:
Some clients of the invalidateNodesAndNotify() function were not updated when its name changed as part of the implementation of the Rerandomize() features.


jsandoe:
:thumbsup:

PR #76, opened 2016-07-29
Fix formulas when attributes are moved between collections

kswenson:
When an attribute is moved between collections, we need to update its formula context as well as the attribute itself. Previously, the attribute's 'collection' property was being updated but the attribute's formula context was still pointing to the original collection. This was what caused caseIndex to fail to evaluate after an attribute that referenced it was dragged between collections.


jsandoe:
👍

PR #77, opened 2016-07-29
Implement variance(), stdDev(), stdErr() functions

kswenson:
- Implement variance(), stdDev(), stdErr() functions
- Fix handling of infinities in aggregate functions
- Increase width of "--- Functions ---" button and the functions popup to better fit their content


jsandoe:
👍

PR #78, opened 2016-08-01
Fix typeahead/autocomplete bug: only cancel and restore original text on escape

kswenson:
Previous logic could cancel/restore at inappropriate times [#127400821].


jsandoe:
👍

PR #82, opened 2016-08-10
Splash screen improvements

sfentress:
This closes the splash screen when the document is ready (either the "What would you like to do?" dialog is showing or the document has loaded), and re-opens it when the user initiates an Open).

There are a couple questionable parts (checking the CFM params to see if it intends to open a document after the initial `"connect"` event, and abuse of `setTimeout(..., 0)` in order to cause a UI refresh) that could use another pair of eyes.

### DIFF
--- a/apps/dg/main.js
+++ b/apps/dg/main.js
@@ -313,7 +313,6 @@ DG.main = function main() {
         message: DialogContents({}), // jshint ignore:line
         onDrop: function () { DG.cfmClient.hideBlockingModal(); }}
       );
-      DG.splash.hideSplash();
     }
   }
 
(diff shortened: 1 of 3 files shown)


jsandoe:
I agree that interrogating the CFM to figure out if it is going to open a file is unfortunate. In addition to knitting CODAP and the CFM together more, it goes pretty deep into CFM internals. I am more worried about the latter issue than the former. Main.js is already the locus of CFM integration anyway. Would it be reasonable for the CFM to support a more abstract API for the splash screen code to manage, eg: 'isManagingDocument'? I don't mean as a part of this PR necessarily, but when we next work on CFM itself.

One thing that struck me as odd is main.js:406, "DG.cfmClient.hideBlockingModal();". I assume this is the grey overlay, or does it include the "What would you like to do?" dialog.  I know the blockingModal was discussed at one point, but the sharing of responsibilities here seems a little muddy. What do you think?

sfentress:
Updated to use the `ready` message, which depends on https://github.com/concord-consortium/cloud-file-manager/pull/30

jsandoe:
Looks great. I think the 'ready' notification made for cleaner code, esp on the CODAP side. 👍

PR #84, opened 2016-08-13
Clear match and dismiss autocomplete menu only on specific keys

kswenson:
- clear internal match variables on menu dismissal rather than keystroke
- fixes [#128266859] "Formula editor autocomplete sometimes replaces more than it should"


jsandoe:
👍 Looks good.

PR #85, opened 2016-08-16
Fix aggregate missing value handling; fix standard deviation adornment

kswenson:
- Clients of getNumericValue() must handle null return values
- Standard deviation adornment should use sample standard deviation rather than population standard deviation (for consistency with stdDev() function among other reasons)
 - Fixes bug Evangeline found while testing [#126931797]


jsandoe:
👍 I am assuming that Evangeline will handle this in the context of continuing testing on [#126931797].

PR #86, opened 2016-08-18
Update CFM; Pass url query parameter to CFM to open

sfentress:
This updates CFM to 510b93348, which brings in the google drive folder bug fix, the readOnly file hashes, and the url hash.

It also updates CODAP to pass the url query parameter to CFM on open.

[#120377399]
[#128470511]
[#121532705]


jsandoe:
👍 Change to main.js looks good. (CFM changes reviewed by others, so I didn't review, here.)

PR #87, opened 2016-08-24
Fix date time function and formula evaluation bugs

kswenson:
- Fix strings error for "Tuesday" string [#128621181]
- Fix inappropriate conversion of 0 values to dates [#128622833]
- Invalidate formulas when attributes move between collections [#128682337][#128896427]
- Invalidate formulas on creation/deletion of attributes [#128698547]


jsandoe:
👍  Looks fine. The isDate defect was surprising :-)

PR #89, opened 2016-09-07
Integrate with new CODAP drive app

sfentress:


jsandoe:
Should the call to 'DG.splash.hideSplash()' be inside the run loop? I am not sure it would matter one way or another, but am curious if there is a reason to place it outside?

sfentress:
You're right, there was no real reason to have it be outside. Moved.

I also updated the PR with the spinner gif on the splash screen. It's not part of the same feature, so I could easily move it out again, but it's a very small addition, and didn't seem worth its own PR.

jsandoe:
I think its fine to include the spinner in this PR.

Looks good. 👍

sfentress:
@jsandoe We should make sure to coordinate on making testing releases and production releases for this. I will need to update the open/new targets in the Google API for each new url that we release to.

jsandoe:
@sfentress, will do. Just to be clear, the open/new targets are the URL patterns that are invoked when a gdrive user clicks on a CODAP document, right? Can there be more than one? Or, were you thinking in terms of a staging -> production sequence?

sfentress:
There cannot be more than one, they are the targets for opening files from Google Drive. However, it won't effect regular saving and opening via CFM, so it's not urgent that we have to set them within seconds of deploying, and no one is yet using/depending on this feature.

I just didn't want you to deploy and try and test, and realize that part wasn't working, because I never got around to updating the targets.

PR #90, opened 2016-09-13
Update CFM: Fix bug in handling of null file names

kswenson:
Incorporate latest version of CFM which fixes handling of null file names (which can be created by doc store)

In the V2 doc store API, file names can be empty or null, as they aren't really used for anything any more. Even though we haven't deployed the V2 API to production, anyone that has been testing the V2 API are liable to have documents with null or empty names in their Concord Cloud accounts. Preexisting code for matching file names to extensions wasn't handling null file names. We add the necessary test to the matchesExtension() function, but there may be other places that requires similar treatment.

Incorporates CFM [PR#40](https://github.com/concord-consortium/cloud-file-manager/pull/40).


dougmartin:
Looks good :+1:

jsandoe:
Just for the record, the CFM integrated in this PR is CFM PR#40, commit id: 26ef683.

PR #91, opened 2016-09-14
Export csv and images via the CFM

sfentress:
This PR removes the copy-and-paste export of table CSV files, and the downloadable graph image snapshots, and replaces them with an export dialog presented by the CFM.

When the user choses to export a table's CSV file, after picking the table the CFM displays a save-as dialog. When the user saves the file, it is saved as raw data with the extension .csv and the correct mime-type.

Likewise, when the user hits the graph snapshot button, the CFM dialog is displayed, and the user can save a PNG file.

Currently, the only place a user can save a file is Google Drive, as the CFM's Save dialog is not merged with the "Download" dialog -- i.e. a download option is not part of the CFM's save dialog options. It looks like there are plans to include Download as an option. It may be that we want to wait on this PR until that is available, as this PR will prevent people who do not use Google Drive from exporting tables or images.

https://www.pivotaltracker.com/story/show/127512941
https://www.pivotaltracker.com/story/show/127513087


kswenson:
I've updated this PR to include the latest CFM PRs so that exported CSVs and graph snapshots can be saved/downloaded locally using the LocalFileProvider.
- includes CFM PRs [#61](https://github.com/concord-consortium/cloud-file-manager/pull/61), [#62](https://github.com/concord-consortium/cloud-file-manager/pull/62), [#63](https://github.com/concord-consortium/cloud-file-manager/pull/63).
- CFM sha 8dc6f87bcbbd72e9b33e980024ad74bbf7b0416e

jsandoe:
👍 Looks good.

PR #98, opened 2016-11-05
Function improvements

kswenson:
- count() with no arguments should count child cases [#132781159]
- String functions handle numeric arguments properly [#133351711]
- Polyfill Math.log10() for PhantomJS


kswenson:
I didn't notice that initially. This is what Fathom did (of course, Fathom didn't support hierarchical collections) and I would argue is the most natural interpretation. It seems to me that aggregate functions should operate in the current collection unless there is evidence to suggest switching to another context, i.e. child collection attribute references among the arguments. Furthermore, the child case interpretation would lead to confusion when used in a child collection where there are no child cases -- should it always return 0 or an error? -- and to being completely useless in the absence of a parent collection. It seems much more sensible to me that the default interpretation should be the current collection and that if another collection's case count is desired, an attribute of that collection can be specified as the argument.

bfinzer:
It’s because of the existence of hierarchical structures that we want the behavior of count() to refer to the number of child cases. With this functionality, the user can, for example, create a top level in which summary statistics are computed, count being one of them. We don’t want the user to have to refer to a child level attribute because that can evaluate, unbeknownst to them, as missing when what they want is the count of the children regardless of that attribute’s value.

When there are no children, the value should be zero. The user drags the attribute to make a new parent and gets the desired result.

kswenson:
Will do.

jsandoe:
The story arose, I think, out of a discussion of making it easier for users to write a function that computed the percent of cases represented by a group.

PR #99, opened 2016-11-09
Fix formula invalidation/evaluation when moving attributes between collections

kswenson:
- invalidate formulas when attributes are moved between collections [#131462435]
- invalidate formulas when cases are created/deleted


kswenson:
@jsandoe Good catch. I fixed the missing JSDoc argument name.

jsandoe:
Thanks!

PR #100, opened 2016-11-10
Eslint configuration

kswenson:
- configure `eslint` for use with CODAP
- fix issues found by `eslint`
- add `package.json` with
  - `dev-dependencies` so that `npm install` will install required tools (`eslint`, `phantomjs`)
  - `lint` script so that `npm run lint` runs `eslint`
  - `test` script so that `npm test` runs unit tests via `phantomjs`

`eslint` has surpassed `jshint` as the linter of choice these days. In working on the story to eliminate unused code, I found it useful to run `eslint` and the unit tests frequently to make sure I hadn't broken anything that could be checked statically. Also, `eslint`'s editor integration made it easier to see errors while I was removing code. Furthermore, while I was working on this Jonathan encountered the issue with `export` being used as a property name which caused `yuicompressor` to barf and took a fair amount of time to track down given the lack of useful error message. When I mentioned to Jonathan that I had `eslint` working locally he immediately tracked down the relevant `eslint` rule (`quote-props`) and I added it to our `eslint` configuration so that this problem will be caught more quickly the next time around.

Note that for now, both `eslint` and `jshint` are currently supported, but that we will likely choose to drop support for `jshint` at some point as `eslint` is clearly superior at this point.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,3 @@
+apps/dg/formula/formulaParser.*
+apps/dg/tinkerplots/**
+apps/dg/libraries/**
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,0 +1,48 @@
+{
+  "env": {
+    "browser": true
+  },
+  "globals": { "DG": false, "Raphael": false, "$": false, "SC": false, "sc_require": false,
+                "sc_super": false, "static_url": false, "NO": false, "YES": false },
+  "extends": [ "eslint:recommended" ],
+  "rules": {
+    // possible errors
+    "no-console": "off",
+    "no-template-curly-in-string": "warn",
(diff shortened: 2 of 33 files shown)


kswenson:
While you're correct that comments are not strictly allowed in JSON files, ESLint is documented as ignoring/supporting comments in eslintrc.json and since ESLint is the primary client of the file I thought (and still think) it made sense to include comments for clarity. (As an aside, I think the lack of support for comments in JSON should be rectified.) The downside is that generic JSON file readers will sometimes complain -- Github highlights them red when showing changes and generic validators would presumably complain. I would argue that the benefits to clarity of having the comments outweighs the drawbacks, but that's a matter of opinion.

jsandoe:
I entirely agree. Thanks.

PR #104, opened 2016-11-16
Extended data interactive API to handle collaboration

dougmartin:
* added getCases to collection to return all serialized cases
* added searchCases to collection to return serialized cases that match a query
* return serialized cases on all change notifications

My editor also trims spaces at the end of lines which accounts for all the "stripes" in the diff of collection_client.js


dougmartin:
@jsandoe I've updated the PR based on our meeting on Friday except that I ended up needing the case info during updates and deletes so I didn't add the "only createCase" conditional around the code.

For updates and deletes we only need the values of the case (specifically the firebase id) so we could reduce the transfer load to only the values for updates and deletes.  However, I thought this might lead to a confusing API so I held off doing it.  Let me know what you think.

dougmartin:
@jsandoe You can see where we need the values here in the collaboration service:

https://github.com/concord-consortium/codap-data-interactives/blob/e4b91c2ce4f4ac160e3609d64f8afee20d1a72b7/CollaborationService/js/CollaborationService.js#L237

PR #108, opened 2016-11-18
Completely specify CFM/File menu in CODAP

kswenson:
Previously, a default menu was provided by CFM which CODAP customized. In discussion with Jonathan we decided that CODAP should completely specify its menu configuration rather than share joint responsibility for it with CFM.

While there, fix the bug in which a document opened from or saved to Google Drive could not be saved locally [#134568775] by changing the behavior of the Save... menu item so that it always brings up the dialog (behaves like Save As...). In cases where autosave was enabled, the previous behavior was essentially a no-op, as the file would be saved every few seconds anyway. In situations where autosave was not enabled, the Save As... dialog would come up anyway. This way autosave handled saving to the current provider while the Save... menu item allows the user to change provider.

### DIFF
--- a/apps/dg/main.js
+++ b/apps/dg/main.js
@@ -134,6 +134,41 @@ DG.main = function main() {
           appBuildNum: DG.BUILD_NUM,
           appOrMenuElemId: iViewConfig.navBarId,
           hideMenuBar: DG.get('hideCFMMenu'),
+          ui: {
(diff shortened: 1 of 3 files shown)


dougmartin:
Is the CFM update the same code in https://github.com/concord-consortium/cloud-file-manager/pull/65 ?

PR #109, opened 2016-11-19
Add if(); remove secondsToDate() functions

kswenson:
- implement spreadsheet-like if() function
- add support for command-line generation of formula parser using PEG.js
- remove keywords (including "if") from formula grammar
- remove secondsToDate() function

Prior to this the CODAP formula parser was generated using the online PEG.js converter, which was a manual, error-prone, and non-repeatable process, particularly now that the required version of PEG.js is several versions old. I added a PT story for updating to the latest version of PEG.js and marked it "technical debt".


kswenson:
Yes, we are bound specifically to version 0.7.0 of PEG.js. I tried updating to a more recent version but our formula grammar doesn't work with anything newer. I have already created the corresponding [PT story](https://www.pivotaltracker.com/story/show/134617233) which describes this in more detail. I should have referenced it in the original PR.

From looking at the generated parser changes it seemed like it was doing what I wanted it to do, namely eliminate support for the commented out keywords. You question made me reconsider the actual meaning of the grammar, which seemed less than perfectly clear. I've come up with an alternate grammar change that eliminates keywords entirely from the grammar, rather than attempting to leave them as an empty set.

PR #111, opened 2016-11-21
Plotted value updates on axis attribute change [#121455949]

kswenson:
On the plotted value side, the simple fix is to invalidate the expression when the axis attribute changes. Getting the notification turned out to be the trickier part. I ended up updating `DG.AxisModel. firstAttributeNameDidChange` because it wasn't triggering before the change.


bfinzer:
Looks good

PR #122, opened 2016-12-07
WIP: React function browser [#133932137]

dougmartin:


kswenson:
@dougmartin Assigning to you as well as Jonathan in case you want to take a look at the changes since the handoff. Feel free to focus your attention more on the React side than on the CODAP integration side.

dougmartin:
Looks good other than you may want to remove the demo files from the PR now that it is integrated.

kswenson:
@dougmartin Given that one of the goals of the function browser code is to remain CODAP-independent for potential reusability, I thought it made sense to leave the demo code in place to enable testing that it runs outside of CODAP. If we ever decided to move the function browser into its own repo the demo code would logically move there, but for now Jonathan didn't object to hosting it in the CODAP repo.

dougmartin:
Sounds good.

PR #125, opened 2016-12-13
Fix recently-introduced bug in DG.MathUtilities.isInRange() function

kswenson:
- missing "return" was flagged by ESLint
- add unit tests for several DG.MathUtilities functions
- add semicolon to eliminate another recently introduced ESLint warning


bfinzer:
Looks good.

PR #130, opened 2016-12-20
Case table add entry row

kswenson:
For flat (non-hierarchical) data sets, cases can be created by entering values in a blank row at the bottom of the case table. Multiple values can be entered into this proto-case and the new case will only be created when the user exits the proto-case entirely, at which point a new blank row will be created at the bottom of the case table.

### DIFF
--- a/apps/dg/components/case_table/case_table_adapter.js
+++ b/apps/dg/components/case_table/case_table_adapter.js
@@ -233,6 +233,15 @@ DG.CaseTableAdapter = SC.Object.extend( (function() // closure
     return columnWidth;
   },
 
+  isCellEditable: function(row, column) {
+    var dataView = this.get('gridDataView'),
+        dataItem = dataView.getItem(row),
+        colInfo = this.gridColumns[column],
+        attr = colInfo && colInfo.attribute,
+        attrIsEditable = attr && attr.get('editable');
+    return (dataItem && dataItem._isProtoCase) || attrIsEditable;
+  },
+
   /**
     Builds the array of column definitions required by SlickGrid from the data context.
     
@@ -254,11 +263,8 @@ DG.CaseTableAdapter = SC.Object.extend( (function() // closure
     }
     
     function updateDynamicColumnProperties( iAttribute, ioColumnInfo) {
-      // only allow 'editable' columns to have editable cells
-      if( iAttribute.get('editable') && !iAttribute.get('hasFormula'))
-        ioColumnInfo.editor = DG.CaseTableCellEditor;
-      else
-        delete ioColumnInfo.editor;
+      // cell-specific editability handled by isCellEditable() method
+      ioColumnInfo.editor = DG.CaseTableCellEditor;
     }
     
     // Build the columnInfo for a single attribute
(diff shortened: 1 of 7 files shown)


apps/dg/components/case_table/case_table_adapter.js
```diff
@@ -533,12 +539,13 @@ DG.CaseTableAdapter = SC.Object.extend( (function() // closure
...
               // We filter because, at least in FireFox, we can get to a row that is one beyond the end
               // which therefore has no case
-              return !SC.none( iCase);
+              return iCase && !iCase._isProtoCase;
             }),
             select: true,
             extend: false
           };
-      tContext.applyChange( tChange);
+      if (tChange.cases.length)
```
jsandoe: This will bypass selection if the selection list has length 0.  I think this is the mechanism to clear existing selections is to send an empty list. I wonder if there is a place where the case table uses it in this way.

apps/dg/libraries/slickgrid/plugins/slick.rowselectionmodel.js
```diff
@@ -34,10 +34,12 @@
 
     function wrapHandler(handler) {
       return function () {
+        var result;
```
jsandoe: There is a modification of slick.rowselectionmodel, but the case table installs its own model, DG.CaseTableRowSelectionModel.

kswenson: I fixed it in both places for consistency in case the original were ever used or referenced down the road. If we don't anticipate using the original perhaps it would be clearer to remove it entirely.

apps/dg/components/case_table/case_table_adapter.js
kswenson: Good thought. I'll investigate.

apps/dg/libraries/slickgrid/plugins/slick.rowselectionmodel.js
jsandoe: I don't think there is any reason to remove it, more concerned that it get in the right version,  if it is depended on.

kswenson: The wrapHandler() in DG.CaseTableView.subscribe() was the one that I needed to fix for things to work. The other(s) would just be for consistency in case they ever are used by clients that require the return value. Since we don't seem to be using Slick.RowSelectionModel, it seems to me it should be removed for all of the reasons we usually remove unused code. There are lots of pieces of SlickGrid that we don't include in CODAP because we're not using them.

apps/dg/components/case_table/case_table_adapter.js
kswenson: I believe I was just following the convention used in selectRowsInRange() immediately above, but further investigation suggests that it doesn't matter. When the context applies a `selectCases` change with `select: true`, it only selects. To deselect, one must pass `select: false`. Given that, I'm inclined to think that not applying the change is clearer than applying a no-op change which some might expect to have an effect.

apps/dg/libraries/slickgrid/plugins/slick.rowselectionmodel.js
kswenson: I added another commit which fixes the wrapHandler() in the DG.CaseTableRowSelectionModel. In my testing, removing Slick.RowSelectionModel seems not to have any effect, so I'm inclined to commit its removal as well.

jsandoe: Makes sense.

apps/dg/components/case_table/case_table_adapter.js
jsandoe: Ah, sorry, I was mistaken. Thanks for double-checking.

PR #140, opened 2017-01-05
Function strings0369

nstclair-cc:
Hi @bfinzer,

This is still a work in progress. I mostly paid attention to how we use "type" and "name" in this iteration. I also added examples where it seemed necessary, paying attention to the Strings functions.

Does the description for concatA look correct to you? There might be a better example to use.

Thanks!

Talya

### DIFF
--- a/apps/dg/resources/en/function_strings.json
+++ b/apps/dg/resources/en/function_strings.json
@@ -490,6 +490,11 @@
         "description": "Will be evaluated in the context of the next case, return the resulting value."
       },
       {
+        "name": "argument",
+        "type": "filter",
+        "description": "A default parameter if using a filter. If there is no default, the last case evaluates as empty."
+      },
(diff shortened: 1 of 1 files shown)


apps/dg/resources/en/function_strings.json
```diff
@@ -490,34 +490,44 @@
         "description": "Will be evaluated in the context of the next case, return the resulting value."
       },
       {
+        "name": "argument",
+        "type": "filter",
+        "description": "A default parameter if using a filter. If there is no default, the last case evaluates as empty."
```
bfinzer: Actually, I believe 'next' and 'prev' have **two** optional arguments: 'default' and 'filter.' The first of these is the value that will be returned if there is no next (or previous) case. The second of these limits the cases that are considered.

PR #141, opened 2017-01-23
Added exportFilesViaPostMessage url option [#137910223]

dougmartin:
This allows for the existing graph camera tool to send the image data to the window parent via postMessage() inside of using the CFM export function.

### DIFF
--- a/apps/dg/core.js
+++ b/apps/dg/core.js
@@ -278,7 +278,7 @@ DG = SC.Application.create((function () // closure
     runKey: getUrlParameter('runKey') || '',
 
     /**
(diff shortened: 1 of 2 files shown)


scytacki:
So @dougmartin based on this feedback, what about adding the handling of this parameter to the CFM and it would basically modify the CFM's implementation of `saveSecondaryFileAsDialog`?  
I agree that this is a special case feature.

dougmartin:
Closing this without merging as this code has been moved to the CFM (https://github.com/concord-consortium/cloud-file-manager/pull/82)

PR #143, opened 2017-02-10
Restore inadvertently removed functionality of `cfmBaseUrl` query parameter

kswenson:
In an earlier effort to consolidate the way CODAP loads its bundled libraries, I switched the loading of the CFM to dynamically-generated script tags rather than using ajax calls. In the process, I inadvertently eliminated support Doug had added for loading the CFM dynamically from an external URL, primarily for debugging purposes. With this PR, if the `cfmBaseUrl` query parameter is not specified, the CFM will continue to be loaded by dynamic script tag. When the `cfmBaseUrl` query parameter is present, however, the previous ajax method will be used. This allows the shipping version of CODAP, for instance, to be tested with an alternate CFM implementation. As part of Doug's original work, the specified `cfmBaseUrl` must be hosted at `codap.concord.org` for security reasons.

### DIFF
--- a/apps/dg/lib/index.rhtml
+++ b/apps/dg/lib/index.rhtml
@@ -99,15 +99,20 @@
     <!-- https://www.html5rocks.com/en/tutorials/speed/script-loading/ -->
     <script>
       (function() {
+        // load the CFM here if cfmBaseUrl is not specified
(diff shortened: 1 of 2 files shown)


dougmartin:
Looks good 👍 I'll not going to merge since you also requested @jsandoe to review the code.

PR #144, opened 2017-02-16
Added support for importing csv data uris [#138395609]

dougmartin:

PR #146, opened 2017-03-22
Added snapshot icon to case table

dougmartin:
This code was removed from https://github.com/concord-consortium/codap/pull/144 into its own PR so 
that it can be evaluated independently.

The use case for this new feature is to have exports of graphs and tables both be in the slideout menu so they can be used with the minimal number of clicks when the sharinator is in use.


dougmartin:
Closing during cleanup of old non-merged PRs.

PR #149, opened 2017-04-14
Updated function_strings.json 

nstclair-cc:
Follow up to PT story: https://www.pivotaltracker.com/story/show/143592331

Also added new PT story: https://www.pivotaltracker.com/story/show/143735903

### DIFF
--- a/apps/dg/english.lproj/function_strings.json
+++ b/apps/dg/english.lproj/function_strings.json
@@ -234,7 +234,7 @@
       {
         "name": "date",
         "type": "string",
-        "description": "The specified date. May be a string or a computed date."
+        "description": "The specified date. May be a string or a computed date. Note that the string needs to be inside quotes."
       }
     ],
     "examples": [
@@ -249,7 +249,7 @@
       {
         "name": "date",
         "type": "string",
-        "description": "The specified date. May be a string or a computed date."
+        "description": "The specified date. May be a string or a computed date. Note that the string needs to be inside quotes."
       }
     ],
     "examples": [
@@ -264,7 +264,7 @@
       {
         "name": "date",
         "type": "string",
-        "description": "The specified date. May be a string or a computed date."
+        "description": "The specified date. May be a string or a computed date. Note that the string needs to be inside quotes."
       }
     ],
     "examples": [
@@ -279,7 +279,7 @@
       {
         "name": "dateTime",
         "type": "string",
-        "description": "The specified date and time. May be a string or a computed date-time."
+        "description": "The specified date and time. May be a string or a computed date-time. Note that the string needs to be inside quotes."
       }
     ],
     "examples": [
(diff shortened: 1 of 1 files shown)


kswenson:
Only literal strings need to be in quotes. In many formulas, attributes and other expressions (which shouldn't be quoted) will be passed as function arguments. I wonder if the proposed changes might sometimes lead to the expectation that quotes are always required.

apps/dg/english.lproj/function_strings.json
```diff
@@ -396,7 +396,7 @@
       {
         "name": "filter",
         "type": "filter",
-        "description": "A true-false expression that limits the cases considered."
+        "description": "A true-false expression that limits the cases considered. Note that the string needs to be inside quotes."
```
bfinzer: There is no mention of a string here, so the note doesn't apply. True of a number of the descriptions that follow this one.

apps/dg/english.lproj/function_strings.json
```diff
@@ -234,7 +234,7 @@
       {
         "name": "date",
         "type": "string",
-        "description": "The specified date. May be a string or a computed date."
+        "description": "The specified date. May be a string or a computed date. Note that the string needs to be inside quotes."
```
bfinzer: I would suggest that instead of adding this sentence, we do it as:
May be a string (in quotes) or a …

apps/dg/english.lproj/function_strings.json
```diff
@@ -432,12 +432,12 @@
       {
         "name": "\"otherDataSet\"",
         "type": "string",
-        "description": "The data set in which you want to look up the value."
+        "description": "The data set in which you want to look up the value. Note that the string needs to be inside quotes."
```
bfinzer: I suggest:
The name of the data set (in quotes) in which you want to look up the value.

apps/dg/english.lproj/function_strings.json
```diff
@@ -432,12 +432,12 @@
...
       },
       {
         "name": "\"attrName\"",
         "type": "string",
-        "description": "The name of the attribute in the other data set with the desired value."
+        "description": "The name of the attribute in the other data set with the desired value. Note that the string needs to be inside quotes."
```
bfinzer: I suggest:
The name of the attribute (in quotes) in the …

apps/dg/english.lproj/function_strings.json
```diff
@@ -1179,12 +1179,12 @@
       {
         "name": "stringToLookIn",
         "type": "string",
-        "description": "The string you want to find."
+        "description": "The string you want to find. Note that the string needs to be inside quotes."
```
bfinzer: Suggest:
A string (in quotes) or an expression that evaluates to a string.

This applies to a number of the changes below this one.

apps/dg/english.lproj/function_strings.json
```diff
@@ -1195,27 +1195,27 @@
   },
   "join": {
     "displayName": "join",
-    "description": "Combines multiple strings into a single string with a delimiter.",
+    "description": "Combines multiple strings into a single string with a delimiter. Note that the string needs to be inside quotes.",
```
bfinzer: This doesn't make sense as part of the description of the function.

PR #150, opened 2017-04-15
Add number toggle UI to graph menu; show values rather than indices for toggles

kswenson:
- toggle labels default to first parent attribute values rather than parent case indices [#143266497]
- improve layout of toggle labels with arbitrary-length string labels
- add "Show/Hide Parent Visibility Toggles" to the graph menu
- Showing/Hiding Parent Visibility Toggles is undoable
- fix bug in which document would show through transparent number toggle view
- fix bug which prevented "Hide All" label from appearing when appropriate
- fix bug which prevented tooltip from appearing for "Show/Hide All" label

Combining the best attributes of the various naming suggestions provided, I considered using "Show/Hide Parent Case Quick Visibility Filter Toggles-inator", but in consultation with Bill we decided to try "Show/Hide Parent Visibility Toggles" for now.

Also in conversation with Bill, this PR removes the rescale button from the number toggle UI, as it is redundant with the rescale button in the inspector which is generally right next to it.

### DIFF
--- a/apps/dg/components/graph/graph_model.js
+++ b/apps/dg/components/graph/graph_model.js
@@ -235,10 +235,12 @@ DG.GraphModel = DG.DataDisplayModel.extend(
 
       this._plots = [];
 
-      if( DG.IS_INQUIRY_SPACE_BUILD || this.get('enableNumberToggle')) {
-        this.set('enableNumberToggle', true);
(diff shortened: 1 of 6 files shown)


bfinzer:
Looks good, including removal of the rescale button.

PR #177, opened 2017-06-09
Fix display of qualitative and color cells [#146961277]

kswenson:
- these were broken by HTML-escaping added in PR #169
- fixed by moving HTML-escaping to the cell formatter
- requires separate HTML-escaping of tooltip values

Note: it occurred to me while working on this that having a client-provided `tooltipFormatter` would be a better architecture, but for now we fix the bug and leave that for a later improvement.


kswenson:
Good catch; I'll make the fix.

PR #193, opened 2017-07-12
Sort cases in case table

kswenson:
Add new case table column header menu items for `Sort Ascending` and `Sort Descending`
- sorting is undoable

Also fixed a couple of issues noticed along the way
- fixed formatting of boolean values
- eliminated SproutCore warning when creating attribute (SC.run())
- fixed bugs in `DG.DataSet.value()` function


apps/dg/utilities/data_utilities.js
```diff
@@ -173,4 +173,96 @@ DG.DataUtilities.toString = function (iValue) {
...
+    kTypeNaN = 4,
+    kTypeError = 5,
+    kTypeSimpleMap = 6, // e.g. boundaries
+    kTypeUnknown = 7,
+    kTypeEmpty = 8;
+
+/**
+  Returns the type code corresponding to the specified value.
+ */
+DG.DataUtilities.typeCode = function(value) {
```
jsandoe: Should DG.DataUtilitities.typeCode have specific handling for Date objects? In the "Four Seals" std document, the "date" field does not sort.

PR #197, opened 2017-09-07
Support POEditor translations; initial Turkish & Hebrew translations

kswenson:
- Master English strings file is now `lang/strings/en-US.json`, which is a strict JSON file except that comments are allowed. (Comments are stripped before use where appropriate.) In other words, code changes should be made to `lang/strings/en-US.json`, _not_ `english.lproj/strings.js` (which is now an output file generated by script).
- Changes to master English strings file can be pushed to POEditor with `npm run poeditor:push`. (POEditor API key must be specified as command-line argument or set as environment variable. It is too sensitive to be committed to Github.)
- Updated translations can be pulled from POEditor with `npm run poeditor:pull`. (POEditor API key must be specified as command-line argument or set as environment variable. It is too sensitive to be committed to Github.) Note that the pull command also copies the master English strings file into its JavaScript form in the project, so it should be used to update the English as well.
- The CFM is now configured to support multiple languages and to auto-detect the correct language, so only one generic copy of the CFM is required rather than a separate copy for each language.
- `function_strings.json` is not currently translatable via POEditor. It requires a modification of the push script to flatten the file before pushing, and then to the pull script to unflatten after pulling.
- Testable at http://codap.concord.org/~kswenson/translate/static/dg/tr/cert/index.html (Turkish) and http://codap.concord.org/~kswenson/translate/static/dg/he/cert/index.html (Hebrew).


jsandoe:
@kswenson Is there a PT story associated with these changes? Couldn't find in any of the commits. If there is none, it would be good to add one to at least the CODAP project, to track QA.

PR #200, opened 2017-09-13
Plugin API improvements

kswenson:
- Fix component deletion
- Support more properties in component `update` calls
  - all `directMapping` properties are supported
  - add specific mapping for `WebView`'s `URL` property
  - return `{ success: false }` when updating a component that has been destroyed

Note: In testing I have seen some intermittent failures of the component deletion code. The plugin API is now calling the same code that is used when clicking the close box of a component, so if there are issues there I believe they are beyond the scope of this PR.

### DIFF
--- a/apps/dg/components/data_interactive/data_interactive_phone_handler.js
+++ b/apps/dg/components/data_interactive/data_interactive_phone_handler.js
@@ -1625,8 +1625,10 @@ DG.DataInteractivePhoneHandler = SC.Object.extend(
           },
 
           'delete': function (iResource) {
-            var component = iResource.component;
-            component.destroy();
+            var component = iResource.component,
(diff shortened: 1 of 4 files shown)


jsandoe:
@kswenson , I know this PR was responsive to issues Dan raised. Do you know if there was a PT story or stories corresponding?

jsandoe:
The bugs addressed by this PR are [#151133909] and [#151133806]

kswenson:
The PT story that drove this was from Inquiry Space 2's Ramp Game [#151097630](https://www.pivotaltracker.com/story/show/151024072).

jsandoe:
Thanks for the link to the IS2 story. I think the work should have CODAP Stories in addition. Otherwise, they are missed for QA tracking.

PR #201, opened 2017-09-20
Support substring matching of the `formatStr` for log monitors

kswenson:
Previously, log monitors could match on topic string (full or substring) or format string (full). This extends the matching to support substring matches on format strings. This is in service of Inquiry Space PT story [#151097630](https://www.pivotaltracker.com/story/show/151097630).


dougmartin:
Looks good 👍 - I'd recommend updating the wiki entry when this merges (https://github.com/concord-consortium/codap/wiki/CODAP-Data-Interactive-Plugin-API#logmessagemonitor)

PR #205, opened 2017-10-09
Add null test for `tColor` in `forEachCategory`

kswenson:
@bfinzer This is a local fix for a broken graph that showed up in an InquirySpace 2 document. I didn't look into it further than making the minimal null test change. You may want to see if there's more that should be done in this situation. Evangeline will probably create a [PT story](https://www.pivotaltracker.com/story/show/151827266) if she hasn't already, but the document is https://codap.concord.org/releases/staging/static/dg/en/cert/index.html#shared=24532.

@bfinzer Ignore the "failing" tests reported below. Jonathan and I were in the middle of setting up the Travis build on Friday and ran into some issues that have yet to be resolved.

### DIFF
--- a/apps/dg/models/attribute_model.js
+++ b/apps/dg/models/attribute_model.js
@@ -226,7 +226,7 @@ DG.Attribute = DG.BaseModel.extend(
         DG.assert(!SC.none( tCategories) && SC.isArray( tCategories.__order));
         tCategories.__order.forEach( function( iCategory, iIndex) {
           var tColor = tCategories[iCategory];
-          tColor = tColor && tColor.colorString;
+          tColor = tColor && tColor.colorString ? tColor.colorString : tColor;
           iFunc( iCategory, tColor, iIndex);
         });
       },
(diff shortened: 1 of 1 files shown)


apps/dg/models/attribute_model.js
```diff
@@ -226,7 +226,7 @@ DG.Attribute = DG.BaseModel.extend(
         DG.assert(!SC.none( tCategories) && SC.isArray( tCategories.__order));
         tCategories.__order.forEach( function( iCategory, iIndex) {
           var tColor = tCategories[iCategory];
-          tColor = tColor.colorString || tColor;
+          tColor = tColor && tColor.colorString;
```
bfinzer: It is frequently the case that tColor is already a string, in which case that's what we want to go with. It is not required that tColor have a property colorString, but if it does, that's what we want to go with.

kswenson:
I think I got it now. If not, feel free to add a commit of your own to the PR.

PR #206, opened 2017-10-10
Fix invokeOnce() warning with addition of SC.run() in ajax callback

kswenson:


jsandoe:
This change was made on a separate branch and  committed there.

PR #209, opened 2017-10-27
Snapshot draw tool

kswenson:
Implement "Copy as Image" [#152123215], [#152123320]
- graph snapshot icon brings up menu with two options
  - Copy as Image (brings up Draw Tool component with Graph image)
  - Export image... (export PNG via CFM, as before)

Requires deployment of CODAP Data Interactives [PR#18](https://github.com/concord-consortium/codap-data-interactives/pull/18).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,8 @@
 sudo: false
 language: node_js
 node_js: node
-before_install:
-- rvm use 2.2.7
 install:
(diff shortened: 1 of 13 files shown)


kswenson:
Vis-a-vis the queuing, GameController's `activeChannel` is only configured upon connection, which may be after the first commands have been issued, so we may not have an iFramePhone available initially.

PR #210, opened 2017-10-31
Fix number toggle bugs

kswenson:
- Fix bugs in handling of 3+ level hierarchies [#152368313]
- Fix bug in handling changes to number toggle label value [#152368622]

### DIFF
--- a/apps/dg/components/graph/plots/number_toggle_model.js
+++ b/apps/dg/components/graph/plots/number_toggle_model.js
@@ -186,8 +186,18 @@ return {
    */
   childrenOfParent: function( iIndex) {
     var tParents = this.get('parentCases' ),
-        tParent = (iIndex < tParents.length) ? tParents[ iIndex] : null;
-    return tParent ? tParent.get('children' ).flatten() : [];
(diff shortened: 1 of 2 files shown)


apps/dg/components/graph/plots/number_toggle_model.js
```diff
@@ -186,8 +186,18 @@ return {
    */
   childrenOfParent: function( iIndex) {
     var tParents = this.get('parentCases' ),
-        tParent = (iIndex < tParents.length) ? tParents[ iIndex] : null;
-    return tParent ? tParent.get('children' ).flatten() : [];
+        tParent = (iIndex < tParents.length) ? tParents[ iIndex] : null,
+        tChildren = tParent ? tParent.get('children') : [];
+    for (var i = 0; i < tChildren.get('length'); ++i) {
```
bfinzer: Is there a reason not to use tChildren.forEach here?

kswenson: `tChildren` is modified inside the loop to achieve the recursive behavior. The behavior of `forEach` is not clear under these circumstances. Your comment points out the need for a comment to that effect, which I will add momentarily.

PR #214, opened 2017-11-05
Fix function info icon layout issues [#152088101]

kswenson:


kswenson:
I don't know that I'd call it a React idiom. I think in this case Doug just figured it was a more concise way of saying the same thing.

PR #215, opened 2017-11-17
Improve number toggle performance [#152949179]

kswenson:
- eliminate bottleneck in `getFirstParentAttribute()`
- replace calls to `flatten()` with calls to `slice()` which should perform slightly better and is clearer

@bfinzer You might consider adding your commit to this PR (use the `number-toggle-performance` branch), which would allow the two to be tested together.


bfinzer:
@kswenson Can you review my changes and merge? (I think I've screwed up the protocol a bit.)

bfinzer:
Yes. Sorry for missing that. I think updateCases should be part of the list, and I’ll make the change and commit it with Jonathan’s review.

PR #231, opened 2018-07-27
Fix text selection on touch systems

kswenson:
Fix touch text selection [#158700634]
- disable touch text selection at the page level
- enable it in components where it's appropriate

@jsandoe This PR depends on #230. Once that is merged I will rebase this one.

### DIFF
--- a/apps/dg/resources/dg.css
+++ b/apps/dg/resources/dg.css
@@ -60,7 +60,7 @@ html, body {
 .sc-view.allow-select {
     -khtml-user-select: text;
     -moz-user-select: text;
-    -ms-user-select: none;
+    -ms-user-select: text;
     -webkit-user-select: text;
     user-select: text;
     cursor:  text;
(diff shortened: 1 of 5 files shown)


apps/dg/resources/dg.css
```diff
@@ -42,6 +42,30 @@
...
+    -ms-user-select: none;
+    user-select: none;
+    -webkit-touch-callout: none;
+}
+
+/* SproutCore inexplicably left out -ms-user-select and -webkit-user-select */
+.sc-view.allow-select {
+    -khtml-user-select: text;
+    -moz-user-select: text;
+    -ms-user-select: none;
```
jsandoe: Why "none" for ms-user-select? Do we not want to allow selection there?

kswenson: Good catch. Copy/paste error.

jsandoe: Why "none" for ms-user-select? Do we not want to allow selection there?

kswenson: Good catch. Copy/paste error.

PR #232, opened 2018-07-27
Add touch-and-hold tooltips to toolbar icons

kswenson:
Add touch-and-hold tooltips to toolbar icons [#158703450]
- uses SproutCore's gesture recognition system for event handling
- uses Popper.js/Tooltip.js for tooltip display

@jsandoe This PR depends on #230 and #231. Once those are merged, I'll rebase this PR.

PR #233, opened 2018-07-27
Fix case table and formula editor touch issues

kswenson:
- Set `dg-wants-touch` on table view
  - makes it easier to bring up column header menu (among other things) [#158121857]
- Set `dg-wants-touch` on formula edit view and formula editor hints
  - makes hints/completions selectable on touch
- Move CodeMirror into `codap-lib-bundle`
- Monkey-patch CodeMirror's `touchend` handler to set edit caret properly on touch [#158700665]

@jsandoe This PR depends on #230, #231, and #232. Once those are merged, I'll rebase this one.


apps/dg/components/case_table/case_table_view.js
```diff
@@ -867,6 +867,14 @@ DG.CaseTableView = SC.View.extend( (function() // closure
                                       if( !ioMenuItem.disabled)
                                         ++enabledItems;
                                    });
+          // add 'dg-wants-touch' class to menus
+          setTimeout(function() {
+            $('.slick-header-menu')
+              .addClass(function(index, classes) {
+                return (!classes || (classes.indexOf('dg-wants-touch') < 0))
+                        ? 'dg-wants-touch' : '';
+              });
```
jsandoe: I think that this code is adding 'dg-wants-touch' if not already present. I think this is already the behavior of $(..).addClass(string). Is there more to this that I am missing?

kswenson: That may be the behavior of `addClass`, but it's not documented as such. The closest the docs get to addressing the question is "It's important to note that this method does not replace a class. It simply adds the class, appending it to any which may already be assigned to the elements." -- which sounds more like appending the class independent of whether it already exists or not.

jsandoe: I agree that its not documented as such, so your implementation makes sense. It would be possible that they could change the behavior and still be true to the documentation. I stand corrected.

PR #235, opened 2018-08-01
Fix drag offsets on iOS Safari

kswenson:
Fix drag offsets on iOS Safari [#159418765]

The fix is actually in the SproutCore submodule. This PR just updates the submodule pointer. Note that this PR currently points the submodule at the PR branch. This PR should be updated to the correct submodule pointer once the submodule PR is merged.


kswenson:
Another option would have been to push a commit to this PR branch with the updated submodule reference and then merge the PR. Six of one, half-dozen of the the other.

PR #239, opened 2018-08-12
Fix production build issues while re-bundling CodeMirror

kswenson:
- Re-bundle CodeMirror while fixing the production build issues that resulted from bundling it the first time.
- Add Gemfile to document gem dependencies and simplify gem installation
- Update .travis.yml to call `bundle install`
- Eliminate need for `webpack-shell-plugin`. (Bill seemed to have issues with this part of the Webpack build.)

IMPORTANT: Developers will need to update their Ruby gems as a result of this change. This can be done by executing `bundle install` or `gem install oily_png` after pulling the contents of this PR.

Back story: I _hate_ going backwards.

In PR #233, I upgraded the version of CodeMirror and at the same time removed it from the CODAP sources and into the bundle built by Webpack. Managing dependencies with npm/yarn and Webpack makes it easier to upgrade to new versions, means we don't have to keep redundant/stale copies of libraries in the CODAP sources, etc. It's a win all the way around. One improvement I added at this time was the ability to `require()` CSS files and have them bundled as well. For CodeMirror, we use three JS files and two CSS files, so moving all of that into the bundle was a significant simplification. Ultimately, we'd like to move all CODAP dependencies into the bundle. I went on to submit several more touch-related PRs, at which point I was feeling pretty good about what I'd accomplished during this iteration on CODAP.

Until Jonathan mentioned that the production build was broken and he had traced it back to the commit in PR #233 that moved CodeMirror into the bundle. !@#$%? Weirdly, the build failures were stochastic. Due to the way SproutCore handles localization, a CODAP production build is essentially comprised of six independent language-specific builds. In this case, the failure would sometimes occur on the second language, and sometimes the fourth or fifth. Occasionally, the build would even complete successfully. Huh? One reason I hadn't noticed the build failure is that the Travis CI build only builds the English version on the assumption that for testing purposes we only need one language and if one language builds surely they'll all build. (May have to rethink that assumption.) The failure occurred deep in the bowels of SproutCore's Ruby build tools, and the error message was a completely uninformative `FATAL ~ data error`. Neither Jonathan nor I could come up with a quick solution and so Jonathan backed out of the CodeMirror bundling changes, taking CodeMirror out of the bundle and putting the files back in the CODAP repository.

I _hate_ going backwards -- and I hate being forced to write inelegant code due to the limitations of the tools we're using. I figured there had to be a way of fixing the build without losing the bundling. Jonathan's initial investigation led to the hypothesis that it was the bundling of the CSS that was the relevant change, so I spent a while playing around with Webpack configurations. Bundling the CSS also meant that it was loaded in a different order, so what if we change the order in which the bundled CSS is processed? What if Webpack output individual CSS files rather than a single bundle? What if Webpack output a file with a `.css.ignore` extension, so SproutCore didn't try to bundle it again? With each of these changes, sometimes the stochastic nature of the failures would change, i.e. it might be more likely to fail earlier or later, but the failures remained.

The next step was to do a deep dive into the SproutCore build tools. After littering SproutCore's Ruby build tools with logging statements I eventually determined that when the failure occurred, it always occurred when processing a sprited image file deep in SproutCore's theming submodule. Huh? What could bundling application CSS possibly have to do with spriting of images in the SproutCore theming module? I still don't know the answer, but I do know that the stack trace for the failure indicated that an exception occurred during image compression in a module called `chunky_png`. SproutCore uses a module called `Compass` to do its spriting, and Compass apparently uses `chunky_png` to do its image encoding/decoding. While reading up on Compass I came across this intriguing note:

>Compass generates PNG files using a pure-ruby library called `chunky_png`. This library can be made faster by installing a simple C extension called `oily_png`.

So I installed `oily_png` -- and have completed dozens of production builds without a single failure since then. My best guess is that there's some sort of thread timing issue that explains the stochasticity of the results. For this PR we add a `Gemfile` which specifies the version of SproutCore we require (1.11) and the dependency on `oily_png`, along with a handful of other minor improvements. In addition to building consistently, the build should also be faster.

I _hate_ going backwards.

@jsandoe I've spent _way_ too much time on this already. I leave it to you to update the CODAP Developer documentation accordingly.


kswenson:
@jsandoe I think of it as replacing a broken module with a working one. The fact that it should be a build optimization as well is just a bonus. I sincerely hope that we will eliminate our dependence on the Ruby build tools before we have to revisit this.

PR #242, opened 2018-08-15
[#141778481] Case table: drag attribute in between hierarchies.

ttsuchiya:
@jsandoe Please review the code. The old dividers should be destroyed more properly now, as we discussed. Thanks.


apps/dg/resources/dg.css
```diff
@@ -1214,19 +1214,23 @@ input.dg-editor-text {
...
+    -webkit-writing-mode: vertical-rl;
+    writing-mode: vertical-rl;
     text-align: center;
-    padding: 0.5em;
     transform: rotate(180deg);
+    display: flex;
+    align-items: center;
+    justify-content: center;
 }
+
```
jsandoe: Taka, I am curious if you verified the vertical-rl writing mode in Chinese. That translation may be affected by the change.

apps/dg/components/case_table/case_table_controller.js
```diff
@@ -161,7 +161,39 @@ DG.CaseTableController = DG.ComponentController.extend(
...
+        if (contentView) {
+          var dividerDropDataCallback = function (iDivider) {
+            var dropData = iDivider.getPath('dropData');
+            if (SC.none(dropData)) {
+              return;
+            }
+            var context = iDivider.getPath('dataContext');
+            var parentTable = iDivider.getPath('leftTable');
+
+            // TODO: Add the rest of UNDO procedures in createCollectionCommand.
```
jsandoe: Is this TODO complete? If so, maybe the comment should be removed.

ttsuchiya: It has not been done yet. I could work on it as a separate story?

PR #243, opened 2018-08-24
[#160034361] Bug fix: Default graph with 13K points takes overlong to…

bfinzer:
… show.

CasePlotView, when no elements have been created, now creates all the elements at once using descriptions rather than adding each element individuallyl.

### DIFF
--- a/apps/dg/components/graph/plots/case_plot_view.js
+++ b/apps/dg/components/graph/plots/case_plot_view.js
@@ -160,53 +160,152 @@ DG.CasePlotView = DG.PlotView.extend(
       /**
        We may clear and draw everything from scratch if required.
        */
+
       drawData: function () {
-        var animatePoints = function () {
-          // If the user closes the graph component while the animation is happening, we're likely to
-          // crash because we no longer have paper to draw points on. Detect and bail!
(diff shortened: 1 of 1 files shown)


apps/dg/components/graph/plots/case_plot_view.js
```diff
@@ -160,53 +160,152 @@ DG.CasePlotView = DG.PlotView.extend(
...
+                  stroke: tStrokeParams.strokeColor,
+                  'stroke-opacity': tStrokeParams.strokeTransparency,
+                  cursor: 'pointer'
+                }).forEach(function (iElement, iIndex) {
+                  tPlottedElements.push(iElement);
+                  iElement.addClass(DG.PlotUtilities.kDotClassName);
+                  iElement.index = iIndex;
+                  iElement.node.setAttribute('shape-rendering', 'geometric-precision');
+                  return true;
+                });
```
jsandoe: I am curious why a number of these things were not set in the json object during createCircleDescription(). Did you try it and find no performance improvement? I ask because it looks to me like Raphael spends a great deal of time in a loop after it gets control and before it starts rendering.

bfinzer: Well, I think you are right. I'll give it a try.

PR #247, opened 2018-09-06
158016370 start first child case parent

mklewandowski:
Changes to ensure that navigation tools allow user to move to first child case of parent case when viewing the case card.  Helper function `getSelectedCaseFirstChildCaseID()` is used by both `moveToNextCase` and `moveToPreviousCase()` to access the proper child case.


apps/dg/components/case_card/case_card.js
```diff
@@ -385,20 +385,51 @@ DG.React.ready(function () {
...
+
           /**
            *
            * @param iCollectionClient
            * @param iCaseIndex  {
            */
           moveToPreviousCase: function (iCollectionClient, iCaseIndex) {
+            var tPrevIndex = null;
+            var tNumCases = iCollectionClient.getPath('collection.cases').length;
+            if (SC.none(iCaseIndex) && iCollectionClient.collection.parent) {
```
kswenson: In SproutCore parlance, this should be `iCollectionClient.getPath('collection.parent')` analogous to the way cases are accessed above. The same pattern occurs below as well.

apps/dg/components/case_card/case_card.js
```diff
@@ -385,20 +385,51 @@ DG.React.ready(function () {
...
 
+          /**
+           * -------------------Get first child case of selected case-----------------
+           */
+          getSelectedCaseFirstChildCaseID: function (iCollection) {
+              var tSelectedCaseFirstChildCaseID,
+                  tContext = this.props.context;
+              var tSelectedCase = tContext.getCollectionByID(iCollection.get('id')).getPath('casesController.selection').toArray();
+              if (tSelectedCase.length) {
+                tSelectedCaseFirstChildCaseID = tSelectedCase[0].children[0].id;
```
kswenson: In SproutCore parlance, this should probably be `tSelectedCase[0].get('children')[0].id`.

mklewandowski:
@bfinzer Changes have been made and tested and new commit has been added based on @kswenson's feedback.

PR #248, opened 2018-09-06
158015541 case card does not show certain

mklewandowski:
Extend display capabilities in case card cell to handle the following new types: Boundary thumbnails, Errors, Color bars, Booleans, Qualitative bars.  Added new CSS instead of recycling existing CSS from non-case card stylesheet (to avoid unexpected changes and allow case-card specific tweaks).  Some of this code has redundancy with similar case table handling in `case_table_adapter.js`.  However, React implementation vs. SlickGrid implementation requires subtle differences between the two.  
[#158015541]

### DIFF
--- a/apps/dg/components/case_card/case_card.js
+++ b/apps/dg/components/case_card/case_card.js
@@ -311,6 +311,8 @@ DG.React.ready(function () {
                 tCase = iShouldSummarize ? null : (iChildmostSelected && iChildmostSelected[0]) || iCases[0],
                 tValue = iShouldSummarize ? '' : tCase && tCase.getValue(tAttrID),
                 tType = iAttr.get('type');
+            if( tValue && tValue.jsonBoundaryObject)
+              tType = 'boundary';								
             this.state.attrIndex++;
             if (isNotEmpty(tUnit))
               tUnitWithParens = ' (' + tUnit + ')';
--- a/apps/dg/components/case_card/case_card.js
+++ b/apps/dg/components/case_card/case_card.js
@@ -8,7 +8,8 @@ DG.React.ready(function () {
       table = React.DOM.table,
       tbody = React.DOM.tbody,
       tr = React.DOM.tr,
-      td = React.DOM.td//,
+      td = React.DOM.td,
+      img = React.DOM.img//,
       // input = React.DOM.input
   ;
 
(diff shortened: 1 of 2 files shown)


apps/dg/components/case_card/case_card.js
```diff
@@ -308,16 +309,66 @@ DG.React.ready(function () {
...
+              tColorValueField = span({
+                className: 'react-data-card-color-table-cell',
+                style: spanStyle
+              });
+            } else if (tType === 'qualitative') {
+              if (tValue === null || tValue === undefined || tValue === "") {
+                tValue = "";
+              } else {
+                var color = DG.PlotUtilities.kDefaultPointColor,
+                    tWidth = tValue;
```
kswenson: `;` => `,`, but according to my editor `tWidth` is not used so the entire line can presumably be removed.

kswenson:
Also, `npm run lint` reports a number of other issues that should be addressed.

apps/dg/components/case_card/case_card.js
mklewandowski: I had a more verbose implementation originally.  `twidth` was meant to be yanked out.

apps/dg/components/case_card/case_card.js
```diff
@@ -308,16 +309,66 @@ DG.React.ready(function () {
...
+                      className: 'react-data-card-thumbnail',
+                      src: tThumb
+                    });
+                tBoundaryValueField = span({}, tBoundaryInternalImage);
+              }
+              else if( tBoundaryObject && (tBoundaryObject.jsonBoundaryObject instanceof Error)) {
+                tValue = tBoundaryObject.jsonBoundaryObject.name + tBoundaryObject.jsonBoundaryObject.message;
+              }
+              tValue = tResult;
+            } else if (DG.isNumeric(tValue) && typeof tValue != 'boolean') {
```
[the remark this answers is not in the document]
mklewandowski: 👍

apps/dg/components/case_card/case_card.js
```diff
@@ -308,16 +309,66 @@ DG.React.ready(function () {
...
+              var tColor = tinycolor( tValue.toLowerCase().replace(/\s/gi,'')),
+                  spanStyle = {
+                    backgroundColor: tColor.toString('rgb'),
+                  };
+              tColorValueField = span({
+                className: 'react-data-card-color-table-cell',
+                style: spanStyle
+              });
+            } else if (tType === 'qualitative') {
+              if (tValue === null || tValue === undefined || tValue === "") {
```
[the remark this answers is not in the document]
mklewandowski: This code pattern also exists elsewhere in the codebase, but that can be for another day.  For now, I'll add `SC.empty()` here to at least keep this chunk a little tidier.

apps/dg/components/case_card/case_card.js
```diff
@@ -308,16 +309,65 @@ DG.React.ready(function () {
                 tHasFormula = iAttr.get('hasFormula'),
                 tFormula = iAttr.get('formula'),
                 tCase = iShouldSummarize ? null : (iChildmostSelected && iChildmostSelected[0]) || iCases[0],
-                tValue = iShouldSummarize ? '' : tCase && tCase.getValue(tAttrID);
+                tValue = iShouldSummarize ? '' : tCase && tCase.getValue(tAttrID),
+                tType = iAttr.get('type');
             this.state.attrIndex++;
```
bfinzer: You also need:
          if( cellValue && cellValue.jsonBoundaryObject)
            type = 'boundary';

mklewandowski: Right.  This seems to fix it (tweaked for local vars):
if( tValue && tValue.jsonBoundaryObject)
  tType = 'boundary';		

Changes have been added.

PR #249, opened 2018-09-07
set default document name if document name is null

mklewandowski:
Set default document ("untitled document") name if document name is null.  This appeared to happen most frequently when opening a shared document.
[#159935849]


apps/dg/controllers/document_controller.js
```diff
@@ -206,6 +206,9 @@ DG.DocumentController = SC.Object.extend(
           return;
         }
         var name = this.get('documentName');
+        if (name === null) {
```
bfinzer: A better test would be SC.empty which checks for null, undefined, and empty string.

mklewandowski: Got it.  Changes added to use SC.empty.

PR #252, opened 2018-09-08
remove "add background image" from map toolbar menu

mklewandowski:
Detect if model type is a map and remove "add background image" from map toolbar menu.
[#159509708]


mklewandowski:
Sounds good.  I'll hold off on any merging until I'm given the green light.

PR #254, opened 2018-09-18
Delete Formula (Keeping Values)

ttsuchiya:
@jsandoe Please review it whenever you have time. Thanks!


apps/dg/components/case_table/case_table_controller.js
```diff
@@ -1018,14 +1021,80 @@ DG.CaseTableController = DG.ComponentController.extend(
...
             tRef = tDataContext && tDataContext.getAttrRefByID( iAttrID),
-            tAttrName = tRef && tRef.attribute.get('name'),
-            tAttrFormula = tRef && tRef.attribute.get('formula');
+            tAttrName = tRef && tRef.attribute.get('name');
+
         DG.assert( tRef && tAttrName, "editAttributeFormula() is missing the attribute reference or attribute name" );
         // for now we use the newAttribute() method which will replace one attribute formula with another
         // if the new attribute has the same name as the old.
+
+        var tAttrFormula = tRef.attribute.get('formula');
```
ttsuchiya: I think I made this minor change for implementing the freeze / unfreeze. It should not affect the overall behaviors.

ttsuchiya: Actually, this line can be put back to the original location. It was needed for automatically unfreezing the formula upon opening the edit-formula window, but obviously I removed the freezing stuff.

apps/dg/controllers/data_context.js
```diff
@@ -1517,6 +1517,7 @@ DG.DataContext = SC.Object.extend((function() // closure
     var newPosition = constrainInt(position, 0, attrCount);
     DG.assert(originalPosition !== -1, 'Moving attribute is found in original collection');
     var casesAffected;
+
```
ttsuchiya: Auto-freezing upon dragging an attribute used to be here.

apps/dg/components/case_table/case_table_controller.js
jsandoe: I agree with putting it back to its original location. Cleaner.

ttsuchiya:
(a) As we discussed in person, it's only retained in the undo history but we could also have the toggle for "Delete Formula..." item with "Recover Deleted Formula" or something. @bfinzer any thoughts?
(b) I checked and it appears those settings are kept with the newly converted attribute.

apps/dg/components/case_table/case_table_controller.js
```diff
@@ -1019,13 +1025,153 @@ DG.CaseTableController = DG.ComponentController.extend(
...
+                attrPropsArray: [{ name: tAttrName, formula: tFormula }]
+              },
+              tResult = tDataContext && tDataContext.applyChange( tChange);
+            if( tResult.success) {
+              tRef.attribute.set('formula', tFormula);
+              tRef.attribute.set('deletedFormula', tPrevFormula);
+              hierTableView.updateColumnInfo();
+
+              var action = "attributeEditFormula";
+              this.log = "%@: { name: '%@', collection: '%@', formula: '%@' }".fmt(
```
ttsuchiya: I just found that the direct call of "Delete..." / "Recover..." functions from the menu item results in "unknown action" as the log output in console. Undo / redoing, however, logs properly as `attributeEditFormula`. Not sure what's happening here...

jsandoe:
Hello, in the PT story, you say you will add a toggle. Will that be part of this PR or another? (In other words, should this story be held open?)

ttsuchiya:
@jsandoe I pushed the changes adding a toggle to this PR branch. Could you review the latest commit?

apps/dg/components/case_table/case_table_controller.js
jsandoe: Usually, the DG.Command object will be created with a `log` property allong with 'name', 'undoString' and 'redoString'. Doesn't appear to be the case here.

jsandoe: Usually, the DG.Command object will be created with a `log` property along with `name`, `undoString` and `redoString`. Doesn't appear to be the case here.

ttsuchiya: It was a simple random scope problem with an unnecessary `.bind(this)`. I don't know where it came from, but logging should be fine now.

PR #255, opened 2018-09-19
Prevent propagation of mouse events on case card to allow proper input handling

mklewandowski:
Prevent propagation of mouse events on case card to allow proper input handling.  Mousedown and mouseup events were being bubbled up to SproutCore listener which was preventing default behaviors (focus, selection, cursor movement, etc.).  If we handle these events specifically on a case card input field, be sure to prevent them from being bubbled up to SproutCore.
[#158852455]

### DIFF
--- a/apps/dg/components/case_card/text_input.js
+++ b/apps/dg/components/case_card/text_input.js
@@ -29,13 +29,16 @@ DG.React.ready(function () {
           },
 
           componentDidMount: function () {
+            window.addEventListener('mousedown', this._onMouseDown, true);
+            window.addEventListener('mouseup', this._onMouseUp, true);
             DG.mainPage.mainPane.addListener({action: 'click', target: this, method: this._onWindowClick});
             DG.mainPage.mainPane.addListener({action: 'touchstart', target: this, method: this._onWindowClick});
           },
-
           componentWillUnmount: function () {
             DG.mainPage.mainPane.removeListener({action: 'click', target: this, method: this._onWindowClick});
             DG.mainPage.mainPane.removeListener({action: 'touchstart', target: this, method: this._onWindowClick});
+            window.removeListener('mousedown', this._onMouseDown, true);
+            window.removeListener('mouseup', this._onMouseUp, true);
           },
 
           componentWillReceiveProps: function (iNewProps) {
@@ -43,6 +46,20 @@ DG.React.ready(function () {
               this.setState({value: iNewProps.value});
             if (iNewProps.unit !== this.state.unit)
               this.setState({unit: iNewProps.unit});
+          },
+
+          _onMouseDown: function (e) {
+            var inputElement = findDOMNode(this);
+            if (e.target === inputElement) {
+              e.stopPropagation();
+            }
+          },
+
+          _onMouseUp: function (e) {
+            var inputElement = findDOMNode(this);
+            if (e.target === inputElement) {
+              e.stopPropagation();
+            }
           },
 
           _onWindowClick: function (event) {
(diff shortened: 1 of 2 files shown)


apps/dg/components/case_card/text_input.js
```diff
@@ -29,13 +29,16 @@ DG.React.ready(function () {
           },
 
           componentDidMount: function () {
+            window.addEventListener('mousedown', this._onMouseDown, true);
+            window.addEventListener('mouseup', this._onMouseUp, true);
```
kswenson: Out of curiosity, did you try adding these handlers directly to the React component? It seems like something like this should work:
```
input({
  ...
  onMouseDown(function(iEvent) {
    iEvent.stopPropagation();
  },
  onMouseUp(function(iEvent) {
    iEvent.stopPropagation();
  }
}) :
```

mklewandowski: I believe we tried this path and the React specific event handler occurred after the SproutCore handler when using this approach.  I'll double check again tomorrow.

kswenson: That's plausible. It's likely that the same may be required for touch events. For the latter case I added support for a `wantsTouch` class to `SC.RootResponder` so clients need only apply the class to let `SC.RootResponder` know to keep its mitts off. Something analogous seems like it would work here.

jsandoe: @mklewandowski, I am not sure of the state of this PR. I don't understand whether the issue @kswenson mentioned is covered under this PR or not. Have we explored whether the case card has an issue with touch?

For my part, I am a bit ambivalent. The approach taken corresponds with the approach taken by the CFM, but may be incomplete. The approach @kswenson seems to be suggesting -- modifying the SproutCore RootResponder to support SC ignoring mouse events when a new dg-wants-mouse class is present in the hierarchy of current element -- is pretty clean, but involves a change to Sproutcore source. I am generally loathe to change Sproutcore if there is an alternative.

kswenson: One possible approach to migrating away from SproutCore would be to gradually replace SproutCore-based components with React-based ones like the data card. In that case, having a consistent means of indicating that the SC.RootResponder should let the component handle its own mouse events would be preferable to having each component do so independently.

jsandoe: Good point. @mklewandowski , do you feel up to exploring this approach? You can find how @kswenson handled the touch events, I believe, by grepping the code for dg-wants-touch.

mklewandowski: @kswenson @jsandoe This approach presented by Kirk also seems to work.  I added a `dg-wants-mouse` class that will allow an element to bypass the SproutCore mouse events (mousedown, mouseup, mousemove, click, dblclick, mousewheel).  This does, however, require changes to SproutCore'S `root_responder.JS`.  It looks like the latest SproutCore that CODAP uses is located in the Concord fork on the branch dg-sc-1.11.
https://github.com/concord-consortium/codap-sproutcore/tree/dg-sc-1.11
To clarify, should I commit my SproutCore changes to the dg-sc-1.11 branch so they can be reviewed?  I want to confirm that I commit them in the expected location.

mklewandowski:
I added changes to CODAP to handle this using a new `dg-wants-mouse` class that can be added to elements.  Waiting for permission to push related changes to the CC SproutCore fork.

mklewandowski:
The CC SproutCore fork has been updated with changes that will now permit this branch to work correctly if it is merged into master.  The SproutCore changes are here:
https://github.com/concord-consortium/codap-sproutcore/commit/e762557d6b1c9b9506d4110bdee9aab0b3940a16

PR #257, opened 2018-09-24
Add component selection to plugin API

mklewandowski:
Add component selection to plugin API.  Call using component name (requires no values).  Brings component to front of displayed component windows.
[#153373624]

### DIFF
--- a/apps/dg/components/data_interactive/data_interactive_phone_handler.js
+++ b/apps/dg/components/data_interactive/data_interactive_phone_handler.js
@@ -1833,27 +1833,23 @@ DG.DataInteractivePhoneHandler = SC.Object.extend(
             }
           },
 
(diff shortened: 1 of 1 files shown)


apps/dg/components/data_interactive/data_interactive_phone_handler.js
```diff
@@ -1833,6 +1833,29 @@ DG.DataInteractivePhoneHandler = SC.Object.extend(
...
           },
 
+          select: function (iResources) {
+            if (!iResources.component || iResources.component.get(
+                'isDestroyed')) {
+              return {success: false, values: {error: 'Component not found'}};
+            }
+            var component = iResources.component;
+
+            if (!component) {
```
kswenson: This test seems redundant with the one above.

PR #266, opened 2018-12-17
Item-level formula evaluation basic implementation

ttsuchiya:


ttsuchiya:
@jsandoe Please review. Thanks!

ttsuchiya:
That is fair. I guess these regenerateCollectionCases could use conditionals with dependent (formula) and dependency (attributes used by formula) checks rather than blind calls, but I couldn't get to the bottom of those yet. With globals concerned, the current code needs some changes in how to determine the type of dependency (simple vs aggregate dependency).

kswenson:
Closing as we're not making these kinds of changes in v2 any more.

ttsuchiya:
Thanks for letting me work on this all those years ago!

kswenson:
> Thanks for letting me work on this all those years ago!

Thanks for working on it! We may be able to incorporate some of these ideas into the ongoing v3 development.

PR #269, opened 2019-03-21
Add CODAP plugin API support for changing the order of cases

kswenson:
The API is handled as a case `update` to the `caseOrder` property, e.g.
```
codapInterface.sendRequest({
  action: "update",
  resource: "dataContext[myDataContext].collection[myCollection].caseByID[myCaseID],
  values: { caseOrder: "last" }
});
```
- Rather than `position`, as we talked about, I decided to use `caseOrder`, as `position` seems like it could refer to coordinate location rather than order of the cases within the collection. 
- Currently, `first` and `last` are the only `caseOrder` values supported, but there are code comments indicating where the API could be extended to support things like `index:n` or `before:id`.
- At the DataContext level, the API allows an array of cases so it would be possible to sort the entire collection by specifying an array of cases to be moved to `last` in the order of the array.
- There is some preexisting support for a `moveCases` notification, so I made use of this existing notification internally.
- There was a preexisting `moveCase()` method in the `CollectionModel` which was sufficiently broken that it seems clear it was not previously used. It is now fixed and in use.
- I added an `iSkipIDMapUpdate` parameter to the `moveCase()` method so that the default behavior is to update the ID map on each call, but clients can disable the update for all but the last call when multiple cases are to be moved. Excessive calls to `updateCaseIDToIndexMap()` has been a performance issue in the past, so this should make it possible to avoid that.

### DIFF
--- a/apps/dg/components/case_table/case_table_controller.js
+++ b/apps/dg/components/case_table/case_table_controller.js
@@ -357,6 +357,7 @@ DG.CaseTableController = DG.ComponentController.extend(
             break;
           case 'createCase':
           case 'createCases':
+          case 'moveCases':
             this.doCreateCases( iChange);
             invalidateAggregates = false; // handled by doDependentCases
             break;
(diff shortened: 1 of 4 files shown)


kswenson:
You're right -- you did point out that we should use "notify" and I forgot about that between the time that we wrote it on the white board and when I actually wrote the code. I'll make the necessary change.

kswenson:
I've added a commit with the requested change.

PR #270, opened 2019-03-26
Fix plugin API issues

kswenson:
Fixes a few issues encountered while developing the CODAP shared table plugin
- Fix `itemSearch` which could return success for `undefined` results
  - replace all instances of `!== null` tests with `!= null`
    - after consideration, all of them made sense to handle `undefined` like null
  - in `resolveResources()` don't dereference items if no items were found
- Fix deleteItem handler
- Call `regenerateCases()` when appropriate after deleting attributes


kswenson:
For posterity, our eslint configuration contains `"eqeqeq": ["error", "allow-null"]` which allows the `== null` and `!= null` as permitted exceptions.

PR #271, opened 2019-04-03
Support disabling top-level collection reorganization

kswenson:
Provides a mechanism for a plugin to indicate that collection reorganization is allowed generally, but not for the top-level collection. This is used by the shared table plugin to guarantee that the share names remain the top-level collection.

@jsandoe After our conversation, I added two more commits to enable rearranging attributes within the top-level collection if there are multiple attributes and to disable adding attributes to the top-level collection in the case card and the case table.

cf. [PT #165199713](https://www.pivotaltracker.com/story/show/165199713)

### DIFF
--- a/apps/dg/components/game/game_controller.js
+++ b/apps/dg/components/game/game_controller.js
@@ -88,14 +88,39 @@ DG.GameController = DG.ComponentController.extend(
        * As of build ~0428 we allow data interactives to come to the front like any other component
        * @type {boolean}
        */
-      preventBringToFront: false,
+      preventBringToFront: function(key, value) {
+        var path = 'model.content.preventBringToFront';
+        if (value != null) {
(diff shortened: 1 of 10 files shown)


apps/dg/components/game/game_controller.js
```diff
@@ -88,14 +88,39 @@ DG.GameController = DG.ComponentController.extend(
...
        * @type {boolean}
        */
-      preventBringToFront: false,
+      preventBringToFront: function(key, value) {
+        var path = 'model.content.preventBringToFront';
+        if (value != null) {
+          this.setPath(path, value);
+        }
+        return this.getPath(path);
+      }.property(),
```
kswenson: Note: I tried using bindings for these, but could not get them to work bidirectionally.

PR #278, opened 2019-04-17
Prevent deletion of plugin-protected cases

kswenson:
PR #275 added support for the plugin to protect individual cases. In that PR, users are prevented from editing protected cells in the case table. This PR extends that protection to deletion of protected cases. This is accomplished by disabling the caseIndex popup menu entirely and by enabling/disabling items in the inspector menu appropriately and then restricting deletion to unprotected cases.

As a result of this PR, the items in the inspector menu are reinterpreted to apply to deletable cases, i.e. "Delete Selected Cases" is interpreted as "Delete [Deletable] Selected Cases" and is enabled if any selected cases are deletable. Likewise, "Delete Unselected Cases" is interpreted as "Delete [Deletable] Unselected Cases" and is enabled if any unselected cases are deletable, and "Delete All [Deletable] Cases" is enabled if any cases are deletable. In all cases, only deletable cases are deleted.

At an implementation level, previously, selected cases were determined in `DataContext.getSelectedCases()`, while unselected cases were determined in `CaseDisplayController._deleteOrSetAsideUnselectedCases()`. With this PR, `DataContext.getSelectedCases()` now optionally returns unselected cases as well, eliminating the need for the `CaseDisplayController` to do so. The `CaseDisplayController` applies the deletability filter based on the plugin-protected status of each case.

PT [#165434514](https://www.pivotaltracker.com/story/show/165434514)

### DIFF
--- a/apps/dg/components/case_display_common/case_display_controller.js
+++ b/apps/dg/components/case_display_common/case_display_controller.js
@@ -175,12 +175,22 @@ DG.CaseDisplayController = DG.ComponentController.extend(
        * The menu is instantiated from the Inspector.
        */
       showDeletePopup: function() {
+        function deletableCounts(dataContext, cases) {
+          var count = cases ? cases.length : 0,
+              deletable = cases && cases.map(function(aCase) {
+                return DG.DataContextUtilities.isCaseEditable(dataContext, aCase);
+              }),
+              deletableCount = deletable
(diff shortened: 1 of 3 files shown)


jsandoe:
Is there a PT story for this?

kswenson:
There is now: [#165434514](https://www.pivotaltracker.com/story/show/165434514)

PR #279, opened 2019-04-17
Add plugin API support for `caseSearch` to match missing values

kswenson:
e.g. `caseSearch[__collaborator__==]` will return all cases with empty (undefined, null, or empty string) values.

PT [#165434555](https://www.pivotaltracker.com/story/show/165434555)


jsandoe:
Is there a PT story?

kswenson:
PT [#165434555](https://www.pivotaltracker.com/story/show/165434555)

PR #282, opened 2019-05-10
Support combining itemSearch and itemOrder to procedurally order items

kswenson:
The shared table plugin endeavors to keep each user's cases/items at the bottom of the table for ease of editing. Previously this was accomplished by moving the single top-level case for a given user to the bottom of the table. Unfortunately, since cases are frequently regenerated from items, case order is inherently ephemeral, resulting in the need to frequently move the case again and leading to visible flashing as the cases are regenerated and then moved.

With this PR, we support specifying an `itemOrder` value to the `itemSearch` resource, which allows the set of items that match the search to be ordered. Currently, the only valid values for the `itemOrder` parameter are `first` and `last`, as is true for the `caseOrder` parameter when moving cases.

Note: the Travis build is failing for unrelated reasons that are fixed in a separate PR (#283).

### DIFF
--- a/apps/dg/components/case_table/case_table_controller.js
+++ b/apps/dg/components/case_table/case_table_controller.js
@@ -381,6 +381,7 @@ DG.CaseTableController = DG.CaseDisplayController.extend(
           case 'createItems':
           case 'updateItems':
           case 'deleteItems':
+          case 'moveItems':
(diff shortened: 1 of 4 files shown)


kswenson:
There is now: [#166487911](https://www.pivotaltracker.com/story/show/166487911)

PR #290, opened 2019-06-25
Can't edit cells for non-editable attributes, even in the input row

kswenson:
Corresponds to PT [#166915590](https://www.pivotaltracker.com/story/show/166915590)


jsandoe:
This bug fix, as implemented, undoes the fix of PT story https://www.pivotaltracker.com/story/show/142880433 . @eireland made the suggestion of omitting the edit row for collections that have been configured to be unalterable, such as the parent table in a shared table configuration. This change would (a) make this bug moot, (b) avoid the awkwardness and user confusion of an empty edit row that contains no cells that can be edited, and (c) avoids the conflict with the previous fix. Discussing with @kswenson, we agreed to take this alternate approach, so am closing this PR.

PR #299, opened 2019-11-25
Added name parameter to CFM importedData message [#168406646]

dougmartin:
This fixes the importer trying to use the data uri as the name.  A seperate fix is needed in building-models to not send the componentType so that the csv import path can be followed in CODAP.

### DIFF
--- a/apps/dg/controllers/app_controller.js
+++ b/apps/dg/controllers/app_controller.js
@@ -361,8 +361,8 @@ DG.appController = SC.Object.create((function () // closure
      * @param {Boolean} iShowCaseTable
      * @return {Deferred|undefined}
      */
-    importTextFromUrl: function (iURL, iShowCaseTable) {
-      var name = this.extractNameFromURLPath(iURL);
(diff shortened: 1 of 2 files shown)


apps/dg/controllers/app_controller.js
```diff
@@ -588,7 +588,7 @@ DG.appController = SC.Object.create((function () // closure
       if (mimeSpec) {
         switch (mimeSpec.group) {
           case 'TEXT':
-            this.importTextFromUrl(iURL);
+            this.importTextFromUrl(iURL, false, iName);
```
jsandoe: If I understand this correctly, then _any_ URL dropped into CODAP will suppress the case table open. Is this right?  Outside the Sage context, its desirable for the case table to be open if a URL is dropped, just as when a CSV file is dropped.

dougmartin: The previous code was not passing a value for the second parameter.  I could set it to undefined to exactly match the previous code if you want but I thought false would be better since it is a boolean parameter and both false and undefined are "falsy".

PR #301, opened 2020-01-08
Fix getNameFromURL function

pjanik:
URL params with `.` were breaking it. E.g. for `/sage/sagemodeler.html?timestep=0.125` it was returning
`sagemodeler.html?timestep=0` while it should have been just `sagemodeler`. It was causing some other failures later in the code, which resulted in the model iframe being 300x200px only (sagemodeler app).

Related PT story: https://www.pivotaltracker.com/story/show/170507635

### DIFF
--- a/apps/dg/controllers/document_controller.js
+++ b/apps/dg/controllers/document_controller.js
@@ -870,14 +870,19 @@ DG.DocumentController = SC.Object.extend(
       },
 
       addGame: function (iParentView, iComponent, isInitialization) {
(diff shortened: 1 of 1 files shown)


pjanik:
Yes, I've linked it above: https://www.pivotaltracker.com/n/projects/1263626/stories/170507635
Although it's a Sage story, as that's where the issue was visible. I'm not sure how to reproduce it using CODAP only.

jsandoe:
OK, thanks.

PR #308, opened 2020-03-04
Improve case table vertical scrolling on touch devices

kswenson:
Testable at https://codap-dev.concord.org/branch/121247039-touch-vertical-scroll/.

- Add `dg-wants-touch` to the table body to keep SproutCore from interfering with touch events there.
- Disable touch drag selection in the body of the case table to make it easier to scroll the table contents vertically.
    - In consultation with @bfinzer we disable drag-row-selection on touch events while leaving single-touch case selection to simplify vertical scrolling by touch in the body of the case table.
- Enable momentum scrolling.

~~Note that this PR also includes a hard-coded version of the Mammals sample document to make it easier to test scrolling. This should be removed before merging.~~

### DIFF
--- a/apps/dg/libraries/jquery/jquery.event.drag-2.2.js
+++ b/apps/dg/libraries/jquery/jquery.event.drag-2.2.js
@@ -37,6 +37,7 @@ drag = $special.drag = {
 		not: ':input', // selector to suppress dragging on target elements
 		handle: null, // selector to match handle target elements
 		relative: false, // true to use "position", false to use "offset"
+		touch: true, // [CC] false to suppress touch dragging, true to allow
 		drop: true, // false to suppress drop events, true or selector to allow
 		click: false // false to suppress click events after dragend (no proxy)
 	},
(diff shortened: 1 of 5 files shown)


apps/dg/libraries/jquery/jquery.event.drag-2.2.js
```diff
@@ -37,6 +37,7 @@ drag = $special.drag = {
 		not: ':input', // selector to suppress dragging on target elements
 		handle: null, // selector to match handle target elements
 		relative: false, // true to use "position", false to use "offset"
+		touch: true, // [CC] false to suppress touch dragging, true to allow
```
jsandoe: Can you elaborate on this change? It looks like you undo the effect of it in the Slick Grid code.

jsandoe: Never mind. On reflection, I understand. This is a _new_ default. Makes sense, now.

kswenson: It seemed like the right approach from the `jquery.event.drag` perspective would be to support touch by default but then to allow clients to disable it if desired.

jsandoe: I agree. I had just misunderstood it, initially.

PR #325, opened 2020-03-16
Implement column resize in case card [#158851772]

kswenson:
I started out trying to use an existing jQuery-based library but found myself having to override a number of things to get it to work in our situation and eventually abandoned it in favor of writing my own React component to accomplish the task, which in the end I'm much happier with.

@bfinzer I've added some comments in the PR to assist the review, but it's still a fair amount of code to review. Let me know if you'd like to review it synchronously.

Testable at https://codap-dev.concord.org/branch/158851772-case-card-column-resize/.

### DIFF
--- a/apps/dg/components/case_card/case_card_model.js
+++ b/apps/dg/components/case_card/case_card_model.js
@@ -25,4 +25,11 @@ sc_require('components/case_table/data_context_base_model');
  @extends DG.DataContextBaseModel
  */
 
-DG.CaseCardModel = DG.DataContextBaseModel.extend();
+DG.CaseCardModel = DG.DataContextBaseModel.extend({
+    /**
+     * Percentage width of attribute name column
+     * 
+     * @property {number}
+     */
+    columnWidthPct: null
+});
--- a/apps/dg/components/case_card/column_resize_handle.js
+++ b/apps/dg/components/case_card/column_resize_handle.js
@@ -1,0 +1,146 @@
+/* global React */
+
+DG.React.ready(function () {
+  var div = React.DOM.div;
+
+  /**
+   * Provides a vertical column-resizing control handle.
+   * Loosely patterned after https://github.com/MonsantoCo/column-resizer.
+   */
+  DG.React.Components.ColumnResizeHandle = DG.React.createComponent(
+    (function () {
+
+      /**
+       * props are
+       *  className: {string} - className attribute applied to the dom element
+       *  enabled: {boolean} - enable/disable resizing
+       *  containerWidth: {number} - width of table in pixels
+       *  columnWidth: {number} - width of column to resize in pixels
+       *  minWidth: {number} - minimum width of column in pixels
+       *  maxWidth: {number} - maximum width of column in pixels
+       *  onResize(width: number): {function} - Called with updated value of width property
+       */
+      var resizeTouchId = null,
+          deltaStart = null;
+
+      function stop(e) {
+        e.preventDefault();
+        e.stopPropagation();
+      }
+
+      return {
+
+        getInitialState: function() {
+          return {
+            isResizing: false,
+            resizeWidth: null
+          };
+        },
+
+        componentDidMount: function() {
+          this.componentDidRender();
+        },
+
+        componentDidUpdate: function() {
+          this.componentDidRender();
+        },
+
+        componentDidRender: function() {
+          if (this.state.resizeWidth !== this.props.columnWidth)
+            this.setState({ resizeWidth: this.props.columnWidth });
+        },
+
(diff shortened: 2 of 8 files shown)


apps/dg/components/case_card/column_resize_handle.js
```diff
@@ -0,0 +1,146 @@
...
+
+        endResize: function() {
+          this.setState({ isResizing: false });
+        },
+
+        handleMouseDown: function(e) {
+          if (!this.props.enabled || (e.button !== 0)) return;
+          this.beginResize(e);
+          document.addEventListener('mousemove', this.handleMouseMove, true);
+          document.addEventListener('mouseup', this.handleMouseUp, true);
```
kswenson: With both mouse and touch events, we install the handlers we need precisely when we need them. This contrasts with the SproutCore/SC.RootResponder approach of installing event handlers centrally and having all events captured by a single entity that then must keep track of a lot of state to determine how/where to dispatch any given event.

apps/dg/components/case_card/column_resize_handle.js
```diff
@@ -0,0 +1,146 @@
...
+              return;
+            }
+          }
+          // resize touch is no longer active
+          this.endResize();
+
+          document.removeEventListener('touchstart', this.handleSubsequentTouch, true);
+          document.removeEventListener('touchmove', this.handleSubsequentTouch, true);
+          document.removeEventListener('touchcancel', this.handleSubsequentTouch, true);
+          document.removeEventListener('touchend', this.handleSubsequentTouch, true);
```
kswenson: Note that we don't `stop()` the subsequent touch events, because it's likely that they're destined for other valid handlers. This allows multiple touches to be active simultaneously. For a good time, try resizing a case card component while actively resizing the column on a touch device. 😉

apps/dg/controllers/document_controller.js
```diff
@@ -1295,26 +1279,27 @@ DG.DocumentController = SC.Object.extend(
               dataContext: iContext
             }),
             tTitle = iTitle || (iContext ? iContext.get('name') : 'DG.DocumentController.caseCardTitle'.loc()),
-            tModel, tRestoredLayout = iLayout;
-        if (!iComponent) {
-          tModel = this.configureComponent(iComponent, {
+            tComponent = iComponent, tRestoredLayout = iLayout;
```
kswenson: I did a little bit of refactoring here just to make the code clearer to me. There was inconsistent use of `iComponent'/`tComponent` and the variable named `tModel` actually held a Component[Model] but the name led me to think it was a `ContentModel`.

apps/dg/resources/case_card.css
```diff
@@ -253,3 +253,23 @@ tr:nth-child(even) td.attr-cell {
...
+        position: absolute;
+        left: 0;
+        width: 50%;
+        height: 100%;
+
+        &.is-resizing {
+            border-right: 1px dotted black;
+        }
+    }
+}
```
kswenson: We never got in the habit of making use of the fact that SproutCore runs all of the CSS files it processes through SASS, so we can use variables and nesting. Now that I've used SASS on other projects I'm glad I don't have to live without it here.

apps/dg/components/case_card/case_card_model.js
```diff
@@ -25,4 +25,11 @@ sc_require('components/case_table/data_context_base_model');
...
  */
 
-DG.CaseCardModel = DG.DataContextBaseModel.extend();
+DG.CaseCardModel = DG.DataContextBaseModel.extend({
+    /**
+     * Percentage width of attribute name column
+     * 
+     * @property {number}
+     */
+    columnWidthPct: null
```
kswenson: This is the ultimate source of truth. It is passed down to the React components that need it as props and callbacks are provided for setting it when resizing occurs.

apps/dg/components/case_card/column_resize_handle.js
```diff
@@ -0,0 +1,146 @@
...
+
+        render: function() {
+          var style = this.state.resizeWidth != null
+                        ? { left: this.state.resizeWidth }
+                        : null;
+          return (this.props.enabled || this.state.isResizing) &&
+            div({
+              className: this.props.className + ' column-resize-handle dg-wants-mouse dg-wants-touch',
+              onMouseDown: this.handleMouseDown,
+              onTouchStart: this.handleInitialTouch,
```
kswenson: The initial `mousedown` and `touchstart` handlers are installed as React handlers, which means that React is responsible for cleaning them up when the component is unmounted. The subsequent handlers are installed directly on the window, so we're responsible for cleaning them up.

PR #335, opened 2020-03-22
Case card should allow editing collection name [#171912998]

kswenson:
Testable at https://codap-dev.concord.org/branch/171912998-case-card-edit-label/.

Note that this doesn't work on touch devices, but rather than hold up this PR I propose fixing that when implementing touch tooltips which will require a refactor of the touch code.


kswenson:
Added I-beam cursor for collection label.

PR #337, opened 2020-03-25
Fix formula handling of white-space-only strings

kswenson:
[[#171908377]](https://www.pivotaltracker.com/story/show/171908377) Formula hours and minutes are not calculated correctly

Unexpectedly, the root of the problem was the handling of white-space-only strings (e.g. `" "`). The bug actually had nothing to do with dates per se, but was related to the appending of `" "` to a date string. Under the hood in JavaScript, white-space-only strings default-convert to `0`, which caused the preceding date-string to default-convert to the numeric interpretation of the date so that the two could be added numerically rather than appending as a string. The fix is to add special-case handling of these white-space-only strings to our conversion code when performing operations on them.

When tweaking the formula engine at this level unit tests are important to making sure that the changes don't break existing functionality. In the process of working on this I discovered several unexpected things about our unit tests.
1. When running the unit tests in the browser there was a broken test in the DG.Component tests. I never figured out the root cause of this failure as none of the relevant code had changed recently. I ended up just commenting the test out.
1. When running the unit tests from the command line an exception is thrown from the test-runner which prevents the tests from running at all. After a bit of investigation I could not determine the root cause of this failure either or when it might have begun to fail and ended up just running my tests from the browser.
1. One of the primary SproutCore-provided functions for running unit tests is the `equals()` function which tests whether two values are equal and passes/fails the unit test on the basis of the result. I was shocked to discover that the default behavior of this function is to use lax equality (`==`) rather than strict equality (`===`) for its value comparisons. Switching to strict equality immediately revealed about a half-dozen tests which were only passing due to this lax equality testing. In this case, I decided to change the implementation of the `equals()` function in the SproutCore submodule ([#4](https://github.com/concord-consortium/codap-sproutcore/pull/4)) and fix the corresponding tests as I don't think it ever makes sense to default to the more lax equality comparison.
1. In some cases the right way to handle these white-space-only strings is to throw an exception. I discovered that SproutCore has a `should_throw()` primitive among its unit-testing functions which we had never made use of before. I added it to our `JSHint` and `ESLint` configurations and used it to test the behavior here.

@jsandoe In the interest of avoiding the double-PR dance, I went ahead and merged the SproutCore PR so that this PR now includes the appropriate submodule reference.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -110,6 +110,7 @@
     "module": false,
     "ok": false,
     "same": false,
+    "should_throw": false,
     "start": false,
     "stop": false,
     "test": false,
--- a/apps/dg/formula/formula.js
+++ b/apps/dg/formula/formula.js
@@ -277,6 +277,9 @@ DG.Formula.functionRegExp = (function() {
   return new RegExp('[%@][%@]*(?=\\()'.fmt(firstChar, otherChars));
 }());
 
+// matches non-empty strings consisting entirely of white space characters
+DG.Formula.whiteSpaceRegExp = /^[\s\uFEFF\xA0]+$/;
+
 /**
   Addition function which handles types by our rules rather than JavaScript's.
   Numbers and values interpretable as numeric (e.g. booleans, some strings)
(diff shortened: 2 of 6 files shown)


kswenson:
I thought perhaps that PhantomJS might be the limitation. Getting unit tests working from the command-line using some other mechanism would be a good task moving forward.

PR #344, opened 2020-04-03
Add touch-tooltips for case card attributes [#171913001]

kswenson:
Addresses [[#171913001](https://www.pivotaltracker.com/story/show/171913001)]

Also changes the default position for touch-tooltips to be above the target element rather than below it. Below the target element frequently means underneath the finger, so above is generally more useful in my experience.

Testable at https://codap-dev.concord.org/branch/171913001-case-card-touch-tooltips/.

### DIFF
--- a/apps/dg/components/case_card/case_card.js
+++ b/apps/dg/components/case_card/case_card.js
@@ -551,7 +551,7 @@ DG.React.ready(function () {
                     isEditing: tAttrID === this.state.attrIdOfValueToEdit,
                     onToggleEditing: toggleEditing,
                     onEscapeEditing: escapeEditing,
-                    editModeCallback: editModeCallback
+                    onEditModeCallback: editModeCallback
                   }
                 });
 
--- a/apps/dg/components/case_card/case_card.js
+++ b/apps/dg/components/case_card/case_card.js
@@ -22,6 +22,14 @@ DG.React.ready(function () {
             // leave some room for collection name (left) and navigation buttons (right)
             kMinColumnWidth = 82;
 
+        var kThresholdDistance2 = 0, // pixels^2
+            tTouchID,
+            tTouchTimer,
+            tPointerIsDown = false,
+            tStartCoordinates,
+            tDragInProgress = false,
+            tDragHandler;
+
         var ChangeListener = SC.Object.extend({
           dependent: null,
           context: null,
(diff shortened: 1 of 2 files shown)


apps/dg/components/case_card/case_card.js
```diff
@@ -22,6 +22,14 @@ DG.React.ready(function () {
             // leave some room for collection name (left) and navigation buttons (right)
             kMinColumnWidth = 82;
 
+        var kThresholdDistance2 = 0, // pixels^2
+            tTouchID,
+            tTouchTimer,
+            tPointerIsDown = false,
+            tStartCoordinates,
+            tDragInProgress = false,
+            tDragHandler;
```
kswenson, line 31: These variables were originally defined inside the `renderAttribute()` function along with the event handling functions that use them. This meant that 1) each attribute had its own copy of them and 2) every time a re-render occurred a new set of them was generated along with a new set of event handler functions. Moving them to the scope of the case card function closure makes them effectively global.

apps/dg/components/case_card/case_card.js
```diff
@@ -509,7 +551,7 @@ DG.React.ready(function () {
                     isEditing: tAttrID === this.state.attrIdOfValueToEdit,
                     onToggleEditing: toggleEditing,
                     onEscapeEditing: escapeEditing,
-                    editModeCallback: editModeCallback
+                    onEditModeCallback: editModeCallback
```
kswenson, line 554: This typo should mean there's a bug in editing case values, unless this doesn't matter for some reason. 🤔

apps/dg/utilities/touch_tooltips.js
```diff
@@ -41,8 +41,9 @@ DG.TouchTooltips = (function() {
                     ? triggerView.get('layer')
                     : triggerView,
           container = DG.mainPage.getPath('mainPane.layer'),
-          _placement = placement || 'bottom-start',
+          _placement = placement || 'top-start',
```
kswenson, line 44: Changing the default placement to above the target element so the tooltip doesn't appear underneath the finger.

apps/dg/components/case_card/case_card.js
```diff
@@ -257,29 +267,61 @@ DG.React.ready(function () {
...
-            function handleMouseUp() {
-              tMouseIsDown = false;
+            function handlePointerUp() {
+              tTouchID = null;
+              tTouchTimer = null;
+              tPointerIsDown = false;
               tStartCoordinates = null;
               tDragInProgress = false;
-              if (tDragHandler)
-                tDragHandler.handleEndDrag();
```
kswenson, line 269: It turns out the drag handler doesn't _have_ a `handleEndDrag()` function. 😉

PR #357, opened 2020-05-29
WIP: Improved data interactive API (plugin) support

kswenson:
This WIP PR includes untested code I was in the middle of when I was called away from CODAP work. I'm submitting it now with the hope that it will make enough sense for someone else to take it the rest of the way.


kswenson:
This functionality was apparently cherry-picked and committed separately, so this PR is no longer needed.

PR #363, opened 2020-07-23
Update to slate-editor 0.5.0 (fix-SC)

kswenson:
- adds improved image dialog with ability to adjust size, placement, etc.
- requires SproutCore change to fix `Array.prototype.find()`

Note that this PR depends on [codap-sproutcore PR#6](https://github.com/concord-consortium/codap-sproutcore/pull/6). This PR should not be merged until either that PR is merged or some other solution is implemented to fix the `Array.prototype.find()` incompatibility described there. Without such a fix, the slate-editor library crashes when editing an existing image. Toward that end, this PR is currently pointing at the PR branch of the `codap-sproutcore` repository for testing purposes. If that PR is merged, this PR should be updated to point to the merged production branch, i.e. `dg-sc-1.11`. Testable at https://codap-dev.concord.org/branch/rte-0.5.0/.

### DIFF
--- a/apps/dg/components/text/text_view.js
+++ b/apps/dg/components/text/text_view.js
@@ -168,9 +168,9 @@ DG.TextView = SC.View.extend((function() {
         renderToolbar: function() {
           var props = {
                 modalPortalRoot: this.reactModalDiv,
-                modalCoverClassName: 'codap-slate-cover' + kWantsEventsClasses,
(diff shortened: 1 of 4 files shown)


jsandoe:
This PR and PR #365 address the same issue and are different solutions to the same problem. After discussion with @kswenson and @bfinzer, it has been decided that the solution in #365 is preferable. Will close, but not merge this PR and close and merge #365.

PR #371, opened 2020-12-29
WIP: Bar chart formula

kswenson:
Turning this over as a work-in-progress per prior agreement.

What's Done:
- New `ComputedBarChartModel` and `ComputedBarChartView` classes are siblings to the preexisting `BarChartModel` and `BarChartView` classes and both derive from new `BarChartBaseModel` and `BarChartBaseView` classes respectively.
- `ComputedBarChartModel` maintains a formula and computes its value across each primary cell.
- `BarChartBaseModel` and `BarChartBaseView` define new `getBarHeight()` virtual method that is overridden by the derived classes to determine the bar height appropriately.
- `BarChartBaseModel` defines a new property, `isBarHeightComputed`, which is true for `ComputedBarChartModel` and false for the original `BarChartModel`.
  - I originally thought that I could add a third option for the `breakdownType` property, but it turns out to be difficult to have a property that is sometimes a simple property (e.g. `count` vs. `percent`) and sometimes a plot-type determining property (e.g. `formula`) as the two are handled completely differently from an undo/redo perspective. Changes to the new property have a one-to-one correspondence with plot type changes.
- New `BarChartFormulaContext` class for use with bar chart formulas.
  - I had originally thought that I might be able to reuse the `PlottedValueFormulaContext` but that turned out not to be the case.
- Inspector UI for switching between computed (`formula`) and simple (`count` or `percent`) bar charts.

What's To Be Done:
- Inspector UI for entering/editing the formula.
  - Formula currently defaults to `mean(num)`
- Formula is currently created on first use rather than being created/updated when plot configuration changes or user enters/edits expression.
- Formula results are computed on demand rather than being cached.
- Axis boundaries are currently hard-coded rather than being determined from computed results.
- As we saw in the demo, computed bars seem to be rendered invisibly in addition to the regular visible bars.
- Verify that some changes made in service of computed bar charts don't adversely affect other plots. If necessary, there may be other ways to achieve the desired results.
  - In `GraphModel.swapPlotForNewPlot()`, calls to `synchAxes()` and `rescaleAxesFromData()` were added.
  - In `ChartModel._buildCache()`, `iSecondaryCell` is defaulted to `0` so that cell counts are computed for the computed bar chart.

### DIFF
--- a/apps/dg/components/graph/graph_model.js
+++ b/apps/dg/components/graph/graph_model.js
@@ -975,7 +975,10 @@ DG.GraphModel = DG.DataLayerModel.extend(
     addPlotObserver: function( iPlot) {
       switch( iPlot.constructor) {
         case DG.BarChartModel:
-          iPlot.addObserver('breakdownType', this, this.propagateBreakdownType);
+        case DG.ComputedBarChartModel:
+          iPlot.addObserver('isBarHeightComputed', this, this.swapComputedBarChartType);
+          if (iPlot.constructor === DG.BarChartModel)
+            iPlot.addObserver('breakdownType', this, this.propagateBreakdownType);
           // fallthrough intentional
         case DG.DotChartModel:
           iPlot.addObserver('displayAsBarChart', this, this.swapChartType);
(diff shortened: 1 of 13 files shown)


kswenson:
Bill finished this work outside of this PR.

PR #372, opened 2021-01-26
Enable row-resizing by dragging from bottom edge of index column cells

kswenson:
[[#174775393]](https://www.pivotaltracker.com/story/show/174775393)

@bfinzer Consider yourself an optional reviewer.

@bfinzer @jsandoe The bug we saw during the demo with resizing rows near the bottom of the window has been fixed.

### DIFF
--- a/apps/dg/resources/dg.css
+++ b/apps/dg/resources/dg.css
@@ -1280,16 +1280,26 @@ body.dg .sc-view.dg-doc-background {
     width: 100px;
 }
 
-/* used by DG.CaseTableCellEditor */
-input.dg-editor-text {
-    width: 100%;
-    height: 100%;
-    border: 0;
-    margin: 0;
-    background: transparent;
-    outline: 0;
-    padding: 0;
+.slick-row-resizer {
+    position: absolute;
+    width: 35px;    // width of index column
+    cursor: row-resize;
+    z-index: 1;
+}
 
+/* used by DG.CaseTableCellEditor */
+input, textarea {
+    &.dg-editor-text {
+        width: 100%;
+        height: 100%;
+        border: 0;
+        margin: 0;
+        background: transparent;
+        outline: 0;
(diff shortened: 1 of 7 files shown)


apps/dg/resources/dg.css
```diff
@@ -1525,6 +1541,14 @@ html.dg-embedded-mode .dg-data-tip-text {
     color: black;
 }
 
+.slick-multiline-cells .slick-cell {
```
jsandoe, line 1544: Curious why -webkit-line-clamp required dynamic setting, since it is easier set in code. Also, I cannot find `display: -webkit-box` documented anywhere. I see that it is comes from the article, but is it on a standards track and is it likely to persist with the same meaning in browsers?

jsandoe (commented):
Looks good. Just have the one question about the CSS.

apps/dg/resources/dg.css
kswenson: For `-webkit-line-clamp`, its value is a numeric value indicating the number of lines to clamp to, which varies with row height and so can't be set statically in CSS. As described in the article, it's not apparently on any standards track at this time, but it has been adopted by all the major browsers. Without it, there basically isn't a clean solution to the problem of putting the ellipsis where it belongs.

jsandoe (approved):
OK, thanks.

apps/dg/resources/dg.css
kswenson: One other note: I used that particular set of CSS because it's what was described in the article, including the comment that "Firefox supports this now, exactly this way (with the -webkit- prefixes and all)." It's conceivable that the line-clamping would work with `display: flex`, for instance (presumably, that's the way it will work eventually), but I didn't try it.

jsandoe: Yes, I saw that in the article. Thanks.

PR #386, opened 2021-05-19
Data entry row implementation tweaks

kswenson:
- fix data entry in first row of table
- fix connecting lines after deleting cases
- allow drag/drop across group boundary
- disable row-resize from data entry row


jsandoe (approved):
As you said, pretty straightforward fixes. 👍

kswenson:
@jsandoe Dan reported one more issue, so expect one additional commit.

apps/dg/components/case_table/case_table_view.js
```diff
@@ -584,8 +584,8 @@ DG.CaseTableView = SC.View.extend( (function() // closure
           this.invokeNext(function() {
             gridDataView.expandGroup(parentCaseID);
             gridAdapter.refresh();
-          });
-        });
+          }.bind(this));
```
jsandoe, line 587: Is this interior bind necessary? The function references are all in the closure.

kswenson: Not strictly, but I went with consistency. It looked curious with only one `bind()`.

jsandoe: OK, that's fine. I was just wondering whether I missed something.

jsandoe (approved):
Looks good 👍

PR #390, opened 2021-06-22
Modifications for transformations plugin

stardust66:
### Changes
- Add `evalExpression` resource, which evaluates a given codap language expression. The request values must include a string `source` field, which is the expression to be evaluated, and a `records` field, which is an array of data items, serving as the environment in which the expression is evaluated. The expression is evaluated once for each record in the array. Response values is an array of results.
- Add `functionNames` resource, which supports the `get` action. Returns an array of function names in the expression language. This enables autocompleting codap functions in our expression editors. 
- Add `create` action to `interactiveFrame` resource, which spawns a new plugin window with the given name and url.
- Add notification for context deletion that indicates the name of the context being deleted.

Please let us know your feedback.

Paul, Thomas, and Jason

### DIFF
--- a/apps/dg/components/data_interactive/data_interactive_phone_handler.js
+++ b/apps/dg/components/data_interactive/data_interactive_phone_handler.js
@@ -2364,7 +2364,7 @@ DG.DataInteractivePhoneHandler = SC.Object.extend(
         get: function() {
           return {
             success: true,
-            values: DG.functionRegistry.get("namesArray"),
+            values: DG.functionRegistry.get("categorizedFunctionInfo"),
           };
         },
         notify: function (_iResources, iValues) {
--- a/apps/dg/components/data_interactive/data_interactive_phone_handler.js
+++ b/apps/dg/components/data_interactive/data_interactive_phone_handler.js
@@ -509,13 +509,31 @@ DG.DataInteractivePhoneHandler = SC.Object.extend(
 
       /**
        *     {{
-       *      action: 'update'|'get'|'delete'
+       *      action: 'create'|'update'|'get'|'delete'
        *      what: {type: 'interactiveFrame'}
        *      values: { {title: {String}, version: {String}, dimensions: { width: {Number}, height: {Number} }}}
        *     }}
        * @return {object} Object should include status and values.
        */
       handleInteractiveFrame: {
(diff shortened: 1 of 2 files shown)


jsandoe:
Hi Paul, Thomas, and Jason,

I have the following overall comments and suggestions. 
* On the changes to the API, in general, I think the capabilities you wish to achieve are appropriate, I would just like to see if we can make the API more generalized and a bit more in line with existing API.
* The create/interactiveFrame addition, though, I don't think is necessary. The ability to create a new Plugin already exists (though, is poorly documented). Try this:
```
{
  "action": "create",
  "resource": "component",
  "values": {
    "name": "x",
    "URL": "https://concord-consortium.github.io/codap-data-interactives//NOAA-weather/index.html",
    "type": "game"
  }
}
```
* Concerning the `functionNames` resource, I was curious that you only need the names. I am assuming you are using this to compose expressions. Would you be able to make use of the entire definition if made available? I mean a structure similar to this (although proper json): https://github.com/concord-consortium/codap/blob/master/apps/dg/resources/json/function_strings.json
* Concerning `evalExpression`, I have a naming quibble and a suggestion.
    * The naming quibble: generally the API has identified resources as nouns. This name has a procedural flavor. 
    * Its not that we do not have procedural calls in the Plugin API. We have couched them as requests on named objects using the "notify" action. Here is my suggestion: The object that the plugin is wanting to reference is the FormulaEngine, so I suggest that that be the resource name instead of "evalExpression". The request for evaluation could be phrased like this:

```
{
  "action": "notify",
  "resource": "FormulaEngine",
  "values": {
    "request": "evalExpression",
    "records": ...
    "source": ...
  }
}
```

*
    * This would be better in line with existing usage. Also, it occurs to me that the `functionNames` resource also refers to the `FormulaEngine` object, perhaps these two resources could be combined. In other words, the 'get' action of the formula engine object would return the function information and the 'notify' would respond to the request to evaluate the expression.

bfinzer:
Could you also modify the [API documentation](https://github.com/concord-consortium/codap/wiki/CODAP-Data-Interactive-Plugin-API) to reflect the new capabilities?

jsandoe (approved):
Looks very good to me. 👍 
Thank you.

PR #391, opened 2021-07-08
Use new CFM version with new InteractiveApiProvider enabled

kswenson:
This PR incorporates the changes from CFM PRs [#216](https://github.com/concord-consortium/cloud-file-manager/pull/216) through [#222](https://github.com/concord-consortium/cloud-file-manager/pull/222) which implement a new CFM provider which makes it possible for CODAP documents (or other embedded content up to 1MB) to be stored the same way other LARA/ActivityPlayer question answers are stored. Furthermore, with the `interactiveApi` url parameter, it can be configured to save CODAP documents as attachments that are not restricted by the Firestore 1MB document limit. This should make it easier/possible to use CODAP in LARA and the ActivityPlayer. This PR should not be merged until the corresponding CFM PRs have been merged and incorporated into an "official" CFM release.


jsandoe (approved):
Looks fine 👍

jsandoe:
@kswenson , I just reviewed this PR. Sorry I did not see that it was ready for review earlier. I did not look at any of the CFM code relating the the Interactive Interface, because I lack the context, and I am sure it has been reviewed elsewhere. 

Are there timing issues on the other side of this interface or can this PR be merged right away?

kswenson:
>Sorry I did not see that it was ready for review earlier.

No worries. It's not holding anything up that I'm aware of, which is why I didn't ping you about it earlier.

>Are there timing issues on the other side of this interface or can this PR be merged right away?

I don't think there's a technical reason to hold up the merge, i.e. there are no interdependencies that would cause something to break. It's more of a CFM-release process question. I don't know how much testing has been done with Sage/SageModeler and I don't know when/whether it will make sense to deploy the new CFM version to production. I also don't know to what extent it makes sense to try to synchronize those things with CODAP's adoption of a new CFM "release".

On the one hand the changes _should_ be limited to the new sharing features, but in fact a lot of code was converted from JavaScript to TypeScript and so the testing surface is broader than the limited new features would suggest. None of which answers your question about when to do the merge, but should provide some context in thinking about it. 😉

PR #393, opened 2021-08-02
Use square brackets when compiling variables

stardust66:
Using the dot syntax causes an issue with variable names containing spaces. The name "Attribute Name" will be compiled to `c.vars.Attribute Name`, which will error. This edit fixes the issue.


bfinzer (approved):
Looks good to me

bfinzer:
👍 Nice catch!

PR #394, opened 2021-08-03
add name to component deletion notification

thomascastleman:
This adds the `name` of a component to the notification that is sent when a component is deleted. Right now it sends the `id` and `title`, but we are using names to track which text components have been created by transformers. This makes it easy for us to listen for text component deletion.

The name seemed readily accessible and simple enough to put into the notification, but let me know if there's an issue with doing this!


jsandoe (approved):
Looks good. 👍 

Thanks!

PR #402, opened 2022-02-16
Changing the default cypress snapshot diff directory to keep tracked …

tejal-shah:
…git files cleaner


jsandoe (approved):
Looks good to me 👍

PR #406, opened 2022-05-18
Disabled CFM dirty sync during document load [#182222132]

dougmartin:
Disabled the "dirty" flag from syncing during document load to avoid showing the "UNSAVED" notice in SageModeler and to avoid an automatic initial save when documents are loaded by the interactiveApi provider.


jsandoe (approved):
Makes sense to me. 👍

PR #409, opened 2022-07-15
V3: Initial v3 project folder

kswenson:
My assumption is that when CODAP `v3` development is complete we will want `codap` to be the GitHub repository used for development moving forward (e.g. towards `v4`) and `https://codap.concord.org` to be the URL for launching CODAP moving forward. In the interim, however, we want `v2` and `v3` development to be kept separate.

To achieve these goals, `v3` is developed as a folder within the existing `codap` GitHub repository. GitHub Actions triggers are used to have pushes to `v2` files result in running the `CODAP v2` workflow (e.g. deploy to `https://codap-dev.concord.org`), while pushes to `v3` files trigger the `CODAP v3` workflow, which deploys to `https://codap3.concord.org`.

When we are ready to retire `v2` development, we can retire the `v2` workflow and adjust the `v3` workflow to be the primary one.

@scytacki I'm mostly interested in your thoughts on the GitHub Actions configuration.


scytacki (commented):
I should have responded to your question on the starter project about this earlier. But yes I do think using the newer deploy process (non s3_website) is better in general.

However its approach to moving a single file to the top level would not work with the SproutCore build of codap 2. Because this requires dynamic paths to resources loaded by the javascript. So probably codap 2 has to stay as is. 

But perhaps codap 3 could use it.

What makes it not work well is if there are multiple "entry points" to run the app. Ie if there are multiple html files created by each build. I think CODAP has only one html file so from that point of view using it should be straight forward.

As far as the github current workflows in this file, they seem fine to me. @tejal-shah has more experience setting up multiple workflows that use paths and branches to decide which workflow to run. I'd be curious if she sees any issues with the paths and paths-ignore of these two files.

bfinzer (approved):
👍 Approved with a few notes: license date, semicolons, cypress 10, etc. as discussed during synchronous review.

tejal-shah:
@scytacki , the `paths-ignore` and `paths` properties in `ci.yml` and `v3.yml` look okay to me.

kswenson:
> I should have responded to your question on the starter project about this earlier. But yes I do think using the newer deploy process (non s3_website) is better in general.

Given the need to migrate the `v2` build to GitHub Actions and support the new `v3` build as well it seemed simpler to me to stick with the tried-and-true. Now that I've got it working I can look into switching `v3` to the newer build model when the next opportunity for project configuration arises.

PR #413, opened 2022-07-21
[#182715129] Feature: Simple CSV files can be dragged and dropped int…

bfinzer:
…o browser window resulting in imported data

* Added DropHandler component
* A few tweaks to other files and lint cjs


kswenson (changes requested):
👍 Code looks fine -- for the CI build to complete you'll need to:
- rename the other `.eslintrc` files to have `.cjs` extensions
- change the reference to the base `.eslintrc` file in the other `.eslintrc` files to reference the new file name in the `extends` property, e.g. `.js` => `.cjs`
- eliminate or comment out the `console.log`s

kswenson:
I went ahead and fixed the remaining ESLint issues and the cypress tests. The build now completes with warnings related to the `console.log`s. I'll let you decide whether to deal with those now or later and whether to do so by deleting them or commenting them out (in case you think they're worth hanging on to).

kswenson (approved):
👍 Changes look good

PR #415, opened 2022-07-22
Import .csv data into DataBroker

kswenson:
Adds a `DataBroker` which can manage multiple `DataSet`s, although initially it is configured to support only a single one. The .csv import code is updated to import the data into a `DataSet` and add it to the `DataBroker`. The code for displaying the result of the import is separated into its own component which re-renders via MobX mechanisms when the `DataBroker` changes.

### DIFF
--- a/v3/src/data-model/data-broker.ts
+++ b/v3/src/data-model/data-broker.ts
@@ -62,4 +62,5 @@ export class DataBroker {
   }
 }
 
-export const gDataBroker = new DataBroker()
+// for MVP we only support a single DataSet
+export const gDataBroker = new DataBroker({ allowMultiple: false })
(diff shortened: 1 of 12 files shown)


bfinzer (approved):
👍 Looks good and I appreciate the many tests!

v3/src/data-model/data-broker.ts
```diff
@@ -0,0 +1,66 @@
...
+  attributes: number;
+  cases: number;
+}
+
+interface IDataBrokerOptions {
+  allowMultiple?: boolean
+}
+export class DataBroker {
+  @observable dataSets = new Map<string, IDataSet>()
+  @observable readonly allowMultiple: boolean
```
bfinzer, line 16: Doesn't make sense to have readonly prop be @observable

PR #417, opened 2022-07-26
V3: Initial case table implementation

kswenson:


v3/src/hooks/use-max-attr-value-widths.ts
```diff
@@ -0,0 +1,35 @@
+import { useEffect, useState } from "react"
+import { IDataSet } from "../data-model/data-set"
+import { measureText } from "./use-measure-text"
+
+interface IWidthEntry {
+  caseId: string
```
bfinzer, line 6: What does caseId have to do with maximum width of value string? Seem unnecessary to me.

bfinzer (changes requested):
A few quibbles.

v3/src/components/case-table/case-table.tsx
```diff
@@ -0,0 +1,52 @@
...
+    return <span>{value}</span>
+  }, [data])
+
+  // column definitions
+  const columns: TColumn[] = useMemo(() => data
+    ? data.attributes.map(({ id, name }) => ({
+        key: id,
+        name,
+        // auto-size each column to fit its maximum string
+        width: Math.max(kDefaultColumnWidth, (maxWidths[id]?.width || 0) + kCellPadding),
```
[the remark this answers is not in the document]
kswenson: Agreed, but until we have user column resizing, not being able to see your data _and_ not being able to do anything about it is too frustrating so I'll leave it for now and we can remove it once we support column resizing.

v3/src/hooks/use-max-attr-value-widths.ts
kswenson: I was thinking ahead to how we'll deal with change notifications. If you don't know the case that contained the longest string, then a change to any value requires re-assessing all values. With it, if the value of that case wasn't changed, then you only have to compare any changed values to the existing max value.

PR #418, opened 2022-07-27
Enable column resizing by user; disable initial auto-sizing

kswenson:
Turns out user column-resizing is supported out of the box by react-data-grid.


bfinzer (approved):
:+1 Looks good. Glad resizing is built-in!

PR #419, opened 2022-07-27
Initial commit of graph. Needs model layer and tests

bfinzer:
* The graph appears in quadrant 4, initially empty.
* On data import, finds first two numeric attributes and plots them as a scatterplot.
* Button allows transition to dot plot
* Points can be selected individually or with marquee
* Selected points drag and return to original place on release
* Movable line in scatterplot shows equation
* Movable value in dot plot shows value
* Axes can be dragged in the same manner as in V2


v3/src/components/graph/axis.tsx
```diff
@@ -0,0 +1,173 @@
...
+      const theAxis = select(axisRef.current)
+        .attr("transform", props.axisProps.transform)
+        // @ts-expect-error null => SVGSVGElement
+        .call(axis(scale))
+
+      // Add three rects in which the user can drag to dilate or translate the scale
+      const tLength = props.axisProps.length !== undefined ? props.axisProps.length : 0,
+        classPrefix = orientation === 'bottom' ? 'h' : 'v',
+        numbering = orientation === 'bottom' ? [0, 1, 2] : [2, 1, 0],
+        classPostfixes = orientation === 'bottom' ? ['lower-dilate', 'translate', 'upper-dilate'] :
```
kswenson, line 96: When there's an axis model, it should be able to some of these distinctions internal to the model.

v3/src/components/graph/graph.scss
```diff
@@ -0,0 +1,149 @@
...
+  font: 10px sans-serif;
+}
+
+.dotArea {
+  background: white;
+}
+
+.background {
+  fill: white;
+}
```
kswenson, line 28: Generally, .scss should be as close to its code as possible, which generally means associated with particular components where possible, and often nested within each other which minimizes the chances of global conflicts. That way, if the case table also wants to use `.background` as a class name the two definitions won't conflict with each other.

v3/src/components/app.tsx
```diff
@@ -1,21 +1,22 @@
...
-import { useDropHandler } from "../hooks/use-drop-handler"
-import { useSampleText } from "../hooks/use-sample-text"
+import {Container} from "./container"
+import {DataSummary} from "./data-summary"
+import {gDataBroker} from "../data-model/data-broker"
+import {DataSet, toCanonical} from "../data-model/data-set"
+import {Text} from "./text"
+import {useDropHandler} from "../hooks/use-drop-handler"
+import {useSampleText} from "../hooks/use-sample-text"
 import Icon from "../assets/concord.png"
```
kswenson: I'm guessing the IntelliJ auto-formatter is removing the spaces before/after the braces. I'd recomment turning that off as I don't think we want to mandate one way or the other and automatic white-space only changes are annoying on review.

v3/src/components/app.tsx
```diff
@@ -1,21 +1,22 @@
...
-import { useDropHandler } from "../hooks/use-drop-handler"
-import { useSampleText } from "../hooks/use-sample-text"
+import {Container} from "./container"
+import {DataSummary} from "./data-summary"
+import {gDataBroker} from "../data-model/data-broker"
+import {DataSet, toCanonical} from "../data-model/data-set"
+import {Graph} from "./graph/graph"
+import {Text} from "./text"
+import {useDropHandler} from "../hooks/use-drop-handler"
+import {useSampleText} from "../hooks/use-sample-text"
```
kswenson, line 10: I'm guessing that the IntelliJ auto-formatter removed the spaces before/after the braces here. I'd recommend turning that off, as I don't think that's something we want to mandate.

kswenson (approved):
👍 This seems like a reasonable starting point. Where I had minor line-level suggestions I just went ahead and made the changes in an additional commit. As already discussed, we're going to want to introduce models and refactor this code in ways that I don't yet see clearly enough to add any useful comments. There are a few other comments/questions/suggestions but they can call be addressed in subsequent PRs so I don't think there's anything that _has_ to be fixed before this goes in.

v3/src/components/graph/movable-value.tsx
```diff
@@ -0,0 +1,92 @@
...
+      line: null, cover: null, valueLabel: null
+    })
+  // Refresh the line
+  useEffect(function refresh() {
+      if (!valueObject.line) {
+        return
+      }
+
+      function refreshValueLabel() {
+        const screenX = x(value) + 60,
```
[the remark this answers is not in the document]
bfinzer: Actually, it shouldn't be present but derived. Definitely a ToDo

v3/src/components/graph/movable-value.tsx
```diff
@@ -0,0 +1,92 @@
...
+          .attr('x2', x(value))
+          .attr('y2', bottom)
+      })
+      refreshValueLabel()
+    }, [valueObject, transform, x, value, bottom, top, xMin, xMax, xRangeMin, xRangeMax]
+  )
+
+  const
+    dragValue = useCallback((event: MouseEvent) => {
+      setValue(x.invert(event.x - 60))
```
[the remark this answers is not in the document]
bfinzer: Actually, it shouldn't be present but derived. Definitely a ToDo

PR #421, opened 2022-07-29
V3: Implement row selection [#181846512]

kswenson:
PT: [[#181846512]](https://www.pivotaltracker.com/story/show/181846512)
Demo: https://codap3.concord.org/branch/v3-row-selection/

Case selection is supported in the `DataSet` model.
Clicking in `index` column of case table selects corresponding row.
Clicking with any modifier key held down extends the selection (all modifier keys use discontiguous selection at this point).

### DIFF
--- a/v3/package-lock.json
+++ b/v3/package-lock.json
@@ -36,6 +36,7 @@
         "@testing-library/jest-dom": "^5.16.4",
         "@testing-library/react": "^12.1.5",
         "@testing-library/react-hooks": "^8.0.1",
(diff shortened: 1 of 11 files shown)


bfinzer (approved):
:+1 Looks good. Too bad MST doesn't have a Set. Took me awhile to understand using Map.

kswenson:
> :+1 Looks good. Too bad MST doesn't have a Set. Took me awhile to understand using Map.

We could always implement it and submit it as a PR.

PR #423, opened 2022-07-29
V3: Support type inference for attributes

kswenson:


v3/src/data-model/attribute.ts
```diff
@@ -48,6 +48,7 @@ export const Attribute = types.model("Attribute", {
   clientKey: "",
   sourceID: types.maybe(types.string),
   name: types.string,
+  userType: types.maybe(types.enumeration(["numeric", "nominal"])),
```
bfinzer, line 51: Note that the user sees categorical, numeric, date, qualitative, boundary and checkbox as allowable types. Let's not use 'nominal', only 'categorical', to eliminate ambiguity compared to V2. I would think we would want to include all possibilities in this enumeration.

bfinzer (changes requested):
See note about full set of attribute type possibilities

v3/src/data-model/attribute.ts
kswenson: I used `nominal` for consistency with CODAP2 because of this comment from `attribute_model.js`:
```
      /**
       * Type of the attribute: 'numeric' or 'nominal'
       */
      type: null,
```
but a closer look at the code suggests that both `nominal` and `categorical` are used internally so I can change it.

bfinzer (approved):
:+1 All good

PR #425, opened 2022-08-01
V3: Implement attribute drag from summary component

kswenson:
Demo: https://codap3.concord.org/branch/v3-attribute-drag/

Dragging from the case table column headers is a little bit tricky because of the interplay between ReactDataGrid and DnDKit. Therefore, this PR implements the ability to drag attributes from the summary component instead. Attributes can be dropped on the Inspector button in the summary component itself, resulting in a report on the attribute's type, or on the case table column header row which is interpreted as a means to change the order of the attributes.

- attributes can be dragged from the summary component
- attributes can be dropped on the Inspector button in the summary component
- attributes can be dropped on the case table column headers to reorder attributes
- `mammals` URL parameter preloads mammals data for easier development/debugging

@bfinzer Probably a good candidate for a sync review.

### DIFF
--- a/v3/.eslintrc.cjs
+++ b/v3/.eslintrc.cjs
@@ -83,7 +83,7 @@ module.exports = {
     "radix": "error",
     "react/jsx-closing-tag-location": "error",
     "react/jsx-handler-names": "off",
-    "react/jsx-no-useless-fragment": "error",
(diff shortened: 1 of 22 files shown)


bfinzer (approved):
:+1 Looks Elegant

PR #426, opened 2022-08-02
Update ReadMe

kswenson:
First PR against the newly created `main` branch.

Note: The top-level `ReadMe` on the `master` branch should also be updated in a different PR, but I'm not sure if I'm the right person or this is the right time.


bfinzer (approved):
:thu Looks good to me

PR #427, opened 2022-08-02
V3: Graph shows selection plus refactoring to use dataset as ultimate source

bfinzer:
* Point selection (click or marquee) selects cases in dataset
* Point selection in graph stays in synch with case selection in dataset
* Values that determine point location are derived from attribute values in dataset
* Circles are removed when a new dataset is discovered rather than left in DOM


v3/src/components/graph/dotplotdots.tsx
```diff
@@ -1,122 +1,114 @@
...
+  plotProps,
+  transitionDuration,
+  InternalizedData,
+  defaultRadius,
+  defaultDiameter,
+  dragRadius
+} from "./graphing-types"
+import {useDragHandlers, useSelection} from "./graph-hooks/graph-hooks"
+import {IDataSet} from "../../data-model/data-set"
+import {observer} from "mobx-react-lite"
```
kswenson, line 13: Move import ⬆️

v3/src/components/graph/dotplotdots.tsx
```diff
@@ -1,122 +1,114 @@
...
     } = props,
-    [dragID, setDragID] = useState(-1),
+    [dragID, setDragID] = useState(''),
     [refreshCounter, setRefreshCounter] = useState(0),
     currPos = useRef({x: 0}),
     target = useRef<any>(),
     [firstTime, setFirstTime] = useState<boolean | null>(true),
-    selectedDataObjects = useRef<{ [index: number]: { x: number } }>({}),
-    [forceRefreshCounter, setForceRefreshCounter]=useState(0)
+    selectedDataObjects = useRef<{ [index: string]: { x: number } }>({}),
```
kswenson, line 36: FYI, this could also be:
```
    selectedDataObjects = useRef<{Record<string, { x: number }>}>({}),
```

v3/src/components/graph/dotplotdots.tsx
```diff
@@ -1,122 +1,114 @@
...
         target.current.transition()
           .attr('r', dragRadius)
         setDragID(() => tItsID)
         currPos.current = {x: event.clientX}
-        setData(selectPoint(data))
+        worldDataRef.current?.selectCases([tItsID])
+        // Record the current values so we can change them during the drag and restore them when done
+        worldDataRef.current?.selection.forEach(anID => {
+          selectedDataObjects.current[anID] = {
+            x: worldDataRef.current?.getNumeric(anID, xAttrID) ?? 0
```
kswenson, line 56: It probably shouldn't ever happen, but I would think that defaulting to `NaN` would make more sense than `0`. Otherwise, it would run the risk of being plotted as data.

v3/src/components/graph/dotplotdots.tsx
```diff
@@ -1,122 +1,114 @@
...
-          selectedDataObjects.current[datum.id] = {x: datum.x}
-        }
-      })
-    }, [data, firstTime, setData, setHighlightCounter]),
+    }, [firstTime, setFirstTime, worldDataRef, graphDataRef]),
 
     onDrag = useCallback((event: MouseEvent) => {
-      if (dragID >= 0) {
+      const xAttrID = graphDataRef.current.xAttributeID
+      if (dragID !== '') {
```
kswenson, line 64: Could also just be:
```
      if (dragID) {
```

v3/src/components/graph/scatterdots.tsx
```diff
@@ -11,159 +12,135 @@ export const ScatterDots = memo(function ScatterDots(props: {
...
-                }
-              })
-              return prevData
-            }
-          )
+          worldDataRef.current?.selection.forEach(anID => {
+            const currX = worldDataRef.current?.getNumeric(anID, xAttrID),
+              currY = worldDataRef.current?.getNumeric(anID, yAttrID)
+            worldDataRef.current?.setValue(anID, xAttrID, Number(currX ?? 0) + deltaX)
+            worldDataRef.current?.setValue(anID, yAttrID, Number(currY ?? 0) + deltaY)
```
kswenson, line 70: Rather than running the risk of changing `NaN`s to numbers in the data, I'd've thought something like this would make more sense:
```
            if (isFinite(currX) && isFinite(currY)) {
              worldDataRef.current?.setValue(anID, xAttrID, Number(currX) + deltaX)
              worldDataRef.current?.setValue(anID, yAttrID, Number(currY) + deltaY)
            }
```
or even:
```
            if (isFinite(currX) && isFinite(currY)) {
              worldDataRef.current?.setCaseValues([{
                __id__: anID, [xAttrID]: currX + deltaX, [yAttrID]: currY + deltaY
              }])
            }
```
which has the advantage of being a single action at the model layer.

v3/src/components/graph/scatterdots.tsx
```diff
@@ -11,159 +12,135 @@ export const ScatterDots = memo(function ScatterDots(props: {
...
-            const sDatum = selectedDataObjects.current[datum.id]
-            if (sDatum) {
-              datum.x = sDatum.x
-              datum.y = sDatum.y
-            }
-          })
-          return prevData
+        worldDataRef.current?.selection.forEach(anID => {
+          worldDataRef.current?.setValue(anID, xAttrID, selectedDataObjects.current[anID].x)
+          worldDataRef.current?.setValue(anID, yAttrID, selectedDataObjects.current[anID].y)
```
kswenson, line 88: As described in a comment above, this could be replaced with a single call to `setCaseValues()`.

v3/src/data-model/data-set.ts
```diff
@@ -590,6 +590,14 @@ export const DataSet = types.model("DataSet", {
         })
       },
 
+      setValue(caseID: string, attributeID: string, value:IValueType) {
```
kswenson, line 593: Let's leave this for now but I've started working on the refactor we discussed which will change this (e.g. `setCaseValue()`).

kswenson (changes requested):
👍  looks good! I made a bunch of suggestions but most of them are points of interest and I leave it to your discretion whether to address them now or not. The one that really should be fixed before the PR goes in is the conditional calling of the `useGetData()` hook, which oddly isn't triggering a rules-of-hooks warning. I also found a few bugs that you may or may not want to address before the PR is merged:

- after initial switch to scatter plot, Y axis isn't draggable and label doesn't appear. Everything works after a redraw is triggered, e.g. by dragging the X axis.
- movable line isn't visible initially with mammals data (height on Y axis so scale only goes to 6.5 but movable line intercept defaults to 10).
- dots are visible below X axis when translating the Y axis (clipping issue).

v3/src/components/graph/graph-hooks/graph-hooks.ts
```diff
@@ -26,55 +29,73 @@ export const useDragHandlers = (target: any, { start, drag, end }: IDragHandlers
...
-// Todo: Any function with more than a few arguments should probably take an object instead.
-export const useGetData = (broker: DataBroker, dataRef: React.MutableRefObject<worldData[]>,
-                           xNameRef: React.MutableRefObject<string>, yNameRef: React.MutableRefObject<string>,
-                           xAxis: ScaleLinear<number, number, never>, yAxis: ScaleLinear<number, number, never>,
-                           setCounter: any) => {
+export interface IUseGetDataProps {
+  broker: DataBroker,
+  dataRef: React.MutableRefObject<InternalizedData>,
+  xNameRef: React.MutableRefObject<string>,
+  yNameRef: React.MutableRefObject<string>,
```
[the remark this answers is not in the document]
bfinzer: Let's talk about how this works.

PR #428, opened 2022-08-03
V3: Case table supports cell editing

kswenson:
Demo: https://codap3.concord.org/branch/edit-cells/

For now, uses single-click to edit cells rather than CODAP2's single-click to select the row and double-click to edit. This behavior was easier to implement and lets us test how value changes are handled.


bfinzer (approved):
:thu Looks good. I'm glad you figured out how to properly patch modules!

PR #430, opened 2022-08-05
V3: Axis models and useNumericAxisScale hook

kswenson:
For simplicity, we limit ourselves to two axis models, the base `AxisModel` and `NumericAxisModel`. For illustrative purposes, the `NumericAxisModel` supports linear or log scales. The `useNumericAxisScale()` hook returns a `D3` scale of the appropriate type and keeps it synchronized with the model via a MobX `autorun`. These are not yet used by the graph code.


bfinzer (approved):
👍🏻 Looks good

PR #431, opened 2022-08-05
V3: Case table responds to model changes; performance improvements

kswenson:
Case table re-renders in response to model changes
Case table uses default number format when rendering numbers
Graph restores model values at end of drag (although not during final animation)

Demo: https://codap3.concord.org/branch/v3-table-rows/?mammals

Note: I'll aim to write some more tests later.

>Please figure out how to restore acceptable performance and fix whatever is causing the stall.

Gauntlet thrown; challenge accepted

I spent a good deal of time analyzing and addressing the performance concerns with the following findings/results:
- Our code was unnecessarily inefficient in several places:
  - Rather than calling `setCaseValues()` ~4000 times on drag (once for each case), build a buffer of ~4000 cases and call `setCaseValues()` once.
  - D3's `end` transition callback is called for each point. Calling `setCaseValues()` ~4000 times in that context results in an n^2 explosion.
- MST model properties incur overhead (action recording, patch generation) when changed and should therefore be avoided in performance-critical paths. MST volatile properties are as efficient as any other MobX properties.
  - DataSet selection moved to volatile property. This allows it to be a proper Set (rather than an MST map). Serialized selection is now a simple array of case IDs which must be generated at serialization time.
  - DataSet supports a `caching` mode in which calls to `setCaseValues()` update a volatile cache rather than the underlying model properties. In `caching` mode, retrieval methods return values in cache when possible.
- React `setState` calls are bottlenecks in performance-critical contexts. With this PR we introduce an `appMode` property which is generally `normal` but can be elevated to `performance` when performance is critical. Clients can use this flag to adjust their behavior accordingly. The case table responds to this by eschewing the normal mechanism of setting react state and then letting RDG handle rendering for the more immediate response of simply making changes directly to the DOM. When `appMode` returns to normal, the normal `setState` calls are made and rendering control is returned to RDG. This performance mode is entered when a drag is started and exited when the drag is completed.

In figuring all of this out I was hampered by the fact that both the Chrome and React profilers were almost completely useless. Somehow the captured data never drilled down to anything useful. The largest chunks of time were generally identified as "Function Call", with no further information available. (Yes, source maps are enabled and our code shows up in the profile as expected.)

In response, I ended up building my own profiler which facilitates the measurement/recording/reporting the number of calls and time spent in various blocks of code. It requires explicitly inserting profiler calls into the code, which isn't ideal, but has the distinct advantages that it measures exactly what we want to measure, including call counts, which can be tremendously informative.

With this PR we introduce a `Start/Stop Profiling` button to the Summary Component, which makes it easy to start profiling, perform whatever operations one wants to profile (drag-select with marquee, drag points around, etc.) and then stop profiling, at which point a report is automatically logged to the console.

With the changes in this PR, the graph is now the major bottleneck in drag performance. The next step is to eliminate React from the critical path so that dragging is mediated entirely by MobX/D3.


bfinzer (changes requested):
The code looks good. 
* But with ~4000 cases, performance becomes unacceptably slow on drag in graph. I'll Slack you my test dataset, abalone.csv
* After having dragged a chunk of abalone, dragging in another csv stalls and doesn't load new data.
* Please figure out how to restore acceptable performance and fix whatever is causing the stall.

bfinzer (approved):
👍🏻 Looks good

PR #432, opened 2022-08-05
V3—Graph Improvements

bfinzer:
Various code improvements plus better positioning of axis labels and a work around for the vertical axis label bug


kswenson (approved):
👍 Changes look good -- I'm not worrying about the changes in the `axis-hook.test.tsx` because I assume that's going away (although I can't remember if I actually deleted it in my PR), but we'll have to figure out whether `min` and `max` should be `types.maybe` in `NumericAxisModel`, for instance.

PR #434, opened 2022-08-08
V3 graph folder reorganization and breakup of graph.scss

bfinzer:
* The graph folder now has folders for adornments, components, hooks, models and utilities
* graph.scss is now broken up into files for marquee, movable-line, movable-value and axis


v3/src/components/graph/graph-tests/graph.test.tsx
```diff
@@ -2,7 +2,7 @@ import { render, screen } from "@testing-library/react"
 import React from "react"
 import { DataBroker } from "../../../data-model/data-broker"
 import { DataSet, toCanonical } from "../../../data-model/data-set"
-import {Graph} from "../graph"
+import {Graph} from "../components/graph"
```
kswenson, line 5: The convention in most CC projects is to keep the tests in the same folder with the component being tested. There's no compelling reason to prefer one approach over the other. Keeping tests in the same folder means that paths are easier (e.g. the test file can import the tested component from ./component.tsx) and it means that the component and its test are neighbors in code reviews. On the other hand, some people prefer collecting the tests in one place, as you seem to be suggesting here. I'm fine with it either way, but if you do opt for the separate folder I'd name it `tests` rather than `graph-tests`, which is more analogous with the other folders.

v3/src/components/graph/graph-utils/graph_utils.ts
```diff
@@ -212,7 +212,7 @@ export function setPointCoordinates(props: IUseRefreshPointsProps) {
     dotsSvgElement = dotsRef.current,
     tTransitionDuration = firstTime ? transitionDuration : 0,
     selection = select(dotsSvgElement).selectAll('circle')
-      .classed('dot-highlighted',
+      .classed('graph-dot-highlighted',
```
kswenson, line 215: I would have expected the folder to simply be named `utilities`.

kswenson (approved):
👍 Looks good -- just one minor .scss nesting suggestion and a couple thoughts on folder names.

v3/src/components/graph/graph-tests/graph.test.tsx
bfinzer: I'm all for keeping them together. This was an oversight.

bfinzer:
Thanks for the suggestions. All are now in place, so I'll merge.

PR #437, opened 2022-08-11
V3—Graphs respond to changes in dataset

bfinzer:
* This is pretty much where we left it with incomplete implementation and nasty layout bugs


kswenson (approved):
👍 @bfinzer Rather than performing a conventional review, I took the opportunity to build on what you/we did, making a number of changes along the way. As a result, you should really review the subsequent three commits I added. Unfortunately, the author of a PR cannot subsequently review that PR, even when others have made subsequent commits. So I'm going to approve this but we should probably go over it together before it gets merged.

kswenson:
> * This is pretty much where we left it with incomplete implementation and nasty layout bugs

I took this as a starting point and continued in the same vein, fixing most of the bugs and making additional improvements:

- GraphLayout consolidates size/dimensions/scales in one place
- GraphLayoutContext makes size/dimensions/scales available through graph
- eliminate spurious react counters
- implement MobX/MST listeners for responding to model changes, etc.
- implement performance mode for scatter plot drags

Things that still need work:
- dot plot (I've spent most of my time on the scatter plot)
- movable line

PR #439, opened 2022-08-18
V3—Graph numeric scale improvements

bfinzer:
Added computeNiceNumericBounds in graph_utils to mimic V2 algorithm for choosing bounds from given data. (Also changed x and y to xScale and yScale in graph.tsx.)

### DIFF
--- a/v3/src/components/graph/hooks/graph-hooks.ts
+++ b/v3/src/components/graph/hooks/graph-hooks.ts
@@ -4,10 +4,11 @@
 import {extent} from "d3"
 import {useEffect, useMemo} from "react"
 import {IAttribute} from "../../../data-model/attribute"
-import { INumericAxisModel } from "../models/axis-model"
-import { useGraphLayoutContext } from "../models/graph-layout"
+import {INumericAxisModel} from "../models/axis-model"
+import {ScaleBaseType, useGraphLayoutContext} from "../models/graph-layout"
 import {InternalizedData} from "../graphing-types"
-import { useDataSetContext } from "../../../hooks/use-data-set-context"
+import {useDataSetContext} from "../../../hooks/use-data-set-context"
+import {computeNiceNumericBounds} from "../utilities/graph_utils"
 
 interface IDragHandlers {
   start: (event: MouseEvent) => void
@@ -33,6 +34,7 @@ export interface IUseGetDataProps {
   xAxis: INumericAxisModel
   yAxis: INumericAxisModel
 }
+
 export const useGetData = (props: IUseGetDataProps) => {
   const {xAxis, yAxis} = props,
     dataset = useDataSetContext(),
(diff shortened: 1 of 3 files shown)


v3/src/components/graph/hooks/graph-hooks.ts
```diff
@@ -41,34 +43,42 @@ export const useGetData = (props: IUseGetDataProps) => {
...
           }
         }
+      },
+
+      setNiceDomain = (values: number[], scale: ScaleBaseType, axis: INumericAxisModel) => {
+        const valueExtent = extent(values, d => d) as [number, number],
+          niceBounds = computeNiceNumericBounds(valueExtent[0], valueExtent[1])
+        scale.domain([niceBounds.min, niceBounds.max])
+        const [min, max] = scale.domain()
+        axis.setDomain(min, max)
```
kswenson, line 72: I see that you're following the pattern of the preexisting code, but it should be sufficient to set the domain on the axis model and let the synchronization to the scale happen via the normal update mechanism which would simplify the code and prove that the system is working as intended.

v3/src/components/graph/hooks/graph-hooks.ts
```diff
@@ -41,34 +43,42 @@ export const useGetData = (props: IUseGetDataProps) => {
...
+            if (xAttrId === '') {
+              xAttrId = iAttr.id
+            } else if (yAttrId === '') {
+              yAttrId = iAttr.id
+            } else {
+              break
+            }
           }
         }
+      },
```
kswenson, line 65: Seems to be some unnecessary indenting here.

v3/src/components/graph/utilities/graph_utils.ts
```diff
@@ -3,8 +3,9 @@ import React from "react"
 import {defaultRadius, Rect, rTreeRect} from "../graphing-types"
 import {between} from "./math_utils"
 import {IDataSet} from "../../../data-model/data-set"
-import { ScaleBaseType } from "../models/graph-layout"
-import { prf } from "../../../utilities/profiler"
+import {ScaleBaseType} from "../models/graph-layout"
+import {prf} from "../../../utilities/profiler"
+import {isInteger} from "lodash"
```
kswenson, line 8: Move this import to the top since it's a third-party/node_modules import.

kswenson (changes requested):
👍 Looks good -- a couple of minor suggestions. The only substantive one is to simplify `setNiceDomain()` so that it only sets the axis. The update to the D3 scale should happen automatically.

kswenson:
Actually, one additional comment -- seems like it might be a useful to move the `setNiceDomain()` to its own utility function so it's independently testable and potentially callable from elsewhere.

v3/src/components/graph/hooks/graph-hooks.ts
bfinzer: Yes, that's true!

v3/src/components/graph/hooks/graph-hooks.ts
bfinzer: I don't think so. IntelliJ's formatter doesn't either.

PR #443, opened 2022-08-20
V3—Fixed dot plot bugs

bfinzer:
* Dot plot click point doesn’t select case
* Selecting cases in dot plot with marquee works but result is not reflected by highlighted points. Similarly selection in case table isn’t reflected in dot plot.


v3/src/components/graph/components/dotplotdots.tsx
```diff
@@ -1,102 +1,207 @@
...
+    const disposer = dataset && onAction(dataset, action => {
+      if (isSelectionAction(action)) {
+        refreshPointSelection()
+      } else if (isSetCaseValuesAction(action)) {
+        // assumes that if we're caching then only selected cases are being updated
+        refreshPointPositions(dataset.isCaching)
+      }
+    }, true)
+    return () => disposer?.()
+  }, [dataset, refreshPointPositions, refreshPointSelection])
```
kswenson, line 204: Not necessary for this PR, but these three `useEffects()` are now virtually identical between the `DotPlotDots` and `ScatterDots`, which suggests that there's the potential for a reusable custom hook that could be used by both.

kswenson (approved):
👍 Looks good -- there are a number of suggestions but none of them are critical.

v3/src/components/graph/components/dotplotdots.tsx
```diff
@@ -1,102 +1,207 @@
...
         currPos.current = {x: event.clientX}
-        dataset?.selectCases([tItsID])
+
+        const [, caseId] = tItsID.split("_")
+        dataset?.selectCases([caseId])
         // Record the current values so we can change them during the drag and restore them when done
         dataset?.selection.forEach(anID => {
           selectedDataObjects.current[anID] = {
-            x: dataset?.getNumeric(anID, xAttrID) ?? NaN
+            x: dataset?.getNumeric(anID, xAttrID) ?? 0
```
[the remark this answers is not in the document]
bfinzer: It can't get into the selection without having a value. But, in spirit of knowing that whatever can go wrong will, I'm going to prevent it from getting in in the unlikely event that it doesn't have a value.

v3/src/components/graph/components/dotplotdots.tsx
bfinzer: I agree. Thinking the same myself.

PR #451, opened 2022-08-25
V3-183016745 empty case table

eireland:
Shows an empty case table on load

### DIFF
--- a/v3/src/components/app.scss
+++ b/v3/src/components/app.scss
@@ -21,6 +21,12 @@
       align-items: center;
       justify-content: center;
     }
+
+    .add-new-table-button {
+      position: absolute;
+      bottom: 10px;
+      left: 10px;
+    }
   }
 }
 
--- a/v3/src/components/app.scss
+++ b/v3/src/components/app.scss
@@ -21,12 +21,6 @@
       align-items: center;
       justify-content: center;
     }
-
-    .add-new-table-button {
-      position: absolute;
-      bottom: 10px;
-      left: 10px;
-    }
   }
 }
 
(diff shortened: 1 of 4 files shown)


v3/src/components/app.tsx
```diff
@@ -47,10 +47,22 @@
     }
   }
 
+  function createNewTable() {
```
bfinzer, line 50: Suggest createNewStarterDataset since the table appears as a side effect

v3/src/components/app.scss
```diff
@@ -21,6 +21,12 @@
       align-items: center;
       justify-content: center;
     }
+
+    .add-new-table-button {
```
bfinzer, line 25: This style isn't used, right?

bfinzer (changes requested):

bfinzer:
To help us distinguish V3 pull requests from V2 pull requests, we start the name with "V3—" and we use the magenta V3 label

v3/src/components/case-table/use-rows.ts
```diff
@@ -55,7 +55,7 @@ export const useRows = (data?: IDataSet) => {
             const strValue = data.getValue(caseId, attr.id)
             const numValue = data.getNumeric(caseId, attr.id)
             const formatStr = attr.format || kDefaultFormatStr
-            const formatted = (numValue != null) && isFinite(numValue) ? format(formatStr)(numValue) : strValue;
+            const formatted = (numValue != null) && isFinite(numValue) ? format(formatStr)(numValue) : strValue
```
kswenson, line 58: FYI, the way the semicolon got in there was that semicolons can be required depending on how the next line begins. That used to be true here, but then the code for the subsequent line was changed without removing the semicolon.

kswenson (approved):
👍 Nothing to add beyond Bill's comments

bfinzer (approved):
👍🏻 LGTM

PR #454, opened 2022-08-30
V3: 183016994 dataset switching popup

eireland:
Adds ability to switch data sets via dropdown


v3/src/components/data-summary.tsx
```diff
@@ -28,9 +30,34 @@ export const DataSummary = observer(({ broker, v2Document }: IProps) => {
...
+        <Select onChange={handleDataSetSelection} value={data?.id}>
+          { dataSetSummaries?.map(summary => {
+              return renderOption(summary.name || `DataSet ${summary.id}`, summary.id)
+            })
+          }
+        </Select>
+      )
+    } else {
+      return null
+    }
```
kswenson, line 54:
```suggestion
    return null
```
You don't need the `else` here.

kswenson (changes requested):
👍 Looks good -- just a couple suggestions

bfinzer (approved):
👍🏻 LGTM

kswenson (approved):
👍 Changes look good

PR #455, opened 2022-08-31
V3: 181503563 component titles

eireland:
Adds an editable component title to table and graph. Note that the title is not saved.


v3/src/components/codap-component.tsx
```diff
@@ -1,14 +1,23 @@
-import React, { Children, cloneElement, ReactElement } from "react"
+import React, { Children, cloneElement, ReactElement, useState } from "react"
 import { DataBroker } from "../data-model/data-broker"
 
 import "./codap-component.scss"
+import { EditableComponentTitle } from "./editable-component-title"
```
kswenson, line 5: Move import ⬆️

kswenson (approved):
👍 Just a couple remaining issues:
- move one import
- if it's straightforward, when starting to edit a title the entire string should start out selected, like in CODAP2. If it's not straightforward we can save it for a future story.

bfinzer (approved):
👍🏻 LGTM
* I agree with Kirk's comment that entire string should start out selected

PR #456, opened 2022-08-31
V3: Adds version and build number to the hello codap area

eireland:


v3/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "codap3",
-  "version": "0.0.1",
+  "version": "v3.0.0-build.700",
```
kswenson, line 3: Version listed here should not have the "V" prefix. It's fine to display it that way to the user if desired, but this field is processed by npm and others. I gather the "700" was an attempt to push it beyond the current v2 build number? If that's a goal I'd probably start at "1000" just to make sure they don't overlap, but I'm not sure that actually matters. I was imagining we'd just start low and build.

kswenson (changes requested):
👍 This captures the basic idea. Need to figure out what we actually want to do with the build number.

bfinzer (approved):
👍🏻 LGTM
To what extent can we automate updating the version information? And when should it be automated? Perhaps the build number gets incremented every time there's a merge to main?

kswenson:
I talked to @bfinzer and he suggested starting the build numbers at `1000`.

kswenson:
> 👍🏻 LGTM To what extent can we automate updating the version information? And when should it be automated? Perhaps the build number gets incremented every time there's a merge to main?

Automating the build number incrementing will require some research and probably shouldn't hold up this PR.

PR #458, opened 2022-09-02
V3: Move axis bounds into GraphLayout

kswenson:


v3/src/components/graph/hooks/use-axis-bounds.ts
```diff
@@ -0,0 +1,40 @@
...
+          left: axisBounds.left - graphBounds.left,
+          top: axisBounds.top - graphBounds.top,
+          width: axisBounds.width,
+          height: axisBounds.height
+        })
+      }
+    })
+    wrapperElt && observer?.observe(wrapperElt)
+
+    return () => observer?.disconnect()
```
bfinzer, line 36: Should also call layout.setAxisBounds(place) so that the bounds are removed from the map

bfinzer (approved):
👍🏻 LGTM
One comment about removing axis bounds from layout when axis goes away

PR #459, opened 2022-09-02
V3: Update to React 18, Chakra 2.3.1, etc.

kswenson:
I had been thinking that it would be useful to stay on React 17 since CLUE is on React 17 and it would make it easier to share code between the two, but @eireland ran into issues with Chakra/cypress and so it seems better to bite the bullet and move ahead.


eireland (approved):
👍

bfinzer (approved):
👍🏻 LGTM

PR #461, opened 2022-09-03
V3: Opens a menu when case table index is clicked. 

eireland:
Menu items only bring up alerts at this time.

### DIFF
--- a/v3/src/components/case-table/use-index-column.tsx
+++ b/v3/src/components/case-table/use-index-column.tsx
@@ -1,9 +1,10 @@
 import React, { useCallback, useEffect, useMemo, useState } from "react"
 import { createPortal } from "react-dom"
-import { Menu, MenuButton, MenuItem, MenuList } from "@chakra-ui/react"
+import { Menu, MenuButton } from "@chakra-ui/react"
 import { IDataSet } from "../../data-model/data-set"
 import { kIndexColumnKey, TColumn, TFormatterProps } from "./case-table-types"
 import { ColumnHeader } from "./column-header"
+import { IndexMenuList } from "./index-menu-list"
 
 interface IHookProps {
   data?: IDataSet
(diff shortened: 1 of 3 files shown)


v3/src/components/case-table/use-index-column.tsx
```diff
@@ -59,10 +59,48 @@ export const IndexCell = ({ caseId, index, onClick }: ICellProps) => {
...
+      for (parent = cellElt; parent; parent = parent.parentElement) {
+        if (parent.classList.contains("codap-component")) {
+          setCodapComponentElt(parent)
+          break
+        }
+      }
+    }
+  }, [cellElt, codapComponentElt])
+
+  const handleMoveDataRow = () => {
```
kswenson, line 75: Better name would be:
```suggestion
  const handleMoveEntryRow = () => {
```

v3/src/components/case-table/use-index-column.tsx
```diff
@@ -59,10 +59,48 @@ export const IndexCell = ({ caseId, index, onClick }: ICellProps) => {
...
+      <MenuButton ref={setNodeRef} className="codap-index-content" data-testid="codap-index-content">
+        {index != null ? `${index + 1}` : ""}
+      </MenuButton>
+      {codapComponentElt && createPortal((
+        <MenuList>
+          <MenuItem onClick={handleMoveDataRow}>Move Data Entry Row Here</MenuItem>
+          <MenuItem onClick={handleInsertCase}>Insert Case</MenuItem>
+          <MenuItem onClick={handleInsertCases}>Insert Cases...</MenuItem>
+          <MenuItem onClick={handleDeleteCase}>Delete Case</MenuItem>
+        </MenuList>
```
kswenson, line 102: Ultimately, we're not going to want all of these handlers in the `useIndexColumn()` hook. One way to achieve that would be to create a separate `IndexMenuList` component which renders the `<MenuList>` and has the handlers so this would just be:
```suggestion
        <IndexMenuList />
```
Also, "Delete Case" should be "Delete Cases" when multiple cases are selected.

kswenson (changes requested):
👍 Looks good -- seems like Chakra is making things as easy as they're supposed to be. A number of suggestions but they should all be straightforward.

kswenson (approved):
👍 Changes look good

bfinzer (approved):
👍🏻 LGTM

PR #462, opened 2022-09-04
V3: Drag from case table column headers; general drag/drop improvements

kswenson:
- Drag from case table column headers
- Add `<CaseTableComponent>` which wraps `<CaseTable>` (analogous to `<GraphComponent>`)
- General drag/drop improvements
  - require three-pixel movement before initiating drags (helps disambiguate clicks from drags)
  - hide tooltips during drags
  - use rectangle intersection for dropping attributes on graph axes
  - eliminate unnecessary dependency from `useEffect()` in `useAxisBounds()`

### DIFF
--- a/v3/src/components/app-state.test.ts
+++ b/v3/src/components/app-state.test.ts
@@ -1,0 +1,30 @@
+import { reaction } from "mobx"
+import { appState } from "./app-state"
+
+describe("AppState", () => {
+  it("works when enabled", () => {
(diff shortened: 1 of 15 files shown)


bfinzer (approved):
👍🏻 LGTM
- I tested on iPad and it worked great there as well.
- There are techniques I don't fully understand, but I'm content to wait until I need them to learn more.

PR #463, opened 2022-09-05
V3—graph-bug-fixes

bfinzer:
Though initially this was meant to be minor bug fixes, it turned out that the fixes required significant refactoring.

* [#183037465] Bug fix: y-axis isn't draggable after transition from dot plot
  - Moved routine for updating drag rects to graph_utils.ts refreshAxisDragRects
* [#183037590] Bug fix: Both dot plots and scatter plots first appear scrunched in lower left

* Improving graph transition animation and fixing graph bugs
## Attribute drop on axis
- Both axis and points should animate to new scale

- Fix this: When we have dotplot and drop in new data the points appear as scatterplot. Requires dragging axis or resizing window to fix
- Fix this: Points that have missing value are being plotted when they shouldn’t be.
* Importing a new dataset with a dot plot was not displaying the points initially.
  • Fixed by updating casesRef.current early in process
  • Some refactoring
* Refactoring updating placement of axis drag rects
* Correct placement of vertical axis label
* Factoring with useGraphModels and useMovables
* Improved recomputation of default movable line and value


kswenson (approved):
👍 Sync review plus a few minor comments

PR #464, opened 2022-09-06
V3: 183170686 attribute menu UI

eireland:


v3/src/components/case-table/attribute-menu.tsx
```diff
@@ -0,0 +1,156 @@
+import { MenuItem, MenuList, useToast } from "@chakra-ui/react"
+import React, { forwardRef } from "react"
+import { CalculatedColumn } from "react-data-grid"
+import { TRow } from "./case-table-types"
+
+interface IProps {
+  disableToolTips?: (show: boolean) => void;
+  column: CalculatedColumn<TRow, unknown>;
```
kswenson, line 8:
```suggestion
  disableToolTips?: (show: boolean) => void
  column: CalculatedColumn<TRow, unknown>
```
Semicolons aren't required here, either.

v3/src/components/case-table/attribute-menu.tsx
```diff
@@ -0,0 +1,156 @@
...
+    console.log("in handleRename")
+
+    e.preventDefault()
+    toast({
+      title: 'Menu item clicked',
+      description: `You clicked on Rename ${column.name}`,
+      status: 'success',
+      duration: 5000,
+      isClosable: true,
+    })
```
kswenson, line 25: Not that important, since all of this code is temporary, but it still seems worth DRYing up the code by having a `handleMenuItemClick` function which contains the redundant code and then each menu item can be implemented with something like:
```suggestion
    handleMenuItemClick("handleRename", `You clicked on Rename ${column.name}`);
```
And then if we wanted to change the logging (for instance) we could do it in one place.

kswenson (approved):
👍 Looks good -- one suggesting about DRYing up the placeholder code a bit.

PR #465, opened 2022-09-15
V3: Adds ability to delete case from index menu

eireland:


kswenson (changes requested):
👍 Basic approach is fine, but as with another PR, for consistency use `useDataSetContext()` hook rather than passing the `DataSet` directly as a prop.

PR #467, opened 2022-09-15
V3: Adds workflow that will automatically increment build number on push to main

eireland:


bfinzer (approved):
- LGTM👍🏻
Took me awhile to understand that there exists a script that accomplishes the increment. Cool!

kswenson (changes requested):
👍 While I'm pleasantly surprised that you found an existing GitHub action for accomplishing the incrementing, by separating the build number from the version in the `package.json` we're now back to the problem of having to manually update the version in `package.json` when appropriate. Maybe for now we should just set the version in `package.json` to something like `3.0.0-mvp.1` while we think about it.

kswenson (approved):
👍 LGTM

PR #468, opened 2022-09-15
V3: 183170694 index insert case

eireland:


bfinzer (approved):
LGTM 👍🏻

v3/src/components/case-table/index-menu-list.tsx
```diff
@@ -1,11 +1,13 @@
 import { MenuItem, MenuList, useToast } from "@chakra-ui/react"
 import React from "react"
+import { IDataSet } from "../../data-model/data-set"
 
 interface IProps {
   caseId: string
   index?: number
+  data?: IDataSet
```
kswenson, line 8: Rather than passing in the `data` as a prop, you should use the `useDataSetContext` hook.

v3/src/components/case-table/use-rows.ts
```diff
@@ -98,6 +98,7 @@ export const useRows = (data?: IDataSet) => {
             // update cache entries for each affected case
             const cases = (action as AddCasesAction | SetCaseValuesAction).args[0] || []
             cases.forEach(({ __id__ }) => rowCache.set(__id__, { __id__ }))
+            resetRowCache()
```
kswenson, line 101: This should not be necessary. The point of the prior code is to make sure that only the rows that change re-render, while this will force all rows to re-render. If the more efficient code didn't work as expected then we should figure out why and fix it.

kswenson (changes requested):
👍 Mostly looks good, but two suggestions:
- use `useDataSetContext` hook rather than passing `DataSet` as prop
- re-render new row(s) without forcing re-render of all rows

v3/src/components/case-table/use-rows.ts
```diff
@@ -86,24 +86,49 @@ export const useRows = (data?: IDataSet) => {
...
-            // remove affected cases from cache
+            // remove affected cases from cache and update cache after deleted case
             const caseIds = (action as RemoveCasesAction).args[0] || []
-            caseIds.forEach(id => rowCache.delete(id))
+            caseIds.forEach(id => {
+              const caseToDeleteIndex = data.caseIndexFromID(id)
+              rowCache.delete(id)
+              const casesToUpdate = getCasesToUpdate([], caseToDeleteIndex)
+              casesToUpdate.forEach(({ __id__ }) => rowCache.set(__id__, { __id__ }))
+            })
```
kswenson, line 131: The approach below only requires a single invalidation pass rather than one pass for each deleted case.
```suggestion
            const lowestIndex = Math.min(...caseIds.map(id => data.caseIndexFromID(id)).filter(index => !!index))
            caseIds.forEach(id => rowCache.delete(id))
            const casesToUpdate = getCasesToUpdate([], lowestIndex)
            casesToUpdate.forEach(({ __id__ }) => rowCache.set(__id__, { __id__ }))
```

kswenson (changes requested):
👍 Almost there -- just a couple of suggested tweaks

kswenson (approved):
👍 LGTM

PR #469, opened 2022-09-15
V3: 183170693 index insert cases

eireland:
Adds a customizable modal to be used for future modals.
Adds ability to insert specified number of cases before or after selected case.


v3/src/components/case-table/index-menu-list.tsx
```diff
@@ -1,12 +1,44 @@
...
+  const [numCasesToInsert, setNumCasesToInsert] = useState(1)
+  const [insertPosition, setInsertPosition] = useState("after")
+
+  const InsertCasesModalContent = () => {
+    const initialRef = useRef(null)
+    return (
+      <>
+        <FormControl display="flex" flexDirection="row">
+          <FormLabel># cases to insert:</FormLabel>
+          <Input size="xs" w="75" ref={initialRef} placeholder="1"
```
bfinzer, line 24: When I erase the default value in the input field, NaN appears. Maybe "Number of cases" instead?

v3/src/components/codap-modal.tsx
```diff
@@ -0,0 +1,63 @@
...
+  onCustomClose?: () => void
+}
+export const CodapModal = <IContentProps,>({ isOpen, onClose,
+  className, Icon, title, Content, contentProps, hasCloseButton, buttons, onCustomClose
+}: IProps<IContentProps>) => {
+
+return (
+    <Modal isOpen={isOpen} onClose={onClose} >
+    <ModalOverlay />
+    <ModalContent w="250">
```
bfinzer, line 39: I'm not clear why this width is hard coded?

v3/src/components/codap-modal.tsx
```diff
@@ -0,0 +1,63 @@
...
+}
+
+interface IProps<IContentProps> {
+  children?: any
+  isOpen: boolean
+  onClose: () => void
+  className?: string
+  title: string
+  Icon?: React.FC<any>
+  Content: React.FC<IContentProps>
```
bfinzer, line 23: For my benefit: How can IContentProps be part of the definition of IContentProps?

v3/src/data-model/data-set.ts
```diff
@@ -72,6 +72,14 @@ export interface IAddCaseOptions {
   // id(s) of case(s) before which to insert new cases
   // if not specified, new cases are appended
   before?: string | string[];
+  after?: string | string[];
```
bfinzer, line 75: Can both 'before' and 'after' be specified? And what does it mean for there to be an array of ids?

bfinzer (changes requested):
One bug and some clarifications needed.

v3/src/components/codap-modal.tsx
eireland: IContentProps is a template type that allows Content to specify what type it will be. https://www.typescriptlang.org/docs/handbook/2/objects.html#generic-object-types

eireland: IContentProps is a template type that allows Content to specify what type it will be. [https://www.typescriptlang.org/docs/handbook/2/objects.html#generic-object-types](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html)

v3/src/components/codap-modal.tsx
```diff
@@ -0,0 +1,63 @@
...
+        {buttons.map((b: any, i)=>{
+          const key = `${i}-${b.className}`
+          return (
+            <Button key={key} size="xs" variant="ghost" ml="5" onClick={b.onClick} data-testid={b.label}>
+              {b.label}
+            </Button>
+          )
+        })}
+      </ModalFooter>
+    </ModalContent>
```
kswenson, line 60: These should be tabbed in.

v3/src/components/codap-modal.tsx
```diff
@@ -0,0 +1,63 @@
...
+    <ModalContent>
+      <ModalHeader h="30" className="codap-modal-header" data-testid="codap-modal-header">
+        <div className="codap-modal-icon-container">
+          {Icon && <Icon />}
+        </div>
+        <div className="codap-header-title">{title}</div>
+        {hasCloseButton && <ModalCloseButton onClick={onClose}/>}
+      </ModalHeader>
+      <ModalBody>
+        <Content {...(contentProps)} />
```
kswenson, line 48:
```suggestion
        <Content {...contentProps} />
```

v3/src/components/case-table/index-menu-list.tsx
```diff
@@ -46,12 +65,76 @@ export const IndexMenuList = ({caseId, index}: IProps) => {
...
+    </>
+
+  )
+}
+
+interface IInsertCasesModalProps {
+  numCasesToInsert: number
+  insertPosition: string
+  onChangeNumCasesToInsert: (value: any) => void
+  onChangeInsertPosition: (value: any) => void
```
kswenson, line 108: It should be possible to type these properly. I suspect they should both be strings while they're being edited.

v3/src/components/case-table/index-menu-list.tsx
```diff
@@ -46,12 +65,76 @@ export const IndexMenuList = ({caseId, index}: IProps) => {
...
+}
+
+export const InsertCasesModalContent: React.FC<IInsertCasesModalProps> =
```
kswenson, line 111: I've been trying to keep our files/modules small and tightly defined. With that in mind, I think as much of the code that specific to the insert cases menu item as possible should be moved into its own file/module.

kswenson (changes requested):
👍 Basically this looks good -- I have a bunch of comments but most of them are minor and/or organizational. We need to sort out what's going on with the invalidation of rows so that we don't have to resort to re-rendering the entire table on every change.

kswenson (approved):
👍 Changes look good

PR #470, opened 2022-09-21
V3: Add FilteredCases class

kswenson:
I started out thinking this would be a more elaborate `DerivedDataSet` or `DataSetProxy` but the attribute filtering and API forwarding aspects of it seemed to extra work for relatively little gain, so I decided to focus more narrowly on the case filtering aspect.


bfinzer (approved):
👍🏻 LGTM

PR #471, opened 2022-09-22
V3-Disable index attribute menu

eireland:


v3/src/components/case-table/column-header.tsx
```diff
@@ -70,7 +70,7 @@ export const ColumnHeader = ({ column }: Pick<THeaderRendererProps, "column">) =
                 openDelay={1000} placement="bottom" bottom="15px" left="15px"
                 isDisabled={dragging} closeOnMouseDown={true}>
               <div className="codap-column-header-content" ref={setCellRef} {...attributes} {...listeners}>
-                <MenuButton className="codap-attribute-button"
+                <MenuButton className="codap-attribute-button" disabled={column?.name==="index"}
```
kswenson, line 73:
```suggestion
                <MenuButton className="codap-attribute-button" disabled={column?.key === kIndexColumnKey}
```
less likely to run into problems with localization.

kswenson (approved):
👍 Looks good -- just suggested keying on the id rather than the name, but feel free to merge after you've made that change.

PR #478, opened 2022-09-27
V3—point_size_and_hover

bfinzer:
* The size of plotted points is determined dynamically, computed using the same algorithm as in V2
* Point data tips are now correctly computed from the current plotted attributes, including a caption attribute. Their display is quite close to that of V2, but done using an npm library based on d3.

### DIFF
--- a/v3/src/components/graph/components/graph.tsx
+++ b/v3/src/components/graph/components/graph.tsx
@@ -1,11 +1,12 @@
 import {useToast} from "@chakra-ui/react"
 import {select} from "d3"
+import {tip as d3tip} from "d3-v6-tip"
 import {observer} from "mobx-react-lite"
 import {onAction} from "mobx-state-tree"
 import React, {MutableRefObject, useEffect, useRef} from "react"
 import {Axis} from "./axis"
 import {Background} from "./background"
-import {defaultRadius, kGraphClass} from "../graphing-types"
+import {kGraphClass} from "../graphing-types"
 import {ScatterDots} from "./scatterdots"
 import {DotPlotDots} from "./dotplotdots"
 import {CaseDots} from "./casedots"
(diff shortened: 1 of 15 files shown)


v3/src/components/graph/components/graph.tsx
```diff
@@ -31,105 +32,133 @@ interface IProps {
   dotsRef: React.RefObject<SVGSVGElement>
 }
 
-const marqueeState = new MarqueeState()
+const marqueeState = new MarqueeState(),
+  dataTip = d3tip().attr('class', 'graph-d3-tip')/*.attr('opacity', 0.8)*/
+    .html((d: string) => {
+      return "<p>" + d + "</p>"
+    })
```
kswenson, line 39: Seems like we might have to worry about sanitization here, e.g. if the content being rendered included `</p>` then we'd end up with invalid html. Doesn't necessarily have to be addressed in this PR, but perhaps a `TODO` comment to remind us to come back to it would be warranted.

v3/src/components/graph/components/graph.tsx
```diff
@@ -31,105 +32,133 @@ interface IProps {
...
+    // We only need to make the following connection once
+    useEffect(function passDotsRefToController() {
+      graphController.setDotsRef(dotsRef)
+    }, [dotsRef, graphController])
+
+    // MouseOver events, if over an element, brings up hover text
+    function showDataTip(event: MouseEvent, d: any) {
+      const
+        target = select(event.target as SVGSVGElement)
+      if (target.node()?.nodeName === 'circle' && dataset) {
```
kswenson, line 106: We'll need to generalize this to more than dots at some point.

v3/src/components/graph/models/axis-model.ts
```diff
@@ -1,6 +1,6 @@
 import {Instance, types} from "mobx-state-tree"
 
-export const AxisPlaces = ["bottom", "left", "right", "top"] as const
+export const AxisPlaces = ["bottom", "left", "right", "top", "caption"] as const
```
kswenson, line 3: We can talk about this more when we discuss the DataConfiguration, but in my current formulation `caption` is a `GraphAttrPlace` rather than an `AxisPlace`.

kswenson (approved):
👍 Looks good -- just a few questions/comments but nothing that need hold up the PR.

v3/src/components/graph/components/graph.tsx
bfinzer: Indeed!

PR #482, opened 2022-09-29
V3: Fix several case table bugs

kswenson:
- horizontal scroll bar not visible
- column headers no longer bold
- no row selection highlighting on odd rows
- clicking on row sometimes scrolls unnecessarily
- selecting case in graph doesn't always scroll corresponding row into view

### DIFF
--- a/v3/src/components/case-table/case-table.scss
+++ b/v3/src/components/case-table/case-table.scss
@@ -1,7 +1,9 @@
+@import "../vars.scss";
+
 .case-table {
   position: relative;
   width: 100%;
-  height: 100%;
+  height: calc(100% - $title-bar-height);
 
   .rdg {
     width: 100%;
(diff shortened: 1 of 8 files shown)


bfinzer (approved):
👍🏻 LGTM

v3/src/components/case-table/case-table.scss
```diff
@@ -14,6 +16,7 @@
         .codap-column-header-content {
           width: 100%;
           height: 100%;
+          font-weight: bold !important;
```
eireland, line 19: Is this for the drag overlay? We are already specifying `fontWeight="bold"` in inline styling.

eireland (approved):
looks good except for maybe redundant styling?

v3/src/components/case-table/case-table.scss
kswenson: Good catch -- probably redundant. I'll check and remove it if so.

PR #483, opened 2022-09-29
V3: Hides attribute when selected from attribute menu

eireland:


bfinzer (approved):
👍🏻 LGTM

v3/src/data-model/data-set.ts
```diff
@@ -422,6 +422,11 @@ export const DataSet = types.model("DataSet", {
         }
       },
 
+      hideAttribute(attributeID: string) {
+        const attrToHide = self.attributes.find(attr => attr.id === attributeID)
+        attrToHide?.setHidden(true)
+      },
```
kswenson, line 428: I'm somewhat ambivalent, but my inclination would be _not_ to add this function here. I don't think it's necessary for _every_ action that can be applied to an attribute to be wrapped at the `DataSet` level. If it were to be wrapped at this level then it would also be necessary to provide a mechanism for clients to show a previously hidden attribute. In this case it seems to me that asking clients to access the attribute directly is not onerous, e.g.:
```typescript
  dataSet.attrFromID(attrID)?.setHidden(true)
  ...
  dataSet.attrFromID(attrID)?.setHidden(false)
```

kswenson (changes requested):
👍 Looks good -- one suggestion to remove the `DataSet` change.

kswenson (approved):
👍 LGTM

PR #484, opened 2022-09-29
V3: 181889611 rename attribute

eireland:
Adds ability to rename attribute in the case table column header. Column header changes into an input component to allow change


bfinzer:
I notice that soon after the name gets focus for editing the tooltip comes up. I think you were trying to fix that, right?

v3/src/components/case-table/column-header.tsx
```diff
@@ -60,26 +65,58 @@
...
+      case "Enter":
+      case "Tab":
+        handleClose(true)
+        e.currentTarget.blur()
+        break
+    }
+  }
+  const handleClose = (accept: boolean) => {
+    const attrId = data?.attrIDFromName(columnNameStr)
+    const trimTitle = editingAttrName?.trim()
```
bfinzer, line 84: I like to declare these in the clause where they are used

v3/src/data-model/data-set.ts
```diff
@@ -419,6 +419,7 @@ export const DataSet = types.model("DataSet", {
         const attribute = attributeID && attrIDMap[attributeID]
         if (attribute) {
           attribute.setName(name)
+          attrNameMap[name] = attributeID
```
bfinzer, line 422: Where does the entry in the attrNameMap for the old name get deleted?

bfinzer (changes requested):
One question needs answering.

kswenson (changes requested):
👍 I was going to suggest a couple of tweaks but then after trying them locally went ahead and added a commit (after showing them to @eireland). I also made a suggestion about clearing out the old attribute name from the dataset (also pointed out by Bill). Finally, as we discussed when we met, I think you should try to remove the inner input highlight so that there isn't a double-highlight when editing and it follows the convention for editing body cells.

kswenson (approved):
👍 Changes look good

PR #485, opened 2022-09-29
V3—Tool shelf with buttons

bfinzer:
* The buttons appear but don't do anything yet. 
* Styling is close to V2, but needs some tweaking, especially the Plugins button.
* Added three dots (…) for Tables and Plugins because it looked less clunky than ▼, but we might want to try harder with the arrow.


v3/src/components/tool-shelf/tool-shelf.scss
```diff
@@ -0,0 +1,26 @@
+@import "../vars.scss/";
+
+.tool-shelf {
+  height: $tool-shelf-height;
+  width: 100%;
+  background-color: #ececec;
+  border-bottom: 1px solid #dcdbd8;
+  font-family: 'MuseoSans-500', 'sans-serif';
```
kswenson, line 8: TODO: guarantee that MuseoSans is available, either by including it in the project bundle or linking to it via CDN.

v3/src/components/tool-shelf/tool-shelf.scss
```diff
@@ -0,0 +1,26 @@
...
+  font-size: 11px;
+  box-shadow: 0 0 0 1px #888888!important;
+
+  .button-stack {
+    margin-top: 5px;
+  }
+  .button-stack:hover {
+    border-right: solid 2px #888888;
+    border-bottom: solid 2px #888888;
```
kswenson, line 17: Should probably use a .scss constant rather than repeating `#888888` (or alternatively use `gray`).

v3/src/components/tool-shelf/tool-shelf.tsx
```diff
@@ -0,0 +1,103 @@
...
+import {Box, Flex, HStack, Tag, useToast, VStack} from "@chakra-ui/react"
+
+import './tool-shelf.scss'
+import GraphIcon from '../../assets/icons/icon-graph.svg'
+import TableIcon from '../../assets/icons/icon-table.svg'
+import MapIcon from '../../assets/icons/icon-map.svg'
+import SliderIcon from '../../assets/icons/icon-slider.svg'
+import CalcIcon from '../../assets/icons/icon-calc.svg'
+import TextIcon from '../../assets/icons/icon-text.svg'
+import PluginsIcon from '../../assets/icons/icon-plug.svg'
```
kswenson, line 11: Move these up with the main body of imports, because they're actually importing React components (after conversion by SVGR).

v3/src/components/tool-shelf/tool-shelf.tsx
```diff
@@ -0,0 +1,103 @@
...
+      icon: PluginsIcon,
+      iconLabel: 'Plugins …',
+      buttonHint: 'Add a plugin to the document',
+      handler: pluginsHandler
+    }
+  ]
+
+  return (
+    <HStack className='tool-shelf' alignContent='center'>
+      <Flex height='100%' borderRight='solid #888888'>
```
kswenson, line 85:
```suggestion
      <Flex height='100%' borderRight='solid gray'>
```

v3/src/components/tool-shelf/tool-shelf.tsx
```diff
@@ -0,0 +1,103 @@
...
+        {buttonDescriptions.map(aDesc => {
+          return (
+            <Box
+              as='button'
+              key={aDesc.iconLabel}
+              bg='white'
+              onClick={aDesc.handler}
+            >
+              <VStack className='button-stack'>
+                {<aDesc.icon height='25px'/>}
```
kswenson, line 95:
```suggestion
                {<aDesc.icon height={25}/>}
```
I would think you should just be able to pass the number here.

kswenson (approved):
👍 Looks good -- just a few comments/suggestions.

PR #491, opened 2022-10-02
V3—Point radius fixes

bfinzer:
The fact that the necessary point radius was being computed in graph-utils where it required a bunch of parameters was making it difficult and confusing to get the right point radius. Most of the changes have to do with refactoring so that the GraphModel coughs up the desired point radius. This makes sense because it has all the information already to make the computation.

### DIFF
--- a/v3/src/components/graph/components/casedots.tsx
+++ b/v3/src/components/graph/components/casedots.tsx
@@ -1,11 +1,11 @@
 import {randomUniform, select} from "d3"
-import React, {memo, useCallback, useRef, useState} from "react"
-import {defaultRadius, dragRadius, transitionDuration} from "../graphing-types"
+import React, {memo, useCallback, useEffect, useRef, useState} from "react"
+import {hoverRadiusFactor, pointRadiusSelectionAddend, transitionDuration} from "../graphing-types"
 import {useDragHandlers, usePlotResponders} from "../hooks/graph-hooks"
 import {useDataSetContext} from "../../../hooks/use-data-set-context"
 import {useInstanceIdContext} from "../../../hooks/use-instance-id-context"
 import {ScaleNumericBaseType, useGraphLayoutContext} from "../models/graph-layout"
-import {computedPointRadius, setPointSelection} from "../utilities/graph_utils"
+import {setPointSelection} from "../utilities/graph_utils"
 import {IGraphModel} from "../models/graph-model"
 
 export const CaseDots = memo(function CaseDots(props: {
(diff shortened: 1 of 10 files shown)


v3/src/components/graph/components/graph.tsx
```diff
@@ -102,21 +105,28 @@ export const Graph = observer((
...
+    function showDataTip(event: MouseEvent) {
       const
         target = select(event.target as SVGSVGElement)
       if (target.node()?.nodeName === 'circle' && dataset) {
+        target.transition().duration(transitionDuration).attr('r', hoverPointRadius)
         const [, caseID] = target.property('id').split("_"),
           attrIDs = graphModel.config.uniqueTipAttributes,
           tipText = getPointTipText(dataset, caseID, attrIDs)
-        dataTip.show(tipText, event.target)
+        tipText !== '' && dataTip.show(tipText, event.target)
```
kswenson, line 116: Assuming `tipText` can't be another falsy value (like numeric `0`), you could also use:
```suggestion
        tipText && dataTip.show(tipText, event.target)
```

v3/src/components/graph/utilities/graph_utils.ts
```diff
@@ -95,21 +87,22 @@ export function getPointTipText(dataset: IDataSet, caseID: string, attributeIDs:
         isNumeric = attribute?.type === 'numeric',
         value = isNumeric ? float(dataset?.getNumeric(caseID, attrID) ?? 0) :
           dataset?.getValue(caseID, attrID)
-      return `${name}: ${value}`
+      return (value && (isNumeric && isFinite(value)) || (!isNumeric && value !== '')) ? `${name}: ${value}` : ''
     }))
   // Caption attribute can also be one of the plotted attributes, so we remove dups and join into html string
-  return Array.from(new Set(attrArray)).join('<br>')
+  return Array.from(new Set(attrArray)).filter(anEntry => anEntry !== '').join('<br>')
```
kswenson, line 93: Assuming `anEntry` can't be another falsy value, could also use:
```suggestion
  return Array.from(new Set(attrArray)).filter(anEntry => !!anEntry).join('<br>')
```

kswenson (approved):
👍 Looks good -- just a few minor comments.

v3/src/components/graph/components/graph.tsx
bfinzer: But tipText can be '', in which case we don't want to show the dataTip because actually displays a small white area. (The call to getPointTipText guarantees that it is a string.)

v3/src/components/graph/utilities/graph_utils.ts
bfinzer: anEntry _will_ be a string, but if it's an _empty_ string we don't want to include it.

v3/src/components/graph/components/casedots.tsx
```diff
@@ -20,33 +20,36 @@ export const CaseDots = memo(function CaseDots(props: {
...
-    randomPointsRef = useRef<Record<string, { x: number, y: number }>>({})
-  dataset?.cases.forEach(({__id__}) => {
-    randomPointsRef.current[__id__] = {x: uniform(), y: uniform()}
-  })
+  useEffect(function initDistribution() {
+    const uniform = randomUniform()
+    dataset?.cases.forEach(({__id__}) => {
+      randomPointsRef.current[__id__] = {x: uniform(), y: uniform()}
+    })
+  }, [dataset?.cases])
```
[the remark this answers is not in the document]
bfinzer: I agree that it should and am also puzzled if it doesn't. How can I set it up so that the useEffect only runs if there is a new dataset, but not when the cases change?

v3/src/components/graph/components/graph.tsx
kswenson: Exactly -- my proposed change should have the same effect with less code.

v3/src/components/graph/utilities/graph_utils.ts
kswenson: Exactly -- my proposed change should have the same effect with less code.

PR #492, opened 2022-10-03
V3—Drag csv bug

bfinzer:
* The new data configuration created when a new dataset is dragged into V3 was not getting linked up with the dataset soon enough. We rearrange things os everything is in place by the time the case plot positions the points.


kswenson (approved):
👍 LGTM

PR #493, opened 2022-10-03
Adjust useEffect dependencies

kswenson:
>>It seems like this would cause the point positions to be rerandomized on insertion/removal of cases, but that doesn't seem to be the case in testing. 🤔
>
>I agree that it should and am also puzzled if it doesn't. How can I set it up so that the useEffect only runs if there is a new dataset, but not when the cases change?


bfinzer (approved):
👍🏻 LGTM
That is a subtle solution!

PR #494, opened 2022-10-04
V3: Fix click handling in case table column header menus

kswenson:
Fix click handling in Chakra menus and dialogs caused by React's event bubbling through portals. See comments in code for details.

Also prevents the dev-server overlay from appearing when there are only warnings in the build. (Errors still trigger the overlay).


bfinzer (approved):
👍🏻 LGTM
Tough problem, elegant fix!

eireland (approved):
LGTM

PR #499, opened 2022-10-06
V3—2D_dot_chart

bfinzer:
Given a dot chart, dropping a categorical attribute on the empty axis splits the dot chart by the new categories.
* Note that the stack of dots grows perpendicular to the _first_ attribute and that therefore the stacks can be made to grow either vertically _or_ horizontally depending on which attribute was first. There are some subtleties brought about by _replacing_ the first attribute with a different categorical which currently forces the growth direction to switch. We may want to change this behavior.
* As with split dot plots the grid lines remain to be implemented

### DIFF
--- a/v3/src/components/graph/components/chartdots.tsx
+++ b/v3/src/components/graph/components/chartdots.tsx
@@ -1,87 +1,116 @@
 import {ScaleBand, select} from "d3"
-import React, {memo, useCallback, useEffect, useRef} from "react"
+import React, {memo, useCallback, useEffect} from "react"
 import {PlotProps, transitionDuration} from "../graphing-types"
 import {usePlotResponders} from "../hooks/graph-hooks"
 import {useDataConfigurationContext} from "../hooks/use-data-configuration-context"
 import {useDataSetContext} from "../../../hooks/use-data-set-context"
-import {ScaleType, useGraphLayoutContext} from "../models/graph-layout"
+import {useGraphLayoutContext} from "../models/graph-layout"
 import {setPointSelection} from "../utilities/graph_utils"
 import {IGraphModel} from "../models/graph-model"
+import {attrPlaceToAxisPlace} from "../models/axis-model"
 
 interface IProps {
-  graphModel:IGraphModel
-  plotProps:PlotProps
+  graphModel: IGraphModel
+  plotProps: PlotProps
 }
+
+type BinMap = Record<string, Record<string, number>>
 
 export const ChartDots = memo(function ChartDots(props: IProps) {
   const {graphModel, plotProps: {dotsRef, enableAnimation}} = props,
     dataConfig = useDataConfigurationContext(),
     dataset = useDataSetContext(),
     layout = useGraphLayoutContext(),
-    xAttrID = dataConfig?.attributeID('x'),
-    xAttributeType = dataConfig?.attributeType('x'),
-    yAttrID = dataConfig?.attributeID('y'),
-    // yAttributeType = dataset?.attrFromID(yAttrID)?.type,
-    primaryPlace = xAttributeType === 'categorical' ? 'bottom' : 'left',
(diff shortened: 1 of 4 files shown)


v3/src/components/graph/components/chartdots.tsx
```diff
@@ -1,126 +1,157 @@
...
+      bins[aValue.primary][aValue.secondary]++
     })
-    return Object.keys(bins).reduce((max, aKey) => Math.max(max, bins[aKey]), 0)
-  }, [attribute])
+    return Object.keys(bins).reduce((hMax, hKey) => {
+      return Math.max(hMax, Object.keys(bins[hKey]).reduce((vMax, vKey) => {
+        return Math.max(vMax, bins[hKey][vKey])
+      }, 0))
+    }, 0)
+  }, [dataConfig?.cases, dataset, primaryAttrID, secondaryAttrID])
```
kswenson, line 56: Use the `const { cases } = dataConfig` trick above so that this just depends on `dataConfig`.

v3/src/components/graph/components/chartdots.tsx
```diff
@@ -1,126 +1,157 @@
...
+      secondaryCategoriesArray: string[] = dataConfig ?
+        Array.from(dataConfig.categorySetForPlace(secondaryAttrPlace)) : [],
       pointDiameter = 2 * graphModel.getPointRadius(),
       selection = select(dotsRef.current).selectAll(selectedOnly ? '.graph-dot-highlighted' : '.graph-dot'),
       duration = enableAnimation.current ? transitionDuration : 0,
-      cellWidth = primaryScale.bandwidth()
+      primaryCellWidth = primaryScale.bandwidth(),
+      primaryHeight = secondaryScale.bandwidth ? secondaryScale.bandwidth() :
+        layout.axisLength(secondaryAxisPlace),
+      categoriesMap: Record<string, Record<string, { cell: { h: number, v: number }, numSoFar: number }>> = {}
```
kswenson, line 77: We haven't talked about this before, and it need not hold up this PR, but it's worth considering whether a JS `Map` would make sense in some of these places where we're using a simple JS object as a map.

v3/src/components/graph/models/data-configuration-model.ts
```diff
@@ -93,6 +93,21 @@ export const DataConfigurationModel = types
...
+      return attrID ? self.cases.map(anID => String(dataset?.getValue(anID, attrID))) : []
+    },
+    categorySetForPlace(place:GraphAttrPlace):Set<string> {
+      const result:Set<string> = new Set(this.attributeValuesForPlace(place))
+      result.delete('')
+      if(result.size === 0) {
+        result.add('__main__')
+      }
+      return result
+    }
```
kswenson, line 109: There should be some tests of these methods as well, but I leave it to your discretion whether that's a task for this PR or not.

kswenson (changes requested):
👍 Generally looks good -- just a few suggestions. Also, there's a lint warning for an undeclared `useEffect` dependency in `DotPlotDots` that's left over from the previous PR that would be good to fix.

v3/src/components/graph/components/chartdots.tsx
bfinzer: Interesting. There certainly have been times in V2 when I had to work around the inability to guarantee stable insertion order.

v3/src/components/graph/models/data-configuration-model.ts
bfinzer: I'll include tests in this PR. Also, I think the detection/elimination of empty values should occur in attributeValuesForPlace, obviating the need to delete them from the Set.

bfinzer: I've added them now.

kswenson (approved):
👍 Changes look good

PR #500, opened 2022-10-07
V3: DataConfigurationModel supports onAction listeners

kswenson:
The `DataConfigurationModel` supports `onAction` listeners. Generally speaking, these are forwarded from the `DataSet` but in the case of `setCaseValues` actions, the `DataConfigurationModel` will notify of `addCases` and `removeCases` changes as well based on any filtering changes that result.


bfinzer (approved):
👍🏻 LGTM

PR #504, opened 2022-10-11
V3: update table styling; fix bugs

kswenson:
Builds on @eireland's `table-styling` branch:
- updates column header formatting
- updates scrolling code to account for row height changes
- fixes bug that caused unnecessary scroll when double-clicking to edit cells
- shares row height constants between .scss and .tsx files
- updates unit tests to work with updated row heights


bfinzer (approved):
👍🏻LGTM
But there's a lot going on here that I don't understand.

PR #506, opened 2022-10-12
183469552 attr plot drop

bacalj:
Adds ability to drop an attribute on the plot.  

#### `dnd-detect-collision.ts`
- If we have a drop collision with the overlay, we accept drops on a droppable element with an id that matches `/graph.+-drop/`.  

#### `axis.ts` 
- `droppableId` in now matches the above pattern

#### `graph-component.tsx`
- the drop Id now ends with `-overlay` to differentiate from all existing and future droppables

#### `graph.tsx`
- Instantiates a `DroppableSvg` and attaches it to the `g` forwarded from `Background`
- Handles drops on the above and sends up a toast with the attribute dropped

#### `background.ts`
- the `Background` props are now typed with an `IProps` interface.
- The `Background` component now receives a ref when it is instantiated so that it can forward this ref as a `g` element that the `DroppableSvg` will use as its `target`

There are also a few spots where I adjusted indentation as I went


v3/src/components/graph/components/graph.tsx
```diff
@@ -103,14 +110,13 @@ export const Graph = observer((
...
-      const
-        target = select(event.target as SVGSVGElement)
+      const target = select(event.target as SVGSVGElement)
       if (target.node()?.nodeName === 'circle' && dataset) {
         target.transition().duration(transitionDuration).attr('r', hoverPointRadius)
         const [, caseID] = target.property('id').split("_"),
           attrIDs = graphModel.config.uniqueTipAttributes,
           tipText = getPointTipText(dataset, caseID, attrIDs)
-        tipText !== '' && dataTip.show(tipText, event.target)
+          tipText !== '' && dataTip.show(tipText, event.target)
```
kswenson, line 119: This indent is incorrect -- it should be at the same level as the `const` (as it was) since it's not one of the values being defined by the `const`.

kswenson (commented):
👍 Looks good -- I see you've already merged this, but for future reference just one indenting issue to note.

v3/src/components/graph/components/graph.tsx
bacalj: Oh, thanks, I see that now.  I'll move it back in my next branch.

PR #509, opened 2022-10-12
V3: Fix drag/drop with keyboard

kswenson:
- fix our custom collision-detection function to work with keyboard sensor
- case table adopts latest droppable id convention

Note that currently drag/drop of attributes from the summary view works fine, but drag/drop from the case table column headers doesn't work because the enter/return/space keys bring up the column header menu rather than initiating a drag. We'll have to figure out some way of enabling both behaviors at some point.

@bacalj I'm assigning this to you for review since you're now more familiar with the dnd-kit API than others.

### DIFF
--- a/v3/src/components/case-table/case-table-component.tsx
+++ b/v3/src/components/case-table/case-table-component.tsx
@@ -12,8 +12,8 @@ interface IProps {
 export const CaseTableComponent = observer(({ broker }: IProps) => {
   const instanceId = useNextInstanceId("case-table")
   const data = broker?.selectedDataSet || broker?.last
-
-  const { setNodeRef } = useDroppable({ id: `${instanceId}-component-drop`, data: { accepts: ["attribute"] } })
+  const id = `${instanceId}-component-drop-overlay`
+  const { setNodeRef } = useDroppable({ id, data: { accepts: ["attribute"] } })
 
   return (
     <DataSetContext.Provider value={data}>
--- a/v3/src/components/case-table/column-header-divider.tsx
+++ b/v3/src/components/case-table/column-header-divider.tsx
@@ -3,6 +3,7 @@ import React, { CSSProperties, useEffect, useState } from "react"
 import { createPortal } from "react-dom"
 import { IMoveAttributeOptions } from "../../data-model/data-set"
 import { useDataSetContext } from "../../hooks/use-data-set-context"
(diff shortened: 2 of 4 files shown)


bacalj:
👍 + a thought + a question

> drag/drop from the case table column headers doesn't work because the enter/return/space keys bring up the column header menu rather than initiating a drag. We'll have to figure out some way of enabling both behaviors at some point.

Maybe adding a "drag this attribute with the keyboard" item to the contextual menu? In terms of accessibility, perhaps we'd want menu items that were like "assign this attribute to the y axis" etc, removing the spatial factor entirely (at least from the user action)

> case table adopts latest droppable id convention

So, for the moment we have a convention of ending with `-overlay` so as to not match `-drop`, correct?  Would we eventually want to enforce this with a constant or a type?

bacalj (approved):
👍 👍  I left a few thoughts/questions in context as a comment.

kswenson:
> > drag/drop from the case table column headers doesn't work because the enter/return/space keys bring up the column header menu rather than initiating a drag. We'll have to figure out some way of enabling both behaviors at some point.
> 
> Maybe adding a "drag this attribute with the keyboard" item to the contextual menu? In terms of accessibility, perhaps we'd want menu items that were like "assign this attribute to the y axis" etc, removing the spatial factor entirely (at least from the user action)

Worth considering when we get around to designing the interaction more fully.

> > case table adopts latest droppable id convention
> 
> So, for the moment we have a convention of ending with `-overlay` so as to not match `-drop`, correct? Would we eventually want to enforce this with a constant or a type?

Eventually, I expect each component to register its own component-specific collision detection algorithm so there won't be a need to enforce consistency, although similarity of approach between the different components could still make things easier for the those working with the code.

ddamelin:
> Maybe adding a "drag this attribute with the keyboard" item to the contextual menu? In terms of accessibility, perhaps we'd want menu items that were like "assign this attribute to the y axis" etc,

One can click on a graph axis to set the attribute rather than using drag-and-drop. With many multiple components/targets for the drop action, it seems like the better approach would be to interact directly with the area where the attribute would be dropped, as we do now in CODAP V2. The area of improvement would be in making more of those drop targets accessible via keyboard interaction. I think right now you can only set the horizontal and left-vertical axis attributes by clicking on them (or potentially using a keyboard to trigger the menu of attributes that happens when you click on it now).

PR #510, opened 2022-10-13
V3: Reimplement DataSet caching

kswenson:
Reimplement DataSet caching to use closure variables instead of volatile properties

This allows `beginCaching()` and other methods that only affect caching to be `views` rather than `actions`.


bfinzer (approved):
👍🏻Nice fix!
Just wondering if we can do anything to protect against mismatched being and end caching, say when an exception occurs?

PR #511, opened 2022-10-13
183534287 observer warning

bacalj:
Addresses persistent console warning: "Cannot update a component (`observerComponent`) while rendering a different component"

`graph-component.tsx`
- moves instantiation of new/replacement `GraphController` to a dedicated function 
- new instance is stored in a dedicated `ref`, whose current value is passed to the `Graph`
- conditional logic for re-creating the object is moved to a `useState` call with the same dependencies as the original memo

`graph-controller`
- model will only set the `axes` and `left` properties to an instance of an `EmptyAxisModel` if the `dotsRef` does not yet hold an svg
- if it does, it will `matchCirclesToData`

`graph.tsx`
- returns an unrelated indentation to its correct level

### DIFF
--- a/v3/src/components/graph/components/graph-component.tsx
+++ b/v3/src/components/graph/components/graph-component.tsx
@@ -42,7 +42,9 @@ export const GraphComponent = observer(({broker, v2Document}: IProps) => {
         graphModel: defaultGraphModel,
         dataset, layout, enableAnimation, instanceId, dotsRef, v2Document
       }),
-      [dataset, layout, instanceId, v2Document])
+      [/* dataset, */ layout, instanceId, v2Document])
+      // removing this dependency works, but dots won't load
+      // experiment with putting graphController in a memo as here, but maybe in a ref instead
 
   useEffect(() => {
     (width != null) && (height != null) && layout.setGraphExtent(width, height)
--- a/v3/src/components/graph/components/graph-component.tsx
+++ b/v3/src/components/graph/components/graph-component.tsx
@@ -11,8 +11,6 @@ import {DataConfigurationModel} from "../models/data-configuration-model"
 import {GraphLayout, GraphLayoutContext} from "../models/graph-layout"
 import {GraphModel} from "../models/graph-model"
 import {Graph} from "./graph"
-import {CodapV2Document} from "../../../v2/codap-v2-document"
-import {useGraphController} from '../hooks/use-graph-controller'
 
 const defaultGraphModel = GraphModel.create({
   axes: {
@@ -25,21 +23,15 @@ const defaultGraphModel = GraphModel.create({
 
 interface IProps {
   broker?: DataBroker
-  v2Document?: CodapV2Document
 }
 
-export const GraphComponent = observer(({broker, v2Document}: IProps) => {
+export const GraphComponent = observer(({broker}: IProps) => {
   const instanceId = useNextInstanceId("graph")
   const layout = useMemo(() => new GraphLayout(), [])
   const {width, height, ref: graphRef} = useResizeDetector({refreshMode: "debounce", refreshRate: 200})
   const enableAnimation = useRef(true)
   const dataset = broker?.selectedDataSet || broker?.last
   const dotsRef = useRef<SVGSVGElement>(null)
-
-  const graphController = useGraphController({
-    graphModel: defaultGraphModel,
-    layout, dataset, enableAnimation, dotsRef, v2Document
-  })
 
   useEffect(() => {
     (width != null) && (height != null) && layout.setGraphExtent(width, height)
@@ -55,7 +47,6 @@ export const GraphComponent = observer(({broker, v2Document}: IProps) => {
       <InstanceIdContext.Provider value={instanceId}>
         <GraphLayoutContext.Provider value={layout}>
           <Graph model={defaultGraphModel}
-            graphController={graphController.current}
             graphRef={graphRef}
             enableAnimation={enableAnimation}
             dotsRef={dotsRef}
--- a/v3/src/components/graph/components/graph-component.tsx
+++ b/v3/src/components/graph/components/graph-component.tsx
@@ -63,7 +63,7 @@ export const GraphComponent = observer(({broker, v2Document}: IProps) => {
       <InstanceIdContext.Provider value={instanceId}>
         <GraphLayoutContext.Provider value={layout}>
           <Graph model={defaultGraphModel}
-            graphController={graphController.current || getNewGraphController()}
+            graphController={graphController.current}
             graphRef={graphRef}
             enableAnimation={enableAnimation}
             dotsRef={dotsRef}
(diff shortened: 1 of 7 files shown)


v3/src/components/graph/components/graph-component.tsx
```diff
@@ -35,14 +35,19 @@ export const GraphComponent = observer(({broker, v2Document}: IProps) => {
...
+  const getNewGraphController = () => {
+    return new GraphController({
+      graphModel: defaultGraphModel,
+      dataset, layout, enableAnimation, instanceId, dotsRef, v2Document
+    })
+  }
+
+  useEffect(() => {
+    graphController.current = getNewGraphController()
+  },[dataset, layout, instanceId, v2Document])
```
kswenson, line 49: I suggest moving all of this into a custom hook, `useGraphController`, which takes as arguments the things it needs from this function (`graphModel`, `enableAnimation`, `dotsRef`, `v2Document`), recovers others from context where possible (`dataset`, `layout`, `instanceId`), and then returns the `graphControllerRef`.

v3/src/components/graph/components/graph-component.tsx
```diff
@@ -58,10 +63,10 @@ export const GraphComponent = observer(({broker, v2Document}: IProps) => {
       <InstanceIdContext.Provider value={instanceId}>
         <GraphLayoutContext.Provider value={layout}>
           <Graph model={defaultGraphModel}
-                 graphController={graphController}
-                 graphRef={graphRef}
-                 enableAnimation={enableAnimation}
-                 dotsRef={dotsRef}
+            graphController={graphController.current || getNewGraphController()}
```
kswenson, line 66:
```suggestion
            graphController={graphController.current}
```
Rather than having two different places to create the graph controller, modify the `Graph` component to handle an `undefined` graph controller.

v3/src/components/graph/models/graph-controller.ts
```diff
@@ -62,22 +62,25 @@ export class GraphController {
   initializeGraph() {
     const {graphModel, layout, dotsRef, enableAnimation, instanceId, v2Document} = this,
       dataConfig = graphModel.config
+
     if (v2Document) {
       this.processV2Document()
     } else {
-      graphModel.setGraphProperties({
-        axes: {bottom: EmptyAxisModel.create({place: 'bottom'}),
+      if (!dotsRef.current){
```
kswenson, line 69: combine this with line above:
```suggestion
      } else if (!dotsRef.current) {
```

v3/src/components/graph/models/graph-controller.ts
```diff
@@ -62,22 +62,25 @@ export class GraphController {
...
+        graphModel.setGraphProperties({
+          axes: {bottom: EmptyAxisModel.create({place: 'bottom'}),
           left: EmptyAxisModel.create({place: 'left'})}, plotType: 'casePlot'
-      })
+        })
+      }
+      else if (dotsRef.current) {
```
kswenson, line 75:
```suggestion
      else {
```
don't need the condition

v3/src/components/graph/models/graph-controller.ts
```diff
@@ -62,22 +62,25 @@ export class GraphController {
...
     const {graphModel, layout, dotsRef, enableAnimation, instanceId, v2Document} = this,
       dataConfig = graphModel.config
+
     if (v2Document) {
       this.processV2Document()
     } else {
-      graphModel.setGraphProperties({
-        axes: {bottom: EmptyAxisModel.create({place: 'bottom'}),
+      if (!dotsRef.current){
+        graphModel.setGraphProperties({
```
kswenson, line 70: The previous code always called `graphModel.setGraphProperties` whenever we created a `GraphController` without a `v2Document`. I'm not sure what the implications of calling `graphModel.setGraphProperties` conditionally like this might be, so I'll leave that for @bfinzer to determine.

kswenson (changes requested):
👍 Nice job tracking down the root of the problem! I've suggested some ways to improve the organization of the code. Functionally, my one concern is the conditional call to `graphModel.setGraphProperties()`, on which @bfinzer should be able to offer some insight.

v3/src/components/graph/models/graph-controller.ts
kswenson: It's also not clear to me whether `!dotsRef.current` is the right condition to be testing here (assuming the conditional call is appropriate).

v3/src/components/graph/models/graph-controller.ts
bacalj: If I do this TypeScript complains `Property 'processV2Document' does not exist on type 'GraphController'.ts(2339)`

v3/src/components/graph/models/graph-controller.ts
bacalj: I found that once I put the instantiation in a ref (to avoid the original warning), when I let it attempt to `setGraphProperties` without `dotsRef.current`, I would end up with a repeating list of errors for each attempt at each axis:


>```Error: [mobx-state-tree] You are trying to read or write to an object that is no longer part of a state tree. (Object type: 'EmptyAxisModel', Path upon death: '/axes/left', Subpath: 'type', Action: ''). Either detach nodes first, or don't use objects after removing / replacing them in the tree.```

v3/src/components/graph/models/graph-controller.ts
kswenson: When looking at the GitHub differences I thought my suggestion was equivalent, but now I see that it's not, so forget what I said. 😉

v3/src/components/graph/components/graph.tsx
```diff
@@ -29,7 +29,7 @@ import {getDragAttributeId, IDropData} from "../../../hooks/use-drag-drop"
 import "./graph.scss"
 
 interface IProps {
-  graphController: GraphController
+  graphController: GraphController | undefined // TODO, better approach?
```
bacalj, line 32: @kswenson this. surprisingly, "works" (along with two `graphController?`s below)...but doesn't look quite right. I guess we need to work out how much of a `Graph` should or should not exist without a `GraphController`?

bfinzer (approved):
Nothing to add to Kirk's comments

v3/src/components/graph/components/graph-component.tsx
bacalj: @kswenson in the latest iteration I moved things into a hook, but was only able to get `instanceId` from context without introducing new errors.

v3/src/components/graph/components/graph.tsx
kswenson: See code review commit.

v3/src/components/graph/components/graph-component.tsx
kswenson: I added a commit which implements what I had in mind.

kswenson (approved):
👍 Looks good -- rather than do another round of requesting changes and re-review I went ahead and added a commit which implements what I had in mind. Take a look and see that it makes sense and double-check that I haven't broken anything before merging.

PR #512, opened 2022-10-13
V3—Graph Legend

bfinzer:
When a categorical legend attribute has been assigned to a graph, the points are colored according to the categories in that attribute.
* We do not yet give a "key" to the legend colors
* Only works with categorical attribute, not numeric attribute
* Note that cases are sorted so that the points of each color are grouped together

### DIFF
--- a/v3/src/data-model/attribute.ts
+++ b/v3/src/data-model/attribute.ts
@@ -31,7 +31,7 @@
   more like `math.js` but that decision can be left for another day.
  */
 
-import { Instance, SnapshotIn, types } from "mobx-state-tree"
+import {IAnyType, Instance, SnapshotIn, types} from "mobx-state-tree"
 import { Formula } from "./formula"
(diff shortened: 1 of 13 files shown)


v3/src/data-model/attribute.ts
```diff
@@ -31,7 +31,7 @@
   more like `math.js` but that decision can be left for another day.
  */
 
-import { Instance, SnapshotIn, types } from "mobx-state-tree"
+import {IAnyType, Instance, SnapshotIn, types} from "mobx-state-tree"
```
kswenson, line 34:
```suggestion
import {Instance, SnapshotIn, types} from "mobx-state-tree"
```
Apparently, `IAnyType` is unused.

kswenson (changes requested):
👍 Looks good -- just a few comments, the most important of which is the handling of the `cases` in `ChartDots`.

PR #513, opened 2022-10-13
V3: Fix delete cases item in column header menu

kswenson:
PT: [[#183533068]](https://www.pivotaltracker.com/story/show/183533068)
Demo: https://codap3.concord.org/branch/delete-cases/?sample=mammals


eireland (approved):
👍

PR #515, opened 2022-10-14
V3—Dropping a categorical attribute in middle of plot colors points according to the attribute's categories

bfinzer:
Previously the only way to get a graph with a legend attribute assigned to it was to import from a V2 document. With this commit users can drag a categorical attribute into the middle of a plot and the points will be colored according to its categories.

### DIFF
--- a/v3/src/components/graph/components/graph.tsx
+++ b/v3/src/components/graph/components/graph.tsx
@@ -16,7 +16,7 @@ import {Marquee} from "./marquee"
 import {DataConfigurationContext} from "../hooks/use-data-configuration-context"
 import {useDataSetContext} from "../../../hooks/use-data-set-context"
 import {useGraphModel} from "../hooks/use-graph-model"
-import {AxisPlace, IAxisModel, attrPlaceToAxisPlace} from "../models/axis-model"
+import {AxisPlace, IAxisModel, attrPlaceToAxisPlace, axisPlaceToAttrPlace} from "../models/axis-model"
 import {useGraphLayoutContext} from "../models/graph-layout"
 import {IGraphModel, isSetAttributeIDAction} from "../models/graph-model"
 import {useInstanceIdContext} from "../../../hooks/use-instance-id-context"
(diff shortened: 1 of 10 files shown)


bacalj (approved):
Interesting to see what the plot drop is for, at least so far! Reading through, I am working to understand better the notion of "place".  At first I thought it was the human-notion of a "spot-on-the-ui" independent of the role of the thing in that spot in the graph (e.g. an axis or a legend), which can take various forms and be in various locations, Then the `axisPlaceToAttrPlace` and `attrPlaceToAxisPlace` allow us to pass data from one paradigm to the other? But I don't know if that's entirely correct and/or missing something.   Anyway, I don't believe my approval is the key approval here but I'm clicking approve.

bfinzer:
Good observations about the ambiguity between AttributePlace and AxisPlace. The former might better be named AttributeRole. We've tried to disambiguate them in the past, but failed. I'll put a story in the `Coding Quandaries` epic to have another stab at it.

v3/src/components/graph/components/graph.tsx
```diff
@@ -44,103 +44,103 @@ const marqueeState = new MarqueeState(),
...
-        const [, caseID] = target.property('id').split("_"),
-          attrIDs = graphModel.config.uniqueTipAttributes,
-          tipText = getPointTipText(dataset, caseID, attrIDs)
-          tipText !== '' && dataTip.show(tipText, event.target)
+  }, [layout.plotHeight, layout.plotWidth, margin.left, xScale])
+
+  const toast = useToast()
+
+  const handleDropAttribute = (place: AxisPlace, attrId: string) => {
+    // TODO: will need to be more sophisticated
```
kswenson, line 82:
```suggestion
```
Can remove this comment now

v3/src/components/graph/models/axis-model.ts
```diff
@@ -1,22 +1,25 @@
 import {Instance, types} from "mobx-state-tree"
 import {GraphAttrPlace} from "./data-configuration-model"
 
-export const AxisPlaces = ["bottom", "left", "right", "top"] as const
+export const AxisPlaces = ["bottom", "left", "right", "top", "plot", "legend"] as const
```
kswenson, line 4: When we last had this conversation, I thought we had settled on the notion that `AxisPlaces` were places where axes could be and `GraphAttrPlaces` were places/roles that attributes could be. Now it seems like the ideas are getting muddled again.

For instance, `GraphModel` contains
```
    // keys are AxisPlaces
    axes: types.map(types.maybe(AxisModelUnion)),
```
so this change now implies that the graph has `plot` and `legend` axes. Maybe we need a new notion like
```
  export const GraphPlaces = [...AxisPlaces, "plot", "legend"]
```

kswenson (approved):
👍 Looks good -- I made several minor comments about things like `@ts-expect-error` that are easy to fix. I think the notion of places should be clarified at some point but that need not hold up this PR if you'd rather deal with it later.

v3/src/components/graph/models/axis-model.ts
bfinzer: I followed your recommendation for GraphPlaces and also renamed PrimaryAttrPlaces to PrimaryAttrRoles, TipAttrPlaces to TipAttrRoles, GraphAttrPlaces to GraphAttrRoles, and GraphAttrPlace to GraphAttrRole

PR #516, opened 2022-10-14
V3—Numeric axis snap to zero behavior

bfinzer:
* When a numeric axis is rescaled by dragging, the zero of the scale will snap to the lower or upper bound when the zero is sufficiently close to that boundary.


bacalj (approved):
👍 Looks good - I tested it and it works nicely with axes with intervals in the 1000s as well as those with intervals in the .1s

PR #518, opened 2022-10-19
V3—Scatterplot grid lines

bfinzer:
When horizontal and vertical axes are numeric, grid lines extend across the plot from each axis. The gridlines located at zero are darker than the other grid lines.


v3/src/components/graph/models/graph-layout.ts
```diff
@@ -56,7 +56,14 @@ export class GraphLayout {
...
   @action setAxisBounds(place: AxisPlace, bounds: Bounds | undefined) {
     if (bounds) {
-      this.axisBounds.set(place, bounds)
+      // Don't let the axis bounds overlap the plot
+      const newBounds = {
+        left: bounds.left,
+        top: place === 'left' ? bounds.top : this.axisLength('left'),
+        width: place === 'left' ? this.graphWidth - this.axisLength('bottom') : bounds.width,
+        height: place === 'left' ? bounds.height : this.graphHeight - this.axisLength('left')
+      }
```
kswenson, line 65: As we discussed on sync review, this seems inelegant. Your intuition that it would make more sense to render these on the background makes sense to me, but given that it didn't work as expected when you tried I'm okay with leaving it for now. I would suggest making the changes we talked about to handle all four axes and it would be good to add a comment about why this is here and what happened when you tried to render the gridlines on the background in case we even decide to revisit the situation.

kswenson (approved):
👍 Looks good -- just a couple of comments, the most important of which we discussed in sync review.

PR #520, opened 2022-10-20
V3: Initial support for localization

kswenson:
PT: [[#183444886]](https://www.pivotaltracker.com/story/show/183444886)
Demo (English): https://codap3.concord.org/branch/localization/?sample=mammals
Demo (Spanish): https://codap3.concord.org/branch/localization/?lang=es&sample=mammals

- scripts for pulling strings are copied to v3 folder and updated
- v2 strings files are imported from POEditor into v3 folder
- master v2 strings file (`en-US.json`) is copied into v3 folder (as `en-US.json5` to preserve comments)
- webpack and jest are configured to support `.json5` files
- V3-only strings (e.g. summary view) are added to `en-US.json5` and `es.json` (at least for now)
- only summary/info views translated for now; table and graph left for others

Assuming we want to continue to use the existing POEditor project, we'll need to figure out how we want to handle adding/changing strings. Currently, there are two copies of the master `en-US.json[5]` file, but ultimately there should be a single source of truth.


bfinzer (approved):
👍🏻LGTM as in our synchronous review

PR #521, opened 2022-10-20
V3—Point and case synchronization

bfinzer:
When something about a graph changes, we almost always have to refresh the point positions. (Case selection is an exception.) But there are two arrays that must be kept in synch:
1. The data configuration's array of cases that have the values needed to position the points in the plot.
2. D3s set of circle elements, each of which carries a caseID.

But calls to refresh point positions can be triggered a number of places and can happen _before_ the synchronization has had a chance to happen. This leads to incorrect displays of points and/or attempts to display points that have invalid coordinates.

It also happens that some user-initiated changes, such as dragging an attribute to an axis, trigger multiple calls to refresh the point positions. We don't want to go through the refresh loop multiple times when once is enough.

So, we "debounce" the calls. This amounts to accumulating the calls, waiting a small amount of time, and then calling refreshPointPositions just once.

### DIFF
--- a/v3/src/components/graph/hooks/graph-hooks.ts
+++ b/v3/src/components/graph/hooks/graph-hooks.ts
@@ -1,7 +1,6 @@
 /**
  * Graph Custom Hooks
  */
-import {debounce} from "lodash"
 import {useCallback, useEffect, useRef} from "react"
 import {reaction} from "mobx"
 import {onAction} from "mobx-state-tree"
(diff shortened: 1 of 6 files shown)


v3/src/components/graph/hooks/graph-hooks.ts
```diff
@@ -50,29 +51,47 @@ export const usePlotResponders = (props: IPlotResponderProps) => {
...
+    timer.current = setTimeout(() => {
+      refreshPointsRef.current(selectedOnly)
+      timer.current = null
+    }, 10)
+  }, [refreshPointsRef])
+
   // respond to axis domain changes (e.g. axis dragging)
   useEffect(() => {
-    refreshPointsRef.current(false)
+    callRefreshPointPositions(false)
```
kswenson, line 74:
```suggestion
```
Preexisting condition, but I think you can eliminate the redundant call here by passing `{ fireImmediately: true }` to the `reaction`.

v3/src/components/graph/hooks/graph-hooks.ts
```diff
@@ -50,29 +51,47 @@ export const usePlotResponders = (props: IPlotResponderProps) => {
...
       }
     )
     return () => disposer()
-  }, [refreshPointsRef, xNumeric?.domain, yNumeric?.domain])
+  }, [callRefreshPointPositions, xNumeric?.domain, yNumeric?.domain])
 
   // respond to axis range changes (e.g. component resizing)
   useEffect(() => {
-    refreshPointsRef.current(false)
+    callRefreshPointPositions(false)
```
kswenson, line 86:
```suggestion
```
You should be able to remove this call entirely. The previous `useEffect` is going to `callRefreshPointPositions` on the initial render, so this `useEffect` shouldn't have to.

v3/src/components/graph/hooks/graph-hooks.ts
```diff
@@ -50,29 +51,47 @@ export const usePlotResponders = (props: IPlotResponderProps) => {
...
+  * have been synched with the data configuration's notion of which cases are plottable. Delaying the actual
+  * plotting of points until the next event cycle ensures that the data configuration's filter process will
+  * have had a chance to take place. */
+  const timer = useRef<any>()
+  const callRefreshPointPositions = useCallback((selectedOnly:boolean) => {
+    if( timer.current) {
+      return
+    }
+    timer.current = setTimeout(() => {
+      refreshPointsRef.current(selectedOnly)
```
kswenson, line 67: We'll need to make sure that this call doesn't happen after the ref has been destroyed. I think the way to do that is to add a `useEffect` to handle cleanup:
```typescript
    useEffect(() => {
      return () => if (timer.current) clearTimeout(timer.current)
    }, [])

kswenson (changes requested):
👍 Looks good -- just a couple of minor changes.

v3/src/components/graph/hooks/graph-hooks.ts
bfinzer: Done.

v3/src/components/graph/hooks/graph-hooks.ts
bfinzer: Done

kswenson (approved):
👍 Changes look good

PR #522, opened 2022-10-21
183559169 attr drag hints

bacalj:
Starting a draft pull request for this - all in place except for vertical placement of hint in cases where drag is to x-axis, plus I have some questions which I'll put in comments on the PR


v3/src/components/graph/components/droppable-svg.scss
```diff
@@ -11,3 +11,33 @@
     background: yellow;
   }
 }
+
+.drop-hint {
+  position:absolute;
+  top: 10px;
+  left: 0px; // calculated on fly
+  z-index:500;
+  background-color:#5b616b; // NOTE: http://web-accessibility.carnegiemuseums.org/design/color/
```
bacalj, line 20: This does not match the original exactly - I thought the contrast looked a little to subtle for accessibility purposes, so found a combination that was close but had enough contrast (according to the source, obviously need to verify)

v3/src/components/graph/components/graph-drop-hint.scss
```diff
@@ -0,0 +1,25 @@
+.drop-hint {
```
bacalj, line 1: This should be `.graph-drop-hint` here and in the jsx

v3/src/components/graph/components/graph-drop-hint.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+  useLayoutEffect(() => {
+    if (hintDiv.current){
+      setHintFoundW(hintDiv.current?.offsetWidth)
+      setHintFoundH(hintDiv.current?.offsetHeight)
+    }
+  }, [hintDiv]);
+
+  function getLeftVal(wrapW: number, hintW: number){
+    return (wrapW * .5) - (hintW * .25)
```
bacalj, line 26: This calculation - it appears to work at the moment, but I am suspicious.  Shouldn't it be this: https://codepen.io/joe_bacal/pen/xxjvpVK

v3/src/components/graph/components/graph-drop-hint.tsx
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+
+  function getTopVal(wrapH: number, hintH: number){
+    if (place && place === "bottom") {
+      return 300 // TODO replace w/ a calc based on graph height or mouse position
```
bacalj, line 32: I will either need to use the mouse position - looks like V2 does that - or bring in graphHeight as a prop and calculate based on that for all cases.

v3/src/hooks/use-drop-hint.ts
```diff
@@ -0,0 +1,76 @@
...
+      numeric: {
+        empty: "addToEmptyPlace",
+        existing: "replaceAttribute"
+      },
+      categorical: {
+        empty: "layoutPlotsVertically",
+        existing: "layoutPlotsVertically"
+      }
+    },
+    // in V2 a drop on empty plot puts attr on X axis ("addToEmptyX")
```
bacalj, line 34: Noticed slightly different behavior in v3 from v2 at the moment - not sure this is meant to be, or if the exact same behavior is expected to be in place down the road.  As noted:

v2 - first drop on empty plot puts the attr on the x axis
v3 - first drop on empty plot puts the attr in the legend

v3/src/hooks/use-drop-hint.ts
```diff
@@ -0,0 +1,76 @@
...
+      priorAttrType = dataConfiguration?.attributeType(thisAxisPlace),
+      priorAttrName = dataSet?.attrFromID(priorAttrId as string)?.name,
+      attrId = active.data.current.attributeId,
+      attrName = dataSet.attrFromID(attrId).name,
+      attrType = dataSet.attrFromID(attrId).type as string,
+      details = priorAttrId ? [priorAttrName, attrName] : [attrName],
+      baseString = determineBaseString(placeString, attrType, priorAttrType),
+      hintText = getStringWithSwaps(baseString, "%@", details);
+    return hintText
+  } //TODO ? handle incompatible state?
```
bacalj, line 75: I was expecting the compiler to have a problem with the fact that this might not return anything?

v3/src/utilities/tokened-string.ts
```diff
@@ -0,0 +1,22 @@
+/* This file is a temporary measure until we implement a more robust localized string assembly system */
```
bacalj, line 1: I imagine we'll have something else doing this job (replacing words in a string template with variable number of identical tokens)

bacalj (commented):
Left questions as comments on the PR

v3/src/components/graph/components/droppable-svg.scss
bfinzer: I agree. The contrast is better.

v3/src/components/graph/components/graph-drop-hint.tsx
bfinzer: Mouse position is preferred since the axis can be quite wide and we don't want the hint appearing far from where the user's attention is focussed.

v3/src/hooks/use-drop-hint.ts
bfinzer: Yes, down the road, in fact soon, if there are no attributes on the x or y axes, the drop in the plot area will go to the x-axis. (Nice observation!)

v3/src/components/graph/components/graph-drop-hint.tsx
bacalj: OK, perfect

kswenson (commented):
Since you haven't requested review yet, I'm just leaving these suggestions as comments.

v3/src/hooks/use-drop-hint.ts
bacalj: makes sense, great

v3/src/utilities/tokened-string.ts
bacalj: Ok,  but -- unless I misunderstand  -- for now I'll leave it in as it is doing the job that may be done down the road by something else.

kswenson: My point is that the _something else_ is already on `main`.

v3/src/components/graph/components/graph-drop-hint.tsx
bacalj: ended up using starting coordinates + deltas to get mouse position and set hint

v3/src/components/graph/components/graph-drop-hint.tsx
```diff
@@ -0,0 +1,45 @@
+import React, { useLayoutEffect, useRef, useState} from "react"
+import { AxisPlace } from "../models/axis-model"
+
+import "./graph-drop-hint.scss"
+
+interface IProps {
+  hintText: string | undefined
+  targetW: number | undefined
+  targetH: number | undefined
```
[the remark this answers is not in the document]
bacalj: I ended up only passing hintTest, everything else is calculated within the hook

v3/src/components/graph/components/graph-drop-hint.tsx
bacalj: now entirely using mouse position

bacalj:
Everything working!

PR #525, opened 2022-10-21
183559592 V3: unique attribute names

eireland:
Checks if attribute name already exists in the dataset and adds a number suffix to the name if it does exist.
It is case sensitive so `mass` != `Mass`, which is the same behavior in V2


bfinzer (approved):
👍🏻LGTM
One minor "remove comment" suggestion

v3/src/components/case-table/column-header.tsx
```diff
@@ -44,7 +44,6 @@ export const ColumnHeader = ({ column }: Pick<THeaderRendererProps, "column">) =
       case "Enter":
       case "Tab":
         handleClose(true)
-        e.currentTarget.blur()
```
kswenson, line 47: I'm curious about this change, which seems somewhat unrelated. Happy to see it go, but do you have some deeper understanding that led to its removal?

v3/src/data-model/data-set.test.ts
```diff
@@ -124,6 +124,18 @@ test("DataSet basic functionality", () => {
   expect(dataset.attributes[0].name).toBe("str2")
   dataset.setAttributeName(strAttrID, "str")
   expect(dataset.attributes[0].name).toBe("str")
+
+  // unique at tribute name
```
kswenson, line 128:
```
// unique attribute name
```

v3/src/data-model/data-set.ts
```diff
@@ -268,6 +268,18 @@ export const DataSet = types.model("DataSet", {
     }
   }
 
+  function getUniqueAttributeName(baseName: string, allowNames: (string | null)[]) {
```
kswenson, line 271: There is already a `uniqueName()` function in `js-utils` that I think could be used to accomplish this. You'd just need to pass an `isValid` function that looked something like:
```
(name: string) => (name === existingName) || !self.attributes.find(attr => name === attr.name)
```

v3/src/data-model/data-set.ts
```diff
@@ -433,9 +445,11 @@ export const DataSet = types.model("DataSet", {
       setAttributeName(attributeID: string, name: string) {
```
kswenson, line 445: What if this were changed so that the `setAttributeName()` function took either a `string` or a function which returned a string, i.e.
```
  setAttributeName(attributeID: string, name: string | (() => string)) {
    ...
```
and then clients that want to enforce uniqueness can do so by passing something like:
```
  data.setAttributeName(attrId, () => uniqueName(name, ...))
```
and pass an appropriate `isValid()` function?

kswenson (changes requested):
👍 Functionality generally looks good. I made one suggestion about the implementation of the uniqueness constraint. I have a more nebulous concern that I'm not sure how I feel about in this case that I well mention for explication purposes. The long-term goal is to have the `DataSet` be a sharable module that can be used in other CC projects and maybe even in non-CC projects. Toward that end, it is important that the `DataSet` _not_ take on any CODAP-specific behavior or functionality. So the question then becomes, is guaranteeing attribute uniqueness in this particular way a general `DataSet` behavior that should be mandated for all potential clients or is it a CODAP-specific behavior that should be implemented via an external utility function, for instance? I made a suggestion below of how to leave more of that decision to clients.

v3/src/components/case-table/column-header.tsx
eireland: It was part of my experimentation with the Enter key that should have been removed when I added `e.stopPropagation()`

kswenson (approved):
👍 Changes look good

PR #526, opened 2022-10-21
V3—Graph responds properly to csv and codap drops

bfinzer:
This used to work, presumably because there were redundancies in the code that caused the display of points to happen. Improvements removed a redundancy that was being relied on. So now we have to be more aware of what is changing when a csv or codap file is dragged in.

### DIFF
--- a/v3/src/components/graph/components/axis.tsx
+++ b/v3/src/components/graph/components/axis.tsx
@@ -7,32 +7,35 @@ import {useDataSetContext} from "../../../hooks/use-data-set-context"
 import {getDragAttributeId, IDropData} from "../../../hooks/use-drag-drop"
 import {useInstanceIdContext} from "../../../hooks/use-instance-id-context"
 import {useAxis} from "../hooks/use-axis"
-import {AxisPlace, IAxisModel, INumericAxisModel} from "../models/axis-model"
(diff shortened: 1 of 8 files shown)


kswenson (changes requested):
👍 Looks good overall -- I think it's worth trying to avoid the explicit dependence of `Axis` on `GraphModel`. See below for a suggestion of how to do that.

v3/src/components/graph/components/axis.tsx
```diff
@@ -7,27 +7,26 @@ import {useDataSetContext} from "../../../hooks/use-data-set-context"
...
+import {AxisPlace, IAxisModel, INumericAxisModel} from "../models/axis-model"
 import {useGraphLayoutContext} from "../models/graph-layout"
 import {AxisDragRects} from "./axis-drag-rects"
-import {IGraphModel} from "../models/graph-model"
 
 import "./axis.scss"
 
 interface IProps {
-  graphModel: IGraphModel
+  getAxisModel: any // todo: Figure out how to declare this properly
```
kswenson, line 17: I was thinking it could just be `() => IAxisModel` and then clients could pass something like `() => graphModel.getAxisModel(place)` and then `place` wouldn't need to be passed explicitly, either.

PR #527, opened 2022-10-22
#183559661 V3: Columns respect attribute editable property

eireland:
Allows user to specify whether an attribute is editable or not. Currently, editable is defaulted to true.


bfinzer (approved):
👍🏻LGTM

kswenson (approved):
👍 LGTM

PR #530, opened 2022-10-24
V3: Use `DataConfiguration`'s `onAction` method rather than MST's

kswenson:


bfinzer (approved):
👍🏻LGTM
(But I don't pretend to fully appreciate the subtlety. 🤔)

PR #531, opened 2022-10-25
Update dependencies, notably react-data-grid

kswenson:
In `react-data-grid` version `beta.17` they changed the way their styles are defined which required some changes to the way we include their styles as well as a few tweaks to our styles to adjust priorities between ReactDataGrid and Chakra.

@eireland Assigning to you for review, mainly to see if you can see any other styles that might have been adversely affected given that you've spent a fair amount of time with the case table at this point.


eireland (approved):
👍

PR #534, opened 2022-10-28
V3—Attribute dropped in empty plot puts attribute on x-axis

bfinzer:
* Use new property of data configuration to detect noAttributesAssigned
* Changed drop hint droppable-plot
* Changed behavior in graph component


kswenson (approved):
👍 Looks good!

PR #535, opened 2022-10-29
Translate variables

eireland:
Adds translation variables to case table menus, modals and hints, and graph hints


kswenson (approved):
👍 LGTM

PR #537, opened 2022-10-31
reimpose size limitations on drop zones to adjust for gridlines

bacalj:
This makes the necessary recalculations to keep axis drop zones in the correct positions when a graph has no attributes, one attribute, or two attributes, and re-limits location once gridlines are present.  It replaces a few inline conditionals with a simplified setup so that we can have one set of decisions/recalculations for the left axis and another for the bottom.  

I did also try just mutating the `bounds` prop in place, but stuck with the creation of the `newBounds` for clarity.  Two more comments in-line.

### DIFF
--- a/v3/src/components/graph/models/graph-layout.ts
+++ b/v3/src/components/graph/models/graph-layout.ts
@@ -1,3 +1,4 @@
+import { ThemeProvider } from "@emotion/react"
 import {ScaleBand, ScaleContinuousNumeric, ScaleOrdinal, scaleOrdinal} from "d3"
 import { action, computed, makeObservable, observable } from "mobx"
 import { createContext, useContext } from "react"
@@ -63,18 +64,18 @@ export class GraphLayout {
       // the plot. We tried work arounds to get gridlines that were _not_ part of the axis element with the result
       // that the gridlines got out of synch with axis tick marks during drag. So we have this inelegant solution
       // that shouldn't affect the top and right axes when we get them but
-      // todo: check to make sure this still works with top and right axes
-      const newBounds = {
-        left: bounds.left,
-        top: place === 'bottom' ?
-          Math.min(bounds.top, this.axisLength('left')) : bounds.top,
-        width: place === 'left' ?
-          Math.min(bounds.width, this.graphWidth - this.axisLength('bottom')) : bounds.width,
-        height: place === 'bottom' ?
-          Math.min(bounds.height, this.graphHeight - this.axisLength('left') - this.legendHeight) :
-          place === 'left' ?
-            Math.min(bounds.height, this.graphHeight - this.legendHeight) : bounds.height
+
+      // given state of the graph, we may need to adjust the drop areas' bounds
+      const newBounds = bounds
+
+      if (place === "bottom"){
+        newBounds.top = this.plotHeight
       }
+
+      if (place === "left" && bounds.width > this.plotWidth){
+         newBounds.width -= this.plotWidth
+      }
+
       this.axisBounds.set(place, newBounds)
     }
     else {
--- a/v3/src/components/graph/models/graph-layout.ts
+++ b/v3/src/components/graph/models/graph-layout.ts
@@ -69,11 +69,13 @@ export class GraphLayout {
       const newBounds = bounds
 
       if (place === "bottom"){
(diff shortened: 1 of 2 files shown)


v3/src/components/graph/models/graph-layout.ts
```diff
@@ -63,18 +64,18 @@ export class GraphLayout {
       // the plot. We tried work arounds to get gridlines that were _not_ part of the axis element with the result
       // that the gridlines got out of synch with axis tick marks during drag. So we have this inelegant solution
       // that shouldn't affect the top and right axes when we get them but
```
bacalj, line 66: I did try to find conditions where the calculations that I removed were still necessary, but could not find them.  If I have missed something, or omitted a calculation that anticipated future changes I can put them back in.  I found it easier to reason about by just including the necessary calculations and separating them by axis.  I imagine that if we were to need to add the other calculations down the line it might look more like this: 

```ts
      if (place === "bottom"){
        newBounds.top = this.plotHeight
        newBounds.height = // some calculation
        newBounds.width = // some calculation
      }
```

I see that the width of the axes needs to stay flexible, so for that reason intentionally let the y axis drop zone overlap with the table next door when the words are long enough.  

Anyway, let me know what you think.

bacalj: I'll add that each of these are basically doing the same thing, re-limiting the size of the dropzone so that it does not follow the size of the svg when that svg extends into the plot because of gridlines.  In the first case, it is not always required, but always valid.  In the second case, it is only valid when the gridlines are on.

v3/src/components/graph/models/graph-layout.ts
```diff
@@ -1,3 +1,4 @@
+import { ThemeProvider } from "@emotion/react"
```
bacalj: this import snuck in but I have just removed it

v3/src/components/graph/models/graph-layout.ts
```diff
@@ -63,18 +63,18 @@ export class GraphLayout {
       // the plot. We tried work arounds to get gridlines that were _not_ part of the axis element with the result
       // that the gridlines got out of synch with axis tick marks during drag. So we have this inelegant solution
       // that shouldn't affect the top and right axes when we get them but
```
kswenson, line 65:
```
      // that the gridlines got out of synch with axis tick marks during drag, so we have this inelegant solution.
      // TODO: figure out if there's a better way to render gridlines on background (or plot) so this isn't necessary.

File: v3/src/components/graph/models/graph-layout.ts [hunk 4]
```diff
@@ -63,18 +63,18 @@
...
+
+      if (place === "bottom"){
+        newBounds.top = this.plotHeight
       }
+
+      if (place === "left" && bounds.width > this.plotWidth){
+         newBounds.width -= this.plotWidth
+      }
+
       this.axisBounds.set(place, newBounds)
```
bfinzer, line 79: I entered a reply about the importance of computations for height, but the reply disappeared Along with the conversation thread. :-(

bfinzer (approved):
👍🏻LGTM
But I bet we're going to be revising axis bounds computations in the future, especially as we make the graph layout more responsive.

v3/src/components/graph/models/graph-layout.ts
bacalj: @bfinzer thanks for pointing that out, I should have noticed the legends were mentioned.  I restored them within the new structure.

bacalj:
> I bet we're going to be revising axis bounds computations in the future, especially as we make the graph layout more responsive.

Great - I hope splitting it into these conditionals separated by place will make for clear reasoning/organizing when we get to that code.

v3/src/components/graph/models/graph-layout.ts
bacalj: >  restrict the drop zone to the graph component in all cases
OK I made a change:

```ts
if (place === "left"){
        newBounds.height = Math.min(bounds.height, this.graphHeight - this.legendHeight)
        newBounds.left = 0
        newBounds.width = this.graphWidth - this.plotWidth - this.margin.left - this.margin.right
        // if gridlines are present, axis will grow to .width + plotWidth, so we recalculate
        if (bounds.width > this.plotWidth){
          newBounds.width -= this.plotWidth
        }
      }
```

PR #539, opened 2022-10-31
183598561 axis drop zone pos

bacalj:
This makes two small adjustments. 
1. Adds a test within `use-drop-hint-string.test.ts` so that numeric attribute drops on empty x has differentiated expected behavior from y
2. No longer forces left side axis drop zone to have `bounds.left = 0` - this is likely to be revisited as graph becomes more responsive


bfinzer (approved):
👍🏻LGTM

kswenson (approved):
👍 LGTM

PR #540, opened 2022-11-01
V3—Graph not dealing properly with CODAP doc containing graph with categorical attr on y-axis

bfinzer:
* The code in ChartDots was making erroneous assumptions, assuming that things would default to having a attribute on the x-axis.


kswenson (approved):
👍 LGTM

PR #542, opened 2022-11-01
V3—graph import codap doc bug fix

bfinzer:
On drag of CODAP doc with dotplot plus legend was not being handled properly.

Added line to set the axis scale.


kswenson (approved):
👍 LGTM

PR #545, opened 2022-11-02
V2: Update to CFM 1.5.10

kswenson:
CFM 1.5.10
- fixes large state bug found in AP/CODAP/StoryQ [[#183681726]](https://www.pivotaltracker.com/story/show/183681726)

At one time it was possible to view the code difference between two CFM versions to verify that the expected changes were present, but with the current minification of the CFM code that seems to no longer be possible. Furthermore, since the deploy to `codap-dev` hasn't been working recently, there's no way to test the deployed branch code.

I tested it against my localhost and it was able to open the document in question without errors, so I'm going to assume that we can proceed to a build. 🤞


eireland (approved):
👍

PR #547, opened 2022-11-04
V3—Category set optimization

bfinzer:
This commit fixes a performance problem easily observed by making a graph with a legend using abalone as the dataset. Each time we wanted to determine the color of a point, the category set for the legend variable was having to be recomputed.

Now we cache category sets in the DataConfigurationModel as a volatile property and create it on demand. When Dataset.setCaseValues is called, with an optional array of attribute ids involved, we either clear the entire volatile property or, if the array of attribute ids is present, just the category sets for those attributes.

So far, the only situation in which we pass the array of attribute ids is when points in a scatterplot or dotplot are being dragged.


kswenson (approved):
👍 Looks good -- as mentioned in standup, it's worth considering at some point whether there's a better solution than `as string` casting the attribute ids returned from the data configuration, but I don't know what it would be off the top of my head and don't think we should hold up the PR to figure it out. Separately, I think some of the logic for determining `primary`/`secondary` could be moved into the `DataConfiguration`, but that can be a topic for another day as well.

PR #550, opened 2022-11-08
V3 181912236 axis menu

bacalj:
This pull request replaces #548 

1. The `axis-attribute-menu` is now instantiated on the `Axis` and is renderable "within" the svg area via a `portal`
2. Menu functionality:
 - change attribute - results in changing attribute
 - remove attribute - results in a toast with params
 - treat attribute as - results in a toast with params
3. On first load, bottom axis menu button requires calculation to arrive at the right spot, otherwise they leverage the new `useOverlayBounds` hook
4. Buttons are now fully invisible
5. Since a single `handleChangeAttribute` function is now set up for handling a removal and a change of an attribute, the instantiated `Axis` and `DroppableSvg` now include: ` onDropAttribute={handleChangeAttribute}`

### DIFF
--- a/v3/src/components/graph/components/axis-attribute-menu.tsx
+++ b/v3/src/components/graph/components/axis-attribute-menu.tsx
@@ -26,7 +26,7 @@ export const AxisAttributeMenu = ({ place, target, portal, onChangeAttribute, on
   const overlayBounds = useOverlayBounds({target, portal})
   const buttonStyles: CSSProperties = { position: "absolute", color: "transparent" }
 
-  if (attrId === undefined && place === "bottom"){
+  if (!attrId && place === "bottom"){
     buttonStyles.top = plotHeight + 4
     buttonStyles.left = ( plotWidth * .5 ) - margin.right - 8 // ~ width of y scale
   }
--- a/v3/src/components/graph/components/axis-attribute-menu.tsx
+++ b/v3/src/components/graph/components/axis-attribute-menu.tsx
@@ -18,31 +18,23 @@ interface IProps {
 export const AxisAttributeMenu = ({ place, target, portal, onChangeAttr, onTreatAttrAs }: IProps ) => {
   const data = useDataSetContext()
   const dataConfig = useDataConfigurationContext()
-  const { graphWidth, plotWidth, plotHeight, margin } = useGraphLayoutContext()
+  const { plotWidth, plotHeight, margin } = useGraphLayoutContext()
   const role = place === "left" ? "y" : "x"
   const attrId = dataConfig?.attributeID(role)
   const attribute = attrId ? data?.attrFromID(attrId) : null
   const treatAs = attribute?.type === "numeric" ? "categorical" : "numeric"
   const overlayBounds = useOverlayBounds({target, portal})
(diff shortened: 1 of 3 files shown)


v3/src/components/graph/components/axis-attribute-menu.tsx
```diff
@@ -0,0 +1,67 @@
...
+  const dataConfig = useDataConfigurationContext()
+  const { plotWidth, plotHeight, margin } = useGraphLayoutContext()
+  const role = place === "left" ? "y" : "x"
+  const attrId = dataConfig?.attributeID(role)
+  const attribute = attrId ? data?.attrFromID(attrId) : null
+  const treatAs = attribute?.type === "numeric" ? "categorical" : "numeric"
+  const overlayBounds = useOverlayBounds({target, portal})
+  const buttonStyles: CSSProperties = { position: "absolute", color: "transparent" }
+
+  if (attrId === undefined && place === "bottom"){
```
kswenson, line 29:
```suggestion
  if (!attrId && place === "bottom"){
```
I would use the simpler test here and avoid explicit mention of `undefined`. Any of the other falsy values would likewise be treated as not having an attribute, e.g. empty string, zero, null, etc.

v3/src/components/graph/components/axis.tsx
```diff
@@ -8,10 +8,12 @@ import {getDragAttributeId, IDropData} from "../../../hooks/use-drag-drop"
 import {useDropHintString} from "../../../hooks/use-drop-hint-string"
 import {useInstanceIdContext} from "../../../hooks/use-instance-id-context"
 import {useAxis} from "../hooks/use-axis"
-import {AxisPlace, axisPlaceToAttrRole, IAxisModel, INumericAxisModel} from "../models/axis-model"
+import {AxisPlace, GraphPlace, axisPlaceToAttrRole, IAxisModel, INumericAxisModel} from "../models/axis-model"
 import {useGraphLayoutContext} from "../models/graph-layout"
 import {AxisDragRects} from "./axis-drag-rects"
+import {AxisAttributeMenu} from "./axis-attribute-menu"
 import t from "../../../utilities/translation/translate"
+import { createPortal } from "react-dom"
```
kswenson, line 16: Move import ⬆️ as we generally list third-party/node_modules imports first.

v3/src/components/graph/components/graph.tsx
```diff
@@ -77,18 +77,20 @@ export const Graph = observer((
 
   const toast = useToast()
 
-  const handleDropAttribute = (place: GraphPlace, attrId: string) => {
+  const handleChangeAttribute = (place: GraphPlace, attrId: string ) => {
+    if (attrId === ""){ // when request is to remove the attribute
```
kswenson, line 81:
```suggestion
    if (!attrId){ // when request is to remove the attribute
```
Same reasoning as above -- simpler test and catches any nulls and undefined that might slip through as well.

kswenson (approved):
👍 Looks good! The behavior seems right and the code is much clearer. I made the usual assortment of minor suggestions.

One note for the next PR: your branch name doesn't follow our convention. The PT number must come at the beginning or the end for it to be stripped off when deploying to `branches/{branch-name}`.

PR #552, opened 2022-11-10
183656334 v3 legend attr drop zone

bacalj:
This renders a drop-zone on the legend. 

- It leverages `DroppableSvg` 
- simplifies `useOverlayBounds()`
- It adds props and handlers, and hooks to `Legend` so that it can instantiate a `DroppableSvg`


v3/src/hooks/use-overlay-bounds.ts
```diff
@@ -1,36 +1,19 @@
...
+  const targetBounds = target?.getBoundingClientRect()
+  if (targetBounds) {
+    return {
+      left: targetBounds.x - (portalBounds?.x ?? 0),
+      top: targetBounds.y - (portalBounds?.y ?? 0),
+      width: targetBounds.width,
+      height: targetBounds.height
+    }
+  } else {
+    return {}
```
bacalj, line 17: @kswenson Should I be returning null? Compiler at calls to this hook will be mad if I don't return something, so this seemed the most compatible, but still doesn't look quite right?

bfinzer (approved):
👍🏻LGTM
Very nice!

v3/src/hooks/use-overlay-bounds.ts
```diff
@@ -1,36 +1,19 @@
...
+  }
```
kswenson, line 18: Yes, `null` is the conventional return value to indicate that a React component should not render. You also don't need the else, so you can replace it with:
```suggestion
  }
  return null
```

kswenson (approved):
👍 Looks good

v3/src/hooks/use-overlay-bounds.ts
bacalj: @kswenson it turns out that since we were returning the value into a variable declared as (React's) type `CSSProperties`, it will not accept `null`, but will accept `{}`.  So I changed it back.  

If you think there is a reason to not type the style variable in `droppable-svg` like this: 

```ts const style: CSSProperties = useOverlayBounds({ target, portal })```

or make some other change, let me know.

kswenson: That's fine -- when I made that suggestion I was thinking it was going to be rendered directly. `{}` is the right default for `CSSProperties`.

PR #555, opened 2022-11-15
V3: Deselects row when in cell edit mode

eireland:
Deselects current selected row when user is in cell edit mode.
Note that "Enter" and "Tab" just accepts changes of the currently selected cell and does not move to the next row/cell, which is the behavior in V2. The current behavior is the default behavior in react-data-grid.


kswenson (approved):
👍 LGTM

PR #556, opened 2022-11-15
Remove attribute menu item implementation

bfinzer:
Getting the **Remove Attribute** to behave properly required a few tweaks to deal with the situation in which an attribute ID is not present where it was before.


bacalj (approved):
👍 Looks good to me - I see we have to have the proper empty axis model in there once the attribute is removed.

PR #557, opened 2022-11-15
V3: Updates missing Spanish translation

eireland:


kswenson (commented):
Where did these translations come from? I pulled the latest from POEditor so I would have thought I had the most recent translations.

kswenson (approved):
👍 LGTM, although I wish I understood how my pull from POEditor resulted in different strings even though I'm not aware of any changes since then. 🤷

PR #559, opened 2022-11-17
v3: 181465538 inspector panel

eireland:
Adds a generic inspector panel that will show specific tools depending on the component.
Temporarily:
1. Graph inspector panel is always showing on top of the graph.
2. Table inspector is shown and hidden by clicking anywhere in the case table.
3. None of the buttons do anything at the moment.

Note: Camera icon is not the same as V2.


bfinzer:
I took a brief look at the V2 architecture and verified that as I mentioned the Inspector knows virtually nothing about what a target component will want. It asks the target for its icons and the hooks that will be triggered by each. I like this because, for one thing, it generalizes nicely to plugins.
Just something to keep in mind.

bfinzer (approved):
👍🏻LGTM
I think we'll revisit the architecture question I raise once we start getting serious about each component's inspector.

v3/src/components/graph/components/graph-inspector.tsx
```diff
@@ -0,0 +1,50 @@
...
+}
+
+const InspectorButton = ({type}:IInspectorButtonProps) => {
+  return (
+    <Button className="inspector-tool-button">
+      {ToolIconMap[type]}
+      {type !== "resize" && <MoreOptionsIcon className="more-options-icon"/>}
+    </Button>
+  )
+}
```
kswenson, line 50: You could make the `InspectorButton` reusable as well with something like the following, which gives more control to the client:
```
interface IInspectorButtonProps {
  children: ReactNode
  moreOptions: boolean
}
const InspectorButton = ({ children, moreOptions }: IInspectorButtonProps) => {
  return (
    <Button className="inspector-tool-button">
      {children}
      {moreOptions && <MoreOptionsIcon className="more-options-icon"/>}
    </Button>
  )
}

kswenson (changes requested):
👍 Almost there -- one more minor refactor to eliminate some redundancy.

kswenson (approved):
👍 Never mind my previous review -- I see you've already moved in that direction in a subsequent PR so let's merge this as is.

PR #561, opened 2022-11-18
V3: 181869900 inspector trash can

eireland:
Adds trash menu list
Adds ability to select all cases
Adds ability to delete selected cases
Adds ability to delete unselected cases
Adds ability to delete all cases

If user deletes all cases it adds back a blank row (not quite a ghost row) for now so it doesn't look strange.

### DIFF
--- a/v3/src/components/case-table/inspector-panel/trash-menu-list.tsx
+++ b/v3/src/components/case-table/inspector-panel/trash-menu-list.tsx
@@ -1,0 +1,75 @@
+import { MenuItem, MenuList } from "@chakra-ui/react"
+import React from "react"
+import { useDataSetContext } from "../../../hooks/use-data-set-context"
+import { IAttribute } from "../../../models/data/attribute"
+import t from "../../../utilities/translation/translate"
+
+export const TrashMenuList = () => {
+  const data = useDataSetContext()
+
+  const createGhostCase = (attrs:IAttribute[]) => {
+    const gCase:Record<string, any> = {}
+    attrs?.forEach(attr => {
+      gCase[attr.name]=""
+    })
+    return gCase
+  }
+
+  const handleSelectAllCases = () => {
+    const caseIDArr:string[] = []
+    if (data) {
+      data.cases.forEach(c => {
+        caseIDArr.push(c.__id__)
+      })
+      data.selectCases(caseIDArr)
+    }
+  }
+
+  const handleDeleteSelectedCases = () => {
+    if (data) {
+      data.removeCases(Array.from(data.selection))
+      if (data.cases.length === 0) {
+        const ghostCase = createGhostCase(data.attributes)
+        data.addCases([ghostCase])
+      }
+    }
+  }
+
+   const handleDeleteUnselectedCases = () => {
+    const unselectedCaseIDArr:string[] = []
+
+    if (data) {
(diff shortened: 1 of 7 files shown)


v3/src/components/case-table/inspector-panel/trash-menu-list.tsx
```diff
@@ -0,0 +1,75 @@
...
+    if (data) {
+      data.cases.forEach(c => {
+        caseIDArr.push(c.__id__)
+      })
+      data.selectCases(caseIDArr)
+    }
+  }
+
+  const handleDeleteSelectedCases = () => {
+    if (data) {
```
kswenson, line 29:
```suggestion
    if (data?.selection.size) {
```

v3/src/components/case-table/inspector-panel/trash-menu-list.tsx
```diff
@@ -0,0 +1,75 @@
...
+   const handleDeleteUnselectedCases = () => {
+    const unselectedCaseIDArr:string[] = []
+
+    if (data) {
+      const casesToRemove = data.cases.filter(c=>
+        !((Array.from(data.selection)).includes(c.__id__)))
+      casesToRemove.forEach(c => {
+        unselectedCaseIDArr.push(c.__id__)
+      })
+      data.removeCases(unselectedCaseIDArr)
```
kswenson, line 47:
```suggestion
      const casesToRemove = data.cases.filter(c => !data.selection.has(c.__id__)).map(c => c.__id__);
      data.removeCases(casesToRemove)
```
Converting to `Array` is inefficient and `Array.includes` is less efficient than `Set.has`.

v3/src/components/case-table/inspector-panel/trash-menu-list.tsx
```diff
@@ -0,0 +1,75 @@
...
+export const TrashMenuList = () => {
+  const data = useDataSetContext()
+
+  const createGhostCase = (attrs:IAttribute[]) => {
+    const gCase:Record<string, any> = {}
+    attrs?.forEach(attr => {
+      gCase[attr.name]=""
+    })
+    return gCase
+  }
```
kswenson, line 16: Now may or may not be the time to deal with it, but this whole ghost case business is a kluge. We shouldn't be adding bogus data to the DataSet. Instead, the case table should be modified to deal with an empty DataSet properly. I let it go initially because it seemed expedient when we were just bootstrapping our way to an initial table state, but much of the code in this component just serves to further cement the kluge, so it may be time to address it properly. It should be the case table's job to render the DataSet in its current form rather than to modify the DataSet so it's easier to render.

kswenson (changes requested):
👍 Generally looks good. I like the InspectorMenu abstraction. I made a couple of minor suggestions to simplify some code. The major concern is that this ghost case business is going in the wrong direction. Let's discuss after the break.

v3/src/components/case-table/inspector-panel/trash-menu-list.tsx
kswenson: Rather than going through another round of changes/review I just went ahead and implemented what I had in mind. 😬

kswenson (approved):
👍 Rather than going through another round of changes/review, I just implemented what I had in mind to short-circuit the process.

PR #562, opened 2022-11-18
V3—Treat attribute as numeric of categorical

bfinzer:
* Mostly making use of existing code
* Data configuration holds onto user-set attribute type
* Have to be sure not to wipe out any existing type


bacalj (approved):
Looks good to me - obviously Kirk has a lot more context for any comments, but I'll say that reading through how this "treat as" is implemented helps me get a better sense of how the graph and data configuration model work together.

PR #564, opened 2022-11-21
V3—axes-stuck-183692921

bfinzer:
Dragging a V2 codap document into V3 was leaving the axes unresponsive to dragging. The required workaround fix is, at this point, a mystery. It involves computing in useAxis whether the axis model has changed and using that computed boolean as a dependent in installDomainSync. Hopefully we'll come back around at some point and figure out how to do without the workaround.


kswenson (approved):
👍 LGTM

PR #565, opened 2022-11-23
V3—Graph point animation fixes

bfinzer:
In transitions from one plot to another and in animation of dragged points back to their starting place, some points stop part way through.

* Part of the problem was that as points passed under the mouse they would trigger showing and hiding of the data tip. But the code for each of these contains point transitions of their own which cancel the transition that is in progress. So in show/hide data tip, we detect that the target is "active" (animating) and forgo the animation.

* But that isn't quite enough to fully fix things, so, at the end of an animation, we iterate through the points again *without* animation to guarantee that all points get to their destination.

Another problem I noticed was that the border of selected points when there was no legend the stroke for selected points during and after a drag was colored red with a width of 2 which was incorrect. The fix for this was to only pass to `setPointCoordinates` a function to get a legend color (`getLegendColor`) if there actually is a legend attribute.

### DIFF
--- a/v3/src/components/graph/components/chartdots.tsx
+++ b/v3/src/components/graph/components/chartdots.tsx
@@ -24,14 +24,14 @@ export const ChartDots = memo(function ChartDots(props: IProps) {
     layout = useGraphLayoutContext(),
     xAttribute = dataConfiguration?.attributeID('x'),
     yAttribute = dataConfiguration?.attributeID('y')
-  const  primaryAttrPlace = xAttribute ? 'x' :
+  const primaryAttrPlace = xAttribute ? 'x' :
       yAttribute ? 'y' : undefined,
     primaryAxisPlace = primaryAttrPlace ? attrRoleToAxisPlace[primaryAttrPlace] : undefined,
     primaryIsBottom = primaryAxisPlace === 'bottom',
     primaryAttrID = primaryAttrPlace ? dataConfiguration?.attributeID(primaryAttrPlace) : '',
     secondaryAttrPlace = primaryAttrPlace === 'x' ? 'y' :
       primaryAttrPlace === 'y' ? 'x' : undefined,
-    secondaryAxisPlace = secondaryAttrPlace ? attrRoleToAxisPlace[secondaryAttrPlace] :undefined,
+    secondaryAxisPlace = secondaryAttrPlace ? attrRoleToAxisPlace[secondaryAttrPlace] : undefined,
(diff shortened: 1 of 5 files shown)


v3/src/components/graph/components/chartdots.tsx
```diff
@@ -121,59 +120,67 @@ export const ChartDots = memo(function ChartDots(props: IProps) {
...
+    computeMaxOverAllCells, primaryAttrID, secondaryAttrID, legendAttrID, dataset, secondaryScale])
 
   useEffect(() => {
     select(dotsRef.current).on('click', (event) => {
       const element = select(event.target as SVGSVGElement)
       if (element.node()?.nodeName === 'circle') {
         const tItsID: string = element.property('id')
         const [, caseId] = tItsID.split("_")
-        dataset?.selectCases([caseId])
+        dataset?.setSelectedCases([caseId])
```
kswenson, line 177: This seems to suggest that every click will deselect all other points, regardless of whether the point is already selected (usually clicking on a selected point doesn't deselect other points) or whether any modifiers are down. Seems like some more logic would be required here.

v3/src/components/graph/components/graph.tsx
```diff
@@ -121,12 +122,16 @@ export const Graph = observer((
...
   function hideDataTip(event: MouseEvent) {
-    const [, caseID] = select(event.target as SVGSVGElement).property('id').split("_"),
-      isSelected = dataset?.isCaseSelected(caseID)
+    const target = select(event.target as SVGSVGElement)
     dataTip.hide()
-    select(event.target as SVGSVGElement)
-      .transition().duration(transitionDuration)
-      .attr('r', isSelected ? selectedPointRadius : pointRadius)
+    if (target.node()?.nodeName === 'circle' && dataset && !active(target.node()) &&
+      !target.property('isDragging')) {
```
kswenson, line 128: Since this same test occurs twice (and as this PR demonstrates, occasionally they have to be modified in tandem), it seems like a utility function might be useful.

kswenson (changes requested):
👍 Generally looks good -- just a couple comments/suggestions. Also, there are some lint warnings to address.

v3/src/components/graph/components/chartdots.tsx
bfinzer: Yes, good catch. This slipped into this pull request by mistake. A subsequent pull request will handle this correctly.

v3/src/components/graph/components/graph.tsx
bfinzer: Done!

PR #566, opened 2022-11-23
V3: 181869977-Inspector Hide-Show menu UI

eireland:
Adds hide/show menu list to case table inspector panel.
Adds ability to show hidden attributes


kswenson (approved):
👍 Looks good after a couple of tweaks I went ahead and made myself. 😉

PR #568, opened 2022-11-27
V3—numeric-axis-animation—183689854

bfinzer:
* By passing enableAnimation to the Axis component and then down to useAxis the decision can be made in refreshAxis about whether to animate.

Note that enableAnimation.current is set to true anytime the plot changes. Thus the rule becomes "Axes animate when points animate."


kswenson (approved):
👍 LGTM

PR #571, opened 2022-12-01
181909355 graph inspector brush

eireland:
Adds graph inspector brush palette.
Adds ability to change point size via slider.
All other functionalities have not been implemented.
This also added the Hide/Show (eye) menu.

### DIFF
--- a/v3/src/components/codap-modal.tsx
+++ b/v3/src/components/codap-modal.tsx
@@ -2,7 +2,7 @@ import { Modal,
   ModalOverlay, ModalContent, ModalHeader, ModalCloseButton, ModalBody, ModalFooter,
   Button,
   Tooltip} from "@chakra-ui/react"
-import React, { forwardRef, ReactNode } from "react"
+import React, { forwardRef } from "react"
 
 import "./codap-modal.scss"
 
@@ -21,7 +21,7 @@ interface IProps<TContentProps> {
   onClose: () => void
   className?: string
   title: string
-  Icon?: ReactNode
+  Icon?: React.FC<any>
   Content: React.FC<TContentProps>
   contentProps?: TContentProps | any
   focusElement?: string
@@ -53,7 +53,7 @@ export const CodapModal = forwardRef(<IContentProps,>({ isOpen, onClose,
       <ModalContent ref={ref} className="codap-modal-content" w={contentProps.modalWidth || "400px"}>
         <ModalHeader h="30" className="codap-modal-header" fontSize="md" data-testid="codap-modal-header">
           <div className="codap-modal-icon-container">
-            {Icon}
+            {Icon && <Icon />}
           </div>
           <div className="codap-header-title">{title}</div>
           {hasCloseButton && <ModalCloseButton onClick={onClose} data-testid="modal-close-button"/>}
--- a/v3/src/components/codap-modal.tsx
+++ b/v3/src/components/codap-modal.tsx
@@ -2,7 +2,7 @@ import { Modal,
   ModalOverlay, ModalContent, ModalHeader, ModalCloseButton, ModalBody, ModalFooter,
   Button,
   Tooltip} from "@chakra-ui/react"
-import React, { forwardRef } from "react"
+import React, { forwardRef, ReactNode } from "react"
 
 import "./codap-modal.scss"
 
@@ -21,7 +21,7 @@ interface IProps<TContentProps> {
   onClose: () => void
   className?: string
   title: string
-  Icon?: React.FC<any>
+  Icon?: ReactNode
   Content: React.FC<TContentProps>
   contentProps?: TContentProps | any
   focusElement?: string
@@ -53,7 +53,7 @@ export const CodapModal = forwardRef(<IContentProps,>({ isOpen, onClose,
       <ModalContent ref={ref} className="codap-modal-content" w={contentProps.modalWidth || "400px"}>
         <ModalHeader h="30" className="codap-modal-header" fontSize="md" data-testid="codap-modal-header">
           <div className="codap-modal-icon-container">
-            {Icon && <Icon />}
+            {Icon}
           </div>
           <div className="codap-header-title">{title}</div>
           {hasCloseButton && <ModalCloseButton onClick={onClose} data-testid="modal-close-button"/>}
--- a/v3/src/components/graph/components/graph-component.tsx
+++ b/v3/src/components/graph/components/graph-component.tsx
@@ -33,8 +33,6 @@ export const GraphComponent = observer(({broker}: IProps) => {
   const dataset = broker?.selectedDataSet || broker?.last
   const dotsRef = useRef<SVGSVGElement>(null)
   const [showInspector, setShowInspector] = useState(false)
-  const [showParentToggles, setShowParentToggles] = useState(false)
-  const [showMeasuresForSelection, setShowMeasuresForSelection] = useState(false)
 
   useEffect(() => {
     (width != null) && (height != null) && layout.setGraphExtent(width, height)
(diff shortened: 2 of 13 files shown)


v3/src/components/graph/components/graph-component.tsx
```diff
@@ -33,6 +33,8 @@ export const GraphComponent = observer(({broker}: IProps) => {
   const dataset = broker?.selectedDataSet || broker?.last
   const dotsRef = useRef<SVGSVGElement>(null)
   const [showInspector, setShowInspector] = useState(false)
+  const [showParentToggles, setShowParentToggles] = useState(false)
+  const [showMeasuresForSelection, setShowMeasuresForSelection] = useState(false)
```
bfinzer, line 37: These two properties will be part of the model, right, since they need to be saved and restored?

v3/src/components/graph/components/graph.tsx
```diff
@@ -165,7 +170,13 @@ export const Graph = observer((
           />
         </svg>
       </div>
-      <GraphInspector show={showInspector} />
+      <GraphInspector graphModel={graphModel}
+                      show={showInspector}
+                      showParentToggles={showParentToggles}
+                      setShowParentToggles={setShowParentToggles}
+                      showMeasuresForSelection={showMeasuresForSelection}
+                      setShowMeasuresForSelection={setShowMeasuresForSelection}
```
bfinzer, line 178: Once these are model level properties, the graphModel will be sufficient.

v3/src/components/graph/components/graph-inspector.tsx
```diff
@@ -1,44 +1,68 @@
...
-        <CameraIcon />
-        </InspectorButton>
-      </InspectorPanel>
+    ? <>
+        <InspectorPanel component="graph">
+          <InspectorButton tooltip={"DG.Inspector.resize.toolTip"} showMoreOptions={false}
+            testId={"graph-resize-button"} >
+            <ScaleDataIcon />
+          </InspectorButton>
+          <InspectorMenu tooltip={"DG.Inspector.displayStyles.toolTip"}
```
bfinzer, line 38: I notice that the tooltips are not currently being shown in English.

bfinzer (approved):
👍🏻LGTM
I like it!
My main question is why not add the parent visibility property and show measures property to the graph model now?

v3/src/components/codap-modal.tsx
```diff
@@ -21,7 +21,7 @@ interface IProps<TContentProps> {
   onClose: () => void
   className?: string
   title: string
-  Icon?: React.FC<any>
+  Icon?: ReactNode
```
kswenson, line 24: The changes in this file do not appear to be used in this PR (or anywhere else as near as I can tell). I would recommend leaving them the way they were. I have in mind a refactoring of this code that would provide a better opportunity to make any such changes.

v3/src/components/graph/components/graph-component.tsx
kswenson: I agree with Bill's comment that it would make more sense to put these properties in the`GraphModel` now rather than have them as React state now and move them to the `GraphModel` later.

v3/src/components/graph/components/graph-inspector.tsx
kswenson: It should be
```suggestion
          <InspectorMenu tooltip={t("DG.Inspector.displayStyles.toolTip")}
```
in each case.

v3/src/components/graph/components/inspector-panel/hide-show-menu-list.tsx
```diff
@@ -0,0 +1,52 @@
...
+      description: `You clicked on ${menuItem}`,
+      status: 'success',
+      duration: 5000,
+      isClosable: true,
+    })
+  }
+
+  const hideSelectedString = data?.selectCases && data?.selectCases.length > 1
+                              ? t("DG.DataDisplayMenu.hideSelectedPlural")
+                              : t("DG.DataDisplayMenu.hideSelectedSing")
```
kswenson, line 30: The default in case of no `data` should be the plural:
```suggestion
  const hideSelectedString = (data?.selectCases?.length ?? 0) === 1
                              ? t("DG.DataDisplayMenu.hideSelectedSing")
                              : t("DG.DataDisplayMenu.hideSelectedPlural")
```

v3/src/components/graph/components/inspector-panel/hide-show-menu-list.tsx
```diff
@@ -0,0 +1,52 @@
...
+  const hideUnselectedString = data && (data.cases.length - data.selectCases.length) > 1
+                              ? t("DG.DataDisplayMenu.hideUnselectedPlural")
+                              : t("DG.DataDisplayMenu.hideUnselectedSing")
+  const parentToggleString = showParentToggles ? t("DG.DataDisplayMenu.disableNumberToggle")
+                                               : t("DG.DataDisplayMenu.enableNumberToggle")
+  const measuresForSelectionString = showMeasuresForSelection ? t("DG.DataDisplayMenu.disableMeasuresForSelection")
+                                               : t("DG.DataDisplayMenu.enableMeasuresForSelection")
```
kswenson, line 37: For consistency:
```suggestion
  const measuresForSelectionString = showMeasuresForSelection
                                               ? t("DG.DataDisplayMenu.disableMeasuresForSelection")
                                               : t("DG.DataDisplayMenu.enableMeasuresForSelection")
```

v3/src/components/graph/components/inspector-panel/point-format-panel.tsx
```diff
@@ -0,0 +1,82 @@
...
+import { IGraphModel } from "../../models/graph-model"
+import { InspectorPalette } from "../../../inspector-panel"
+import StylesIcon from "../../../../assets/icons/icon-styles.svg"
+
+import "./point-format-panel.scss"
+
+interface IProps {
+  graphModel: IGraphModel
+  showFormatPalette: boolean
+  setShowFormatPalette: (show: boolean) => void;
```
kswenson, line 14:
```suggestion
  setShowFormatPalette: (show: boolean) => void
```

v3/src/components/inspector-panel.scss
```diff
@@ -50,3 +50,78 @@
     }
   }
 }
+
+.codap-inspector-palette {
+  background-color: #EEE;
+  border: 0px #444 solid;
+  position: absolute;
+  left: 5px !important;
+  inset: 0px;
```
kswenson, line 59: If you reverse the order you don't need the `!important`:
```suggestion
  inset: 0;
  left: 5px;
```

v3/src/components/inspector-panel.scss
```diff
@@ -50,3 +50,78 @@
...
+  max-width: 270px;
+  height: 251px;
```
kswenson, line 61:
```suggestion
```
I think if you eliminate this it will encompass its children -- as is it's not tall enough to do so.

v3/src/components/graph/components/graph-component.tsx
eireland: I was just adding the state management, but don't have to be added right now. They do have to be part of the model.

v3/src/components/graph/components/graph-inspector.tsx
eireland: Good catch! I forgot to pass it through the translate function

v3/src/components/inspector-panel.tsx
```diff
@@ -53,3 +54,43 @@ export const InspectorMenu = ({children, icon, tooltip, testId, onButtonClick}:I
...
+interface IInspectorPalette {
+  children: ReactNode
+  Icon?: ReactNode
+  title?: string
+  paletteTop?: number
+}
+
+export const InspectorPalette =({children, Icon, title, paletteTop}:IInspectorPalette) => {
+  const paletteRef = useRef(null)
+  const [paletteHeight, ] = useState(251)
```
kswenson, line 67: If this is never changed, why is it in state? Also, it doesn't appear to be the correct value given that the same value in the .scss file doesn't appear to encompass all of the content.

kswenson (changes requested):
👍 Looks good -- the usual assortment of comments and suggestions.

kswenson (approved):
👍 Looks good!

PR #572, opened 2022-12-01
V3: Fixes bug to drag attribute from case table after opening attribute menu

eireland:
Passes disabled property specifically into the useDraggableAttribute

### DIFF
--- a/v3/src/hooks/use-drag-drop.ts
+++ b/v3/src/hooks/use-drag-drop.ts
@@ -29,7 +29,7 @@ export interface IUseDraggableAttribute extends Omit<UseDraggableArguments, "id"
   prefix: string
   attributeId: string
 }
-export const useDraggableAttribute = ({ prefix, attributeId, ...others }: IUseDraggableAttribute) => {
+export const useDraggableAttribute = ({ prefix, attributeId, disabled, ...others }: IUseDraggableAttribute) => {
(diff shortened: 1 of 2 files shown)


v3/src/hooks/use-drag-drop.ts
```diff
@@ -29,7 +29,7 @@ export interface IUseDraggableAttribute extends Omit<UseDraggableArguments, "id"
   prefix: string
   attributeId: string
 }
-export const useDraggableAttribute = ({ prefix, attributeId, ...others }: IUseDraggableAttribute) => {
+export const useDraggableAttribute = ({ prefix, attributeId, disabled, ...others }: IUseDraggableAttribute) => {
```
kswenson, line 32: >Passes disabled property specifically into the useDraggableAttribute

This removes the `disabled` property from the props so that it is _never_ passed to the underlying `useDraggable`, i.e. it means that it's always enabled. The `disabled` flag is used to disable column header drag while the menu is open. I think the proper fix will require understanding why the `disabled` property is getting the incorrect value rather than by eliminating the function of the `disabled` property altogether.

kswenson (changes requested):

eireland:
Investigated how V2 worked and realized we were not implementing it the same way in V3. We are now allowing column header drag even when attribute menu is open, but we close the attribute menu when the drag starts.

kswenson (approved):
👍 Looks good!

PR #575, opened 2022-12-02
V3: Simplify drop handlers

kswenson:
Simplifies the implementation of drop handlers using an approach that @tealefristoe demonstrated when he implemented DnDKit in CLUE.

@bacalj Assigning to you for review since you've worked with the DnDKit APIs.


bacalj (approved):
This is great.  If I am reading right, you implement `useDropHandler` -  It only needs the dropId, which it can then give to the dndMonitor who has all the data we care about filed under the key we present to it.  We avoid having to ask every droppable location to accept the data inline, and whatever we have registered to `onDrop` still gets called.  I sanity checked it and it works.

PR #578, opened 2022-12-05
V3—Split dot plot bug fix—183961539

bfinzer:
* One line fix that substitutes the correct band width in computation of overlap.


bacalj (approved):
👍 Looks good - pulled it down and it works - not sure what original bug was, but plot acting as expected.

PR #588, opened 2022-12-14
V3—Categorical axis improvements—183961343

bfinzer:
This commit adds grid lines to categorical axes and properly places the labels according to whether they should be parallel or perpendicular to the axis, and center or left/bottom justified.

### DIFF
--- a/v3/src/components/axis/hooks/use-axis.ts
+++ b/v3/src/components/axis/hooks/use-axis.ts
@@ -1,5 +1,5 @@
 import {
-  axisBottom, axisLeft, scaleLinear, scaleLog, scaleOrdinal, select, selection
+  axisBottom, axisLeft, ScaleBand, scaleLinear, scaleLog, scaleOrdinal, select, selection
 } from "d3"
 import {autorun, reaction} from "mobx"
 import {MutableRefObject, useCallback, useEffect, useRef} from "react"
@@ -17,12 +17,13 @@ export interface IUseAxis {
   titleRef: MutableRefObject<SVGGElement | null>
   label?: string
   enableAnimation: MutableRefObject<boolean>
-  showGridLines: boolean
+  showScatterPlotGridLines: boolean
+  centerCategoryLabels: boolean
 }
 
 export const useAxis = ({
                           axisModel, axisElt, titleRef, label = "",
-                          showGridLines, enableAnimation
+                          showScatterPlotGridLines, centerCategoryLabels, enableAnimation
                         }: IUseAxis) => {
   const layout = useAxisLayoutContext(),
     place = axisModel?.place ?? 'bottom',
@@ -37,8 +38,8 @@ export const useAxis = ({
     dataConfiguration = useDataConfigurationContext(),
     axisPlace = axisModel?.place ?? 'bottom',
     attrRole = graphPlaceToAttrPlace(axisPlace),
-    [xMin, xMax] = scale?.range() || [0, 100],
-    halfRange = Math.abs(xMax - xMin) / 2,
+    [rangeMin, rangeMax] = scale?.range() || [0, 100],
+    halfRange = (rangeMax !== undefined && rangeMin !== undefined) ? Math.abs(rangeMax - rangeMin) / 2 : 50,
     type = axisModel?.type ?? 'empty',
     attributeID = dataConfiguration?.attributeID(attrRole)
   previousAxisModel.current = axisModel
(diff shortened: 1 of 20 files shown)


v3/webpack.config.js
```diff
@@ -22,7 +22,7 @@ module.exports = (env, argv) => {
       client: {
         overlay: {
           errors: true,
-          warnings: false,
+          warnings: true,
```
kswenson, line 25: Personally, I don't think this change should be checked in, but we can discuss what the default should be.

v3/src/components/axis/hooks/use-axis.ts
```diff
@@ -17,124 +17,239 @@ export interface IUseAxis {
...
+                  case true:
+                    translation = `translate(${-bandWidth}, ${textHeight / 3})`
+                    rotation = `rotate(-90)`
+                    textAnchor = 'end'
+                    break
+                  case false:
+                    translation = `translate(${-bandWidth}, ${textHeight / 3})`
+                    textAnchor = 'start'
+                    break
+                }
```
kswenson, line 199: Since the translations are the same, this could be:
```suggestion
                translation = `translate(${-bandWidth}, ${textHeight / 3})`
                switch (collision) {
                  case true:
                    rotation = `rotate(-90)`
                    textAnchor = 'end'
                    break
                  case false:
                    textAnchor = 'start'
                    break
                }
```

v3/src/components/axis/hooks/use-axis.ts
```diff
@@ -17,124 +17,239 @@ export interface IUseAxis {
...
+          .selectAll('g').remove()
+        // select(axisElt).selectAll('line').remove()
+        select(axisElt).append('g')
+          .transition().duration(duration)
+          .attr('transform', `translate(${axisIsVertical ? 0 : bandWidth / 2}, ` +
+            `${axisIsVertical ? bandWidth / 2 : 0})`)
+          .call(axis(scale).tickSizeInner(-tickLength))
+          .selectAll('.domain').remove()
+        let translation = '', rotation = '', textAnchor = 'none'
+        switch (axisPlace) {
```
kswenson, line 142: Given the length and complexity of this `drawCategoricalAxis()` function and this `useAxis()` hook, I'd be inclined to move this logic into an external utility function that takes the necessary inputs and returns a structure of the form `{ translation, rotation, textAnchor }`. It could be a custom hook (e.g. `useLabelStyle()`), but it doesn't appear to use any other hooks so it doesn't need to be.

v3/src/components/axis/hooks/use-axis.ts
```diff
@@ -17,124 +17,239 @@ export interface IUseAxis {
...
+        break
+      case 'categorical':
+        drawCategoricalAxis()
+        break
     }
-  }, [axisElt, axisPlace, axis, layout, showGridLines, scale, enableAnimation,
-    place, titleRef, label, halfRange])
+    drawAxisTitle()
+  }, [axisElt, axisPlace, axis, layout, showScatterPlotGridLines, scale, enableAnimation,
+    place, titleRef, label, halfRange, type, bandWidth, collisionExists, centerCategoryLabels,])
```
kswenson, line 252: Not necessarily for this PR, but vis-a-vis our conversation about dependencies curation, virtually any change that affects the axis invalidates the `refreshAxis()` function, which in turn triggers all of the effects below, which could explain some of the extraneous axis re-renders. At some point it's worth thinking about whether the `refreshAxis()` function could be made more stable, so that the effects below could run more judiciously.

kswenson (changes requested):
👍 Looks good -- made the usual assortment of suggestions, some of which are straightforward and others may be worth addressing in a later PR.

v3/src/components/axis/hooks/use-axis.ts
```diff
@@ -17,124 +17,239 @@ export interface IUseAxis {
...
     // GraphController:processV2Document installs a new axis model.
     // Todo: Revisit and figure out whether we can remove the workaround.
     previousAxisModel = useRef<IAxisModel>(),
     axisModelChanged = previousAxisModel.current !== axisModel,
     dataConfiguration = useDataConfigurationContext(),
     axisPlace = axisModel?.place ?? 'bottom',
-    attrRole = graphPlaceToAttrPlace(axisPlace),
-    [xMin, xMax] = scale?.range() || [0, 100],
-    halfRange = Math.abs(xMax - xMin) / 2,
+    attrRole = graphPlaceToAttrRole(axisPlace),
```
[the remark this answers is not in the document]
bfinzer: Agreed. To be addressed at some point.

v3/src/components/axis/hooks/use-axis.ts
```diff
@@ -17,124 +17,239 @@ export interface IUseAxis {
...
+    [rangeMin, rangeMax] = scale?.range() || [0, 100],
+    halfRange = (rangeMax !== undefined && rangeMin !== undefined) ? Math.abs(rangeMax - rangeMin) / 2 : 50,
```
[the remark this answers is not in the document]
bfinzer: Unfortunately scale?.range() evaluates to an empty array in early stages of starting up and thus both rangeMax and rangeMin come out as undefined. I was able to slightly improve things with:
range = scale?.range() || [0, 100],
[rangeMin, rangeMax] = range.length === 2 ? range : [0, 100],
halfRange = Math.abs(rangeMax - rangeMin) / 2,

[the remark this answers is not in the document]
kswenson: Ah, then there's a bug in the TypeScript types for D3, because the existing types suggest that `rangeMin` and `rangeMax` are always numeric, i.e. can't be `undefined`.

v3/webpack.config.js
bfinzer: Yeah, it was an accident. I'll revert the change and hope it comes out in the squash.

v3/src/components/axis/hooks/use-axis.ts
bfinzer: I agree. Changes in halfRange, bandWidth and collisionExists will happen any time the graph is resized, thus triggering four of the useEffects and four calls to refreshAxis where we logically only want one.

v3/src/components/axis/hooks/use-axis.ts
bfinzer: I'm not going to try to do this restructuring now before the merge. I certainly agree that the useAxis() hook is ungainly and will consult with you about how best to improve it.

PR #589, opened 2022-12-16
V3: Refactors codap modal 

eireland:
The refactor moves the render of the modal content to the child instead of the modal level

### DIFF
--- a/v3/src/index.scss
+++ b/v3/src/index.scss
@@ -11,12 +11,12 @@ body {
   }
 
   /* Show outline for keyboard :focus-visible */
-  .chakra-button {
-    &:focus-visible {
-      // chakra uses box-shadow for focus highlight rather than outline
-      box-shadow: var(--chakra-shadows-outline) !important;
-    }
-  }
+  // .chakra-button {
+  //   &:focus-visible {
+  //     // chakra uses box-shadow for focus highlight rather than outline
+  //     box-shadow: var(--chakra-shadows-outline) !important;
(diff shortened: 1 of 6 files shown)


v3/src/index.scss
```diff
@@ -11,12 +11,12 @@ body {
...
-      // chakra uses box-shadow for focus highlight rather than outline
-      box-shadow: var(--chakra-shadows-outline) !important;
-    }
-  }
+  // .chakra-button {
+  //   &:focus-visible {
+  //     // chakra uses box-shadow for focus highlight rather than outline
+  //     box-shadow: var(--chakra-shadows-outline) !important;
+  //   }
+  // }
```
kswenson, line 19: Is this change related to this PR? This was done to prevent distracting highlighting when clicking while not disabling highlighting for keyboard focus. Do you have reason to believe it's no longer necessary?

kswenson (changes requested):
👍 Dialog changes look good -- some of the resulting dialog code is quite lengthy and so we might want to look into ways to break it up a bit, but that need not be part of this PR. The one question/suggestion is about the commenting out of the `focus-visible` rule.

v3/src/index.scss
eireland: Highlighting wasn't coming up on keyboard focus. But you're right, it shouldn't be part of this PR

kswenson (approved):
👍 LGTM

PR #590, opened 2022-12-16
Fixes bug with closing the inspector palette

eireland:
Closes the brush and ruler panels when the inspector palette is closed, and when user clicks outside of the palette.


kswenson (approved):
👍 LGTM -- should probably also close the palette or inspector on escape key, but that can be another story.

PR #591, opened 2022-12-16
V3: Document, DocumentContent, and Tile models and friends

kswenson:
Terminology:
`Tile` => CLUE/CODAP component (avoids ambiguity between CODAP/React "components")
`Row` => Container for holding tiles (components)
`FreeTileRow` => Container for holding tiles laid out free-form (e.g. a CODAP document)
`LegacyTileRow` => Container for holding tiles laid out in a simple horizontal row
`MosaicTileRow` => Container for holding tiles laid out in a mosaic/dashboard (e.g. https://nomcopter.github.io/react-mosaic/)
`SharedModel` => MST model that can be shared by multiple tiles (e.g. `DataSet`)
`DocumentContent` => Container for holding one or more rows

A CLUE document contains multiple rows with tiles laid out horizontally in each row. The document can be scrolled vertically but the rows are auto-sized so that no horizontal scrolling is required.

A CODAP v2/v3 document contains a single "row" containing an arbitrary number of tiles laid out in a free-form layout  with components that can overlap, etc. A CODAP v4 document could contain multiple "rows" presented as a tabbed multi-page document. In theory, a user could choose to adopt a mosaic layout on one such page.

CLUE contains a history component that makes use of MST's patch-generation to automate undo/redo that is not brought over as part of this PR but works with the `DocumentContent`, including the `SharedModel` system.

@scytacki Most of this code was taken from CLUE (obviously) and so doesn't warrant a line-by-line review. Perhaps just worth noting the places where changes have been made. I tried to keep the folder hierarchies parallel so you can use folder differencing to assist in finding the changes. Notably, the changes include:
- the new `TileRow` infrastructure for supporting alternative tile layouts
- trimmed-down API for `DocumentModel` and `DocumentContentModel`
  - I tried to keep the properties of the models as compatible as possible but there was a lot of cruft in the API for each that it was easier to eliminate than to try to migrate
- crude removal of the history storage/firebase code which actually went more smoothly than I expected (after I whined about it). I ended up being able to keep the history code more intact than I had feared initially.
- note the presence of `eslint-disable` lines for things like circular imports (still one remaining) and in some of the tests due to stricter lint settings in the CODAP V3 project.


v3/src/models/document/document-content.ts
```diff
@@ -0,0 +1,199 @@
...
+        if (row && !row.isSectionHeader && !self.isPlaceholderRow(row)) {
+          return i + 1
+        }
+      }
+      // if no tiles have content, insert after the first non-header row
+      for (let i = 0; i < self.rowCount; ++i) {
+        const row = self.getRowByIndex(i)
+        if (row?.acceptDefaultInsert) return row
+        if (row && !row.isSectionHeader) {
+          return i + 1
```
scytacki, line 116: I'm not totally sure, but it seems like this should just be `return i` like it used to be.

v3/src/models/document/document-content.ts
```diff
@@ -0,0 +1,199 @@
...
+        const row = self.getRowByIndex(i)
+        if (row?.acceptDefaultInsert) return row
+        if (row && !row.isSectionHeader && !self.isPlaceholderRow(row)) {
+          return i + 1
+        }
+      }
+      // if no tiles have content, insert after the first non-header row
+      for (let i = 0; i < self.rowCount; ++i) {
+        const row = self.getRowByIndex(i)
+        if (row?.acceptDefaultInsert) return row
```
scytacki, line 114: If I'm reading this right, sometimes this function returns a row and sometimes it returns an index. Is that intentional?

scytacki (approved):
This looks good to me.

I found a couple of things in document-content that look wrong, but I might not be understanding some changes.

v3/src/components/tiles/tile-component.tsx
```diff
@@ -0,0 +1,15 @@
+import React from "react"
+import { ITileModel } from "../../models/tiles/tile-model"
+
+import "./tile-component.scss"
+
+export interface ITileProps {
+  tile: ITileModel
+}
+export const TileComponent = ({ tile }: ITileProps) => {
```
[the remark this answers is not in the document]
kswenson: This was meant to be just a quick-and-dirty placeholder, but I agree that even a placeholder should be an `observer` so I made the change.

v3/src/models/document/document-content.ts
kswenson: Yes, that's intentional. In its original form this function always returned an index and the client was always supposed to create a new row to insert at that index. With support for the CODAP-like `FreeRowLayout` it is generally the case that new tiles should be inserted in an existing row instead, but the index-only API made it impossible for clients to know when to create a new row and when to insert in an existing row. I added a comment to clarify this.

v3/src/models/document/document-content.ts
kswenson: I came to the conclusion that this was a bug in the original implementation. The two `for` loops are completely analogous and in both cases they are documented as returning the index of the row _after_ the one that satisfies certain conditions, so I think it should be `i + 1` in both cases.

PR #592, opened 2022-12-19
183895168 v3 dnd rerender

bacalj:
These are a few changes that have the effect of reducing the number of elements that re-render as a result of their parents re-render.  The parents are those topmost components who re-render as a result of changes within DndKit Context state that occur during drag.

- The `handleIsActive` function definition that is defined in `axis.tsx`, `droppable-plot.tsx`, `legend.tsx` is moved from the body of the component function to the file scope, so that it is a constant. 
- `AxisAttributeMenu` is memoized, and this prevents it from re-rendering with its parent. 
- `DroppableSvg` is memoized, and this removes the majority of its re-renders. 
- `DroppablePlot` is memoized.  It is now a closer match to its sibling, `Legend`.  Both are memoized, however both still re-render with changing drag state.   (More info on that in pivotal)


kswenson (approved):
👍 Looks good

bfinzer (approved):
👍 LGTM

PR #594, opened 2022-12-19
V3—Refactoring of refreshAxis

bfinzer:
* Pull big switch statement (getCategoricalLabelPlacement) out of refreshAxis into its own file
* Fine tuning of collision detection of categorical labels to guarantee a bit more space between them

### DIFF
--- a/v3/src/components/axis/axis-utils.ts
+++ b/v3/src/components/axis/axis-utils.ts
@@ -1,0 +1,10 @@
+
+
+export const getCategoricalLabelPlacement = (place: AxisPlace) => {
+  switch (place) {
+    case "bottom":
+      return "bottom"
+    case "left":
+      return "left"
+  }
+}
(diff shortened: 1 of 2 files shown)


v3/src/components/axis/axis-utils.ts
```diff
@@ -0,0 +1,69 @@
+import {AxisPlace} from "./axis-types"
+
+
+export const getCategoricalLabelPlacement = (
+  axisPlace: AxisPlace, centerCategoryLabels: boolean, collision: boolean, bandWidth: number, textHeight: number) => {
```
kswenson, line 5: With more than about three arguments I generally switch to passing an object rather than a list of comma-separated arguments, but it's a judgment call.

kswenson (approved):
👍 Looks good

v3/src/components/axis/axis-utils.ts
bfinzer: Thanks for pushing me on this. I was being lazy!

PR #595, opened 2022-12-20
183822145 v3 chakra menu dismiss

bacalj:
This PR allows the Chakra menu on the Axis Attribute Menu to close when the user clicks anywhere else in CODAP.  Previously, it would not close when the user clicked on the drag rects or on the graph plot.

- `use-codap-outside-click.ts` is a modified copy of Chakra's `useOutsideClick` hook.  Lines 67 and 68 add listeners  for `pointerdown` and `pointerup` in addition to the existing `mousedown` and `mouseup` listeners.

- We use the hook in  `axis-attribute-menu.tsx` and replace the default `isOpen` with our own local component state to control the menu.  

This was implemented because click events on the drag-rect and plot elements — which have d3 drag behaviors defined on them — were bubbling up with properties that were not being listened to by Chakra's uppermost listeners, nor were they recognized by Chakra’s `useOutsideClick` hook.   (On inspection of those events, we can see they return a different result for `e.composedPath()`that includes an additional `__on` property with an array that contains a value like `{type: 'mousedown', name: 'drag', ... }`)

The import included in `use-codap-outside-click.ts` is giving us this lint message, but I didn't want to install it without checking. 

```'@chakra-ui/react-use-callback-ref' should be listed in the project's dependencies. Run 'npm i -S @chakra-ui/react-use-callback-ref' to add it```

### CODE REVIEW

**Review by bfinzer — APPROVED** [review]
👍🏻LGTM
But I don't pretend to understand all the ins and outs of Chakra event handling.

**Review by kswenson — APPROVED** [review]
👍 Looks good -- I added a commit with a few changes to avoid going back-and-forth on review.

File: v3/src/hooks/use-outside-pointer-down.ts [hunk 1]
```diff
@@ -0,0 +1,61 @@
...
+
+    const onPointerDown: any = (e: PointerEvent) => {
+      if (isValidEvent(e, ref)) {
+        savedHandler(e)
+      }
+    }
+
+    // listen for pointerdown because D3 drag handlers intercept subsequent mouse events
+    const doc = getOwnerDocument(ref.current)
+    doc.addEventListener("pointerdown", onPointerDown, true)
```
bacalj, line 39: Nice.  So the mouse and other events from the original were extraneous.

kswenson: In this particular instance, the issue was that the mouse and other events were never getting to the hook in the case we cared about, so by definition the handlers for them were extraneous. As near as I can tell, most of the subtle logic in the Chakra hook was to detect the down event but then only call the callback (e.g. close the dialog) when the up event occurred. I don't see why we care about that distinction or whether we close on a `pointerdown` or other event. `pointerdown` happens to be the first event sent, however, and the only one that isn't swallowed by D3's drag handlers, so that's the one we use.

PR #601, opened 2022-12-22
183808952 v3 legend menu

bacalj:
This adds an attribute menu to the Legend label.   

`legend-attribute-menu`
- modeled after `axis-attribute-menu`
- receives as props functions defined at Graph level for handling the change to a new attribute and handling how the attribute is treated  
   - changing from one categorical attribute to another works as expected
   - treating a categorical attribute as numeric results in removing the legend colors
   - removing the attribute results in the legend going away and in removing the legend colors
   - I don't believe anything else is implemented at the Graph level that should be working here, but wanted to point this out. 
- uses existing positioning `useOverlayBounds()` for the Legend label to position the button over the label

`legend.tsx`
- Now has `onTreatAttributeAs` as a prop in order to pass it down from graph to legend attribute menu
- Instantiates a `<LegendAttributeMenu>` in a portal as a sibling of `<AttributeLabel>`
- I had assumed the change below was to an`any` that I had left in:
   ```diff
   -  onDropAttribute: (place: any, attrId: string) => void
   +  onDropAttribute: (place: GraphPlace, attrId: string) => void
   ```
   but on closer inspection I see it was there and I changed it to `GraphPlace`.  Flagging it in case it is supposed to be `any` for now for any reason. 

`axis-attribute-menu`
- Click handler on the menu list wrapper calls close rather than toggle


kswenson (approved):
👍 Looks good, particularly after I made a few tweaks. 😉

bacalj:
Certainly less duplication!

PR #603, opened 2022-12-24
Build 3.0.0-pre.1085

tejal-shah:


v3/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "codap3",
-  "version": "3.0.0-mvp.1",
+  "version": "3.0.0-1085",
```
kswenson, line 3: I think we should use a version number that makes it clear that this is not an official release -- something like `3.0.0-pre.1085` or `3.0.0-alpha.1085`.

kswenson (changes requested):
👍 Generally looks good, but I think we should use a version number that makes it clear that this is not an official release -- something like `3.0.0-pre.1085` or `3.0.0-alpha.1085` as suggested below.

tejal-shah:
@kswenson , I've updated the version number to 3.0.0-pre.1085

kswenson (approved):
👍 LGTM

PR #604, opened 2022-12-25
V3—Improve_y-axis_space allocation

bfinzer:
* Before computing desired extent of vertical numeric axis, format the numbers using the scale's formatting.
* Also substitute `measureText` in place of getBoundingClientRect where appropriate

### DIFF
--- a/v3/src/components/axis/hooks/use-axis.ts
+++ b/v3/src/components/axis/hooks/use-axis.ts
@@ -1,6 +1,5 @@
 import {
-  axisBottom, axisLeft, ScaleBand, scaleLinear, scaleLog, scaleOrdinal, select, selection
-} from "d3"
+  axisBottom, axisLeft, ScaleBand, scaleLinear, scaleLog, scaleOrdinal, select} from "d3"
 import {autorun, reaction} from "mobx"
 import {MutableRefObject, useCallback, useEffect, useRef} from "react"
 import {AxisBounds, axisGap, isVertical, ScaleNumericBaseType} from "../axis-types"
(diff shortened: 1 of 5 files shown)


v3/src/components/axis/hooks/use-axis.ts
```diff
@@ -73,16 +69,19 @@ export const useAxis = ({
...
-    let ticks: number[] = []
     let desiredExtent = labelHeight + 2 * axisGap
+    let ticks: string[] = []
     switch (type) {
-      case 'numeric':
-        ticks = (scale.ticks?.()) ?? []
-        desiredExtent += axisPlace === 'left' ? Math.max(getLabelBounds(String(ticks[0])).width,
-          getLabelBounds(String(ticks[ticks.length - 1])).width) : getLabelBounds().height
+      case 'numeric': {
+        const format = scale.tickFormat && scale.tickFormat()
```
kswenson, line 76:
```suggestion
        const format = scale.tickFormat?.()
```

v3/src/components/graph/utilities/graph-utils.ts
```diff
@@ -12,18 +12,17 @@ import {
...
 /**
  * Utility routines having to do with graph entities
  */
 
 export const maxWidthOfStringsD3 = (strings: Iterable<string>) => {
-  const text = selection().append('svg').append('text').style('font', '12px sans-serif')
   let maxWidth = 0
   for (const aString of strings) {
-    maxWidth = Math.max(maxWidth, text.text(aString).node()?.getComputedTextLength() ?? 0)
+    maxWidth = Math.max(maxWidth, measureText(aString, '12px sans-serif'))
```
kswenson, line 24: This font string occurs multiple times -- seems like a named constant would be a good idea, e.g. `kGraphD3Font` or something like that, so that if we ever need to change it there's one place to do so.

kswenson (approved):
👍 Looks good -- just a couple suggested tweaks

v3/src/components/graph/utilities/graph-utils.ts
bfinzer: Agreed

PR #605, opened 2022-12-26
V3—Caseplot no initial response

bfinzer:
On restart of V3 the case plot was not responding to clicks on points or marquee select. It appears that this was due to usePlotResponders being unable to detect a change in the dataset from undefined to an object during the startup process. A fix that worked was to pass the dataset directly from the plots making use of usePlotResponders.

### DIFF
--- a/v3/src/components/graph/hooks/use-plot.ts
+++ b/v3/src/components/graph/hooks/use-plot.ts
@@ -8,7 +8,7 @@ import {useCurrent} from "../../../hooks/use-current"
 import {IGraphModel} from "../models/graph-model"
 import {matchCirclesToData} from "../utilities/graph-utils"
 import {useInstanceIdContext} from "../../../hooks/use-instance-id-context"
-import {IDataSet} from "../../../models/data/data-set"
+import {useDataSetContext} from "../../../hooks/use-data-set-context"
 
 interface IDragHandlers {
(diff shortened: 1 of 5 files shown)


v3/src/components/graph/hooks/use-plot.ts
```diff
@@ -43,10 +45,9 @@ export interface IPlotResponderProps {
 
 export const usePlotResponders = (props: IPlotResponderProps) => {
   const {
-      graphModel, primaryAttrID, secondaryAttrID, legendAttrID, enableAnimation,
+      graphModel, dataset, primaryAttrID, secondaryAttrID, legendAttrID, enableAnimation,
```
kswenson, line 48: Rather than passing it in, did you try `useDataSetContext()` instead of `graphModel.config.dataset` below? That should be sufficient to trigger a re-render without requiring clients to pass it in.

kswenson (changes requested):
Seems like this shouldn't be necessary -- I made an alternative suggestion below.

v3/src/components/graph/hooks/use-plot.ts
bfinzer: Yes, that worked! Thanks. Cleared up a mystery for me.

kswenson (approved):
👍 Looks good

PR #606, opened 2022-12-28
184098347 v3 slider drag

bacalj:
This adds drag and animation functionality to the slider.

**Drag Functionality**
- ~~The rounding for dragged values is now configured by a property on the model called `resolution`, hardcoded to `.01`~~
-  The `multipleOf` property ~~still~~ applies to incrementing the slider with the play button, ~~but does not apply to~~ dragging ~~or~~ and keyboard input of values (same as v2)
- The above are differentiated by distinguishing two methods on the model: `setValue` which rounds to the `resolution`, and `setValueRoundedToMultipleOf`, which rounds to the user-changeable `multipleOf` value. (same as v2)
- ~~D3 also has a method called `nice()` that could apply here, I think, but I have not tried it yet.~~

~~I'm aware that  the changes described above may be based on incomplete or incorrect assumptions about how `multiplesOf` is intended to work, so I wanted to be sure to flag them.~~
 
**Animation**
- ~~The motion of the slider across values is implemented with a CSS transition~~
- There is an `isDragging` piece of state on the thumbnail. I added some placeholder styles (a "pick up" effect with a drop shadow -- just a placeholder for the moment.

**clsx**
I've taken this opportunity to add the [clsx](https://www.npmjs.com/package/clsx) library to CODAP, which can be used to simplify adding classes to elements conditionally, for instance.

### DIFF
--- a/v3/.vscode/settings.json
+++ b/v3/.vscode/settings.json
@@ -7,6 +7,7 @@
   "typescript.tsdk": "./node_modules/typescript/lib",
   "cSpell.words": [
     "Chakra",
+    "clsx",
     "papaparse",
     "portaling",
     "ulid"
--- a/v3/package-lock.json
+++ b/v3/package-lock.json
@@ -14,6 +14,7 @@
         "@dnd-kit/core": "^6.0.5",
         "@emotion/react": "^11.10.4",
         "@emotion/styled": "^11.10.4",
+        "clsx": "^1.2.1",
         "d3": "^7.6.1",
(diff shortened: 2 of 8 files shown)


kswenson (approved):
👍 Since @bacalj is out of town, I've taken the liberty of adding a commit which tweaks things such that it now passes my review. 😉 @bfinzer You should review it as well.

bfinzer (approved):
👍🏻LGTM

PR #610, opened 2022-12-31
V3—Fix for plot background response to window size change—184125835

bfinzer:
Pull request #599 attempted to fix an unwanted clipping of graph points dragged to the right of the plot area. Unfortunately, those changes resulted in the bug described in #184125835.
* This pull request reinstates the Graph:setupPlotArea useEffect
* It also avoids use of stale xScale and yScale in casedots


v3/src/components/graph/components/casedots.tsx
```diff
@@ -29,9 +29,7 @@ export const CaseDots = memo(function CaseDots(props: {
     dragPointRadius = graphModel.getPointRadius('hover-drag'),
     [dragID, setDragID] = useState(''),
     currPos = useRef({x: 0, y: 0}),
-    target = useRef<any>(),
-    xScale = layout.getAxisScale('bottom') as ScaleNumericBaseType,
-    yScale = layout.getAxisScale('left') as ScaleNumericBaseType
+    target = useRef<any>()
```
kswenson, line 32: Should eliminate unused import above now that `ScaleNumericBaseType` is no longer used.

v3/src/components/graph/components/graph.tsx
```diff
@@ -116,10 +129,12 @@ export const Graph = observer((
                      onTreatAttributeAs={handleTreatAttrAs}
           />
 
+          <svg ref={plotAreaSVGRef} className='graph-dot-area'>
```
kswenson, line 132: Do we want/need both of these `<svg>`s to have the same `className` (`graph-dot-area`)?

kswenson (approved):
👍 LGTM -- just a couple of questions/suggestions

v3/src/components/graph/components/casedots.tsx
bfinzer: Done

v3/src/components/graph/components/graph.tsx
bfinzer: Nope. Good catch.

PR #614, opened 2023-01-05
V3: Update deployment instructions in README.md and some build improvements

tejal-shah:
This PR has the following changes:

1. Updating `README.md` with deployment instructions
2. Upgrading actions/checkout@v2 in github workflows to actions/checkout@v3 (to see if this fixes the build number increment failures)
3. Making sure on:push: events for v3 tags don't trigger v2 builds
4. Making sure we lock the build number and don't increment it when release branches are merged into main so that the version number matches the build number

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,13 +5,15 @@ on:
     paths-ignore: # don't run this workflow if it only contains v3 changes
       - 'v3/**'   # https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#example-excluding-paths
       - '.github/workflows/v3.yml'
+    tags-ignore:
+      - '3.*'
 
 jobs:
   build_test:
     name: Build and Run Jest Tests
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - uses: actions/setup-node@v2
       - name: Setup Ruby
         uses: actions/setup-ruby@v1
@@ -79,7 +81,7 @@ jobs:
       # - cypress
     runs-on: ubuntu-18.04
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - uses: actions/setup-node@v2
       - name: Setup Ruby
         uses: actions/setup-ruby@v1
--- a/v3/README.md
+++ b/v3/README.md
@@ -77,17 +77,20 @@ Other branches are deployed to https://codap3.concord.org/branch/{branch-name}/,
 
 To deploy a production release:
 
-1. Increment version number in package.json
-2. Create new entry in CHANGELOG.md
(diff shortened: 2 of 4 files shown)


v3/README.md
```diff
@@ -77,17 +77,20 @@ Other branches are deployed to https://codap3.concord.org/branch/{branch-name}/,
...
-2. Create new entry in CHANGELOG.md
+1. Update the version number in `package.json` and `package-lock.json`
+    - `npm version --no-git-tag-version [patch|minor|major]`
+2. Create new entry in `CHANGELOG.md` with a description of the new version
 3. Run `git log --pretty=oneline --reverse <last release tag>...HEAD | grep '#' | grep -v Merge` and add contents (after edits if needed to CHANGELOG.md)
 4. Run `npm run build`
 5. Copy asset size markdown table from previous release and change sizes to match new sizes in `dist`
-6. Create `release-<version>` branch and commit changes, push to GitHub, create PR and merge
+6. Create `v3-release-<version>` branch and commit changes, push to GitHub, create PR and merge
 7. Checkout `master` and pull
```
kswenson, line 87:
```suggestion
7. Checkout `main` and pull
```

kswenson (approved):
👍 Looks good -- there's one typo to fix (`master` => `main`) which isn't part of this PR but I noticed on review and seems worth fixing. Separately, we should really adopt @scytacki's new `s3-deploy` action rather than continuing to use the old `s3_website` gem approach, but that can be a PR for another day.

tejal-shah:
@kswenson , I fixed the typo you suggested. Also, codap v3 deployments already use the `s3-deploy-action` - this was done as part of https://github.com/concord-consortium/codap/pull/523 , although v2 deployments will need to make this switch. I'll create a separate story for this on the v2 board.

kswenson:
> @kswenson Also, codap v3 deployments already use the `s3-deploy-action` - this was done as part of #523 

That's what I thought, but a couple of things threw me off, like the continuing presence of `s3_deploy.sh` (it should be removed from `main`) and these lines in the `v3/README`:
```
10. Checkout `v3-production`
11. Run `git merge main --no-ff`
12. Push `v3-production` to GitHub
```
which I don't think are relevant to the `s3-deploy-action` model, in which case they should presumably also be removed.

.github/workflows/ci.yml
```diff
@@ -5,13 +5,15 @@ on:
     paths-ignore: # don't run this workflow if it only contains v3 changes
       - 'v3/**'   # https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#example-excluding-paths
       - '.github/workflows/v3.yml'
+    tags-ignore:
+      - '3.*'
```
kswenson, line 9: Can we also ignore the `main` branch? I don't think we ever want to trigger a v2 workflow on the `main` branch, nor do we ever want to trigger a v3 workflow on the `master` branch.

PR #617, opened 2023-01-07
V3—choropleth-legend—184029172

bfinzer:
With this pull request assigning a numeric attribute as a legend colors points according to which of 5 quintiles their legend attribute values lie in.

Note that there is a function `neededSignificantDigitsArray` in math-utils that is not currently being used. But I'd like to keep it around, at least for awhile.

### DIFF
--- a/v3/src/components/graph/components/legend/choroplethLegend/choroplethLegend.ts
+++ b/v3/src/components/graph/components/legend/choroplethLegend/choroplethLegend.ts
@@ -13,7 +13,7 @@ import {
   scaleLinear,
   format,
   // range,
-  select, range, min, max
+  select, range, min, max, ScaleQuantile
 } from "d3"
 import {kChoroplethHeight} from "../../../graphing-types"
 import {neededSigDigitsArrayForQuantiles} from "../../../../../utilities/math-utils"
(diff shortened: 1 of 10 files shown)


v3/src/components/graph/components/legend/choroplethLegend/choroplethLegend.ts
```diff
@@ -0,0 +1,112 @@
...
+export function choroplethLegend(scale: any, choroplethElt: SVGGElement, props: ChoroplethLegendProps) {
+  const {
+      tickSize = 6, transform = '', width = 320, marginTop = 0, marginRight = 0, marginLeft = 0,
+      ticks = 5, clickHandler, casesInQuantileSelectedHandler
+    } = props,
+    minValue = min(scale.domain()),
+    maxValue = max(scale.domain())
+
+  let tickFormat: any = '.2r',
+    tickValues: any = []
```
kswenson, line 43: I think these can be typed as:
```suggestion
  let tickFormat: string | ((i: number) => string) = '.2r',
    tickValues: number[] = []
```

v3/src/components/graph/components/legend/numeric-legend.tsx
```diff
@@ -1,18 +1,93 @@
...
+      },
+      () => {
+        refreshScale()
+      }, {fireImmediately: true}
+    )
+    return () => disposer()
+  }, [layout, computeDesiredExtent, dataConfiguration, refreshScale])
+
+  useEffect(function respondToSelectionChange() {
+    const disposer = onAction(dataset, action => {
```
kswenson, line 72: It shouldn't make any difference functionally at this point, but I think it would make more sense to restrict your reactions to the `DataConfigurationModel`:
```suggestion
    const disposer = dataConfiguration.onAction(action => {
```
We could conceivably filter out notifications for cases that don't pass the filter, for instance.

kswenson (changes requested):
👍 Looks good -- the usual assortment of relatively minor suggestions

v3/src/components/graph/models/data-configuration-model.ts
```diff
@@ -219,14 +250,47 @@ export const DataConfigurationModel = types
...
+          thresholds = self.legendQuantileScale.quantiles(),
+          min = quantile === 0 ? -Infinity : thresholds[quantile - 1],
+          max = quantile === thresholds.length ? Infinity : thresholds[quantile],
+          selection: string[] = legendID && legendID !== ''
+            ? self.cases.filter((anID: string) => {
+              const value = dataset?.getNumeric(anID, legendID)
+              return value !== undefined && value >= min && value < max
+            })
+            : []
+        return !!(selection.length > 0 &&selection?.every((anID: string) => dataset?.isCaseSelected(anID)))
```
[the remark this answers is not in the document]
bfinzer: Yup

PR #622, opened 2023-01-12
Implement default free layout

kswenson:
Add `layout=free` url parameter to enable free (CODAP v2-like) component layout.
Once we have basic component creation, moving, and resize we can make free layout the default.


bfinzer (approved):
LGTM

PR #631, opened 2023-01-13
V3: Store DataSets as shared models in document

kswenson:
When I embarked on this story I was expecting to have to replace/rewrite much of the `DataBroker` functionality with `SharedModelManager` mechanisms. When I got into it I realized I could leave the `DataBroker` essentially in tact and just have it also add any `DataSet`s to the document as shared models. I doubt the `DataBroker` will survive into production v3, at least not in its current form, but for now we can put off any decisions about it for another day.

While we have not yet implemented document save/restore (export/import), there are jest tests which verify that snapshots of the document contain the contents of the `DataSet` and that there is only one copy of each `DataSet`, i.e. that the `DataBroker` and `SharedModelManager` both have access to the same `DataSet` object rather than separate instances.

This PR also extends our existing patch of `mobx-state-tree` to also disable the warning for jest tests as well.


bfinzer (approved):
👍🏻LGTM
My difficulty understanding what a shared model is reminds me that we should discuss as a team what we want in the way of overview documentation.

PR #637, opened 2023-01-16
V3: calculator

eireland:
Adds a simple calculator. To open the calculator, click on the Calculator button in the toolshelf.
Note differences to V2 calculator:
1. Cannot type directly into text field.
2. Result from previous calculation does not carry forward to next calculation (i.e. if you do 1+1=, with 2 as the result, you cannot type +3, to get 5)


eireland:
Calculator is now always visible in the Hello CODAP quadrant. It also has the double CODAP component header.

v3/src/components/calculator/calculator.tsx
```diff
@@ -0,0 +1,106 @@
...
+      if (calcValue === "") { setCalcValue(strToInsert) }
+      else  { setCalcValue(calcValue.concat(strToInsert)) }
+    } else {
+      setCalcValue(calcValue.concat(strToInsert))
+    }
+  }
+  const evaluate = () => {
+    // eslint-disable-next-line no-useless-escape
+    const input = calcValue.replace(/[^0-9()+\-*\/.]/g, "") //sanitize input
+    if (calcValue.includes("(") && !calcValue.includes(")")) {
```
bfinzer, line 38: This is one of a great number of possible syntax errors you've singled out. Is it part of a broader plan? Also noting that expressions like (8*9)+(7*8 are not reported with a toast but are still an error of missing a closing paren.
I guess I would recommend postponing any error handling beyond the try/catch until we have a general strategy.

bfinzer (approved):
👍🏻LGTM
Just one suggestion about waiting to handle any errors before try/catch.

v3/src/components/calculator/calculator.tsx
```diff
@@ -0,0 +1,106 @@
...
+    {"4": ()=>insert("4")},
+    {"5": ()=>insert("5")},
+    {"6": ()=>insert("6")},
+    {"-": ()=>insertSymbol("-")},
+    {"1": ()=>insert("1")},
+    {"2": ()=>insert("2")},
+    {"3": ()=>insert("3")},
+    {"+": ()=>insertSymbol("+")},
+    {"0": ()=>insert("0")},
+    {".": ()=>insertSymbol(".")}
```
kswenson, line 79: Should use the Unicode symbols for button labels rather than the ASCII ones:
+, −, ×, ÷

kswenson (approved):
👍 Looks reasonable so far -- I'm not really reviewing any of the evaluation infrastructure because that seems like a work in progress. I added a commit which tweaks the layout mainly to demonstrate how to do so. I think the double title bars and evaluation infrastructure can be figured out later. I did make one suggestion about using Unicode characters for button labels rather than ASCII ones.

PR #639, opened 2023-01-18
V3 fixing tagged builds

tejal-shah:


tejal-shah:
Adding branches-ignore for master in v3.yml broke the tagged builds completely

kswenson (approved):
👍 LGTM

>Adding `branches-ignore` for `master` in v3.yml broke the tagged builds completely

Do we know why this broke tagged builds of v3, which should be based on the `main` branch? Something seems odd here but maybe fixing it is more important than understanding it. 🤷

tejal-shah:
I couldn't fully understand it but it might have something to do with the fact that if you define both branches/branches-ignore and paths, the workflow will only run when both filters are satisfied. There is a lot of weirdness in github workflows around the conditions of when a workflow can get triggered. When you get a chance, maybe you can see if you are able to unentangle it - https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions

PR #644, opened 2023-01-25
184211652 free component UI

eireland:
Adds header component to component tiles that shows minimize and close icons (but no functionality).
Adds case card/case table icon to case table header. Clicking on the icon shows the message re switching from one view to the other, but does not actually switch views.
Adds the resize handle to the component tile. Currently, the resize handle is always visible, but but when tile focus is implemented, the resize handle should be hidden until tile is selected.


v3/src/components/component-header.tsx
```diff
@@ -0,0 +1,65 @@
...
+    setShowCaseCard(!showCaseCard)
+  }
+
+  const cardTableToggleString = showCaseCard
+                                  ? t("DG.DocumentController.toggleToCaseTable")
+                                  : t("DG.DocumentController.toggleToCaseCard")
+
+  return (
+    <Flex className="component-header">
+      {tileType === "codap-case-table" &&
```
bfinzer, line 41: I'm uncomfortable with embedding tile specific knowledge in this general-purpose component. Is there any way around that that you can think of?

bfinzer (approved):
👍🏻If you'll change ComponentHeader to ComponentTitleBar as well as other references to "header" to "title-bar" I'm fine with this PR. I also wish there were a way to ask some other entity for the any special purpose widgetry rather than embedding the knowledge in the ComponentTitleBar, but work on that could come later.

v3/src/components/component-header.tsx
kswenson: I agree with Bill on this one. The common-code title bar should not have tile-specific code in it. Perhaps the client could pass in contents, e.g. as `children`. Take a look at how this is handled in the v2 code.

v3/src/components/data-summary/data-summary.scss
```diff
@@ -6,6 +6,8 @@
   height: calc(100% - vars.$title-bar-height);
   background-color: aliceblue;
   padding: 5px;
+  border-bottom-left-radius: 6px;
+  border-bottom-right-radius: 6px;
```
kswenson, line 10: Since these constants occur in multiple places, perhaps they should be constants in `vars.scss`.

v3/src/components/codap-component.scss
```diff
@@ -9,4 +9,18 @@
...
     }
   }
+
+  .component-resize-handle {
+    position: absolute;
+    bottom: 0px;
+    right: 0px;
+    width: 22px;
+    height: 22px;
+    z-index: 999;
```
kswenson, line 19: Does it need to be that high? We don't want it in front of dialogs or inspectors.

v3/src/components/component-header.scss
```diff
@@ -0,0 +1,125 @@
+@import "./vars.scss";
+
+.component-header {
+  flex-direction: row;
+  height: $title-bar-height !important;
```
kswenson, line 5: Why the `!important` here?

v3/src/components/component-header.tsx
```diff
@@ -0,0 +1,65 @@
...
+import CardIcon from "../assets/icons/icon-case-card.svg"
+
+import "./component-header.scss"
+
+interface IProps {
+  tileType: string
+  datasetName?: string
+}
+
+export const ComponentHeader = ({tileType, datasetName}: IProps) => {
```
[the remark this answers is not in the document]
kswenson: Also note that the title isn't always the data set name, e.g. for a slider it's the slider name.

kswenson (changes requested):
👍 Generally looks good but a few suggestions, primarily Bill's concern about keeping tile-specific code out of common code.

bfinzer (approved):
👍🏻LGTM

v3/src/components/case-table/case-table-title-bar.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { Box, CloseButton, Flex } from "@chakra-ui/react"
+import t from "../../utilities/translation/translate"
+import MinimizeIcon from "../../assets/icons/icon-minimize.svg"
+import TableIcon from "../../assets/icons/icon-table.svg"
+import CardIcon from "../../assets/icons/icon-case-card.svg"
+import { useDataSetContext } from "../../hooks/use-data-set-context"
+
+export const CaseTableTitleBar = () => {
+  const dataset = useDataSetContext()
+  const [title, setTitle] = useState(dataset?.name || "Dataset")
```
kswenson, line 12: If the user changes the title that probably needs to be saved somewhere, so this value should be set in the model not just in React state. But that can be a feature for another day if you don't want to deal with it now.

v3/src/components/component-title-bar.scss
```diff
@@ -0,0 +1,134 @@
...
+      height: 30px;
+      background-color: rgba(255, 255, 255, 0.90);
+      border-radius: 6px;
+      color: black;
+      padding: 5px 10px;
+      z-index: 5;
+      &:hover {
+        color: #3366CC
+      }
+    }
```
kswenson, line 76: Can we move the tile-specific styles out of this generic file as well?

kswenson (approved):
👍 Looks good -- a couple of comments/questions, but nothing that need hold up the PR.

v3/src/components/case-table/case-table-title-bar.tsx
eireland: I'll write a story for it so we remember to implement it later. Story is [here](https://www.pivotaltracker.com/story/show/184374934)

PR #645, opened 2023-01-26
Removes double header on slider

eireland:


kswenson (commented):
The v2 behavior is that it shows the name of the variable rather than the name of the dataset/document. I was thinking that addressing that would be part of this story as well.

eireland:
Addressed it in the free component ui.

PR #646, opened 2023-01-27
184228103 close component

eireland:
Adds ability to close component tiles.
Free layout works as expected, with each tile closing when user clicks on the close button.
~~Mosaic layout does not quite work as expected. If you close all the tiles in a row, all the tiles close. (ie. Close the graph, the case table tile expands correctly. Close the case table. All the tiles disappear.)~~ Fixed


bfinzer (approved):
👍🏻LGTM

kswenson (changes requested):
Actually, there's one other thing we should do -- remove the tile from the document as well.

v3/src/components/mosaic-tile-row.tsx
```diff
@@ -89,14 +92,14 @@ export const MosaicNodeComponent = observer(({ node, direction, pctExtent, ...ot
...
+  content?: IDocumentContentModel
 }
-export const MosaicTileComponent = observer(({ row, tile, direction, pctExtent }: IMosaicTileProps) => {
+export const MosaicTileComponent = observer(({ content, tile, direction, pctExtent }: IMosaicTileProps) => {
   const style = styleFromExtent({ direction, pctExtent })
   const tileType = tile.content.type
   const info = getTileComponentInfo(tileType)
   const handleCloseTile = (tileId: string) => {
-    row?.removeTile(tileId)
+    content?.deleteTile(tileId)
```
kswenson, line 102: I think for now you probably have to do both:
```typescript
    content?.deleteTile(tileId)
    row?.removeTile(tileId)
```
With some additional work we could make it the case that the row models respond to content changes, but that's not in place right now, so manually keeping the two in sync will presumably be required for the time being.

v3/src/components/mosaic-tile-row.tsx
```diff
@@ -13,13 +14,14 @@ import "./mosaic-tile-row.scss"
  */
 interface IMosaicTileRowProps {
   row: IMosaicTileRow
+  content?: IDocumentContentModel
   getTile: (tileId: string) => ITileModel | undefined
 }
-export const MosaicTileRowComponent = ({ row, getTile }: IMosaicTileRowProps) => {
+export const MosaicTileRowComponent = ({ content, row, getTile }: IMosaicTileRowProps) => {
```
kswenson, line 20: This should probably be an `observer()` as well, so that it re-renders if `row.root` changes, for instance.

v3/src/components/free-tile-row.tsx
```diff
@@ -8,13 +9,14 @@ import { CodapComponent } from "./codap-component"
...
+  content?: IDocumentContentModel
   row: IFreeTileRow
   getTile: (tileId: string) => ITileModel | undefined
 }
-export const FreeTileRowComponent = observer(({ row, getTile }: IFreeTileRowProps) => {
+export const FreeTileRowComponent = observer(({ content, row, getTile }: IFreeTileRowProps) => {
   const handleCloseTile = (tileId: string) => {
     if (!tileId) return
-    row?.removeTile(tileId)
+    content?.deleteTile(tileId)
```
kswenson, line 19: I think for now you probably have to do both:
```typescript
    content?.deleteTile(tileId)
    row?.removeTile(tileId)
```
With some additional work we could make it the case that the row models respond to content changes, but that's not in place right now, so manually keeping the two in sync will presumably be required for the time being.

eireland: Doesn't this do the same thing? 
https://github.com/concord-consortium/codap/blob/849725659143f07758ca41f1b448ef5ad767c00b/v3/src/models/document/document-content.ts#L168

kswenson: Right you are! My bad. 🤦

kswenson (approved):
👍 Looks good

PR #648, opened 2023-02-02
184210645 resize components

eireland:
~~Allows free layout components to be resized horizontally and vertically from bottom right corner.~~
~~Allows mosaic layout components to be resized vertically from bottom right corner.~~
~~Things to fix when full resizing feature (resize from both bottom corners and side borders) is implemented is removing the resize handle that css automatically adds when `resize` css is used.~~

Allows free layout components to be resized from left, right, and bottom sides, and from bottom left and bottom right corners.
Mosaic resize is not implemented

### DIFF
--- a/v3/src/components/codap-component.tsx
+++ b/v3/src/components/codap-component.tsx
@@ -2,7 +2,7 @@ import { observer } from "mobx-react-lite"
 import React from "react"
 import { DataSetContext } from "../hooks/use-data-set-context"
 import { gDataBroker } from "../models/data/data-broker"
-import { ITileBaseProps } from "./tiles/tile-base-props"
+import { ITileBaseProps, ITileTitleBarProps } from "./tiles/tile-base-props"
 import { ITileModel } from "../models/tiles/tile-model"
 import ResizeHandle from "../assets/icons/icon-corner-resize-handle.svg"
 
@@ -10,9 +10,10 @@ import "./codap-component.scss"
 
 export interface IProps extends ITileBaseProps {
   tile: ITileModel
-  TitleBar: React.ComponentType<ITileBaseProps>;
+  TitleBar: React.ComponentType<ITileTitleBarProps>;
   Component: React.ComponentType<ITileBaseProps>;
   tileEltClass: string;
+  onCloseTile: (tileId: string) => void
   onBottomRightPointerDown: (e: React.PointerEvent) => void
   onBottomLeftPointerDown: (e: React.PointerEvent) => void
   onRightPointerDown: (e: React.PointerEvent) => void
@@ -21,14 +22,14 @@ export interface IProps extends ITileBaseProps {
 }
 
 export const CodapComponent =
-    observer(({ tile, TitleBar, Component, tileEltClass, onBottomRightPointerDown, onBottomLeftPointerDown,
+    observer(({ tile, TitleBar, Component, tileEltClass, onCloseTile, onBottomRightPointerDown, onBottomLeftPointerDown,
       onRightPointerDown, onBottomPointerDown, onLeftPointerDown }: IProps) => {
   const dataset = gDataBroker?.selectedDataSet || gDataBroker?.last
 
   return (
     <DataSetContext.Provider value={dataset}>
       <div className={`codap-component ${tileEltClass}`}>
-        <TitleBar tile={tile}/>
+        <TitleBar tile={tile} onCloseTile={onCloseTile}/>
         <Component tile={tile} />
         <div className="codap-component-border right" onPointerDown={onRightPointerDown}/>
         <div className="codap-component-border bottom" onPointerDown={onBottomPointerDown}/>
--- a/v3/src/components/codap-component.tsx
+++ b/v3/src/components/codap-component.tsx
@@ -13,8 +13,9 @@ export interface IProps extends ITileBaseProps {
   TitleBar: React.ComponentType<ITileBaseProps>;
   Component: React.ComponentType<ITileBaseProps>;
   tileEltClass: string;
+  onPointerDown: (e: React.PointerEvent) => void
 }
-export const CodapComponent = observer(({ tile, TitleBar, Component, tileEltClass }: IProps) => {
+export const CodapComponent = observer(({ tile, TitleBar, Component, tileEltClass, onPointerDown }: IProps) => {
   const dataset = gDataBroker?.selectedDataSet || gDataBroker?.last
 
   return (
@@ -22,7 +23,9 @@ export const CodapComponent = observer(({ tile, TitleBar, Component, tileEltClas
       <div className={`codap-component ${tileEltClass}`}>
         <TitleBar tile={tile}/>
         <Component tile={tile} />
-        <ResizeHandle className="component-resize-handle"/>
+        <div className="resize-handle-wrapper" onPointerDown={onPointerDown}>
+          <ResizeHandle className="component-resize-handle"/>
+        </div>
       </div>
     </DataSetContext.Provider>
   )
--- a/v3/src/components/codap-component.tsx
+++ b/v3/src/components/codap-component.tsx
@@ -14,11 +14,11 @@ export interface IProps extends ITileBaseProps {
   Component: React.ComponentType<ITileBaseProps>;
   tileEltClass: string;
   onCloseTile: (tileId: string) => void
-  onBottomRightPointerDown: (e: React.PointerEvent) => void
-  onBottomLeftPointerDown: (e: React.PointerEvent) => void
-  onRightPointerDown: (e: React.PointerEvent) => void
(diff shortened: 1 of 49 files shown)


bfinzer (approved):
👍🏻LGTM
I like being able to resize in the free layout especially!

kswenson (changes requested):
Interesting that you can do that much with just CSS, but we need to update the models as well. In the free layout that means updating the width and height and in the mosaic it means updating the split percent.

v3/src/components/free-tile-row.tsx
```diff
@@ -14,24 +14,95 @@ interface IFreeTileRowProps {
...
+
+          resizingWidthLeft = pointerMoveEvent.pageX
+          break
+        case "bottom":
+          resizingHeight = startHeight - startPosition.y + pointerMoveEvent.pageY
+          break
+        case "right":
+          resizingWidth = startWidth - startPosition.x + pointerMoveEvent.pageX
+          break
+      }
```
kswenson, line 65: I think these would make more sense if expressed in terms of `xDelta` and `yDelta`:
```suggestion
      const xDelta = pointerMoveEvent.pageX - startPosition.x
      const yDelta = pointerMoveEvent.pageY - startPosition.y
      switch (direction) {
        case "bottom-right":
          resizingWidth = startWidth + xDelta
          resizingHeight = startHeight + yDelta
          break
        case "bottom-left":
          resizingWidth = startWidth - xDelta
          resizingHeight = startHeight + yDelta
          resizingLeft = startLeft + xDelta
          break
        case "left":
          resizingWidth = startWidth - xDelta
          resizingLeft = startLeft + xDelta
          break
        case "bottom":
          resizingHeight = startHeight + yDelta
          break
        case "right":
          resizingWidth = startWidth + xDelta
          break
      }
```

v3/src/components/free-tile-row.tsx
```diff
@@ -14,24 +14,95 @@ interface IFreeTileRowProps {
...
+
+      setResizingTileStyle({left: resizingWidthLeft, top: tile.y, width: resizingWidth, height: resizingHeight})
+      tempLeft.current = resizingWidthLeft
+      tempWidth.current = resizingWidth
+      tempHeight.current = resizingHeight
+    }
+    const onPointerUp = () => {
+      document.body.removeEventListener("pointermove", onPointerMove)
+      document.body.removeEventListener("pointerup", onPointerUp)
```
kswenson, line 74: I think we should capture the events so as to minimize the chance that any other event handlers can interfere:
```suggestion
      document.body.removeEventListener("pointermove", onPointerMove, { capture: true })
      document.body.removeEventListener("pointerup", onPointerUp, { capture: true })
```

v3/src/components/free-tile-row.tsx
```diff
@@ -14,24 +14,95 @@ interface IFreeTileRowProps {
...
+      tile.setSize(tempWidth.current, tempHeight.current)
+      tile.setPosition(tempLeft.current, tile.y)
+      setResizingTileId("")
+    }
+
+    document.body.addEventListener("pointermove", onPointerMove)
+    document.body.addEventListener("pointerup", onPointerUp)
```
kswenson, line 81: I think we should capture the events so as to minimize the chance that any other event handlers can interfere:
```suggestion
    document.body.addEventListener("pointermove", onPointerMove, { capture: true })
    document.body.addEventListener("pointerup", onPointerUp, { capture: true })
```

v3/src/models/tiles/tile-model.ts
```diff
@@ -63,7 +63,9 @@ export const TileModel = types
     // whether to restrict display to certain users
     display: DisplayUserTypeEnum,
     // e.g. "TextContentModel", ...
-    content: TileContentUnion
+    content: TileContentUnion,
+    height: types.maybe(types.number),
+    width: types.maybe(types.number)
```
kswenson, line 68: These should not be properties of the tile model -- layout information should only be in the layout. Since the only reference to the new methods appears to be in the nascent mosaic implementation, I think this file can be left unchanged as well.

v3/src/components/free-tile-row.tsx
```diff
@@ -14,24 +14,95 @@ interface IFreeTileRowProps {
   getTile: (tileId: string) => ITileModel | undefined
 }
 export const FreeTileRowComponent = observer(({ content, row, getTile }: IFreeTileRowProps) => {
+  const [resizingTileStyle, setResizingTileStyle] =
+    useState<{left: number, top: number, width: number, height: number}>()
+  const [resizingTileId, setResizingTileId] = useState("")
+  const tempWidth = useRef<number>(0)
+  const tempHeight = useRef<number>(0)
+  const tempLeft = useRef<number>(0)
```
kswenson, line 22: When I suggested using refs I was thinking of them as an alternative to the state. Since you're primarily using the state values for the same purpose, I think you can probably get rid of the refs and just use the state values throughout.

v3/src/components/codap-component.tsx
```diff
@@ -14,17 +14,30 @@ export interface IProps extends ITileBaseProps {
   Component: React.ComponentType<ITileBaseProps>;
   tileEltClass: string;
   onCloseTile: (tileId: string) => void
+  onBottomRightPointerDown: (e: React.PointerEvent) => void
+  onBottomLeftPointerDown: (e: React.PointerEvent) => void
+  onRightPointerDown: (e: React.PointerEvent) => void
+  onBottomPointerDown: (e: React.PointerEvent) => void
+  onLeftPointerDown: (e: React.PointerEvent) => void
```
kswenson, line 21: Make these optional since they won't all be appropriate for other layouts, e.g. mosaic
```suggestion
  onBottomRightPointerDown?: (e: React.PointerEvent) => void
  onBottomLeftPointerDown?: (e: React.PointerEvent) => void
  onRightPointerDown?: (e: React.PointerEvent) => void
  onBottomPointerDown?: (e: React.PointerEvent) => void
  onLeftPointerDown?: (e: React.PointerEvent) => void
```

v3/src/components/codap-component.tsx
```diff
@@ -14,17 +14,30 @@ export interface IProps extends ITileBaseProps {
...
         <TitleBar tile={tile} onCloseTile={onCloseTile}/>
         <Component tile={tile} />
-        <ResizeHandle className="component-resize-handle"/>
+        <div className="codap-component-border right" onPointerDown={onRightPointerDown}/>
+        <div className="codap-component-border bottom" onPointerDown={onBottomPointerDown}/>
+        <div className="codap-component-border left" onPointerDown={onLeftPointerDown}/>
+        <div className="codap-component-corner bottom-left" onPointerDown={onBottomLeftPointerDown}/>
+        <div className="codap-component-corner bottom-right" onPointerDown={onBottomRightPointerDown}>
+          <ResizeHandle className="component-resize-handle"/>
+        </div>
```
kswenson, line 40: Make these conditional on whether the handler was provided -- if we don't have a handler then we don't need the DOM element. For mosaic only a subset of them will be necessary.

v3/src/components/mosaic-tile-row.tsx
```diff
@@ -1,10 +1,11 @@
 import { clsx } from "clsx"
 import { observer } from "mobx-react-lite"
-import React from "react"
+import React, { useRef, useState } from "react"
```
kswenson, line 3: Since you're not actually implementing resize for mosaic, I would leave this file unchanged -- or perhaps add a comment about cribbing from the `FreeTileRow` resize implementation. Adding code that doesn't work just seems confusing.

kswenson (changes requested):
👍 This looks (and works) great! I made a bunch of suggestions mostly just to simplify the implementation in a few places and to remove the incomplete mosaic implementation.

v3/src/components/free-tile-row.tsx
```diff
@@ -14,24 +14,87 @@ interface IFreeTileRowProps {
...
+  const handleResizePointerDown = (e: React.PointerEvent, tile: IFreeTileLayout, direction: string) => {
+    const startWidth = tile.width
+    const startHeight = tile.height
+    const startPosition = {x: e.pageX, y: e.pageY}
+
+    let resizingWidth = startWidth, resizingHeight = startHeight, resizingLeft = tile.x
+    // Because user can start drag 8px within the border, the component's startPosition.x moves by number of pixels
+    // the pointer down event location, which moves the entire component to the right by the same number of pixels.
+    // So we force it to always be the left position of the component
+    const startLeft = startPosition.x > tile.x ? tile.x : startPosition.x
```
kswenson, line 35: Given the use of deltas below, I think this should just be `startLeft = tile.x` at this point. Did that not work as expected?

kswenson (approved):
👍 Looks good -- just one lingering question about a possible simplification.

PR #653, opened 2023-02-03
V3—remove y attribute—184382193

bfinzer:
Feature: When the user removes the attribute first added to the vertical axis of a scatterplot, any already added additional attributes adjust appropriately
* Also made it possible to import a V2 document in which the graph has multiple y-attributes on the left vertical axis


kswenson (approved):
👍 LGTM

PR #657, opened 2023-02-09
V3—Fixes bug involving rescaling on y-attribute removal—184382193

bfinzer:
[#184382193] Feature: When the user removes the attribute first added to the vertical axis of a scatterplot, any already added additional attributes adjust appropriately
* Previously the y-axis would not rescale appropriately on attribute removal. This commit fixes that


kswenson (approved):
👍 LGTM

PR #658, opened 2023-02-10
V3: Hierarchical data sets infrastructure

kswenson:
No user-visible changes in this PR, but the infrastructure is in place and there are jest tests for all of the new APIs.
- collections are sets of attributes used to group cases
- grouping occurs on demand in `DataSet` and results are cached
- grouped cases are called pseudo-cases and several `DataSet` APIs have been updated to handle pseudo-cases
  - values can be retrieved from pseudo-cases
  - values can be set for pseudo-cases (which sets the value for all constituent cases)
  - a pseudo-case is selected if all of its constituent cases are selected

Other tangentially related changes:
- moved some types out of `data-set.ts` to `data-set-types.ts`
- in `DataSet`, moved the `caseIDMap` from a private closure to a volatile property for greater visibility (may do the same for other cached values at some point)


bfinzer (approved):
👍🏻LGTM
Very minor requests. I think it's going to be amazing!

PR #659, opened 2023-02-10
V3—rescale bug fix—184431651

bfinzer:
[#184431651] Bug fix: Points disappear from graph for certain configurations
* This was a stale value issue in DotPlotDots and I fixed by using references for attribute IDs instead of simple consts.
* For good measure I made comparable changes in ScatterDots and ChartDots.


v3/src/components/graph/components/chartdots.tsx
```diff
@@ -26,19 +26,23 @@ export const ChartDots = function ChartDots(props: PlotProps) {
   const primaryAttrRole = dataConfiguration?.primaryRole,
     primaryAxisPlace = primaryAttrRole ? attrRoleToAxisPlace[primaryAttrRole] : undefined,
     primaryIsBottom = primaryAxisPlace === 'bottom',
-    primaryAttrID = primaryAttrRole ? dataConfiguration?.attributeID(primaryAttrRole) : '',
+    primaryAttrIDRef = useRef<string>(''),
+    secondaryAttrIDRef = useRef<string>(''),
```
kswenson, line 30: FYI, you don't need `<string>` here as the type can be inferred from the default value:
```suggestion
    primaryAttrIDRef = useRef(''),
    secondaryAttrIDRef = useRef(''),
```

v3/src/components/graph/components/scatterdots.tsx
```diff
@@ -19,12 +19,12 @@ export const ScatterDots = function ScatterDots(props: PlotProps) {
     instanceId = useInstanceIdContext(),
     dataConfiguration = useDataConfigurationContext(),
     dataset = useDataSetContext(),
+    primaryAttrIDRef = useRef<string>(''),
+    secondaryAttrIDsRef = useRef<string[]>([]),
```
kswenson, line 23: You do need the `<string[]>` in the second case because the type can't be inferred from the default value.

kswenson (approved):
👍 LGTM

PR #662, opened 2023-02-15
Handle new collection drop in table

kswenson:
- Implements a new drop zone in the table to handle attribute drops.
- Eventually, this will be used to create new collections, i.e. hierarchical data.
- For now, dropping an attribute just puts up a toast.
- Under the hood: implements a table-specific collision detection method for use with DnDKit


kswenson:
Merged after sync review with @bfinzer.

PR #663, opened 2023-02-16
feat: Display multiple collections and support dragging attributes between collections

kswenson:
Bodies of tables (cases) are still just the same old flat ungrouped cases. Hierarchical display of cases is left for the next PR.


bfinzer (approved):
👍🏻LGTM
Awesome! I don't pretend to fully understand the subtleties. Wish I did.

PR #664, opened 2023-02-17
184464241—V3—flip axes

bfinzer:
[#184464241] Feature: When placing an attribute on x or y axis, if that attribute is already present on the other axis, the attribute placement flips
* The difficult part is getting the animation to work correctly!
* Cleaned up handleAttributeAssignment


kswenson (approved):
👍 LGTM

PR #665, opened 2023-02-19
feat: display case data in hierarchical case table

kswenson:
Basic rendering of hierarchical case data.


bfinzer (approved):
👍🏻LGTM
The collectionGroups method took me quite awhile to understand. But I stepped through it in the debugger and it looks very good!

PR #667, opened 2023-02-22
184535012-crash on treat as categorical-v3

bfinzer:
In GraphController:handleAttributeAssignment, we were getting attribute types directly from the dataset instead of from the data configuration where user setting of type takes place.


kswenson (approved):
👍 Looks good

PR #668, opened 2023-02-23
refactor: CaseTableModel enhancements => SharedCaseMetadata

kswenson:
- support for hidden attributes in `CaseTableModel` (rather than `DataSet`)
- support for collapsed cases in hierarchical table
- cleans up some loose types in the case table code
- uses symbols for special properties in `IGroupedCase` and `TRow` interfaces to avoid conflicts with user/attribute values and eliminate need for `__foo__` type properties. (`__id__` remains -- for now).

@hmorgancode Assigning to you more for education purposes than for review.


bfinzer (approved):
👍🏻LGTM
Review done synchronously

PR #669, opened 2023-02-24
184514913-drop target behavior-V3

bfinzer:
The meat is in DataConfigurationModel:graphPlaceCanAcceptAttributeIDDrop, but each of the different types of drop targets had to be massaged in their own way.


kswenson (changes requested):
👍 Functionality looks good -- I made a couple of minor comments about use of `clsx`, for instance, but the major comment is that it seems like there's a lot of passing around `isDropAllowed` props to various components which all end up calling the same method on the data configuration and it seems like things could be simplified by just calling the data configuration directly in at least some of the cases, as we discussed in standup.

PR #672, opened 2023-03-01
Fix React Hooks linter errors not applying when mobx-state-tree observerComponent is not given a named function

hmorgancode:
Found by converting components into named components for React Dev Tools.

This PR will need to be updated in response to some of Kirk's incoming changes.

### DIFF
--- a/v3/src/components/codap-component.tsx
+++ b/v3/src/components/codap-component.tsx
@@ -22,9 +22,8 @@ export interface IProps extends ITileBaseProps {
 }
 
 export const CodapComponent =
-    observer(function CodapComponent(
-      { tile, TitleBar, Component, tileEltClass, onCloseTile, onBottomRightPointerDown, onBottomLeftPointerDown,
-      onRightPointerDown, onBottomPointerDown, onLeftPointerDown }: IProps) {
+    observer(({ tile, TitleBar, Component, tileEltClass, onCloseTile, onBottomRightPointerDown, onBottomLeftPointerDown,
+      onRightPointerDown, onBottomPointerDown, onLeftPointerDown }: IProps) => {
   const dataset = gDataBroker?.selectedDataSet || gDataBroker?.last
 
   return (
--- a/v3/src/components/codap-component.tsx
+++ b/v3/src/components/codap-component.tsx
@@ -22,8 +22,9 @@ export interface IProps extends ITileBaseProps {
 }
 
 export const CodapComponent =
-    observer(({ tile, TitleBar, Component, tileEltClass, onCloseTile, onBottomRightPointerDown, onBottomLeftPointerDown,
-      onRightPointerDown, onBottomPointerDown, onLeftPointerDown }: IProps) => {
+    observer(function CodapComponent(
+      { tile, TitleBar, Component, tileEltClass, onCloseTile, onBottomRightPointerDown, onBottomLeftPointerDown,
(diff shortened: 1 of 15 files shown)


v3/src/components/mosaic-tile-row.tsx
```diff
@@ -17,7 +17,8 @@ interface IMosaicTileRowProps {
   content?: IDocumentContentModel
   getTile: (tileId: string) => ITileModel | undefined
 }
-export const MosaicTileRowComponent = observer(({ content, row, getTile }: IMosaicTileRowProps) => {
+export const MosaicTileRowComponent = observer(function MosaicTileRowComponent(
+  { content, row, getTile }: IMosaicTileRowProps) {
```
kswenson, line 21:
```suggestion
export const MosaicTileRowComponent = observer(
  function MosaicTileRowComponent({ content, row, getTile }: IMosaicTileRowProps) {
```
Here and elsewhere in this file, as a matter of personal preference, I think I'd be inclined to keep the function declaration together in this situation, but others may fee differently.

v3/src/components/graph/components/graph.tsx
```diff
@@ -40,8 +40,8 @@ interface IProps {
 
 const marqueeState = new MarqueeState()
 
-export const Graph = observer((
-  {graphRef, enableAnimation, dotsRef, showInspector, setShowInspector}: IProps) => {
+export const Graph = observer(function Graph(
+  {graphRef, enableAnimation, dotsRef, showInspector, setShowInspector}: IProps) {
```
kswenson, line 44: Bill's current PR will conflict with this, so I think you should just back out of this change and you or Bill can come back to it later.

v3/src/components/codap-component.tsx
```diff
@@ -22,8 +22,9 @@ export interface IProps extends ITileBaseProps {
 }
 
 export const CodapComponent =
-    observer(({ tile, TitleBar, Component, tileEltClass, onCloseTile, onBottomRightPointerDown, onBottomLeftPointerDown,
-      onRightPointerDown, onBottomPointerDown, onLeftPointerDown }: IProps) => {
+    observer(function CodapComponent(
+      { tile, TitleBar, Component, tileEltClass, onCloseTile, onBottomRightPointerDown, onBottomLeftPointerDown,
+      onRightPointerDown, onBottomPointerDown, onLeftPointerDown }: IProps) {
```
kswenson, line 27: Evangeline's open PR will conflict with this change, so I think you should back out this change and come back to it later.

kswenson (changes requested):
👍 Thanks for pointing these out! Changes look good -- there will be conflicts with both @bfinzer and @eireland's PRs. In the interest of not making things difficult for their more complicated PRs, I'd recommend either waiting until those PRs are merged and then rebasing this PR on top of them, or the simpler approach would be to back out the changes that are likely to conflict and then come back to those components later once the dust has settled. I had thought that there would also be conflicts with my PR, but you didn't make changes in the case table in this PR, so there won't be any conflicts, and I've made the corresponding changes in my PR.

v3/src/components/graph/components/graph.tsx
bfinzer: I will make this change in my current PR to avoid the conflict.

bfinzer (approved):
👍🏻LGTM

PR #676, opened 2023-03-02
Restoration of proper GraphLayoutContext

bfinzer:
* We restore GraphLayoutContext as no longer returning a new GraphLayout
* In GraphComponent we properly create a new GraphLayout to be used as the value given to GraphLayoutContext.Provider
* This had some ramifications down the line

### DIFF
--- a/v3/src/components/axis/hooks/use-axis-bounds.ts
+++ b/v3/src/components/axis/hooks/use-axis-bounds.ts
@@ -1,10 +1,9 @@
 import {useEffect, useState} from "react"
 import {AxisPlace} from "../axis-types"
-import {useGraphLayoutContext} from "../../graph/models/graph-layout"
 import {useAxisLayoutContext} from "../models/axis-layout-context"
 
 export const useAxisBounds = (place: AxisPlace) => {
-  const layout = useGraphLayoutContext()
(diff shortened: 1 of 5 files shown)


v3/src/components/axis/hooks/use-axis-bounds.ts
```diff
@@ -1,14 +1,15 @@
 import {useEffect, useState} from "react"
 import {AxisPlace} from "../axis-types"
 import {useGraphLayoutContext} from "../../graph/models/graph-layout"
+import {useAxisLayoutContext} from "../models/axis-layout-context"
```
kswenson, line 4: The `useGraphLayoutContext` hook should not longer be required, so its import can be removed.

kswenson (approved):
👍 LGTM -- just one stray import to remove

v3/src/components/axis/hooks/use-axis-bounds.ts
bfinzer: Actually it was still, wrongly, being used in useAxisBounds. So I changed useAxisBounds to refer to usAxisLayoutContext instead of useGraphLayoutContext.

PR #678, opened 2023-03-03
refactor: add CollectionProps model; add jest tests

kswenson:
In v2 there is a `Collection` for each section of a hierarchical table. While the parent `Collection`s indicate which attributes should be grouped within the corresponding section -- and thus represent explicit user action that must be serialized, the attributes in the last `Collection` are simply those that are not included in any other `Collection`. Storing them explicitly would therefore be redundant and since our MST models define our serialization format, would also result in bloating our serialized documents and introduce the burden of keeping the redundant information properly synchronized. Therefore, in v3 we have been representing parent `Collection`s explicitly in the `DataSet` but not the last `Collection`.

Unfortunately, this means that we don't have a place to store the user-defined `name` or `title` of the child-most section. 🤔 To solve this problem without introducing the redundancy issues described above, we introduce the `CollectionPropsModel` which contains the metadata for collections, e.g. `name`, `title`, etc., without the `attributes`. The `CollectionPropsModel` serves as the base model for the `CollectionModel` and there is a single instance of it (currently named `ungrouped`, but there may be a better name that hasn't occurred to me yet) in the `DataSet`. 

While in the neighborhood I also took the opportunity to add some unrelated jest tests to increase the coverage of some of our models.


hmorgancode (approved):
LGTM

bfinzer (approved):
👍🏻LGTM

PR #679, opened 2023-03-03
Add basic scrolling to case table

hmorgancode:
Updated to just the one-line functionality change

### DIFF
--- a/v3/src/components/case-table/case-table.scss
+++ b/v3/src/components/case-table/case-table.scss
@@ -12,6 +12,7 @@ $table-body-font-size: 8pt;
     width: 100%;
     height: 100%;
     display: flex;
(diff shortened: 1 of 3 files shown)


v3/src/components/case-table/case-table.scss
```diff
@@ -12,6 +12,7 @@ $table-body-font-size: 8pt;
     width: 100%;
     height: 100%;
     display: flex;
+    overflow-x: scroll;
```
kswenson, line 15:
```suggestion
    overflow-x: auto;
```
We don't need the scroll bar unless there's something to scroll.

kswenson (changes requested):
👍 Looks like the right place for the change, but I think `auto` would be preferable to `scroll`.

PR #680, opened 2023-03-04
184210651 new tiles

eireland:
Adds ability to add new graph, table, and slider components to the document from the toolshelf.
Note that both graph and table open with data from the current data set only. The table opens a table component directly instead of through a menu. 
Components open staggered from the first component of its type instead of finding an open space in the document.


v3/src/components/tool-shelf/tool-shelf.tsx
```diff
@@ -11,7 +17,24 @@ import CalcIcon from '../../assets/icons/icon-calc.svg'
 import TextIcon from '../../assets/icons/icon-text.svg'
 import PluginsIcon from '../../assets/icons/icon-plug.svg'
 
-export const ToolShelf = () => {
+import './tool-shelf.scss'
+
+const kCalcHeight = 162
```
bfinzer, line 22: Constants that only pertain to a particular component should live in its handler.

bfinzer, line 22: Rather than import information from various files, I think it would be better to consolidate all the information for tiles of different types in the registry and pull the information out with getTileComponentInfo.

v3/src/components/tool-shelf/tool-shelf.tsx
```diff
@@ -21,14 +44,83 @@ export const ToolShelf = () => {
...
-    tableHandler = () => notify('table'),
-    graphHandler = () => notify('graph'),
     mapHandler = () => notify('map'),
-    sliderHandler = () => notify('slider'),
-    calcHandler = () => notify('calculator'),
     textHandler = () => notify('text'),
     pluginsHandler = () => notify('plugins')
 
+  const row = content?.getRowByIndex(0)
+
```
bfinzer, line 52: I wonder if the handlers for different types of tiles could be stored in the global registry? Basically, could the ToolShelf iterate through the registry without having to know anything about individual types of components itself? That way, adding a new type of component could be nearly completely confined to adding a new entry to the registry?
I'm not sure about all this. Might be worth an architecture conversation that includes at least Kirk and maybe Henry.
Meanwhile, we don't need to prevent this PR in lieu of the conversation.

v3/src/models/document/document-content.ts
```diff
@@ -49,6 +49,10 @@ export const DocumentContentModel = types
     getTile(tileId: string) {
       return tileId && self.tileMap.has(tileId) ? self.tileMap.get(tileId) : undefined
     },
+    hasTileOfType(type: string) {
+      const tileMapEntries = Array.from(self.tileMap.values())
+      return tileMapEntries.filter(tile => tile.content.type === type)
```
bfinzer, line 54: Rather than return an array, the function name suggests we want a boolean.

bfinzer (approved):
👍🏻Looks good for now
I made a few simple suggestions.
But we should have a larger architecture discussion about the Component/Tile registry and the possibility of localizing more information there. It's possible the toolshelf might not need to know much of anything about how to create a particular type of tile.

kswenson (changes requested):
I only took a quick look at the PR but I agree with @bfinzer's comments that there should be some decoupling of the tool shelf from the requirements of the individual tiles. Perhaps we should get together and figure out how best to do that.

v3/src/components/tool-shelf/tool-shelf.tsx
```diff
@@ -20,64 +26,96 @@ export const ToolShelf = () => {
...
+  const createTile = (tileType: string, componentInfo: ITileComponentInfo) => {
+    if (row) {
+      const newTile = createDefaultTileOfType(tileType)
+      if (newTile) {
+        const tileOptions = { width: componentInfo.width, height: componentInfo.height }
+        content?.insertTileInRow(newTile, row, tileOptions)
+        const rowTile = row.tiles.get(newTile.id)
+        if (componentInfo.width && componentInfo.height) {
+          rowTile?.setSize(componentInfo.width,  componentInfo.height + kHeaderHeight)
```
kswenson, line 51:
```suggestion
          rowTile?.setSize(componentInfo.width, componentInfo.height + kHeaderHeight)
```

v3/src/models/tiles/tile-component-info.ts
```diff
@@ -27,7 +34,12 @@ export function registerTileComponentInfo(tileComponentInfo: ITileComponentInfo)
...
 // undefined is supported so callers do not need to check the id before passing it in
+// toLowerCase() for legacy support of tool names
 export function getTileComponentInfo(type?: string) {
-  // toLowerCase() for legacy support of tool names
   return type ? gTileComponentInfoMap.get(type.toLowerCase()) : undefined
 }
+
+export function getTileComponentIcon(type?: string) {
+  const cIcon = type && gTileComponentInfoMap.get(type.toLowerCase())?.Icon
+  return cIcon
```
kswenson, line 44:
```suggestion
  return getTileComponentInfo(type)?.Icon
```
and then you can put the `toLowerCase()` comment back where it was above.

v3/src/components/tool-shelf/tool-shelf.tsx
```diff
@@ -90,7 +128,7 @@ export const ToolShelf = () => {
               as='button'
               key={aDesc.iconLabel}
               bg='white'
-              onClick={aDesc.handler}
+              onClick={()=> createComponentHandler(aDesc.tileType)}
```
kswenson, line 131:
```suggestion
              onClick={() => createComponentHandler(aDesc.tileType)}
```

kswenson (approved):
👍 Looks good -- just a couple of small suggestions

v3/src/components/tool-shelf/tool-shelf.tsx
```diff
@@ -20,64 +26,96 @@ export const ToolShelf = () => {
...
-    tableHandler = () => notify('table'),
-    graphHandler = () => notify('graph'),
-    mapHandler = () => notify('map'),
-    sliderHandler = () => notify('slider'),
-    calcHandler = () => notify('calculator'),
-    textHandler = () => notify('text'),
-    pluginsHandler = () => notify('plugins')
+    toast = useToast()
+
+  const row = content?.getRowByIndex(0) as IFreeTileRow
```
kswenson, line 31: I forgot to mention, but we should at least nominally continue to support the mosaic layout, so rather than casting here we should probably do a dynamic type test.

PR #681, opened 2023-03-06
Fix linting for root-level config files

hmorgancode:
Everything except cypress.config.ts should lint properly now- that file needs another pass to fix it up.

### DIFF
--- a/v3/.eslintrc.cjs
+++ b/v3/.eslintrc.cjs
@@ -138,7 +138,8 @@ module.exports = {
       }
     },
     { // eslint configs
-      files: [".eslintrc*.js"],
+      // *.config.ts should be added after cypress.config.ts lint fixes
+      files: [".eslintrc*.cjs", "./*.config.js", "why-did-you-render.ts"],
       env: {
         node: true
       }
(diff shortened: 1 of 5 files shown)


v3/.eslintrc.cjs
```diff
@@ -138,7 +138,8 @@ module.exports = {
       }
     },
     { // eslint configs
-      files: [".eslintrc*.js"],
+      // *.config.ts should be added after cypress.config.ts lint fixes
```
kswenson, line 141: The comment in line 140 should be updated now that more than eslint configs are addressed.

v3/postcss.config.js
```diff
@@ -1,5 +1,7 @@
+/* eslint-disable @typescript-eslint/no-require-imports */
```
kswenson, line 1: Is there a reason this can't be a single-line disable at the point of the `require`? That would be more conventional.

kswenson (approved):
👍 Looks good -- just a couple comment-related suggestions

v3/postcss.config.js
hmorgancode: Good point, I switched over.

v3/.eslintrc.cjs
hmorgancode: 👍

PR #683, opened 2023-03-07
Make title bars show name of sample used

hmorgancode:
Per #184540309, makes title bars show the name of the sample.

@kswenson raised the possibility of merging the disparate *-title-bar components into one, but there's just enough individuality on display that I'd rather that work be done as part of making the title bars editable. (This PR makes them uneditable, but the framework Evangeline put into place is still there, waiting to be used.)

I have e2e tests for Graph and Table, should I add new test files for Calculator, DataSummary, Hello, Slider, and Placeholder? None of these appear to have e2e runs associated with them and this would be a good opportunity to add them.


v3/src/components/calculator/calculator-title-bar.tsx
```diff
@@ -6,11 +6,12 @@ import MinimizeIcon from "../../assets/icons/icon-minimize.svg"
 import { ITileTitleBarProps } from "../tiles/tile-base-props"
 
 export const CalculatorTitleBar = ({tile, onCloseTile}: ITileTitleBarProps) => {
-  const [title, setTitle] = useState("Calculator")
+  const [customTitle, setCustomTitle] = useState<string | null>(null)
+  const title = customTitle ?? "Calculator"
```
kswenson, line 10: Here and in the other components, instead of using React state here, store it in the `TileModel`:
```suggestion
  const title = tile.title || t("DG.DocumentController.calculatorTitle")
```
Also note that this:
1. retrieves the default title from our localized strings
2. uses `||` instead of `??` so that empty strings will be replaced with the default

v3/src/components/case-table/case-table-title-bar.tsx
```diff
@@ -12,13 +12,14 @@ import "./case-table-title-bar.scss"
 
 export const CaseTableTitleBar = ({tile, onCloseTile}: ITileTitleBarProps) => {
   const dataset = useDataSetContext()
-  const [title, setTitle] = useState(dataset?.name || "Dataset")
+  const [customTitle, setCustomTitle] = useState<string | null>(null)
   const [showSwitchMessage, setShowSwitchMessage] = useState(false)
   const [showCaseCard, setShowCaseCard] = useState(false)
+  const title = customTitle ?? (dataset?.name || "Dataset")
```
kswenson, line 18:
```suggestion
  const title = customTitle ?? (dataset?.name || t("DG.AppController.createDataSet.name"))
```

kswenson (changes requested):
👍 Generally looks good -- I made a bunch of suggestions, but they generally boil down to a few types:
1. Use the `TileModel`'s `title` property rather than React state for user-edited titles -- it's required for those changes to be serialized with the document.
2. Use localized strings from the string table for the default names. The use of hard-coded strings obviously precedes your arrival, but since you're in the neighborhood this seems like a good time to fix it.
3. Use `||` instead of `??` when defaulting the title so that we will show the default title preferentially over empty strings.

tejal-shah (approved):
Reviewed only for cypress tests and they seem fine to me 👍

PR #684, opened 2023-03-07
Make case-table tile expansion have a consistent background color

hmorgancode:
Follow-up on #184553678, which was fixed by the scroll: auto; I added but which left a transparent background when a case table was expanded beyond its bounds.

### DIFF
--- a/v3/src/components/case-table/case-table.scss
+++ b/v3/src/components/case-table/case-table.scss
@@ -13,6 +13,7 @@ $table-body-font-size: 8pt;
     height: 100%;
     display: flex;
     overflow-x: auto;
+    background-color: #FFFFFF;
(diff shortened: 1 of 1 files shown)


v3/src/components/case-table/case-table.scss
```diff
@@ -13,6 +13,7 @@ $table-body-font-size: 8pt;
     height: 100%;
     display: flex;
     overflow-x: auto;
+    background-color: #FFFFFF;
```
kswenson, line 16:
```suggestion
    background-color: white;
```
A bit more readable than hex colors.

kswenson (approved):
👍 LGTM -- one minor readability suggestion

PR #685, opened 2023-03-08
refactor: improve CollectionPropsModel integration

kswenson:
Instead of providing a `CollectionModel` or `undefined`, the `CollectionContext` now provides either a `CollectionModel` or a `CollectionPropsModel` for the child-most collection.


hmorgancode (approved):
LGTM- I see you moved some types, I have a related question on the pluralize package in my react-data-grid PR. I assume from your PR that we'd prefer to keep @types/ in devDependencies, and reconfigure or silent eslint to not warn unnecessarily.

PR #686, opened 2023-03-08
feat: implement focused tile; title bar, inspector, and resize UI based on focused tile

kswenson:
Implements a new `UIState` class which maintains an observable `focusTile` property which tracks the actively focused tile. Clicking on a tile or focusing an element within a tile (e.g. tabbing into a tile) sets the property accordingly. Rendering (or not) of focus-only UI such as inspectors, resize widgets, etc. is now determined by this `focusTile` property. The focused tile is also automatically brought to the front in the free tile layout. Also cleaned up the Hello tile somewhat so it works better as a potentially overlapping tile.

### DIFF
--- a/v3/src/components/app.scss
+++ b/v3/src/components/app.scss
@@ -13,14 +13,6 @@
   .app {
     width: 100%;
     height: 100%;
-
-    .hello-codap3 {
-      width: 100%;
-      height: 100%;
-      display: flex;
-      align-items: center;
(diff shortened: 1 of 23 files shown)


hmorgancode (approved):
LGTM (though I'm assuming the use of the @observable tags is correct)

bfinzer (approved):
👍🏻LGTM

PR #687, opened 2023-03-08
Upgrade react-data-grid to 7.0.0-beta.26

hmorgancode:
There's one question RE: where to put @types/ packages in package.json that I left in a comment.


v3/src/components/case-table/collection-table.tsx
```diff
@@ -1,5 +1,8 @@
 import React, { useRef } from "react"
 import DataGrid, { DataGridHandle } from "react-data-grid"
+// eslint wants us to add @types/pluralize to our deps instead of devDependencies
```
hmorgancode, line 3: Looking for feedback on this

v3/src/components/case-table/collection-table.tsx
```diff
@@ -1,5 +1,8 @@
...
+// Are we okay with moving types into our main deps, or do we just want to suppress this warning?
+// eslint-disable-next-line import/no-extraneous-dependencies
```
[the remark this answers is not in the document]
kswenson: I did not encounter any lint issues when I moved it to `devDependencies` in my PR. Types are logically `devDependencies` because they're not relevant at run-time. My preference would be to leave it in `devDependencies` and figure out why the linter is behaving differently for you than for me.

v3/src/components/case-table/use-selected-rows.ts
```diff
@@ -119,7 +119,8 @@ export const useSelectedRows = ({ gridRef }: UseSelectedRows) => {
   // anchor row for shift-selection
   const anchorCase = useRef<string | null>(null)
 
-  const handleRowClick = useCallback(({ __id__: caseId }: TRow) => {
+  const handleCellClick = useCallback(
+  ({ row: { __id__: caseId } }: CellClickArgs<TRow>, event: MouseEvent<HTMLDivElement>) => {
```
kswenson, line 123: Two things:
1. Where RDG templatized types are concerned, the convention is to add new types to `case-table-types.ts`, so I would add `TCellClickArgs = CellClickArgs<TRow>` there for consistency.
2. Since they're passing in the event now, the `isExtending` definition below should be based on the event. The existing code was written to work around the fact that the event wasn't available in earlier versions of the API.

kswenson (changes requested):
👍 Looks good -- I suggested a couple minor changes related to the `HandleCellClick` API change and we should figure out what's going on with the spurious linter warnings.

v3/src/components/case-table/case-table-types.ts
```diff
@@ -1,5 +1,12 @@
 import {
-  CalculatedColumn, Column, EditorProps, FormatterProps, HeaderRendererProps, RowRendererProps, RowsChangeData
+  CalculatedColumn,
```
kswenson, line 2: FYI, I've never liked this kind of formatting because it wastes so much vertical real estate. I'd use something like:
```
  CalculatedColumn, CellClickArgs, Column, EditorProps, FormatterProps, HeaderRendererProps,
  RowRendererProps, RowsChangeData
```

kswenson (approved):
👍 LGTM

hmorgancode:
(cypress passed locally, and the remote flake was why codecov was low. so, I'm comfortable merging this)

PR #690, opened 2023-03-13
feat: typed ids for more readable document export

kswenson:
Replaces 16-char random unique ids with 16-char unique ids with 4-char prefixes and 12-char unique suffixes. This decreases entropy somewhat but makes exported documents more developer-friendly.


bfinzer (approved):
👍🏻LGTM
I like it!

PR #691, opened 2023-03-14
184667117 find open area - V3

eireland:
Finds an empty space in the document that will fit the new component. If there are no empty spaces that will fit the component, the new component will open in the center.
New components opened in the center are staggered regardless of type.


bfinzer (approved):
👍🏻LGTM
Glad to see the V2 algorithm in use!

kswenson (approved):
👍 Looks good -- I made a few minor suggestions and I think the staggering algorithm could use some tweaking but this seems like a reasonable start.

PR #692, opened 2023-03-14
Removes resize handles from calculator.

eireland:
Adds default behavior for use resizable in tile model

### DIFF
--- a/v3/src/components/calculator/calculator-model.ts
+++ b/v3/src/components/calculator/calculator-model.ts
@@ -6,7 +6,6 @@ export const CalculatorModel = TileContentModel
   .named("CalculatorModel")
   .props({
     type: types.optional(types.literal(kCalculatorTileType), kCalculatorTileType),
-    isUserResizable: false
   })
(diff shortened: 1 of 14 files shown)


v3/src/components/calculator/calculator-model.ts
```diff
@@ -5,7 +5,8 @@ import { kCalculatorTileType } from "./calculator-defs"
 export const CalculatorModel = TileContentModel
   .named("CalculatorModel")
   .props({
-    type: types.optional(types.literal(kCalculatorTileType), kCalculatorTileType)
+    type: types.optional(types.literal(kCalculatorTileType), kCalculatorTileType),
+    isUserResizable: false
```
kswenson, line 9: I know you haven't asked for review yet, but I think this should be registered with the component info rather than stored with the model given that it doesn't change from one tile instance to another.

kswenson (changes requested):
👍 Looks good -- unfortunately, now that I see `isUserResizable` in action, I think it would be better to make it an optional `isFixedSize` property so that it's `true` for the calculator and can be left undefined for the other tiles. Sorry 😬

PR #695, opened 2023-03-17
feat: import calculator; better support for non-resizable tiles

kswenson:
- supports import of calculator tiles from v2 documents
- free tile layout does not impose sizes (imported or otherwise) on non-resizable tiles
- non-resizable tiles render at their "natural" size without constraints imposed from above

### DIFF
--- a/v3/src/components/free-tile-component.tsx
+++ b/v3/src/components/free-tile-component.tsx
@@ -2,7 +2,6 @@ import { useDndContext } from "@dnd-kit/core"
 import React, { useState } from "react"
 import { getDragTileId, IUseDraggableTile, useDraggableTile } from "../hooks/use-drag-drop"
 import { IFreeTileLayout, IFreeTileRow, isFreeTileRow } from "../models/document/free-tile-row"
-import { getTileComponentInfo } from "../models/tiles/tile-component-info"
 import { ITileModel } from "../models/tiles/tile-model"
 import { CodapComponent } from "./codap-component"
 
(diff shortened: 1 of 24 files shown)


v3/src/v2/codap-v2-types.ts
```diff
@@ -154,6 +161,13 @@ export interface ICodapV2BaseComponent {
   savedHeight: number | null
 }
 
+export interface ICodapV2CalculatorComponent extends ICodapV2BaseComponent {
+  type: "DG.Calculator"
+  componentStorage: ICodapV2CalculatorStorage
+}
+export const isV2CalculatorComponent = (component: ICodapV2BaseComponent): component is ICodapV2CalculatorComponent =>
```
hmorgancode, line 168: I hadn't encountered type predicates / the `is` keyword before... neat!

hmorgancode (approved):
LGTM with one small change.

v3/src/components/free-tile-component.tsx
```diff
@@ -50,26 +48,17 @@
       setResizingTileId(mtile.tileId)
```
bfinzer, line 48: In V2 resizing components is not continuous but has a delta that aids the user in lining things up nicely. Is this the place that increment should be used?

v3/src/v2/codap-v2-types.ts
```diff
@@ -84,6 +84,13 @@ export interface IGuidLink<T extends string> {
   id: number
 }
 
+export interface ICodapV2CalculatorStorage {
+  title: string
+  name: string
+  userSetTitle: boolean
+  cannotCLose: boolean
```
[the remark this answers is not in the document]
kswenson: Good catch -- will fix.

v3/src/components/free-tile-component.tsx
kswenson: Yes -- do you happen to recall the v2 grid size?

kswenson: Never mind; I found it. The v2 grid size is 5px, specified in `view_utilities.js`.

kswenson: In the interest of not holding up this PR I created a new story for the gridding behavior.

PR #697, opened 2023-03-17
Update node to use version 16, Add timeout on npm start to avoid cyress job to timeout

tejal-shah:


kswenson (approved):
👍 LGTM, particularly if the timeout means fewer cypress test failures. Worth noting that in an upcoming PR (#698) @hmorgancode proposes updating the node version to 18. I don't have a particular preference, but if there's a reason to prefer 16 over 18 that should factor into the review of that PR.

tejal-shah:
@hmorgancode , @kswenson , I was only trying to get rid of the node 12 warnings which can be achieved with versions higher than 12. I'll update to use 18 as @hmorgancode suggests.

PR #700, opened 2023-03-20
Update dependencies; eliminate vulnerability warnings

kswenson:
Updated all dependencies except two:
- TypeScript 5 is incompatible with some of our other dependencies at this point (notably Chakra) and so we will have to wait for those dependencies to be updated.
- Cypress 12 -- our tests fail under cypress 12 and I did not spend time investigating further so I made a story about it.

Note: I also had to add `fs-extra` as an explicit dependency as we use it in `cypress.config.ts` but were apparently inadvertently relying on a transitive dependency which no longer worked after updating.


v3/package.json
```diff
@@ -83,95 +83,96 @@
...
     "sass-loader": "^13.2.0",
     "script-loader": "^0.7.2",
-    "style-loader": "^3.3.1",
-    "ts-jest": "^29.0.3",
+    "style-loader": "^3.3.2",
+    "ts-jest": "^29.0.5",
     "ts-loader": "^9.4.2",
     "ts-node": "^10.9.1",
-    "typescript": "^4.9.4",
+    "typescript": "~4.9.5",
```
kswenson, line 144: Switching to `~` here as TypeScript doesn't following semantic versioning and so "minor" versions can contain breaking changes.

v3/package.json
```diff
@@ -83,95 +83,97 @@
...
+    "@emotion/react": "^11.10.6",
+    "@emotion/styled": "^11.10.6",
     "clsx": "^1.2.1",
-    "d3": "^7.8.0",
+    "d3": "^7.8.2",
     "d3-v6-tip": "^1.0.9",
     "escape-string-regexp": "^5.0.0",
     "expr-eval": "^2.0.2",
-    "framer-motion": "^8.1.7",
+    "framer-motion": "^10.6.0",
```
hmorgancode, line 161: I doublechecked because of the jump in major versions- it doesn't look like this is imported anywhere in v3. Can we remove it?

hmorgancode (approved):
LGTM with one possible change.

v3/package.json
kswenson: Good question! `framer-motion` and `@emotion/*` were added following the [@chakra-ui installation instructions](https://chakra-ui.com/getting-started), but since they're listed as `peerDependencies`, they're added automatically, so there's no reason for us to list them explicitly among our dependencies. I've removed them in the most recent commit.

PR #706, opened 2023-03-22
Dynamic inspector palette top

eireland:
Dynamically changes inspector palette top depending on whether the palette will be in bounds.

### DIFF
--- a/v3/src/components/inspector-panel.tsx
+++ b/v3/src/components/inspector-panel.tsx
@@ -2,7 +2,7 @@ import { forwardRef, Box, Button, Menu, MenuButton } from "@chakra-ui/react"
 import React, { ReactNode, RefObject, useEffect, useRef, useState } from "react"
 import MoreOptionsIcon from "../assets/icons/arrow-moreIconOptions.svg"
 import { useOutsidePointerDown } from "../hooks/use-outside-pointer-down"
-import { isWithinBounds } from "../utilities/view-utils"
+import { isWithinBounds, getPaletteTopPosition } from "../utilities/view-utils"
 
 import "./inspector-panel.scss"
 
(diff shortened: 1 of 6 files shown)


kswenson (changes requested):
👍 Looks good -- I like getting rid of the `paletteTop` props in favor of a utility function. As I suggested in one of the comments below, I do think it's worth considering `position: relative` on the `.codap.container` before merging.

v3/src/components/inspector-panel.tsx
```diff
@@ -72,20 +72,27 @@
...
   const [paletteWidth, setPaletteWidth] = useState(0)
   const paletteRef = useRef<HTMLDivElement>(null)
+  const pointerRef = useRef<HTMLDivElement>(null)
   const viewportEl = document.getElementsByClassName("free-tile-row")?.[0]
   const [inBounds, setInBounds] = useState(isWithinBounds(paletteWidth, panelRect))
+  const paletteHeight = paletteRef.current?.offsetHeight
+  const tempPaletteTop = paletteRef.current?.getBoundingClientRect().top
+  const pointerTop = buttonTop - panelTop - 5
+  const pointerHeight = pointerRef.current?.offsetHeight
+  const pointerMidpoint = pointerHeight ? pointerTop + pointerHeight/2 : 0
```
bfinzer, line 93: I believe that if pointerHeight === 0, pointerMidpoint will be assigned the value of 0. Is that what you want?

bfinzer (approved):
👍🏻LGTM
Just one quibble.

v3/src/utilities/view-utils.ts
```diff
@@ -100,3 +100,12 @@ export const isWithinBounds = (dimension: number, rect?: DOMRect) => {
   const viewportRect = viewportEl.getBoundingClientRect()
   return (rect.right + dimension + 10) < viewportRect?.right
 }
+
+export const getPaletteTopPosition = (tempTop: number, paletteHeight: number, pointerMid: number) => {
+  const top = pointerMid-paletteHeight/2
+  if ((tempTop+top) > kCodapAppHeader + kGap) {
```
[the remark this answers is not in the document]
kswenson: My suggestion of adding `position: relative` to `.codap-container` was based on the assumption that the issue was that we had to position the palette in window coordinates. I now see that the issue is that `tempTop` and `pointerMid` are in window coordinates. I'm wondering if at some point it would make more sense to pass DOM elements around rather than coordinates in different coordinate systems, but probably not worth doing anything about it right now.

v3/src/components/inspector-panel.tsx
```diff
@@ -72,20 +72,27 @@ interface IInspectorPalette {
...
+  const paletteHeight = paletteRef.current?.offsetHeight
+  const tempPaletteTop = paletteRef.current?.getBoundingClientRect().top
+  const pointerTop = buttonTop - panelTop - 5
+  const pointerHeight = pointerRef.current?.offsetHeight
+  const pointerMidpoint = pointerHeight ? pointerTop + pointerHeight/2 : 13
+  const paletteTop = (tempPaletteTop && paletteHeight) &&
+    getPaletteTopPosition(tempPaletteTop, paletteHeight, pointerMidpoint)
 
   useEffect(()=> {
     const observer = new ResizeObserver(entries => {
```
kswenson, line 98: Two things here:
1. I think you need to deal with the possibility of `viewportEl` being `undefined` or `null`.
2. You should dispose of the observer in the return value of the effect.

Something like:
```
    const observer = viewportEl && new ResizeObserver(entries => {
      entries.forEach(entry => {
        if (panelRect) {
          setInBounds(isWithinBounds(paletteWidth, panelRect))
        }
      })
    })
    observer?.observe(viewportEl)
    return () => observer?.disconnect()
```

kswenson (changes requested):
👍 Looks good -- one more suggestion about the `ResizeObserver`

v3/src/components/inspector-panel.tsx
```diff
@@ -72,30 +72,38 @@ interface IInspectorPalette {
...
   useEffect(()=> {
-    const observer = new ResizeObserver(entries => {
+    const observer = viewportEl && new ResizeObserver(entries => {
       entries.forEach(entry => {
         if (panelRect) {
           setInBounds(isWithinBounds(paletteWidth, panelRect))
         }
       })
     })
     observer.observe(viewportEl)
```
kswenson, line 105:
```suggestion
    observer?.observe(viewportEl)
```

kswenson (approved):
👍 Changes look good -- just one more optional chaining question mark and it's good to merge.

PR #709, opened 2023-03-23
ci: Update to use 's3-deploy-action' & 'ubuntu-latest' (PT-184551773)

433eros:
Addresses PT-184551773

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -54,24 +54,345 @@
         "webpack-shell-plugin": "^0.5.0"
       }
     },
-    "node_modules/@babel/code-frame": {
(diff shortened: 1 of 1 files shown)


kswenson (commented):
The v2 build of CODAP is not like other CC projects. It has a very specialized build procedure that requires ruby and multiple steps. The ci build of v2 hasn't been working for some time, so I would love it for that problem to be addressed, but it will require a deep engagement with and understanding of the v2 build process. The straightforward approach used with most CC projects will not be sufficient in this case.

kswenson:
Superseded by #2379.

PR #712, opened 2023-03-24
Adds snap to grid using dnd-kit's modifiers

eireland:
Uses a custom version based on dnd-kits' snap to grid modifier that specifies grid size depending on item being dragged. Specifically, it dragged item is a tile, then grid size is 5 pixels. All other items should have grid size of 1 pixel.


v3/src/components/codap-dnd-context.tsx
```diff
@@ -17,9 +18,10 @@ export const CodapDndContext = ({ children }: IProps) => {
                     useSensor(KeyboardSensor, { coordinateGetter: customCoordinatesGetter }),
                     // mouse sensor can be enabled for cypress tests, for instance
                     urlParams.mouseSensor !== undefined ? useMouseSensor : null)
-
+  const gridSize = 5 // pixels
+  const snapToGridModifier = createSnapModifier(gridSize)
```
kswenson, line 22: This is going to apply the snapping behavior to all drags. We only want to apply the snapping behavior to tile drags. `Modifiers` take a whole bunch of arguments that can be used to adjust the returned transform: https://github.com/clauderic/dnd-kit/blob/master/packages/core/src/modifiers/types.ts.

kswenson (commented):
There was some confusion over `lockFileVersion`. @tejal-shah's latest PR updated us to `lockFileVersion: 3` and then this PR was bringing us back down to `lockFileVersion: 2`. I force-pushed a single commit to your PR branch which makes the minimal changes and keeps us at `lockFileVersion: 3` and the ci build passes. I made my changes with node 16 in case that matters, although you may want to update your node version either to the latest version of 16 (v16.19.1) or 18 (v18.15) just in case.

bfinzer (approved):
👍🏻LGTM
Note that the fully PT story (just modified by me) includes a snap-to-grid for resize as well as move. If it's easy, you might want to take care of that in this pull request as well.

kswenson (approved):
👍 Changes look good -- make sure you use squash merge since you've added back the trouble-making commits that I eliminated.

kswenson:
>Note that the fully PT story (just modified by me) includes a snap-to-grid for resize as well as move. If it's easy, you might want to take care of that in this pull request as well.

I changed the PT story to suggest that gridding resize should be a separate story, since the implementation is unrelated to the implementation here.

PR #713, opened 2023-03-24
feat: add GlobalValueManager; fix slider

kswenson:
- define new `GlobalValueManager`
- each document has a single `GlobalValueManager` instance
- each new slider creates a new `GlobalValue` and registers it with the `GlobalValueManager`
- `GlobalValueManager` is serialized with other shared models in the document
- v3 sliders import from v2 documents correctly
- cleaned up `SliderModel`
- added tests for calculator and slider tile imports
- re-enabled `case-table-component.test.ts`
- show slider name in title bar of slider component


bfinzer (approved):
👍 LGTM
The architecture behind shared data models is amazing!

PR #718, opened 2023-03-27
V3: Refactors close and minimize buttons to the component title bar

eireland:
Encloses minimize icon into a button
Changes most of the components to use the generic Component Title Bar, except for slider and table.


kswenson (approved):
👍 Looks great -- the one wrinkle that I was going to suggest was to move the title access into the `ComponentTitleBar` so that only it re-renders on a title change. If the parent accesses the title and then passes it to the `ComponentTitleBar` then the entire parent component re-renders on a title change. I wanted to prove to myself that the suggestion made sense, however, so I ended up implementing it along with a jest test to verify that it works as expected. The key was to pass in a `getTitle()` function so that the model access occurs in the `ComponentTitleBar`.

PR #721, opened 2023-03-28
Use SWC loader to improve webpack performance

hmorgancode:
**This appears to be breaking in CI because I didn't adapt the changes to code coverage. Still, try this branch out and see if your iteration time goes down- if it's noticeable, we can move forward with this.

SWC is a Rust-based, native JS compiler with significant performance gains (~17x faster) over Webpack. We can't commit fully to SWC because the ecosystem just isn't there yet, but we can use the SWC webpack loader within our current setup to compile our typescript files faster.

I did not notice a decrease to overall build time, but I did observe this reducing the iteration time for a ts/tsx change from 4 seconds down to .3 seconds, so please tell me if this is an improvement for you.

https://github.com/swc-project/swc-loader
https://swc.rs/docs/configuration/swcrc


hmorgancode:
Closing this as Kirk pointed out SWC provides no type checking. Will review when https://github.com/dudykr/stc (rust type checker) is further along, or if dev becomes slow enough that we run SWC and typecheck separately

PR #722, opened 2023-03-28
perf: combine category moves if possible

kswenson:
This optimization was discussed during sync review of #715 with @bfinzer, but not implemented at that time. If the user changes the order of a single category multiple times in succession, we combine those moves into a single move operation.


bfinzer (approved):
👍🏻LGTM

PR #723, opened 2023-03-28
Case Table expand/collapse button UI

hmorgancode:
Still due for some Cypress tests but I think it's ready for review.


kswenson (changes requested):
👍 Mostly looks good -- I made a number of suggestions several of which revolve around
1. simplifying the `CollectionTableSpacer` implementation
2. using localized strings where possible
Separately, the yellow drag feedback message is interacting with the buttons and the column headers in undesirable ways, but perhaps that is best dealt with in a subsequent PR.

PR #724, opened 2023-03-28
V3: Makes slider variable name and value inputs behave more like V2

eireland:
Refactors slider input to use Chakra inputs to handle styling and controls.
A long variable name wraps the variable value to the next line like in V2. The variable name still goes beyond the slider component like in V2
The one difference is the name and value inputs are two separate inputs instead of one. You can tab from one input to the other.

### DIFF
--- a/v3/src/components/slider/slider-component.tsx
+++ b/v3/src/components/slider/slider-component.tsx
@@ -1,13 +1,12 @@
 import React, { CSSProperties, useEffect, useMemo, useRef, useState } from "react"
 import { useResizeDetector } from "react-resize-detector"
-import { Flex, Center } from "@chakra-ui/react"
+import { Flex, Editable, EditablePreview, EditableInput, Button } from "@chakra-ui/react"
 import { observer } from "mobx-react-lite"
 import PlayIcon from "../../assets/icons/icon-play.svg"
 import PauseIcon from "../../assets/icons/icon-pause.svg"
 import { SliderAxisLayout } from "./slider-layout"
 import { isSliderModel } from "./slider-model"
 import { kSliderClass, kSliderClassSelector } from "./slider-types"
-import { measureText } from "../../hooks/use-measure-text"
 import { Axis } from "../axis/components/axis"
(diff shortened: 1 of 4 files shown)


v3/src/components/slider/slider-component.tsx
```diff
@@ -67,58 +64,45 @@ export const SliderComponent = observer(function SliderComponent({ tile } : ITil
...
-                        style={{width: `${titleM + 2 + (titleM * .25)}px`, paddingLeft: "3px"}}
-                      />
-                    : <div onClick={() => setIsEditingName(true)}>
-                        {sliderModel.name}
-                      </div>
-                  }
-
+              <Flex>
+                <Button className={`play-pause ${ running ? "running" : "paused"}`} onClick={toggleRunning}>
+                    { running ? <PauseIcon /> : <PlayIcon /> }
```
kswenson, line 79:
```suggestion
                  { running ? <PauseIcon /> : <PlayIcon /> }
```

v3/src/components/slider/slider-component.tsx
```diff
@@ -67,58 +64,45 @@ export const SliderComponent = observer(function SliderComponent({ tile } : ITil
...
+                </Button>
+              </Flex>
+              <Flex className="slider-inputs">
+                <Flex>
+                  <Editable value={sliderModel.name} className="name-input" submitOnBlur={true}
+                      onChange={handleSliderNameInput}  data-testid="slider-variable-name">
```
kswenson, line 85:
```suggestion
                      onChange={handleSliderNameInput} data-testid="slider-variable-name">
```

kswenson (approved):
👍 Code seems fine. Having the name and value be separate inputs probably makes sense for accessibility if for no other reason. If it were up to me, I'd want the width of the name input to be smaller so that the text to the right doesn't jump so much when editing the name. Perhaps the input could be about the size of the existing value and then auto-size as the user types. Ultimately, I'll let @bfinzer decide whether this UI is suitable for now.

bfinzer (approved):
👍🏻LGTM

PR #725, opened 2023-03-28
Adds default heights to slider and calculator

eireland:
Fixes height and width for newly opened sliders and calculators to a default value


kswenson (changes requested):
I removed these defaults intentionally. For fixed-size components, the sizes should be defined by the component, not by the registration. If there's something you're trying to achieve here we should talk about it.

bfinzer (approved):
I defer to Kirk wrt default sizes

v3/src/components/slider/slider-registration.ts
```diff
@@ -44,7 +44,10 @@ registerTileComponentInfo({
   tileEltClass: kSliderTileClass,
   Icon: SliderIcon,
   defaultWidth: 300,
-  isFixedHeight: true
+  isFixedHeight: true,
+  // must be in sync with rendered size for auto placement code
+    defaultHeight: 73
```
kswenson, line 49:
```suggestion
  defaultHeight: 73
```

kswenson (approved):
👍 @eireland and I looked at this together and determined that it is appropriate for the `defaultWidth` and `defaultHeight` to be specified in registration for fixed-size components because the auto-placement code requires it. Instead, the `FreeTileComponent` no longer imposes a dynamic width/height on fixed-size components.

PR #726, opened 2023-03-30
184814780 & 184077414

bfinzer:
Two fixes of small, annoying bugs.
* Extra vertical space at the bottom of the graph is removed by getting rid of references to the title bar height
* By styling the svg where points are plotted with overflow: visible, we prevent clipping of points near the right edge of the plot


kswenson (approved):
👍 LGTM

PR #728, opened 2023-03-30
animation fix

bfinzer:
Graph animation was occurring when it shouldn't so we turn on animation using a function that will turn it back off after 2 seconds as a failsafe


kswenson (approved):
👍 Looks good -- would `startAnimation()` be a better name than `turnOnAnimation()`?

bfinzer:
Yes, that's a better name. Making it so …

PR #729, opened 2023-03-30
184805436-Numeric axis crowding fix

bfinzer:
* D3 numeric scales don't do a good job a choosing an appropriate number of tick values for a horizontal axis. So we do it ourselves.


kswenson (approved):
👍 Seems fine, but could use a few comments

PR #731, opened 2023-03-31
184805909-changing graph attributes issues-v3

bfinzer:
Bug fix: Incorrect responses to changing attributes on graph
* Using an observable changeCount property in MultiScale, we detect changes to the domain of an axis and refresh subaxis
* Change the default numeric axis formatting to 9 significant digits


kswenson (approved):
👍 LGTM

PR #732, opened 2023-03-31
184847177-axis-reversal-bug-v3

bfinzer:
* There was no protection against dragging the lower end of the numeric scale above the upper end (or vice versa). So we added some.


kswenson (approved):
👍 LGTM

PR #734, opened 2023-04-03
fix: cypress tests

kswenson:
Some of the axis-related cypress tests verify the particular set of axis tick labels that are rendered. In #729, @bfinzer changed the way axis tick labels are chosen which broke a number of these tests.


tejal-shah (approved):

tejal-shah:
LGTM 👍

PR #735, opened 2023-04-03
184210591 slider animation

eireland:
Adds animation to slider. Based on user settings in inspector palette, slider will animate low to high, high to low, or back and forth. Animation can go non-stop, or once only.


eireland:
Not quite ready to be merge but could use another pair of eyes to see if there's a way to make the logic neater.

kswenson (commented):
Some of this discussion happened on slack, including @bfinzer's suggested refactor from ChatGPT. I added a couple of inline comments, but I'll copy my broader comment from slack:

>There are two dimensions of variability, continuous/once and animation mode. ChatGPT created functions for the continuous/once dimension and maintained the if structure for the animation mode. My recommendation would be to use functions for the separate animation modes and then handle the continuous/once distinction within each of those functions. I think the logic would be clearer that way.

In combination with the use of a `useInterval()` hook as suggested elsewhere, I think that will be an improvement.

kswenson:
Other possible improvements to consider:

- The animation code could be moved to a separate `useSliderAnimation()` custom hook just to break up the code.
- Rather than simple functions the separate animation modes could be implemented as structures or classes which could potentially contain state and multiple methods.
- A utility method like this on SliderModel could be useful:
```
  type FixValueFn = (value: number) => number
  validateValue(value: number, belowMin: FixValueFn, aboveMax: FixValueFn) {
    if (value < self.axis.min) return belowMin(value)
    if (value > self.axis.max) return aboveMax(value)
    return value
  }
```
Then the separate functions/classes for each mode could call it with their unique functions.
- Rather than incrementing the value in the model and then correcting the model, it would be better to increment/validate a local variable and then update the model once.

bfinzer (approved):
LGTM
- Leaving remaining issues to be worked out between Evangeline and Kirk

v3/src/components/slider/inspector-panel/slider-settings-panel.tsx
```diff
@@ -46,7 +46,7 @@ export const SliderSettingsPalette =
           <Flex className="palette-row">
             <FormLabel className="form-label">{t("DG.Slider.multiples")}
               <NumberInput className="slider-input multiples" size="xs" defaultValue={sliderModel.multipleOf}
-                  min={0} step={0.5} onChange={handleMultiplesOfChange}>
+                  min={0.5} step={0.5} max={sliderModel.axis.max} onChange={handleMultiplesOfChange}>
```
bfinzer, line 49: I don't think the "multiples" input should have a minimum or maximum. V2 doesn't. You may want tiny increments if the axis is way zoomed in. And there isn't any reason to restrict the user's desire for a max either. I think the same holds for the max frames per second field, but for some reason I'm not able to leave a comment there.

bfinzer (approved):
LGTM
But please remove the min/max restrictions on the two numeric input fields.

eireland:
The last commit removes all limits to the input fields except for the increment amount.
It also does animation correctly even when axis min and max have changed.

bfinzer (approved):
👍🏻LGTM

v3/src/components/slider/slider-component.tsx
```diff
@@ -22,6 +22,7 @@ export const SliderComponent = observer(function SliderComponent({ tile } : ITil
   const layout = useMemo(() => new SliderAxisLayout(), [])
   const {width, height, ref: sliderRef} = useResizeDetector()
   const [running, setRunning] = useState(false)
+  const [isEditingName, setIsEditingName] = useState(false)
```
kswenson, line 25:
```suggestion
  const [isEditingName, setIsEditingName] = useState(false)
```
Remove unused variables.

v3/src/components/slider/use-slider-animation.tsx
```diff
@@ -0,0 +1,114 @@
...
+  setRunning: (running: boolean) => void
+}
+
+export const useSliderAnimation = ({sliderModel, running, setRunning}: IUseSliderAnimationProps) => {
+  const tickTime = 1000/sliderModel.animationRate
+  const direction = sliderModel.animationDirection
+  const mode = sliderModel.animationMode
+  const prevDirectionRef = useRef("")
+  const maxMinHitsRef = useRef(0)
+  const tempValue = sliderModel.value
```
kswenson, line 17:
```suggestion
```
Eliminate this and just use `sliderModel.value` where necessary.

v3/src/components/slider/use-slider-animation.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+  const resetSlider = useCallback((val?: number) => {
+    const dir = sliderModel.animationDirection
+    const testValue = val || sliderModel.value
+    if (dir === "lowToHigh" && testValue >= sliderModel.axis.max) sliderModel.setValue(sliderModel.axis.min)
+    if (dir === "highToLow" && testValue <= sliderModel.axis.min) sliderModel.setValue(sliderModel.axis.max)
```
kswenson, line 23:
```suggestion
    if (dir === "lowToHigh" && testValue > sliderModel.axis.max) sliderModel.setValue(sliderModel.axis.min)
    if (dir === "highToLow" && testValue < sliderModel.axis.min) sliderModel.setValue(sliderModel.axis.max)
```
We don't need to set the value in the `=` cases.

v3/src/components/slider/use-slider-animation.tsx
```diff
@@ -0,0 +1,114 @@
...
+          break
+        case "backAndForth":
+          handleBackAndForthAnimation(newValue)
+      }
+    }
+  }, tickTime)
+
+  const moveSlider = () => {
+    const incrementModifier = direction === 'highToLow' || prevDirectionRef.current === 'highToLow' ? -1 : 1
+    return tempValue + sliderModel.increment * incrementModifier
```
kswenson, line 50:
```suggestion
    return sliderModel.value + sliderModel.increment * incrementModifier
```

v3/src/components/slider/use-slider-animation.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+    if (mode === "onceOnly") {
+      if (maxMinHitsRef.current > 1 && reachedLimit) {
+        setRunning(false)
+        maxMinHitsRef.current = 0
+      }
+      sliderModel.setValue(sliderModel.validateValue(newValue, belowMin, aboveMax))
+    } else {
+      sliderModel.setValue(sliderModel.validateValue(newValue, belowMin, aboveMax))
+    }
```
kswenson, line 112:
```suggestion
    if (mode === "onceOnly" && maxMinHitsRef.current > 1 && reachedLimit) {
      setRunning(false)
      maxMinHitsRef.current = 0
    }
    sliderModel.setValue(sliderModel.validateValue(newValue, belowMin, aboveMax))
```
More compact and should be equivalent.

v3/src/components/slider/slider-thumb.tsx
```diff
@@ -71,6 +69,8 @@ export const CodapSliderThumb = observer(function CodapSliderThumb({sliderContai
     setIsDragging(true)
   }
 
+  useSliderAnimation({sliderModel, running, setRunning})
```
kswenson, line 72: Move this to the `SliderComponent` since there is no interaction with the `SliderThumb`.

kswenson (changes requested):
👍 Looks good -- just a few suggested tweaks.

v3/src/components/slider/slider-thumb.tsx
eireland: `useSliderAnimation` has a lot of references to `sliderModel` which would require a lot of checking and accomodating for when `sliderModel` is `undefined`

v3/src/components/slider/use-slider-animation.tsx
```diff
@@ -0,0 +1,112 @@
...
+      maxMinHitsRef.current = 1 //prevents first time run from doing 3 laps
+    }
+
+    if (mode === "onceOnly" && maxMinHitsRef.current > 1 && reachedLimit) {
+      setRunning(false)
+      maxMinHitsRef.current = 0
+      sliderModel.setValue(sliderModel.validateValue(newValue, belowMin, aboveMax))
+    } else {
+      sliderModel.setValue(sliderModel.validateValue(newValue, belowMin, aboveMax))
+    }
```
kswenson, line 110: You're still making the same `sliderModel` call at the end of both paths, so this can be shortened to:
```suggestion
    if (mode === "onceOnly" && maxMinHitsRef.current > 1 && reachedLimit) {
      setRunning(false)
      maxMinHitsRef.current = 0
    }
    sliderModel.setValue(sliderModel.validateValue(newValue, belowMin, aboveMax))
```
which is what I meant to type before.

kswenson (approved):
👍 Looks good -- one more minor tweak.

PR #736, opened 2023-04-04
V3: Slider default values and precision restrictions

eireland:
This fixes the actual default values for the slider.
Changes default value of multiple of.
Removes precision restriction on slider value
Limits animation frames/sec to max out at 1.
Limits multiple of increments to 0.5

The previous PR just addressed the inspector palette values should be blank on initial open.


bfinzer (approved):
LGTM

kswenson (approved):
👍 LGTM

PR #737, opened 2023-04-04
V3: Close inspector palette on escape key press

eireland:
Focuses inspector palette on open so it closes when Escape key is pressed


kswenson (approved):
👍 LGTM

bfinzer (approved):
👍🏻LGTM

PR #740, opened 2023-04-05
Fix marquee selection in presence of > 1 graph

bfinzer:
[#184805552] Bug fix: Marquee selection and point dragging in two separate graphs with case plots are not independent
* Each graph now gets its own instance of MarqueeState
* The graph dot area gets a class name based on the graph's instance ID to distinguish it from others for the purpose of selecting points

### DIFF
--- a/v3/src/components/graph/components/graph.tsx
+++ b/v3/src/components/graph/components/graph.tsx
@@ -1,6 +1,6 @@
 import {observer} from "mobx-react-lite"
 import {onAction} from "mobx-state-tree"
-import React, {MutableRefObject, useEffect, useRef} from "react"
+import React, {MutableRefObject, useEffect, useMemo, useRef} from "react"
 import {select} from "d3"
 import {GraphController} from "../models/graph-controller"
 import {DroppableAddAttribute} from "./droppable-add-attribute"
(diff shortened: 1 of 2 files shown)


v3/src/components/graph/components/graph.tsx
```diff
@@ -161,7 +159,7 @@ export const Graph = observer(function Graph({graphController, graphRef}: IProps
           {renderGraphAxes()}
 
           <svg ref={plotAreaSVGRef}>
-            <svg ref={dotsRef} className='graph-dot-area'>
+            <svg ref={dotsRef} className={`graph-dot-area ${  instanceId}`}>
```
kswenson, line 162:
```suggestion
            <svg ref={dotsRef} className={`graph-dot-area ${instanceId}`}>
```

v3/src/components/graph/components/graph.tsx
```diff
@@ -34,14 +34,12 @@ interface IProps {
...
 
-const marqueeState = new MarqueeState()
-
-
 export const Graph = observer(function Graph({graphController, graphRef}: IProps) {
   const graphModel = useGraphModelContext(),
     { enableAnimation, dotsRef } = graphController,
     {plotType} = graphModel,
     instanceId = useInstanceIdContext(),
+    marqueeState = useMemo<MarqueeState>(() => new MarqueeState(), []),
```
kswenson, line 42: This is fine, but just to point out another option, it could also be a `volatile` property on the model. If we were ever to support a view-in-window type of functionality, that might have some advantage.

bfinzer: Interesting. I didn't think of making it a model property at all.

PR #741, opened 2023-04-06
Fix problems importing codap3 graphs

bfinzer:
* Modification to GraphController:initializeGraph to properly install category set in categorical axis
* (Unrelated) Setting overflow back to hidden


kswenson (approved):
👍 Looks good!

PR #742, opened 2023-04-06
fix: v2 graph import

kswenson:
The architecture for importing v2 components was changed in [PR #693](https://github.com/concord-consortium/codap/pull/693) in ways that broke the previous v2 graph import.


bfinzer (approved):
👍🏻LGTM

PR #743, opened 2023-04-06
V3: Refactor toolshelf to generate buttons from tile component registry.

eireland:
Toolshelf component buttons are defined in the component registry, and positioned based on position number in the registry. This allows for different behaviors when button is clicked (namely menu vs opening a component).
Placeholder component registers for map, text, and plugins were created so that their toolshelf buttons are generated. They are currently non-functional and no longer have the notification appear when they are clicked.
Note: Table toolshelf menu is not fully implemented. Initial work was done to test that different toolbar buttons are rendered.


bfinzer (approved):
👍🏻LGTM

kswenson (approved):
👍 This looks great -- I was going to suggest some tweaks to the API but wanted to test them out first so I ended up just making them and adding a commit. In summary:
- put tool shelf-related properties under a `shelf` property in the component info
- don't specify tool shelf button component type for tiles using the default
- consistent use of `tool-shelf` or `ToolShelf` because `toolshelf` and `Toolshelf` trigger spell-check warnings

v3/src/components/plugin/plugin-registration.ts
```diff
@@ -0,0 +1,29 @@
+import { registerTileComponentInfo } from "../../models/tiles/tile-component-info"
+// import { registerTileContentInfo } from "../../models/tiles/tile-content-info"
+import { PlaceholderTileComponent } from "../tiles/placeholder/placeholder-tile"
+import { ComponentTitleBar } from "../component-title-bar"
+import PluginsIcon from '../../assets/icons/icon-plug.svg'
+
+export const kPluginIdPrefix = "PLIN"
```
kswenson, line 7: I'd be inclined to go with `PLUG`

v3/src/components/map/map-registration.ts
```diff
@@ -0,0 +1,29 @@
+import { registerTileComponentInfo } from "../../models/tiles/tile-component-info"
+// import { registerTileContentInfo } from "../../models/tiles/tile-content-info"
+import { PlaceholderTileComponent } from "../tiles/placeholder/placeholder-tile"
+import { ComponentTitleBar } from "../component-title-bar"
+import MapIcon from "../../assets/icons/icon-map.svg"
+
+export const kMapIdPrefix = "_MAP"
```
kswenson, line 7: `MAP_` will lead to more readable ids.

PR #744, opened 2023-04-06
Three minor bug fixes

bfinzer:
[#184891112] Bug fix: Graph with legend shows spurious green rectangle
* Green rectangle caused by a reuse of `legend` as className

[#184891232] Bug fix: Categorical axis not getting correct size with split plot
* Categorical axis sizing fixed by adding reaction to number of repetitions of a multiscale changing

[#184814723] Bug fix: Some points don't make it all the way to their position on drop attribute
* Point animation fixed by detecting that enableAnimation.current is true and not showing the data tip if so


kswenson (approved):
👍 LGTM

PR #746, opened 2023-04-07
Point clipping fix

bfinzer:
[#184083861] Bug fix: Plot points should not appear in x- or y-axis areas
* Points were showing up in the x and y axis areas because overflow was allowed to be visible so that points in the plot area would not be clipped. Setting overflow back to hidden fixed this problem.

* The clipping problem was due to setting x (left) and y (top) coordinates of the svg that contains the points to zero and (wrongly) attempting to us a translation of the svg to offset the plot area. Setting the x and y coordinates according to the computed left and top fixed this problem.

* Using the proper offset for the svg necessitated removing transforms of points in the various plot components


kswenson (approved):
👍 Looks good! Nice simplification as well as a nice fix.

PR #747, opened 2023-04-07
Correct numeric drop behavior on top and right

bfinzer:
[#184874124] Bug fix: Same numeric attribute plotted on right and left y axes do not plot the correct scatter plot

* There are two droppable areas on the top and two on the right. They occupy the same space. One is for splitting by a categorical attribute and the other is for adding a numeric y attribute. The categorical drop area was covering up the numeric one. We fix this by only displaying the drop area if the appropriate type of attribute is being dragged.


kswenson (approved):
👍 LGTM

PR #750, opened 2023-04-10
Fix for bug where drop on one graph influences another

bfinzer:
[#184910642] Bug fix: Dragging an attribute into one graph tile influences another graph tile

* `droppableID` has to be unique from one graph to another, so we include the graph's instance id


kswenson (approved):
👍 Looks good -- I tweaked the implementation with another commit so that:
- droppable id follows convention of other tiles (`graph-...-drop`)
- call `useInstanceIdContext()` rather than passing down the id
- default the `InstanceIdContext` to a string so clients don't have to deal with `undefined`

See what you think.

bfinzer:
Nice tweaks! 🙏🏼

PR #751, opened 2023-04-11
chore: update dependencies and eslint configuration

kswenson:
Update dependencies (all except TypeScript 5, because cypress still has incompatibilities).

After @hmorgancode's recent improvements, `eslint` became the bottleneck among our build steps. The following changes improve the `eslint` performance:
1. turn on caching
2. disable `eslint-webpack-plugin` during production builds since we already perform a lint step as part the production build


bfinzer (approved):
👍🏻LGTM

PR #752, opened 2023-04-12
V3: Fixes circle in slider icon from acting like a graph circle

eireland:
The bug was a long click on the circle in the slider icon made the circle big. On pointerup, CODAP would crash with a console error that case ID could not be found, making it seem like it was confusing it with a graph circle
This PR makes the slider icon circle ignore pointer events on it, but still clickable as part of the toolshelf button.


bfinzer:
As we discussed in standup, the root of the problem is in the graph's assignment of behavior to circles. Rather than patch the slider's svg, I suggest rewriting the story to mention where the problem lies and assigning it to me.

PR #753, opened 2023-04-12
 V3: Fixes large slider value inputs

eireland:
Limits the number of digits that can be entered into the value input field

Remove extraneous divs.


kswenson (approved):
👍 LGTM

PR #754, opened 2023-04-12
V3: Restyles slider axis

eireland:
Adds end caps, and changes max, and min values to match V2.
Restyles slider thumb to be closer in size to V2.
This also covers keeping the slider thumb within the axis.[[#184742338](https://www.pivotaltracker.com/story/show/184742338)]

### DIFF
--- a/v3/src/components/slider/slider-component.tsx
+++ b/v3/src/components/slider/slider-component.tsx
@@ -31,7 +31,7 @@ export const SliderComponent = observer(function SliderComponent({ tile } : ITil
   // width and positioning
   useEffect(() => {
     if ((width != null) && (height != null)) {
-      layout.setParentExtent(width - kAxisMargin - 4, height)
+      layout.setParentExtent(width - kAxisMargin, height)
     }
   }, [width, height, layout])
 
--- a/v3/src/components/slider/slider-component.tsx
+++ b/v3/src/components/slider/slider-component.tsx
@@ -16,6 +16,8 @@ import { EditableSliderValue } from "./editable-slider-value"
 
 import './slider.scss'
 
+const kAxisMargin = 30
+
 export const SliderComponent = observer(function SliderComponent({ tile } : ITileBaseProps) {
   const sliderModel = tile?.content
   const instanceId = useNextInstanceId("slider")
@@ -29,12 +31,12 @@ export const SliderComponent = observer(function SliderComponent({ tile } : ITil
   // width and positioning
   useEffect(() => {
     if ((width != null) && (height != null)) {
-      layout.setParentExtent(width - 30, height)
+      layout.setParentExtent(width - kAxisMargin, height)
     }
   }, [width, height, layout])
 
   const axisStyle: CSSProperties = {
-    width: width ? width-30 : width,
+    width: width ? width - kAxisMargin : width,
   }
 
   // control slider value with play/pause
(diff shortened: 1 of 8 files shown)


v3/src/components/slider/slider-component.tsx
```diff
@@ -29,15 +29,12 @@ export const SliderComponent = observer(function SliderComponent({ tile } : ITil
   // width and positioning
   useEffect(() => {
     if ((width != null) && (height != null)) {
-      layout.setParentExtent(width, height)
+      layout.setParentExtent(width - 30, height)
```
kswenson, line 32: Use a named constant for the `30` value rather than repeating it.

kswenson (approved):
👍 looks good -- just one suggestion

v3/src/components/slider/slider-model.ts
```diff
@@ -22,7 +22,7 @@ export const SliderModel = TileContentModel
     animationMode: types.optional(types.enumeration([...AnimationModes]), kDefaultAnimationMode),
     // clients should use animationRate view defined below
     _animationRate: types.maybe(types.number),  // frames per second
-    axis: types.optional(NumericAxisModel, { place: 'bottom', min: 0, max: 12 })
+    axis: types.optional(NumericAxisModel, { place: 'bottom', min: -0.9, max: 11.78 })
```
kswenson, line 25: Out of curiosity, are these the actual v2 defaults?

v3/src/components/slider/slider-thumb.tsx
```diff
@@ -25,8 +25,10 @@ export const CodapSliderThumb = observer(function CodapSliderThumb({sliderContai
   const downOffset = useRef(0)
 
   useEffect(() => {
+    if (sliderModel.value <= sliderModel.axis.min) sliderModel.setValue(sliderModel.axis.min)
+    if (sliderModel.value >= sliderModel.axis.max) sliderModel.setValue(sliderModel.axis.max)
```
kswenson, line 29: This doesn't seem like the right place to do this. Validating/updating the model value should be handled by the model not by the React component at render time. What are you trying to fix/achieve here?

kswenson (changes requested):
👍 Looks good but I have a question about the model validation being done by the slider thumb.

kswenson (changes requested):
👍 Looks pretty good -- the `setValue()` function needs a little more work.

v3/src/components/slider/slider-model.ts
eireland: No, Bill has specified them, and has since corrected the spec.

v3/src/components/slider/slider-model.ts
```diff
@@ -50,18 +50,25 @@ export const SliderModel = TileContentModel
...
-      else if (self.multipleOf) {
+      const keepValueInBounds = (num: number) => {
+        if (num < self.axis.min) return self.axis.min
+        else if (num > self.axis.max) return self.axis.max
+        else return num
+      }
+
+      if (self.multipleOf) {
         n = Math.round(n / self.multipleOf) * self.multipleOf
+        n = keepValueInBounds(n)
```
kswenson, line 61: I don't think this should hold up this PR, but perhaps it warrants another story. If `multipleOf` is 5, `self.axis.max` is 9, and `value` is >= 8, this will set the value to 9, which is not a multiple of 5. Essentially, this says that the axis bounds constraint takes precedence over the `multipleOf` constraint and as I think about it that same assumption is present in the animation PR. Ultimately, it seems to me that to get this right we need to find a value that satisfies both the `multipleOf` _and_ and the axis bounds constraint. Perhaps @bfinzer has another idea of how these constraints should work together in the meantime that can be described in the next story.

kswenson (approved):
👍 Looks good -- there's a consistency issue when the `multipleOf` and axis bounds constraints are applied, but I don't think that should hold up this PR.

bfinzer (approved):
👍🏻LGTM

PR #758, opened 2023-04-17
Fixes for three bugs

bfinzer:
[#184918759] Bug fix: Slider thumb in Slider icon from toolshelf expands

* Prevent unwanted behavior by making sure the circle in question has case data
* Restore proper tooltip text by adapting to changes in dataset getNumeric
* Restore proper marquee selection


v3/src/components/graph/components/background.tsx
kswenson, line 1: `'Point' is defined but never used`

kswenson (approved):
👍 LGTM -- one unused variable warning

PR #760, opened 2023-04-18
Refactor of GraphAxis and Axis to improve labels, menus & drop areas

bfinzer:
[#184902952] Bug fix: A graph with one empty axis should not display "click here" tag unless the graph is selected

* Use `AttributeLabel` to display axis labels
* The AxisOrLegendAttributeMenu component is now properly installed on top of the attribute label, except for legend label.
* Legend label now gets the appropriate menu
* When legend goes away, its space goes away too
* Axes as droppable areas now encompass the attribute label
* When axes go away their space gets reused
* Label refreshes on attribute change even when layout is not affected
* Drop area is only added on top and right if there is not already an axis there


v3/src/components/axis/components/axis.scss
```diff
@@ -1,60 +1,64 @@
-.axis {
-  background-color: white;
+.graph-plot {
```
kswenson, line 1: Wrapping these rules inside `.graph-plot` means they only apply to the graph and won't be applied to the slider. Was that the goal? If so, they should probably be in the graph folder rather than the axis folder.

v3/src/components/axis/components/axis.scss
```diff
@@ -1,60 +1,64 @@
...
-    cursor: grab;
+  .zero .tick line {
+    stroke: #444444;
+    stroke-opacity: 0.8;
+    shape-rendering: crispEdges;
+    pointer-events: none;
   }
+}
+  // We can't put dragRect inside axis for some reason
+  .dragRect {
```
kswenson, line 30: Note that before my recent `strictFunctionTypes` PR, the `.dragRect` class was applied to both the individual drag rects _and_ their parent/wrapper, which could be related to the comment. I renamed the parent class to `.dragRectWrapper` so you could try embedding it inside `.axis` again.

kswenson (changes requested):
👍 Looks good -- most of my comments are related to the organization of the code and the extent to which we should try to keep the axis code from depending on graph code. Some (or all) of that refactoring can be left for a subsequent PR if you don't want to deal with it now.

I added a couple of commits which tuned up some of the types and also rebased the PR to `main`. I'll push those changes once the review is submitted.

v3/src/components/axis/components/axis.scss
bfinzer: Moved these to graph.scss

v3/src/components/axis/components/axis-drag-rects.tsx
```diff
@@ -6,7 +6,7 @@ import t from "../../../utilities/translation/translate"
 import {useAxisLayoutContext} from "../models/axis-layout-context"
 import {INumericAxisModel} from "../models/axis-model"
 import {MultiScale} from "../models/multi-scale"
-import {isVertical} from "../axis-types"
+import {isVertical} from "../../graph/graphing-types"
```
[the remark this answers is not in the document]
bfinzer: I don't have a good idea about how to win this particular battle.

v3/src/components/axis/components/axis.scss
bfinzer: Not attempting now

v3/src/components/graph/components/graph.scss
```diff
@@ -1,11 +1,29 @@
 @import "../../vars.scss";
 
+$graph-label-font: 400 14px Montserrat-Regular, sans-serif;
+$graph-empty-label-font: 12px Montserrat-Regular, sans-serif;
+
+:export {
+  graphLabelFont: $graph-label-font;
+  graphEmptyLabelFont: $graph-empty-label-font;
+}
```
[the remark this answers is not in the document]
bfinzer: Postponed

v3/src/components/graph/components/legend/legend.tsx
```diff
@@ -26,17 +25,16 @@ export const Legend = function Legend({
                                         legendAttrID, graphElt,
                                         onDropAttribute, onTreatAttributeAs, onRemoveAttribute
                                       }: ILegendProps) {
+  useMemo(() => legendAttrID ? [legendAttrID] : [], [legendAttrID])
```
kswenson, line 28: I think this can just be removed given that the function has no side effects and the return value isn't being used.

v3/src/components/graph/models/graph-layout.ts
```diff
@@ -129,34 +129,24 @@ export class GraphLayout implements IAxisLayout {
...
+      newBounds: Record<GraphPlace, Bounds> = {
+        left: {left: 0, top: topAxisHeight, width: leftAxisWidth, height: plotHeight},
+        top: {left: leftAxisWidth, top: 0, width: graphWidth - leftAxisWidth - rightAxisWidth, height: topAxisHeight},
+        plot: {left: leftAxisWidth, top: topAxisHeight, width: plotWidth, height: plotHeight},
+        bottom: {left: leftAxisWidth, top: topAxisHeight + plotHeight, width: plotWidth, height: bottomAxisHeight},
+        legend: {left: 6, top: graphHeight - legendHeight, width: graphWidth - 6, height: legendHeight},
+        rightNumeric: {left: leftAxisWidth + plotWidth, top: topAxisHeight, width: v2AxisWidth, height: plotHeight},
+        rightCat: {left: leftAxisWidth + plotWidth, top: topAxisHeight, width: rightAxisWidth, height: plotHeight},
+        // TODO: fix this (was missing before)
+        yPlus: {left: 0, top: topAxisHeight, width: leftAxisWidth, height: plotHeight}
```
kswenson, line 143: I should have pointed this out before, but I added this so every `GraphPlace` would have a definition for typing purposes, but I just gave it the same value as `left`, so you may want to take a look at providing a real value here at some point.

kswenson (approved):
👍 Just a couple of small tweaks

PR #761, opened 2023-04-19
chore: enable `strictFunctionTypes` in tsconfig.json

kswenson:
"Fixes" fall into three broad categories:
- actual fixes to the typing issues in question
- adding `@ts-expect-error` at the point of the call in places where the actual fix wasn't clear
- ~~adding a cast in some places where `@ts-expect-error` didn't work or was annoying~~

Update: I figured out the root of the issues we were having with D3 types. The trick is to type the initial selection correctly because that determines the types of all of the arguments from the chained callbacks. Also, the functions passed to `join()` must return the appropriate type to enable further chaining.

### DIFF
--- a/v3/src/components/graph/components/background.tsx
+++ b/v3/src/components/graph/components/background.tsx
@@ -18,8 +18,7 @@ interface IProps {
 
 const prepareTree = (areaSelector: string, circleSelector: string): typeof RTree => {
     const selectionTree = RTree(10)
-    select(areaSelector).selectAll(circleSelector)
-      // @ts-expect-error strictFunctionTypes
+    select<HTMLDivElement, unknown>(areaSelector).selectAll<SVGCircleElement, InternalizedData>(circleSelector)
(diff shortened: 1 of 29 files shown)


v3/src/components/graph/components/background.tsx
```diff
@@ -123,6 +124,7 @@ export const Background = forwardRef<SVGGElement, IProps>((props, ref) => {
         (enter) => {
           enter.append('rect')
             .attr('class', 'graph-background')
+            // @ts-expect-error strictFunctionTypes
```
emcelroy, line 127: FWIW: in CLUE, this wasn't needed after I changed line 109 to `dragBehavior = drag<SVGRectElement, number>()`.

emcelroy (approved):
Looks good 👍

bfinzer (approved):
👍🏻LGTM
Amazing!

emcelroy (approved):
Looks even better!

PR #762, opened 2023-04-20
184846919 slider value update

eireland:
Formats the slider value depending on the axis scale
Fixes slider thumbnail position when slider axis bounds are changed
Fixes global value warning on slider create.

### DIFF
--- a/v3/src/components/axis/models/multi-scale.ts
+++ b/v3/src/components/axis/models/multi-scale.ts
@@ -71,6 +71,17 @@ export class MultiScale {
     return this.scale.domain()
   }
 
+  @computed get numericSignificantDigits() {
+    if (["linear"].includes(this.scaleType)) {
+      const numericScale = this.scale as ScaleLinear<number, number>,
+        onePixelDelta = numericScale.invert(1) - numericScale.invert(0)
+        console.log("onePixelDelta", onePixelDelta)
+      return Math.max(0, Math.ceil(-Math.log10(onePixelDelta)))
+    } else {
+      return NaN
+    }
+  }
+
   @action setScaleType(scaleType: IScaleType) {
     this.scaleType = scaleType
     this.scale = scaleTypeToD3Scale(scaleType)
@@ -120,5 +131,4 @@ export class MultiScale {
     }
     return {data: NaN}
   }
-
 }
--- a/v3/src/components/axis/models/multi-scale.ts
+++ b/v3/src/components/axis/models/multi-scale.ts
@@ -1,5 +1,6 @@
 import {action, computed, makeObservable, observable} from "mobx"
 import {
+  format,
   NumberValue, ScaleBand, scaleBand, ScaleLinear, scaleLinear, scaleLog, ScaleOrdinal, scaleOrdinal
 } from "d3"
 import {AxisScaleType, IScaleType, ScaleNumericBaseType} from "../axis-types"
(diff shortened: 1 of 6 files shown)


v3/src/components/axis/models/multi-scale.ts
```diff
@@ -121,4 +122,29 @@
     return {data: NaN}
   }
 
+  formatValueForScale(value: number) {
```
bfinzer, line 125: How about the following comment for this function?
/** To display values for a numeric axis we use just the number of significant figures required to distinguish the value for one screen pixel from the value for the adjacent screen pixel. **/

bfinzer (approved):
LGTM
* Just add the suggested comment to the multiscale function

v3/src/components/axis/models/multi-scale.ts
```diff
@@ -121,4 +122,29 @@ export class MultiScale {
...
+
+    function formatNumber(n: number, dom: [number, number], range: [number, number]): string {
```
kswenson, line 127: FYI, you could define a type for this (or even two types, AxisDomain and AxisRange, in case that was helpful):
```suggestion
    export type AxisExtent = [number, number] // might need to move this outside the class definition
    
    function formatNumber(n: number, domain: AxisExtent, range: AxisExtent): string {
```

v3/src/components/slider/editable-slider-value.tsx
```diff
@@ -1,26 +1,30 @@
...
 
-const kDecimalPlaces = 2
-const d3Format = format(`.${kDecimalPlaces}~f`)
-
-const formatValue = (model: ISliderModel) => d3Format(model.globalValue.value)
-
-export const EditableSliderValue = observer(function EditableSliderValue({sliderModel} : IProps) {
+export const EditableSliderValue = observer(function EditableSliderValue({ sliderModel, domain, multiScale}: IProps) {
   const [candidate, setCandidate] = useState("")
+  const formatValue = (model: ISliderModel) => multiScale.formatValueForScale(model.globalValue.value)
```
kswenson, line 17: This function appears to only be used in the `useEffect()` below so you can define it there. Otherwise, you need to wrap in `useCallback()` and add it as dependency in the `useEffect()` below.

v3/src/components/slider/editable-slider-value.tsx
```diff
@@ -1,26 +1,30 @@
...
 import React, {useState, useEffect} from "react"
-import { ISliderModel } from "./slider-model"
+import {ISliderModel} from "./slider-model"
+import {MultiScale} from "../axis/models/multi-scale"
 
 import './slider.scss'
 
 interface IProps {
   sliderModel: ISliderModel
+  domain: any
```
kswenson, line 11: Is this still necessary? It's not used except as a dependency below.

v3/src/components/slider/slider-component.tsx
```diff
@@ -25,6 +25,8 @@ export const SliderComponent = observer(function SliderComponent({ tile } : ITil
   const {width, height, ref: sliderRef} = useResizeDetector()
   const [running, setRunning] = useState(false)
   const animationRef = useRef(false)
+  const multiScale = layout.getAxisMultiScale("bottom")
+  const domain = layout.getAxisBounds("bottom")
```
kswenson, line 29: Is this still necessary?

kswenson (changes requested):
👍 Looks good -- needs a few tweaks

v3/src/components/slider/slider-component.tsx
eireland: Oddly, when I don't have it in the dependency, slider value shows `NaN`

v3/src/components/slider/editable-slider-value.tsx
eireland: Removed the useEffect that was using the function instead.

kswenson (approved):
👍 Looks good just add a comment about the curious `domain` dependency and a suggestion to tweak the `encompassValue()` functionality.

v3/src/components/slider/editable-slider-value.tsx
```diff
@@ -33,14 +37,11 @@ export const EditableSliderValue = observer(function EditableSliderValue({slider
     const inputValue = parseFloat(e.target.value)
     if (isFinite(inputValue)) {
       sliderModel.encompassValue(inputValue)
+      setCandidate(multiScale.formatValueForScale(sliderModel.value))
+      sliderModel.setValue(inputValue)
```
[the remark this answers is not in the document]
eireland: removed setValue call in encompassValue.

PR #764, opened 2023-04-21
V3: 184771920 blank default document

eireland:
Adds ability to open a blank document, and opens to specified document based on URL parameters.
By default, CODAP will load a blank document with no data set.
If `dashboard` query param is added, then CODAP will load with an empty data set, and all components open.
If `sample=[sample-dataset-name]` query param is added, CODAP will load a blank document, but the specified data set is attached to the document. Currently, we have no way of opening a case table for the dataset, but can verify the dataset exist by opening a graph.
If `sample=[sample-dataset-name]&dashboard` query param is added, CODAP will load dataset with all the components opened.

### DIFF
--- a/v3/src/components/app.tsx
+++ b/v3/src/components/app.tsx
@@ -122,11 +122,9 @@ export const App = observer(function App() {
     if (gDataBroker.dataSets.size === 0) {
       const sample = sampleData.find(name => urlParams.sample === name.toLowerCase())
       if (sample) {
-        console.log("in if sample")
         importSample(sample, handleImportDataSet)
-        addDefaultComponents()
       }
-      else if (getSearchParams().includes("dashboard")) {
+      if (urlParams.dashboard !== undefined) {
         createNewStarterDataset()
         addDefaultComponents()
       }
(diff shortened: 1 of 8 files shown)


tejal-shah (approved):
LGTM 👍

v3/src/components/app.tsx
```diff
@@ -117,8 +115,10 @@ export const App = observer(function App() {
       const sample = sampleData.find(name => urlParams.sample === name.toLowerCase())
       if (sample) {
         importSample(sample, handleImportDataSet)
-      } else {
+      }
+      if (urlParams.dashboard !== undefined) {
         createNewStarterDataset()
+        addDefaultComponents()
       }
```
kswenson, line 122: Do we need the call to `createNewStarterDataSet` if we're not importing a sample? I was expecting something more like:
```
      if (sample) {
        importSample(sample, handleImportDataSet)
      } else {
        createNewStarterDataset()
      }
      if (urlParams.dashboard !== undefined) {
        addDefaultComponents()
      }
```

kswenson (commented):
👍 Looks good -- just one question I'm unclear on

bfinzer (approved):
👍🏻LGTM

PR #765, opened 2023-04-22
Prevent errors during graph dispose

bfinzer:
* GraphAxis has a useEffect for a cleanup needed under normal circumstances. But it gets triggered when the graph component is closed as well, during which accesses to MST model properties of already disposed objects are attempted. So we test that the graphModel is still alive before making the attempt.


kswenson (approved):
👍 Looks good but there a a couple of lint warnings to address before merging

PR #766, opened 2023-04-24
Graph point selection fix

bfinzer:
Starting in build 1201 selection of cases is no longer reflected by point highlighting in graphs.

This PR fixes this bug

### DIFF
--- a/v3/src/components/graph/utilities/graph-utils.ts
+++ b/v3/src/components/graph/utilities/graph-utils.ts
@@ -382,7 +382,7 @@ export function setPointSelection(props: ISetPointSelection) {
     .style('stroke-width', defaultStrokeWidth)
     .style('stroke-opacity', defaultStrokeOpacity)
 
-  const selectedDots = dots.selectAll('.graph-dot-highlighted')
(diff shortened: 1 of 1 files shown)


v3/src/components/graph/utilities/graph-utils.ts
```diff
@@ -382,7 +382,7 @@ export function setPointSelection(props: ISetPointSelection) {
     .style('stroke-width', defaultStrokeWidth)
     .style('stroke-opacity', defaultStrokeOpacity)
 
-  const selectedDots = dots.selectAll('.graph-dot-highlighted')
+  const selectedDots = select(dotsRef.current).selectAll('.graph-dot-highlighted')
```
kswenson, line 385:
```suggestion
  const selectedDots = selectDots(dotsRef.current, true)
```
In this case I think it would make sense to use the new API, which was probably what I originally intended to do and then just forgot to do so.

kswenson (changes requested):
👍 Looks good -- just a suggested tweak to the API used.

kswenson (approved):
👍 LGTM

PR #769, opened 2023-04-26
Graph split plot bug fix

bfinzer:
[#184980432] Bug fix: Graph split plots do not redraw when switching some categorical variables

* Add a useEffect in usePlotResponders that responds to attribute changes to include all GraphAttrRoles
* Also get rid of a useEffect that was relying on a list attribute IDs as depedents

### DIFF
--- a/v3/src/components/graph/hooks/use-plot.ts
+++ b/v3/src/components/graph/hooks/use-plot.ts
@@ -8,6 +8,8 @@ import {IGraphModel} from "../models/graph-model"
 import {matchCirclesToData, startAnimation} from "../utilities/graph-utils"
 import {useInstanceIdContext} from "../../../hooks/use-instance-id-context"
 import {useDataSetContext} from "../../../hooks/use-data-set-context"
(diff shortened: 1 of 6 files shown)


v3/src/components/graph/hooks/use-plot.ts
```diff
@@ -145,7 +150,7 @@ export const usePlotResponders = (props: IPlotResponderProps) => {
         })
         callRefreshPointPositions(false)
       }
-    })
+    }) || (() => true)
```
kswenson, line 153:
```suggestion
    })
```
Rather than this, you should be able to just add a `?.` to line 154 below:
```typescript
    return () => disposer?.()
```

kswenson (approved):
👍 Nice cleanup -- make sure you run lint and fix the extraneous semicolon and the `useEffect` dependency issues as well as the other suggestion I made.

v3/src/components/graph/hooks/use-plot.ts
bfinzer: Nope. TS1109: Expression expected.
Leave as is.

PR #770, opened 2023-04-26
fix logic for opening a blank document

eireland:
We don't want a data set when CODAP is first opened unless `dashboard` query param is present


kswenson (approved):
👍 LGTM -- you've convinced me 😉

PR #773, opened 2023-04-27
Properly import graphs with multiple y attributes

bfinzer:
[#184599671] Bug fix: Graphs with multiple numeric y attributes import correctly

* The DataConfigurationModel, when assigned a new dataset, needs to make sure there are filtered cases for each of the y attributes and for the y2 attribute, if any.


kswenson (approved):
👍 LGTM

PR #776, opened 2023-04-28
V3: 185024367 slider NaN on load

eireland:
Changes the dependencies to the useEffect for updating the value formatting. The useEffect was not running after component mount and so didn't update the format when the slider model gets its final axis max and min.

### DIFF
--- a/v3/src/components/slider/editable-slider-value.tsx
+++ b/v3/src/components/slider/editable-slider-value.tsx
@@ -1,6 +1,6 @@
 import {NumberInput, NumberInputField} from "@chakra-ui/react"
 import {observer} from "mobx-react-lite"
-import React, {useState, useEffect, useCallback} from "react"
+import React, {useState, useEffect} from "react"
 import {ISliderModel} from "./slider-model"
 import {MultiScale} from "../axis/models/multi-scale"
 import { AxisBounds } from "../axis/axis-types"
--- a/v3/src/components/slider/editable-slider-value.tsx
+++ b/v3/src/components/slider/editable-slider-value.tsx
@@ -5,6 +5,7 @@ import {ISliderModel} from "./slider-model"
 import {MultiScale} from "../axis/models/multi-scale"
 
 import './slider.scss'
+import { autorun } from "mobx"
 
 interface IProps {
   sliderModel: ISliderModel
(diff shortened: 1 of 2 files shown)


v3/src/components/slider/editable-slider-value.tsx
```diff
@@ -3,25 +3,26 @@ import {observer} from "mobx-react-lite"
...
+  const axisMax = axisModel?.max
 
-  // when `domain` is not included in the dependency, slider value shows NaN
+  // when `multiScale.cellLength` is not included in the dependency, slider value shows NaN
   useEffect(() => {
     if (sliderModel) {
       setCandidate(multiScale.formatValueForScale(sliderModel.value))
     }
-  }, [domain, multiScale, sliderModel, sliderModel.axis, sliderModel.value])
+  }, [axisMin, axisMax, multiScale.cellLength, multiScale, sliderModel, sliderModel.value])
```
kswenson, line 25: This really wants to be a MobX reaction so that model dependencies are tracked automatically rather than having to be hand-crafted as `useEffect()` dependencies. I suspect something like this would work:
```typescript
  useEffect(() => {
    return autorun(() => {
      // trigger update on domain change
      const domain = sliderModel.domain
      if (sliderModel) {
        setCandidate(multiScale.formatValueForScale(sliderModel.value))
      }
    })
  }, [...])
```

v3/src/components/slider/editable-slider-value.tsx
```diff
@@ -38,8 +39,8 @@ export const EditableSliderValue = observer(function EditableSliderValue({ slide
     const inputValue = parseFloat(e.target.value)
     if (isFinite(inputValue)) {
       sliderModel.encompassValue(inputValue)
-      setCandidate(multiScale.formatValueForScale(sliderModel.value))
       sliderModel.setValue(inputValue)
+      setCandidate(multiScale.formatValueForScale(sliderModel.value))
```
kswenson, line 43: This might not be necessary with the above change because the autorun will take care of it.

kswenson (changes requested):
👍 I think we can do even better with a MobX autorun.

bfinzer (approved):
* On reflection, I think Kirk's suggestion for using an autorun model dependencies to trigger formatting is good.

kswenson (approved):
👍 Looks good -- just a couple refinements to consider.

PR #778, opened 2023-04-28
fix: fix bug when swapping attributes on axes for dot plot

kswenson:


bfinzer (approved):
👍🏻LGTM
Stale function is a subtle concept!

PR #783, opened 2023-05-02
Fix rotated Korean chars

eireland:
Adds css handling of Korean characters like Chinese characters for case table drop area hint


kswenson (approved):
👍 LGTM

PR #784, opened 2023-05-03
feature: use @concord-consortium/mobx-state-tree 5.1.8-cc.1

kswenson:
- add `onAnyAction()` function which listens for any actions, not just outermost actions
- use `onAnyAction()` throughout graph code
- note that client code should continue to import from `"mobx-state-tree"`
  - projects configuration is used to redirect `"mobx-state-tree"` to `"@concord-consortium/mobx-state-tree"`


scytacki (approved):
This looks good to me.

I just noted an issue with the comment referring to a missing file.

PR #785, opened 2023-05-03
V3: 184771808 table toolshelf menu functionality

eireland:
Adds the various functionalities of the case table menu items except for `copy from clipboard`:
- `new` adds a new data set to the shared model and opens a new case table in the document
- Selecting an existing data set opens (or focuses if already open) the case table for the data set
- Click on trash next to an existing data set removes the data set from the shared model and closes the table.


v3/src/models/document/shared-model-document-manager.ts
```diff
@@ -80,6 +83,24 @@ export class SharedModelDocumentManager implements ISharedModelDocumentManager {
...
+    }
+    const sharedDatasets = this.document.getSharedModelsByType<typeof SharedDataSet>("SharedDataSet")
+    const sharedMetadatas = this.document.getSharedModelsByType<typeof SharedCaseMetadata>("SharedCaseMetadata")
+    const dataSetToDeleteIndex = sharedDatasets.map(model => model.dataSet.id).indexOf(dataSetId)
+    const metadataToDeleteIndex = sharedMetadatas.map(model => model.data?.id).indexOf(dataSetId)
+    const sharedModelToDelete = sharedDatasets[dataSetToDeleteIndex]
+    const sharedCaseMetadataToDelete = sharedMetadatas[metadataToDeleteIndex]
+    // remove table associated with data set from document
+    const tileToRemove = this.getSharedModelTiles(sharedModelToDelete)
+                              .find(tile => tile.content.type === kCaseTableTileType)
```
kswenson, line 100: In theory, there could be multiple case table linked to this shared model, so I think this should be:
```suggestion
    const tilesToRemove = this.getSharedModelTiles(sharedModelToDelete)
                              .filter(tile => tile.content.type === kCaseTableTileType)
```
and then you'd pass the resulting `tilesToRemove` to the `removeSharedModelsAndTiles` function.

v3/src/components/case-table/case-table-tool-shelf-button.tsx
```diff
@@ -1,29 +1,195 @@
...
+  const [dataSetIdToDeleteId, setDataSetIdToDelete] = useState("")
+
+  if (!row) return null
+
+  const handleCreateNewDataSet = () => {
+    const tile = createDefaultTileOfType(kCaseTableTileType)
+    if (!tile) return
+    const newData = [{ AttributeName: "" }]
+    const ds = DataSet.create({ name: "New Dataset" })
+    ds.addAttribute({ name: "AttributeName" })
```
kswenson, line 45: These should use localizable strings.

v3/src/components/case-table/case-table-tool-shelf-button.tsx
```diff
@@ -1,29 +1,195 @@
...
+    const tile = createDefaultTileOfType(kCaseTableTileType)
+      if (!tile) return
+      manager?.addTileSharedModel(tile.content, model, true)
+      manager?.addTileSharedModel(tile.content, caseMetadata, true)
+
+      const width = kDefaultTableSize.width
+      const height = kDefaultTableSize.height
+      const {x, y} = getPositionOfNewComponent({width, height})
+      content?.insertTileInRow(tile, row, {x, y, width, height})
+      uiState.setFocusedTile(tile.id)
```
kswenson, line 68: Indentation; also there's redundancy here. It seems like:
```
    manager?.addTileSharedModel(tile.content, sharedData, true)
    manager?.addTileSharedModel(tile.content, caseMetadata, true)

    const width = kDefaultTableSize.width
    const height = kDefaultTableSize.height
    const {x, y} = getPositionOfNewComponent({width, height})
    content?.insertTileInRow(tile, row, {x, y, width, height})
    uiState.setFocusedTile(tile.id)
```
could be put into a reusable function.

v3/src/components/case-table/case-table-tool-shelf-button.tsx
```diff
@@ -1,29 +1,195 @@
...
+  }
 
+  const handleOpenDataSetTable = (dataset: ISharedDataSet) => {
+    const model = manager?.getSharedModelsByType("SharedDataSet").find(m =>  m.id === dataset.id) as ISharedDataSet
```
kswenson, line 72:
```suggestion
    const model = manager?.getSharedModelsByType("SharedDataSet")
                          .find(m => m.id === dataset.id) as ISharedDataSet | undefined
```
Types should acknowledge the fact that it might not be found.

v3/src/components/case-table/case-table-tool-shelf-button.tsx
```diff
@@ -1,29 +1,195 @@
...
+      const tiles = manager?.getSharedModelTiles(model)
+      const tableTile = tiles?.find(tile => tile.content.type === "CodapCaseTable")
+      if (tableTile) { // a case table for the data set already exists in document
+        uiState.setFocusedTile(tableTile.id)
+      } else { // open a table and add it to the shared model manager
+        openTableForDataset(model, caseMetadata)
+      }
+    } else {
+      openTableForDataset(model, caseMetadata)
+    }
```
kswenson, line 86: Isn't this the same as this simpler version:
```suggestion
    if (existingTableTiles?.length !== 0) {
      const tiles = manager?.getSharedModelTiles(model)
      const tableTile = tiles?.find(tile => tile.content.type === "CodapCaseTable")
      if (tableTile) { // a case table for the data set already exists in document
        uiState.setFocusedTile(tableTile.id)
        return
      }
      // open a table and add it to the shared model manager
      openTableForDataset(model, caseMetadata)
    }
```

kswenson (changes requested):
👍 Looks good -- the usual assortment of tweaks and suggestions.

v3/src/components/case-table/case-table-tool-shelf-button.tsx
eireland: The first `else` allows for more than one table to be open in the document for multiple data sets.
The second `else` prevents multiple tables for the same data set to be opened.

kswenson: I'm not trying to change the logic; I just think the code can be simplified while still achieving the same result. Your confusion is warranted, however, since my first suggestion wasn't correct. I think this should be equivalent:
```suggestion
    if (existingTableTiles?.length !== 0) {
      const tiles = manager?.getSharedModelTiles(model)
      const tableTile = tiles?.find(tile => tile.content.type === "CodapCaseTable")
      if (tableTile) { // a case table for the data set already exists in document
        uiState.setFocusedTile(tableTile.id)
        return
      }
    }
    // open a table and add it to the shared model manager
    openTableForDataset(model, caseMetadata)
```
It's essentially a two-by-two table where `openTableForDataSet` is being called in three of the four paths, so the early return in the fourth path allows us to eliminate the redundant calls.

eireland: Was able to make the whole logic simpler by removing the check for existing tables.

bfinzer (approved):
LGTM - deferring to Kirk

v3/src/components/case-table/case-table-tool-shelf-button.tsx
```diff
@@ -1,29 +1,186 @@
...
+    ds.addCases(toCanonical(ds, newData))
+    const tile = createDefaultTileOfType(kCaseTableTileType)
+    if (!tile) return
+    const { sharedData, caseMetadata } = gDataBroker.addDataSet(ds, tile.id)
+    openTableForDataset(sharedData, caseMetadata)
+  }
+
+  const handleOpenDataSetTable = (dataset: ISharedDataSet) => {
+    const model =
+            manager?.getSharedModelsByType("SharedDataSet").find(m =>  m.id === dataset.id) as ISharedDataSet | undefined
```
kswenson: Address line length warning.

kswenson (approved):
👍 Looks good -- rather than go another round of review I just added a commit which made a couple of final tweaks. Take a look and make sure my changes are acceptable and that everything still works. 😉

PR #786, opened 2023-05-04
chore: add "engines" to package.json (PT-185071892)

433eros:


kswenson (changes requested):
The `engineStrict` field is no longer used, so its presence here would only be confusing: https://docs.npmjs.com/cli/v6/configuring-npm/package-json#enginestrict.

kswenson (approved):
👍 LGTM

PR #791, opened 2023-05-10
fix: link/unlink data sets and case metadata

kswenson:
- add `dataSet` to attribute drags
- use `dataSet` in attribute drags to link graphs to shared models on drops
- add a set of utility functions in `shared-data-utils.ts`

### DIFF
--- a/v3/src/components/graph/models/data-configuration-model.ts
+++ b/v3/src/components/graph/models/data-configuration-model.ts
@@ -448,6 +448,10 @@ export const DataConfigurationModel = types
           typeToDropIsNumeric = !!idToDrop && dataSet?.attrFromID(idToDrop)?.type === 'numeric',
           xIsNumeric = self.attributeType('x') === 'numeric',
           existingID = self.attributeID(role)
+        // only drops on left/bottom axes can change data set
+        if (dataSet?.id !== self.dataset?.id && !['left', 'bottom'].includes(place)) {
+          return false
+        }
         if (place === 'yPlus') {
           return xIsNumeric && typeToDropIsNumeric && !!idToDrop && !self.yAttributeIDs.includes(idToDrop)
         } else if (place === 'rightNumeric') {
--- a/v3/src/components/graph/models/data-configuration-model.ts
+++ b/v3/src/components/graph/models/data-configuration-model.ts
@@ -443,9 +443,9 @@ export const DataConfigurationModel = types
           primaryRole = self.primaryRole
         return primaryRole === role || !['left', 'bottom'].includes(place)
       },
-      graphPlaceCanAcceptAttributeIDDrop(place: GraphPlace, idToDrop?: string) {
+      graphPlaceCanAcceptAttributeIDDrop(place: GraphPlace, dataSet?: IDataSet, idToDrop?: string) {
         const role = graphPlaceToAttrRole[place],
-          typeToDropIsNumeric = !!idToDrop && self.dataset?.attrFromID(idToDrop)?.type === 'numeric',
+          typeToDropIsNumeric = !!idToDrop && dataSet?.attrFromID(idToDrop)?.type === 'numeric',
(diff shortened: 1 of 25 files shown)


v3/src/components/graph/models/data-configuration-model.ts
```diff
@@ -443,9 +443,9 @@
           primaryRole = self.primaryRole
         return primaryRole === role || !['left', 'bottom'].includes(place)
       },
-      graphPlaceCanAcceptAttributeIDDrop(place: GraphPlace, idToDrop?: string) {
+      graphPlaceCanAcceptAttributeIDDrop(place: GraphPlace, dataSet?: IDataSet, idToDrop?: string) {
```
bfinzer, line 446: If the incoming dataset is different than the current dataset, the logic will be quite different. For example, it doesn't make sense to drop an attribute from a different dataset on the top or rightCat axes. On the other hand we don't want to forbid dropping an attribute from a different dataset on x or y even though the consequences will be to remove current attributes. Bottomline: we need to test whether the incoming dataset is the same as the current one and, if not, allow drops only on 'x' and 'y'.

v3/src/components/graph/models/graph-controller.ts
```diff
@@ -80,8 +81,9 @@ export class GraphController {
     }
   }
 
-  handleAttributeAssignment(graphPlace: GraphPlace, attrID: string) {
-    const {graphModel, layout, dataset} = this,
+  handleAttributeAssignment(graphPlace: GraphPlace, dataSetID: string, attrID: string) {
+    const {graphModel, layout} = this,
+      dataset = getDataSetFromId(graphModel, dataSetID),
       dataConfig = graphModel?.config
```
bfinzer, line 87: If the incoming dataset is different than this.dataset we need to clear existing attributes and start fresh.

v3/src/components/graph/models/graph-model.ts
```diff
@@ -117,7 +113,13 @@
     removeAxis(place: AxisPlace) {
       self.axes.delete(place)
     },
-    setAttributeID(role: GraphAttrRole, id: string) {
+    setAttributeID(role: GraphAttrRole, dataSetID: string, id: string) {
+      const dataSet = getDataSetFromId(self, dataSetID)
+      if (dataSet && isTileLinkedToOtherDataSet(self, dataSet)) {
```
bfinzer, line 118: Yes, this is the right thing to do. But I don't know how it will interact with the GraphController's handleAttributeAssignment. Maybe OK?

bfinzer (approved):
👍🏻LGTM
Just a couple questions about drops of "foreign" attributes from datasets on graphs. Might be that it all works fine.

v3/src/components/graph/models/data-configuration-model.ts
kswenson: Good point -- I'll add a commit to address this.

v3/src/components/graph/models/graph-controller.ts
kswenson: `handleAttributeAssignment` is called in two paths:
1. treat attribute as type, which doesn't change the data set
2. as a handler called in response to `GraphModel.setAttributeID()`, in which case the necessary changes to the `DataConfigurationModel` have already occurred. It may be that the logic within the `handleAttributeAssignment()` method will need to change, but I don't have a clear picture of what those changes look like so I'm inclined to leave them for a subsequent PR.

v3/src/components/graph/models/graph-model.ts
kswenson: See comment in `GraphController` for response.

v3/src/components/graph/models/graph-controller.ts
bfinzer: Agreed.

PR #792, opened 2023-05-12
fix: linking of data sets

kswenson:
@bfinzer This fixes the coasters problem we saw. There may still be something else going on, but this should be a start.

- eliminates defaulting of data set
- wait until sample data is available to create default tiles in dashboard mode
- eliminate redundant data set from graph controller
- synchronizes data configuration data set/metadata on link/unlink

Note: some of these changes were made originally in CLUE and were brought over in this PR.


bfinzer (approved):
👍🏻LGTM
So much better!

PR #793, opened 2023-05-13
chore: add wrapSerialization function

kswenson:
- eliminate `prepareSnapshots` and `completeSnapshots` from `DataBroker`
- replace them with `wrapSerialization` function in `shared-data-utils`


eireland (approved):
👍 LGTM

PR #794, opened 2023-05-13
184540510 squigglies

eireland:
Adds curved splines betweenparent and child collection tables.
Known bugs:
- Splines for collapsed first parent case are not drawn correctly.
- Scroll down, collapse a case, expand that case - splines are not drawn correctly until table is redrawn.
- Spline for first parent case are is not drawn correctly after expand.


kswenson (approved):
👍 Looks good -- I added a commit to tweak a couple of things, including fixing a bug and reducing the number of re-renders by eliminating some state. Take a look and see what you think.

bfinzer (approved):
👍🏻LGTM
Trusting to Kirk

kswenson (approved):
👍 LGTM -- just one lint warning to fix

PR #795, opened 2023-05-26
perf: use value.toString() instead of JSON.stringify() in Attribute.importValueToString()

kswenson:
@scytacki asked a question on slack about converting values for the `DataSet` and I pointed him to the `Attribute.importValueToString()` function. His suggested alternative used `value.toString()` instead of `JSON.stringify()`, which seemed like it would probably generate the same results with better performance. I whipped up a quick test to verify that (the improvement is ~40%) which also verifies that the results are the same.


eireland (approved):
👍 LGTM

PR #796, opened 2023-05-26
fix: configure shared models in afterAttachToDocument()

kswenson:
- new graphs auto-link to DataSet if there's only one available

This is the fix from the `v3-retreat-demo` branch.


bfinzer (approved):
👍🏻LGTM
Much appreciated!

PR #798, opened 2023-05-26
perf: efficient index column spanning for collapsed groups

kswenson:
Rendering of collapsed rows is handled by the `colSpan()` function of the index column. Previously, a MobX `reaction` was used to recompute the set of visible attributes and collapsed rows whenever they changed, which resulted in a new instance of the index column definition, which resulted in re-rendering the entire table. Thus, expanding/collapsing a single row would result in re-rendering the entire table. With these changes, the `colSpan()` function is stable and so the index column definition is stable across expanding/collapsing a single row.

To accomplish this we:
- ~~replace the MobX `reaction` with an `autorun`~~
- eliminate the MobX `reaction`
- replace the `colSpan()` function with one defined outside the component
- use a ref instead of state to store the index column definition

### DIFF
--- a/v3/src/components/case-table/case-table-types.ts
+++ b/v3/src/components/case-table/case-table-types.ts
@@ -1,6 +1,6 @@
 import {
-  CalculatedColumn, CellClickArgs, Column, EditorProps, FormatterProps, HeaderRendererProps, RowRendererProps,
-  RowsChangeData,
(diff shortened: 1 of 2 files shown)


eireland (approved):
👍 LGTM

bfinzer (approved):
👍🏻LGTM

PR #800, opened 2023-05-26
V3: Adds ability to minimize components

eireland:
Adds functionality to the minimize component button.

### DIFF
--- a/v3/src/components/free-tile-component.tsx
+++ b/v3/src/components/free-tile-component.tsx
@@ -1,5 +1,6 @@
 import { useDndContext } from "@dnd-kit/core"
-import React, { useState } from "react"
+import { clsx } from "clsx"
+import React, { useCallback, useState } from "react"
 import { getDragTileId, IUseDraggableTile, useDraggableTile } from "../hooks/use-drag-drop"
 import { IFreeTileLayout, IFreeTileRow, isFreeTileRow } from "../models/document/free-tile-row"
 import { getTileComponentInfo } from "../models/tiles/tile-component-info"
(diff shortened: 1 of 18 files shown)


v3/src/components/slider/slider-component.tsx
```diff
@@ -17,6 +17,8 @@ import { EditableSliderValue } from "./editable-slider-value"
 import './slider.scss'
 
 const kAxisMargin = 30
+const kComponentTitleHeight = 25
+const kSliderComponentHeight = 98
```
bfinzer, line 21: There is slider-component-height in slider.scss. I know there's some way to make use of that constant here rather than having to redundantly define it.

v3/src/components/free-tile-component.tsx
```diff
@@ -6,6 +6,8 @@ import { getTileComponentInfo } from "../models/tiles/tile-component-info"
 import { ITileModel } from "../models/tiles/tile-model"
 import { CodapComponent } from "./codap-component"
 
+const kComponentTitleHeight = 25
```
bfinzer, line 9: It would be great if this constant were defined just once.

bfinzer (approved):
👍🏻LGTM
But see notes about redundant constants.

v3/src/components/slider/slider-component.tsx
eireland: `slider.scss` no longer has that constant. I had to remove the height styling from the css to be able to change the slider height dynamically.

kswenson (approved):
👍 Looks good -- I added a commit to move the minimize flag from the TileModel to the FreeTileLayout.

PR #802, opened 2023-05-30
Adding component automation tests

tejal-shah:
This PR has the following changes:
1. Adding component level tests like closing, opening, changing title, checking tooltips, etc
2. Adding tests for Calculator and Slider tiles
3. Adding component level tests for Graph and Case table tiles
4. Adding `data-testids` for some elements used by the tests
5. Formatting cypress files so they use 2-space indentation used everywhere else in the project instead of 4-space indentation

### DIFF
--- a/v3/src/components/case-table/case-table-tool-shelf-button.tsx
+++ b/v3/src/components/case-table/case-table-tool-shelf-button.tsx
@@ -91,15 +91,20 @@ export const CaseTableToolShelfMenuList = observer(function CaseTableToolShelfMe
           const tileTitle = caseMetadata?.title ? caseMetadata?.title
                                                 : tableTile?.title ? tableTile?.title : dataset.dataSet.name
           return (
-            <MenuItem key={`${dataset.dataSet.id}`} onClick={()=>handleOpenDataSetTable(dataset)}>
+            <MenuItem key={`${dataset.dataSet.id}`} onClick={()=>handleOpenDataSetTable(dataset)}
+              data-testid={`toolshelf-table-${tileTitle}`}>
               {tileTitle}
-              <TrashIcon className="tool-shelf-menu-trash-icon"
+              <TrashIcon className="toolshelf-menu-trash-icon"
                   onClick={() => handleOpenRemoveDataSetModal(dataset.dataSet.id)} />
             </MenuItem>
           )
         })}
-        <MenuItem>{t("DG.AppController.caseTableMenu.clipboardDataset")}</MenuItem>
-        <MenuItem onClick={handleCreateNewDataSet}>{t("DG.AppController.caseTableMenu.newDataSet")}</MenuItem>
+        <MenuItem data-testid="toolshelf-table-new-clipboard">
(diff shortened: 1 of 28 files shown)


v3/src/components/case-table/case-table-tool-shelf-button.tsx
```diff
@@ -91,15 +91,20 @@ export const CaseTableToolShelfMenuList = observer(function CaseTableToolShelfMe
           const tileTitle = caseMetadata?.title ? caseMetadata?.title
                                                 : tableTile?.title ? tableTile?.title : dataset.dataSet.name
           return (
-            <MenuItem key={`${dataset.dataSet.id}`} onClick={()=>handleOpenDataSetTable(dataset)}>
+            <MenuItem key={`${dataset.dataSet.id}`} onClick={()=>handleOpenDataSetTable(dataset)}
+              data-testid={`toolshelf-table-${tileTitle}`}>
               {tileTitle}
-              <TrashIcon className="tool-shelf-menu-trash-icon"
+              <TrashIcon className="toolshelf-menu-trash-icon"
```
kswenson, line 97: I would rather we standardize on the hyphenated form, which doesn't trigger warnings from spell-check extensions, for instance.

v3/src/components/component-title-bar.tsx
```diff
@@ -43,10 +43,11 @@ export const ComponentTitleBar = observer(function ComponentTitleBar(
       </Editable>
       <Flex className="header-right">
         <Button className="component-minimize-button">
-          <MinimizeIcon className="component-minimize-icon" title={t("DG.Component.minimizeComponent.toolTip")}/>
+          <MinimizeIcon className="component-minimize-icon" title={t("DG.Component.minimizeComponent.toolTip")}
+          data-testid="component-minimize-icon"/>
```
kswenson, line 47: Is it important for the `data-testid` to be on the icon? It would be more consistent with the close button if it were on the `Button`.

kswenson (approved):
👍 Looks good -- just a couple of suggestions about naming and placement of `data-testid`s.

v3/src/components/component-title-bar.tsx
tejal-shah: @kswenson , the reason I added the `data-testid` on the minimize icon is because unlike the close button where the `title` is on the button, for the minimize button, the `title` is on the icon and I'm checking these titles for the tooltips. Please let me know if it makes sense to move the `title` to the button for minimize and I can change both the `title` and `data-testid` on it.

v3/src/components/case-table/case-table-tool-shelf-button.tsx
tejal-shah: I've pushed a commit for this.

v3/src/components/component-title-bar.tsx
kswenson: I would think that the `title` _and_ the `data-testid` should be on the button -- it's the button that has the behavior that the tooltip is describing, not the icon.

PR #803, opened 2023-05-31
184540601 floating new attribute button v3

eireland:
Adds `+` button to collection title area for each collection.
Note that this still does not fix the bug re scrolling and focusing on the newly added attribute in edit mode.

### DIFF
--- a/v3/src/components/case-table/collection-table.tsx
+++ b/v3/src/components/case-table/collection-table.tsx
@@ -11,12 +11,16 @@ import { useCollectionContext } from "../../hooks/use-collection-context"
 import { useDataSetContext } from "../../hooks/use-data-set-context"
 import { IDataSet } from "../../models/data/data-set"
 import { useCaseTableModel } from "./use-case-table-model"
-import { CollectionTitle } from './collection-title'
+import { CollectionTitle } from "./collection-title"
 
 import styles from "./case-table-shared.scss"
 import "react-data-grid/lib/styles.css"
 
-export const CollectionTable = observer(function CollectionTable() {
+interface IProps {
+  isTileInFocus: boolean
+}
+
+export const CollectionTable = observer(function CollectionTable({ isTileInFocus }: IProps) {
   const data = useDataSetContext()
   const collection = useCollectionContext()
   const tableModel = useCaseTableModel()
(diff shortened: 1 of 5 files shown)


v3/src/components/case-table/collection-table.tsx
```diff
@@ -11,12 +11,16 @@ import { useCollectionContext } from "../../hooks/use-collection-context"
...
 import { useCaseTableModel } from "./use-case-table-model"
-import { CollectionTitle } from './collection-title'
+import { CollectionTitle } from "./collection-title"
 
 import styles from "./case-table-shared.scss"
 import "react-data-grid/lib/styles.css"
 
-export const CollectionTable = observer(function CollectionTable() {
+interface IProps {
+  isTileInFocus: boolean
```
kswenson, line 20: Rather than passing this in as a prop, the child can just get it using `useTileModelContext()` which returns the current selection state.

v3/src/components/case-table/collection-title.tsx
```diff
@@ -4,13 +4,20 @@ import { Editable, EditableInput, EditablePreview } from "@chakra-ui/react"
...
+import { uniqueName } from "../../utilities/js-utils"
 import { useDataSetContext } from "../../hooks/use-data-set-context"
 import { useCollectionContext } from "../../hooks/use-collection-context"
 import { getCollectionAttrs } from "../../models/data/data-set-utils"
 import t from "../../utilities/translation/translate"
+import AddIcon from "../../assets/icons/icon-add-circle.svg"
 
-export const CollectionTitle = observer(function CollectionTitle() {
+interface IProps {
+  isTileInFocus: boolean
```
kswenson, line 17: Rather than passing this in as a prop, it can be retrieved from `useCollectionModelContext()`.

v3/src/components/case-table/collection-title.tsx
```diff
@@ -68,7 +78,15 @@ export const CollectionTitle = observer(function CollectionTitle() {
...
 
+  const handleAddNewAttribute = () => {
+    const newAttrName = uniqueName("newAttr",
+      (aName: string) => !data?.attributes.find(attr => aName === attr.name)
+     )
+    data?.addAttribute({name: newAttrName})
+  }
+
   const casesStr = t(caseCount === 1 ? "DG.DataContext.singleCaseName" : "DG.DataContext.pluralCaseName")
+  const addIconClass = clsx("add-icon", { focused: isTileInFocus})
```
kswenson, line 89:
```suggestion
  const addIconClass = clsx("add-icon", { focused: isTileInFocus })
```
for consistency

kswenson (changes requested):
👍 Looks good -- one suggestion to use an available context rather than passing down selection status as a prop. Also, the table should scroll horizontally to the newly created attribute, if necessary.

bfinzer (changes requested):
* The button should have a hover text that says "Add a new attribute to this table" as in V2.
* I've added a story about making the new attribute name visible and immediately editable and assigned it to you.

bfinzer (approved):
👍🏻LGTM

kswenson (commented):
Code looks fine, but the behavior is rather mysterious if you don't scroll to the newly created column.

kswenson (approved):
👍 Code looks fine -- I would suggest scrolling to the newly-created column should be part of this PR and auto-editing can be part of the next PR, but I leave that to @bfinzer's discretion. The current behavior is potentially confusing without the auto-scroll.

eireland:
Auto scroll to newly added attribute stories are https://www.pivotaltracker.com/story/show/184432143
and
https://www.pivotaltracker.com/story/show/185315714

PR #805, opened 2023-06-01
V3: Fixes table scroll bug on focus change

eireland:
Removes duplicate ref on the data grid
Makes grid element scroll match the table scroll top because it was getting lost on grid re-render.


bfinzer (approved):
👍🏻LGTM

kswenson (approved):
👍 Looks good -- after extensive investigation to see if there might be a way to prevent the miscreant scroll from occurring in the first place I came to the conclusion that your approach seems like the best we can do. I took the liberty of extending it to cover horizontal scrolling as well.

PR #807, opened 2023-06-02
V3: Slows component minimization to be closer to V2

eireland:


v3/src/components/free-tile-row.scss
```diff
@@ -10,7 +10,7 @@
     position: absolute;
     box-shadow: 0 1px 3px 0 rgb(0 0 0 / 25%);
     border-radius: 6px;
-    transition: height 0.25s;
+    transition: height 0.5s ease-in-out 0s;
```
bfinzer, line 13: In V2 the duration of the transition is 0.4s. Unless there's a reason, let's use that.

bfinzer (approved):
👍🏻LGTM
* Just suggestion to use same value as in V2 if that seems right.

v3/src/components/free-tile-row.scss
kswenson: The slider has its own `transition` statement. Whatever values are decided upon, it should probably be put in `vars.scss` so it's not duplicated in multiple places.

kswenson (changes requested):
👍 Looks good -- just a suggestion to DRY the transition definition.

kswenson (approved):
👍 LGTM

PR #809, opened 2023-06-05
Invalid point positions for dot plots when brought in from out of bounds

bfinzer:
[#185335006] Bug fix: Plot points appear in invalid places

* NaN was being returned as a screen coordinate when points were not to be plotted because the axis had been rescaled to no longer include them. But a transition (with duration zero!) was scheduled to move the point to a valid coordinate when the axis rescale allowed. The stored NaN caused an exception during the transition that prevented the point from reaching the desired coordinate.
**Conclusion**: Don't return `NaN` for screen coordinates. Use `null` instead.


kswenson (approved):
👍 LGTM

PR #810, opened 2023-06-06
Axis drop area fix

bfinzer:
[#184764820] Bug fix: x-axis attribute drop area is after/below the legend drop area instead of before/above it

* Unnecessarily complicated mechanism replaced with something much more straightforward and less bug-prone, relying on GraphLayout being able to compute axis bounds


v3/src/components/axis/components/droppable-axis.tsx
```diff
@@ -4,8 +4,8 @@ import React, { CSSProperties } from "react"
 import { createPortal } from "react-dom"
 import {clsx} from "clsx"
 import { AxisPlace } from "../axis-types"
-import { useAxisBounds } from "../hooks/use-axis-bounds"
 import { DropHint } from "../../graph/components/drop-hint"
+import {useGraphLayoutContext} from "../../graph/models/graph-layout"
```
kswenson, line 8: I was initially concerned that this was an additional dependency from `axis` to `graph`. Upon closer inspection, however, it appears that `DroppableAxis` is _only_ used by the graph, so I think the entire `DroppableAxis` component should be moved into the `graph` folder.

v3/src/components/graph/components/graph-axis.tsx
```diff
@@ -44,16 +46,16 @@ export const GraphAxis = observer(function GraphAxis(
...
   useDropHandler(droppableId, active => {
     const { dataSet, attributeId: droppedAttrId } = getDragAttributeInfo(active) || {}
     dataSet && droppedAttrId && isDropAllowed(place, dataSet, droppedAttrId) &&
       onDropAttribute?.(place, dataSet, droppedAttrId)
   })
 
+  useEffect(() => {
+    setParentElt(wrapperElt?.closest(kGraphClassSelector) as HTMLDivElement ?? null)
+  }, [wrapperElt])
+
```
kswenson, line 58: This isn't necessarily a problem, but note that this will require three renders before everything is set up:
- first render calls `setWrapperElt()`, which triggers the second render
- after the second render, this effect will fire, which calls `setParentElt()`, which triggers the third render
- third render will have everything in place

This could be reduced to two by determining the parent element during the first call to `setWrapperElt()`.

If you were interested in pursuing this, it would look something like this:
```typescript
  const parentEltRef = useRef<HTMLDivElement | null>(null)
  const [wrapperElt, _setWrapperElt] = useState<SVGGElement | null>(null)
  const setWrapperElt = useCallback((elt: SVGGElement | null) => {
    parentEltRef.current = elt?.closest(kGraphClassSelector) as HTMLDivElement ?? null
    _setWrapperElt(elt)
  }, [])
```

kswenson (approved):
👍 Looks good with two suggestions:
- move the `droppable-axis.tsx` module into the `graph` folder, since it's only used by the graph
- consider eliminating the third startup render in `GraphAxis` as suggested below

PR #811, opened 2023-06-06
chore: Update to Google Analytics 4 (PT-184463878)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/184463878

These changes update CODAP to use the latest version of Google Analytics. The previous version will no longer work after June 30.

### DIFF
--- a/apps/dg/resources/scripts.rhtml
+++ b/apps/dg/resources/scripts.rhtml
@@ -1,14 +1,13 @@
 <% content_for :page_javascript do %>
 <% if config.enable_google_analytics %>
+  <script async src="https://www.googletagmanager.com/gtag/js?id=G-ZYRHG9RN3K"></script>
   <script>
     if (DG.GOOGLE_ANALYTICS_ID) {
-      (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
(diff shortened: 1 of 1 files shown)


apps/dg/resources/scripts.rhtml
```diff
@@ -1,14 +1,13 @@
 <% content_for :page_javascript do %>
 <% if config.enable_google_analytics %>
+  <script async src="https://www.googletagmanager.com/gtag/js?id=G-ZYRHG9RN3K"></script>
```
emcelroy, line 3: I want to note for posterity that I'm aware hard-coding the Google Analytics ID value here is a little at odds with our use of `DG.GOOGLE_ANALYTICS_ID` below. Unfortunately, there doesn't seem to be a straightforward way to use a dynamic value for this first `script` tag. It will at least do no harm to CODAP functionality or to the Concord Consortium's tracking of CODAP usage via GA.

eireland (approved):
👍

kswenson:
Note that this PR is based on `main`, which is used for `v3` development. `v2` is deployed from `master`, so you probably want to make a PR for that as well.

emcelroy:
> Note that this PR is based on `main`, which is used for `v3` development. `v2` is deployed from `master`, so you probably want to make a PR for that as well.

@kswenson Thanks for pointing that out 👍 New PR is [here](https://github.com/concord-consortium/codap/pull/812).

PR #812, opened 2023-06-07
chore: Update to Google Analytics 4 (PT-184463878)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/184463878

These changes update CODAP to use the latest version of Google Analytics. The previous version will no longer work after June 30.


eireland (approved):
👍

PR #815, opened 2023-06-10
Right axis drop hint fix

bfinzer:
[#184758581] Bug fix: Duplicate droppable areas shown in right y axis

* This bug was already fixed except that the drop hint was incorrect. We fix that.


eireland (approved):
Looks good.

PR #817, opened 2023-06-12
Points were being plotted that shouldn't have been

bfinzer:
[#185384939] Bug fix: Problems plotting points in graph

* GraphController:handleAttributeAssignment now needs to call matchCirclesToData
* Couple of code cleanups


kswenson (approved):
👍 LGTM

PR #819, opened 2023-06-14
Categorical axis refresh bug

bfinzer:
[#185393868] Bug fix: Categorical axis failure to refresh

* Adjusting use-sub-axis to respond properly to category set changes
* Memoize SubAxis so that it only rerenders when it changes
* Define getAxisModel as a callback in GraphAxis so it won't trigger rerender of SubAxis every time
* In use-sub-axis install a useRef to be able detect when the category set reaction is specious


kswenson (approved):
👍 LGTM

PR #821, opened 2023-06-19
Drag from Graph axes and legend

bfinzer:
[#185386878] Feature: Users can drag attributes from graph axes and legend

* Added AttributeDragOverlay to GraphComponent
* AxisOrLegendAttributeMenu is now draggable
* Moved attribute-drag-overlay to components/utility-components
* Stubbed out handleDropAnimation in AttributeDragOverlay but haven't yet implemented


v3/src/components/case-table/case-table.tsx
```diff
@@ -1,7 +1,7 @@
 import { useDndContext } from "@dnd-kit/core"
 import { observer } from "mobx-react-lite"
 import React, { CSSProperties, useEffect, useRef } from "react"
-import { AttributeDragOverlay } from "./attribute-drag-overlay"
+import { AttributeDragOverlay } from "../utility-components/attribute-drag-overlay"
```
kswenson, line 4: How about `src/components/drag-drop/attribute-drag-overlay` instead? We may want to create a separate `tiles` folder for `graph`, `case-table`, etc. because `components` currently contains both tiles and general-use components, but `utility-components` doesn't seem quite right.

kswenson (approved):
👍 Looks good -- my only quibble is with the location of `attribute-drag-overlay.*` files.

PR #823, opened 2023-06-26
feat: The user can add a movable line (PT-181923044)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/181923044


v3/src/components/graph/adornments/adornment-models.ts
```diff
@@ -52,27 +52,28 @@ export const MovableValueModel = Adornment
...
 export const MovableLineModel = Adornment
   .named('MovableLineModel')
   .props({
-    type: 'line',
-    intercept: types.number,
-    slope: types.number,
-    pivot1: types.optional(PointModel, kInfinitePoint),
-    pivot2: types.optional(PointModel, kInfinitePoint)
+    type: 'Movable Line',
+    lines: types.map(MovableLineParams)
```
emcelroy, line 66: Should `lines` be something more generic like `instances` so the same property name can be used by all the adornment models?

v3/src/components/graph/adornments/adornments.tsx
```diff
@@ -0,0 +1,124 @@
...
+      top,
+      width: `${width}px`,
+    }
+  let cellsRendered = 0
+
+  const renderAdornments = () => {
+    const adornmentNodes = []
+    while (cellsRendered < cellsRequired) {
+      adornmentNodes.push(
+        <div key={`graph-adornments-cell-${cellsRendered}`} className="graph-adornments__adornment">
```
emcelroy, line 106: This `className` value here isn't quite right since multiple adornments could be rendered within it. I'll need to adjust it.

v3/src/components/graph/adornments/adornments.tsx
```diff
@@ -0,0 +1,124 @@
...
+    cellsRequired = (colCount > 0 ? colCount : 1) * (rowCount > 0 ? rowCount : 1),
+    gridTemplateColumns = `repeat(${colCount}, 1fr)`,
+    gridTemplateRows = `repeat(${rowCount}, 1fr)`,
+    gridStyle = {
+      gridTemplateColumns,
+      gridTemplateRows,
+      height: `${height}px`,
+      left,
+      top,
+      width: `${width}px`,
```
kswenson, line 98:
```suggestion
      height,
      left,
      top,
      width,
```
should work as well, because `px` are assumed when passing numeric values to React style props.

v3/src/components/graph/adornments/adornments.tsx
```diff
@@ -0,0 +1,124 @@
...
+        </div>
+      )
+      cellsRendered++
+    }
+    return adornmentNodes
+  }
+
+  return (
+    <div className={kGraphAdornmentsClass} style={gridStyle}>
+      {renderAdornments()}
```
kswenson, line 121: This is a matter of stylistic preference, but I would eliminate the `renderAdornments` function and just put its contents inline here. The `Adornments` functional component renders its contents by looping over the adornments, having it defer to another function seems unnecessary.

v3/src/components/graph/adornments/movable-line.tsx
```diff
@@ -3,72 +3,124 @@ import {autorun} from "mobx"
...
+  plotIndex: number
+  plotWidth: number
+  transform?: string
   xAxis: INumericAxisModel
   yAxis: INumericAxisModel
-  transform: string
 }) => {
-  const {model, xAxis, yAxis, transform} = props,
+  const {lineKey='', model, plotHeight, plotIndex, plotWidth, transform, xAxis, yAxis} = props,
+    data = useDataConfigurationContext(),
```
kswenson, line 24: We frequently use `data` to refer to the `DataSet`; `dataConfig` would be more conventional.

v3/src/components/graph/adornments/movable-line.tsx
```diff
@@ -89,55 +141,65 @@ export const MovableLine = (props: {
...
+        fixHandles(lineObject.handleLower, 1)
+        fixHandles(lineObject.handleMiddle, 2)
+        fixHandles(lineObject.handleUpper, 3)
         refreshEquation()
       })
       return () => disposer()
-    }, [pointsOnAxes, lineObject, transform, xScale, yScale, model, xAxis, yAxis, xRange, yRange]
+    }, [pointsOnAxes, lineKey, lineObject, plotHeight, plotWidth, transform, 
+        xScale, yScale, model, xAttrName, xSubAxesCount, xAxis, yAttrName, 
+        ySubAxesCount, yAxis, xRange, yRange]
```
kswenson, line 152: This long list of dependencies says that if any of these values change, the `autorun` will be disposed and reinstalled. It would be better to let the existing `autorun` respond to any changes instead, which can be done by dereferencing the relevant values inside the `autorun` rather than outside the `useEffect()`. Just something to think about moving forward.

v3/src/components/graph/adornments/movable-line.tsx
```diff
@@ -166,16 +235,37 @@ export const MovableLine = (props: {
...
+    newLineObject.handleMiddle = selection.append('rect')
+        .attr('class', 'movable-line-handle movable-line-middle-handle')
+    newLineObject.handleUpper = selection.append('rect')
+        .attr('class', 'movable-line-handle movable-line-upper-handle')
+
+    // Set up the corresponding equation box
+    // Define the selector that corresponds with the adornment container
+    // TODO: Make this value more specific or find another way to target the correct element. It's currently 
+    // causing a bug when there is more than one graph present -- the equation box can be added to the wrong 
+    // graph because the selector is too generic.
```
kswenson, line 249: You can use the instance id (`useInstanceIdContext()`) to scope to a particular tile instance.

v3/src/components/graph/components/graph.tsx
```diff
@@ -188,6 +189,7 @@ export const Graph = observer(function Graph({graphController, graphRef, dotsRef
           />
         </svg>
         {renderDroppableAddAttributes()}
+        {graphModel.adornments.length > 0 && <Adornments />}
```
kswenson, line 192:
```suggestion
        <Adornments />
```
The `Adornments` component should be responsible for deciding what to render when there are no adornments.

kswenson (changes requested):
👍 Looks good -- I made a bunch of suggestions, some of them simple and probably worth doing now, others more substantive and can wait for a subsequent PR. I leave that to your judgment. None of them are necessarily show-stoppers, but the pivot-related changes seem most pressing given that they affect the file format.

v3/src/components/graph/adornments/adornments.tsx
emcelroy: @kswenson Is the change I made for this what you had in mind? I believe the while loop has to be outside of the return statement, so I couldn't put the full contents of `renderAdornments` here. I'm not sure if that's exactly what you meant, though.

emcelroy: @kswenson Is [the change I made for this](https://github.com/concord-consortium/codap/blob/23bee14cd3834b0b9df1cbcc1844dcb3ff38dda7/v3/src/components/graph/adornments/adornments.tsx#L63-L87) what you had in mind? I believe the while loop has to happen outside/before the return, so I couldn't put the full contents of `renderAdornments` there exactly, but I'm not sure that's exactly what you meant.

v3/src/components/graph/adornments/adornment-models.test.ts
kswenson, line 1: Nice! Good to have some tests out of the box.

v3/src/components/graph/adornments/adornment-models.ts
```diff
@@ -26,23 +26,43 @@ export const PointModel = types.model("Point", {
...
 export const kInfinitePoint = {x:NaN, y:NaN}
 
-const Adornment = types.model("Adornment", {
-  id: types.optional(types.identifier, () => typedId("ADRN")),
-  type: types.optional(types.string, () => {
-    throw "type must be overridden"
-  }),
-  isVisible: true
-})
+export const Adornment = types.model("Adornment", {
```
kswenson, line 29:
```suggestion
export const AdornmentModel = types.model("AdornmentModel", {
```
for consistency

v3/src/components/graph/adornments/adornment-models.ts
```diff
@@ -26,23 +26,43 @@ export const PointModel = types.model("Point", {
...
+      }
+    },
+    setClassNameFromKey(key: string) {
+      const className = key.replace(/\{/g, '')
+        .replace(/\}/g, '')
+        .replace(/: /g, '-')
+        .replace(/, /g, '-')
+
+      return className
     }
```
kswenson, line 58: These methods don't change the model, so they should be `views`.

v3/src/components/graph/adornments/movable-line.tsx
```diff
@@ -185,8 +282,15 @@ export const MovableLine = (props: {
   }, [])
```
kswenson, line 282: Address the rule-of-hooks warning. If this is intended to run only on mount, then the warning can simply be disabled along with a comment as to why.

v3/src/components/graph/adornments/adornments.tsx
kswenson: I'm not sure what I had in mind would have worked exactly, but this seems like an improvement.

kswenson (approved):
👍 Looks good -- just a few suggested tweaks.

PR #825, opened 2023-06-27
Categorical axis reordering

bfinzer:
[#181912254] Feature: Categories displayed on a categorical axis can be reordered through dragging

* No longer relying on d3 to display categorical axis so that we can more easily control user reordering categories through dragging
* Refactoring so that the categorical axis does not need to be rebuilt from scratch with each rerender
* Removed axis-utils.test.ts because the only test it contained was no longer relevant


emcelroy (approved):
Looks good! The bugs I saw before are now gone.

There are some warnings about missing spaces after instances of `switch` in axis-elements.ts but I don't think it's critical to fix those.

PR #826, opened 2023-06-29
feat: Serialization model for adornments (PT-185406920)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/185406920

This is currently just for getting feedback before I continue.

### DIFF
--- a/v3/src/components/graph/adornments/adornment-component-info.ts
+++ b/v3/src/components/graph/adornments/adornment-component-info.ts
@@ -1,0 +1,25 @@
+import React from "react"
+
+export interface IAdornmentComponentInfo {
+  adornmentEltClass: string
+  Component: React.ComponentType<any>
+  hint: string
+  label: string
+  order: number
+  type: string
+}
+
+const gAdornmentComponentInfoMap = new Map<string, IAdornmentComponentInfo>()
+
+export function registerAdornmentComponentInfo(info: IAdornmentComponentInfo) {
+  gAdornmentComponentInfoMap.set(info.type, info)
+}
+
+export function getAdornmentComponentKeys() {
+  return Array.from(gAdornmentComponentInfoMap.keys())
+}
+
+// undefined is supported so callers do not need to check the type before passing it in
+export function getAdornmentComponentInfo(type?: string) {
+  return type ? gAdornmentComponentInfoMap.get(type) : undefined
+}
--- a/v3/src/components/graph/adornments/adornment-types.ts
+++ b/v3/src/components/graph/adornments/adornment-types.ts
@@ -1,0 +1,2 @@
+export const kGraphAdornmentsClass = "graph-adornments-grid"
+export const kGraphAdornmentsClassSelector = `.${kGraphAdornmentsClass}`
--- a/v3/src/components/graph/adornments/adornment.tsx
+++ b/v3/src/components/graph/adornments/adornment.tsx
@@ -1,9 +1,9 @@
 import React from "react"
-import { IAdornmentModel, isMovableLine } from "./adornment-models"
+import { IAdornmentModel } from "./adornment-models"
 import { useGraphLayoutContext } from "../models/graph-layout"
-import { MovableLine } from "./movable-line"
 import { useGraphModelContext } from "../models/graph-model"
 import { INumericAxisModel } from "../../axis/models/axis-model"
+import { getAdornmentComponentInfo } from "./adornment-component-info"
 
 interface IProps {
   adornment: IAdornmentModel,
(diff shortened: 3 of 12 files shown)


v3/src/components/graph/adornments/adornment-component-info.ts
```diff
@@ -0,0 +1,27 @@
+import React from "react"
+
+export interface IAdornmentComponentInfo {
+  adornmentEltClass: string
+  Component: React.ComponentType<any>
+  inspectorList: {
+    position: number
+    label: string
+    hint: string
+  }
```
emcelroy, line 10: Does it make sense to include information like this here for use in building the adornments list in the Inspector panel?

v3/src/components/graph/adornments/adornment.tsx
```diff
@@ -28,20 +28,18 @@ export const Adornment = ({ adornment, index, xCategories, yCategories}: IProps)
...
-        plotIndex={index}
-        plotWidth={subPlotWidth}
-        xAxis={graphModel.getAxis('bottom') as INumericAxisModel}
-        yAxis={graphModel.getAxis('left') as INumericAxisModel}
-      />
-    )
-  }
-  return null
+  const componentInfo = getAdornmentComponentInfo(adornment.type)
+  const Component = componentInfo?.Component as any
```
emcelroy, line 32: I realize this use of `any` will need to be fixed.

v3/src/components/graph/adornments/adornment-component-info.ts
bfinzer: I don't think so. I think the inspector panel knows about the adornments, but not vice versa.

kswenson: I'm inclined to say this should be part of the registration system. We may want to  call it something more generic than `inspectorList` -- I might eliminate the hierarchy entirely and just have something like:
```typescript
export interface IAdornmentComponentInfo {
  type: string
  adornmentEltClass: string
  Component: React.ComponentType<any>
  order: number
  label: string
  hint: string
}
```
Otherwise, the inspector code would need to import all of the adornment constants individually, which can be avoided if the required information is registered in this way.

v3/src/components/graph/adornments/adornment-component-info.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export function registerAdornmentComponentInfo(info: IAdornmentComponentInfo) {
+  gAdornmentComponentInfoMap.set(info.type, info)
+}
+
+export function getAdornmentComponentKeys() {
+  return Array.from(gAdornmentComponentInfoMap.keys())
+}
+
+// undefined is supported so callers do not need to check the id before passing it in
```
kswenson, line 24:
```suggestion
// undefined is supported so callers do not need to check the type before passing it in
```
no doubt a copy-paste error that originated in the source

v3/src/components/graph/adornments/adornment-types.ts
```diff
@@ -0,0 +1,6 @@
+export const kGraphAdornmentsClass = "graph-adornments-grid"
+export const kGraphAdornmentsClassSelector = `.${kGraphAdornmentsClass}`
+
+export const kMovableLineClass = "movabe-line"
+export const kMovableLineType = "Movable Line"
+export const kMovableLinePrefix = "ADRN"
```
kswenson, line 6: Ultimately, movable line-specific constants should be in a movable line-specific file, e.g. `movable-line-constants.ts` or `movable-line-types.ts` so that the movable line code can be self-contained.

v3/src/components/graph/adornments/movable-line-registration.test.ts
```diff
@@ -0,0 +1,20 @@
+import { getAdornmentComponentInfo } from "./adornment-component-info";
+import { getAdornmentContentInfo } from "./adornment-content-info";
```
kswenson, line 2: Lint warnings about semicolons

v3/src/components/graph/register-adornment-types.ts
```diff
@@ -0,0 +1,9 @@
+import "./adornments/movable-line-registration"
+
+const gAdornmentRegistration: Record<string, () => void> = {
+  "MovableLine": () => import(/* webpackChunkName: "MovableLine" */"./adornments/movable-line-registration")
+};
+
+export function registerAdornmentTypes(adornmentTypeIds: string[]) {
+  return Promise.all(adornmentTypeIds.map(id => gAdornmentRegistration[id]?.()))
+}
```
kswenson, line 9: The point of using dynamic imports with tiles is that in CLUE currently (and potentially down the road in CODAP), Webpack is configured to package individual tiles into their own bundles and so, for instance, all of the code for the Dataflow tile can be included/excluded depending on the unit configuration. Dynamic loading isn't a concern for individual adornments, so the dynamic imports approach isn't necessary. For CODAP's purposes, all that is required is that the desired adornment registration files be imported.
```suggestion
```

v3/src/components/graph/components/graph.tsx
```diff
@@ -53,6 +54,8 @@ export const Graph = observer(function Graph({graphController, graphRef, dotsRef
     xAttrID = graphModel.getAttributeID('x'),
     yAttrID = graphModel.getAttributeID('y')
 
+  registerAdornmentTypes([])
```
kswenson, line 57:
```suggestion
```

v3/src/components/graph/components/graph.tsx
```diff
@@ -29,6 +29,7 @@ import {IDataSet} from "../../../models/data/data-set"
 import {useDataTips} from "../hooks/use-data-tips"
 import {onAnyAction} from "../../../utilities/mst-utils"
 import { Adornments } from "../adornments/adornments"
+import { registerAdornmentTypes } from "../register-adornment-types"
```
kswenson, line 32:
```suggestion
```
It should be sufficient to put this in `graph-component.tsx`:
```typescript
import "../register-adornment-types"
```

kswenson (commented):
👍 Looks good -- a few suggestions, some of which are simplifications.

v3/src/components/graph/adornments/adornments.tsx
```diff
@@ -16,7 +16,7 @@ export const Adornments = observer(function Adornments() {
     layout = useGraphLayoutContext(),
     adornments = graphModel.adornments
 
-  if (adornments.length === 0) return null
+  if (adornments?.length === 0) return null
```
kswenson, line 19:
```suggestion
  if (!adornments?.length) return null
```
Otherwise, the test would fail if `adornments` is `undefined`.

v3/src/components/graph/adornments/movable-line-types.ts
```diff
@@ -0,0 +1,5 @@
+export const kMovableLineClass = "movabe-line"
```
kswenson, line 1:
```suggestion
export const kMovableLineClass = "movable-line"
```

kswenson (approved):
👍 LGTM -- just a couple of minor suggestions.

v3/src/components/graph/adornments/adornments.tsx
emcelroy: Ah, right. Thanks 👍

PR #827, opened 2023-07-01
feat: synchronized table scrolling

kswenson:
- implements synchronized table scrolling on user scroll
- synchronizes table scroll on new collection drop
- uses `scroll-behavior: smooth` rather than programmatically animating scroll position
- fixes collection title height bug
- fixes floating new attribute button placement and z-index
- fixes bug which inadvertently disabled case table-specific collision detection on drag


bfinzer (approved):
👍🏻LGTM

PR #828, opened 2023-07-05
feat: Movable line equation box can be repositioned (PT-185472784)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/185472784

Adds the option to unpin a movable line's equation box from the line so that it can be placed in a specific spot within the associated plot's container. The equation box can be unpinned by clicking on it and dragging it away from the line.

These changes also include a minor styling tweak. When the user's mouse cursor hovers over the equation box, the line's stroke width increases from 1 to 2.


v3/src/components/graph/adornments/movable-line.tsx
```diff
@@ -56,6 +56,23 @@
     xSubAxesCount = layout.getAxisMultiScale('bottom')?.repetitions ?? 1,
     ySubAxesCount = layout.getAxisMultiScale('left')?.repetitions ?? 1
 
+  const setEquationContainerClass = useCallback(() => {
```
bfinzer, line 59: I'm confused about the naming. Nothing is being set, right? Why not just name it equationContainerClass or getEquationContainerClass?

bfinzer, line 59: Similar comment for setEquationContainerSelector possibly better named just equationContainerSelector?

v3/src/components/graph/adornments/movable-line.tsx
```diff
@@ -206,7 +224,25 @@
...
+    moveEquation = useCallback((event: { x: number, y: number, dx: number, dy: number }) => {
+      if (event.dx !== 0 || event.dy !== 0) {
+        const equationContainerSelector = setEquationContainerSelector(),
+          equation = select(`${equationContainerSelector} p`).node() as Element,
+          equationWidth = equation?.getBoundingClientRect().width || 0,
+          equationHeight = equation?.getBoundingClientRect().height || 0,
+          left = event.x - equationWidth / 2,
+          top = event.y - equationHeight / 2
+
+        select(`${equationContainerSelector} p`)
```
bfinzer, line 238: Whether the equation is pinned or not is part of the model state since when the document is saved and restored we don't want a pinned equation to start moving when the line is dragged.

v3/src/components/graph/adornments/movable-line.tsx
```diff
@@ -250,34 +291,33 @@
...
-      .style('opacity', 0)
+      .attr('data-testid', `movable-line-equation-${model.setClassNameFromKey(lineKey)}`)
+      .on('mouseover', () => { newLineObject.line.style('stroke-width', 2) })
+      .on('mouseout', () => { newLineObject.line.style('stroke-width', 1) })
+
+    setLineObject(newLineObject)
+
+    equationDiv.style('opacity', 0)
       .transition()
       .duration(1000)
```
bfinzer, line 316: There's a constant for this: transitionDuration defined in graphing-types.ts

bfinzer (approved):
👍🏻LGTM
A few small suggestions and one note about pinned/unpinned needing to be part of the model state that you can add to the story about save and restore of equation position.

v3/src/components/graph/adornments/movable-line.tsx
emcelroy: @bfinzer I think it is setting something in that it's building a value out of other values and according to a certain condition. I think `get` would imply the function fetches a value from another source.

I think we generally append `set` (or `get`) to function names so they don't clash with associated variable names. If the function was named `equationContainerClass`, you couldn't do `const equationContainerClass = equationContainerClass()`.

I suppose we could have the function named `equationContainerClass` and forgo the use of the variable. In other words, use `equationContainerClass()` where we're now using `equationContainerClass`, but I'm not sure that's a pattern we generally follow?

In any case, I'm not super happy with these functions in general. Maybe there's a more concise way to target the elements without having to build these somewhat complex selector values. I wasn't able to come up with an alternative previously, but I could give it another try if you think it's worth doing?

bfinzer: I think using equationContainerClass() in place of the variable is a good solution even though it's somewhat unusual.

v3/cypress/e2e/graph.spec.ts
```diff
@@ -98,3 +98,41 @@ context("Graph UI", () => {
     })
   })
 })
+
+context("Graph adornments", () => {
```
kswenson, line 102: Nice job including cypress tests at this stage! I suspect this section will get quite long, so I think it would be better to start an `adornments.spec.ts` file.

v3/src/components/graph/adornments/movable-line.tsx
kswenson: I concur with @bfinzer -- these functions are computing/returning a value, not `set`ting something (e.g. state).

v3/src/components/graph/adornments/movable-line.tsx
```diff
@@ -99,19 +116,20 @@ export const MovableLine = (props: {
...
             .style('height', `${plotHeight}px`)
-          select(`${lineContainerClass} p`)
-            .style('left', `${screenX}px`)
-            .style('top', `${screenY}px`)
-            .html(string)
+          equation.html(string)
+          // The equation may have been unpinned from the line if the user
+          // dragged it away from the line. Only move the equation if it
+          // is still pinned.
+          if (!equation.classed('unpinned')) {
```
kswenson, line 129: As @bfinzer mentions elsewhere, this is essentially using the DOM's `unpinned` class as the model, which will prevent it from being serialized. `pinned`/`unpinned` should be a property of the cell or model, and then the class should be determined from the cell/model.

kswenson (changes requested):
👍 LGTM -- just a couple of suggestions, including @bfinzer's naming recommendations.

bfinzer (approved):
👍🏻LGTM
Nice job!

v3/src/components/graph/adornments/adornment-models.ts
```diff
@@ -83,6 +84,9 @@ export const MovableLineParams = types.model("MovableLineParams", {
     pivot2: PointModel.create()
   }))
   .actions(self => ({
+    setUnpinnedEquationCoords(coords: any) {
```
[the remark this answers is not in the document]
emcelroy: @bfinzer Thanks for pointing that out. I initially tried `Point` but got an error I didn't understand, so switched it to `any` planning to revisit it later, but then forgot to. I've worked out the problem and changed the type to `Point` in the latest commit.

v3/src/components/graph/adornments/adornment-models.ts
```diff
@@ -83,6 +84,9 @@ export const MovableLineParams = types.model("MovableLineParams", {
     pivot2: PointModel.create()
   }))
   .actions(self => ({
+    setEquationCoords(coords: Point) {
```
kswenson, line 87: 👍 `Point` is equivalent to the snapshot type for `PointModel`.

v3/src/components/graph/adornments/movable-line.tsx
```diff
@@ -56,6 +56,16 @@ export const MovableLine = (props: {
...
 
+  const equationContainerClass = useCallback(() => {
+    const classFromKey = model.setClassNameFromKey(lineKey)
+    return `movable-line-equation-container${lineKey && lineKey !== '' ? `-${classFromKey}` : ''}`
+  }, [lineKey, model])
+
+  const equationContainerSelector = useCallback(() => {
+    const gridContainerClass = `graph-adornments-grid.${instanceId}`
+    return `.${gridContainerClass} .${equationContainerClass()}`
+  }, [instanceId, equationContainerClass])
```
kswenson, line 67: Minor point, but since these don't rely on many values from the surrounding component, my inclination would be to define them globally outside the component so you don't have to use `useCallback`. It could look something like:
```typescript
function equationContainer(model: IMovableLineModel, instanceId: string, lineKey?: string) {
   ... string manipulation code ...
   return { equationContainerClass, equationContainerSelector }
}
```
and then the calling code would just be:
```typescript
  const { equationContainerClass, equationContainerSelector } = equationContainer(model, instanceId, lineKey)
```

v3/src/components/graph/adornments/movable-line.tsx
```diff
@@ -56,6 +56,16 @@ export const MovableLine = (props: {
     xSubAxesCount = layout.getAxisMultiScale('bottom')?.repetitions ?? 1,
     ySubAxesCount = layout.getAxisMultiScale('left')?.repetitions ?? 1
 
+  const equationContainerClass = useCallback(() => {
+    const classFromKey = model.setClassNameFromKey(lineKey)
```
kswenson, line 60: This has come up before, but this naming makes it seems like this method is setting a property on the model. It should be named something like `classNameFromKey()` instead.

kswenson (approved):
👍 Looks good -- I made a couple of suggestions but they don't need to hold up the PR.

PR #832, opened 2023-07-06
fix: axis attribute menu placement

kswenson:
PT: [[#185495543]](https://www.pivotaltracker.com/story/show/185495543)

>Notice that the menu that comes up is clipped by the bottom of the window instead of being forced to be (as much as possible) visible in the window.

We add `boundary="scrollParent"` to the `<Menu>` component, since the Graph's `scrollParent` is the document.


bfinzer (approved):
👍🏻LGTM

PR #833, opened 2023-07-06
Fix graph rehydration bug

bfinzer:
[#185496498] Bug fix: Graph rehydration bug

See the PT story for details on how to reproduce the bug

There were two problems:
1. When the categorical axis was replaced by an empty axis the stale scale was holding onto invalid ticks.
2. The axis wasn't responding to a change in the array of categories.

* In GraphController:initializeGraph handle empty axis explicitly
* In renderSubAxis handle empty axis separately from numeric axis
* Make the array of categories and explicit dependent of useAxis:computeDesiredExtent


kswenson (approved):
👍 There's a lint warning to be addressed and there are some larger architectural issues that could be addressed later if you didn't want to get into them now.

PR #834, opened 2023-07-07
fix: selecting a case should scroll it (and its parent/children) into view

kswenson:
PT: [[#185556092]](https://www.pivotaltracker.com/story/show/185556092)


bfinzer (approved):
👍🏻LGTM

PR #835, opened 2023-07-07
chore: update dependencies (e.g. TypeScript 5.1, Cypress 12.17, etc.)

kswenson:
TypeScript 5.1 update has been on hold due to lack of @typescript-eslint support (which ultimately resulted from a bug introduced in TS 5.1.0 and then rolled back in a TS 5.1.x update).

Also includes an update to ReactDataGrid which obviates the need for a CSS override for Firefox support.

Note: there are security vulnerability warnings from npm which are due to transitive dependencies on `semver`. The issue has been fixed in the most recent version of `semver` (v7) but many projects (e.g. `babel`) are still using older versions due to the need to support older node versions, etc. Efforts are underway to backport the fix to older versions of `semver` which should resolve the warnings.


bfinzer (approved):
👍🏻LGTM

PR #839, opened 2023-07-17
feat: Adornments should fade in, out (PT-185546266)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/185546266

These changes add fade-in and fade-out animations for adornments when they are added to or removed from a graph. Adornments are no longer truly removed from the graph model when their checkbox is unchecked. Instead, their `isVisible` property is switched to `false`. The `isVisible` value is used to determine when the adornment should be shown or hidden. This makes it simple to bring back adornments in their previous state if they're added, removed, then added again, which is the desired behavior.


v3/src/components/graph/adornments/movable-point/movable-point-types.ts
```diff
@@ -2,4 +2,3 @@ export const kMovablePointClass = "movable-point"
 export const kMovablePointType = "Movable Point"
 export const kMovablePointPrefix = "ADRN"
 export const kMovablePointLabelKey = "DG.Inspector.graphMovablePoint"
-export const kMovablePointHint = ""
```
emcelroy, line 5: I meant to remove this unused variable as part of the [Add Movable Point PR](https://github.com/concord-consortium/codap/pull/838) but missed it. This change doesn't affect the fade in/out at all.

v3/src/components/graph/adornments/movable-value/movable-value.tsx
```diff
@@ -10,13 +10,14 @@ import {valueLabelString} from "../../utilities/graph-utils"
...
+interface IProps {
   model: IMovableValueModel
   axis: INumericAxisModel
   transform: string
-}) => {
-  const {model, axis, transform} = props,
-    layout = useAxisLayoutContext(),
+}
+
+export function MovableValue ({model, axis, transform}: IProps) {
```
kswenson, line 19: This should presumably be an `observer` as well, but you can wait until you're actually working on the `MovableValue` to fix it.

kswenson (approved):
👍 Looks good!

v3/src/components/graph/adornments/adornment.tsx
```diff
@@ -12,35 +17,63 @@ interface IProps {
...
+
+  const adornmentWrapperClass = clsx(
+    'adornment-wrapper',
+    `${adornmentKey}-wrapper`,
+    {
+      'fadeIn': adornment.isVisible && !isFadeInComplete.current,
+      'fadeOut': !adornment.isVisible && !isFadeOutComplete.current,
+      'hidden': !adornment.isVisible && isFadeOutComplete.current
+    }
+  )
```
pjanik, line 57: @emcelroy, this code starts to fail after upgrading mobx-react-lite from v3 to v4.

I think that rendering based on the current ref value should be generally considered a bug, as updates to refs won't trigger re-renders and it's kinda random what we'll see in the DOM in the end.

Also, what's the purpose of `isFadeInComplete` and `isFadeOutComplete`? I can see they're updated once in the useEffect, but never again.

PR #840, opened 2023-07-18
fix: csv import; `sample` without `dashboard` shows case table

kswenson:
[PT-185567221](https://www.pivotaltracker.com/story/show/185567221)

The main issues were that the `DataBroker` needed to have its `sharedModelManager` property reset when the document changed and a new case table needs to be created when a `.csv` file is imported. Along the way I did some refactoring and rearranging of files that makes the diffs seems more complex than the changes required to fix the bug.


emcelroy (approved):
👍 LGTM and works as expected in the browser. One small thing I'm wondering is why not import `createTileOfType` into `ToolShelf` instead of passing it down as a prop?

kswenson:
@emcelroy I'm reassigning for review since I've refactored quite a bit to make the operations more amenable to undo/redo. I might add some more tests but I don't expect the basic implementation to change.

>One small thing I'm wondering is why not import `createTileOfType` into `ToolShelf` instead of passing it down as a prop?

The general principle is that importing a module directly couples one module to another, whereas passing in a prop pushes that coupling higher up the component tree. It can lower the import load of a module and make it easier to test because passing alternate props is easier than potentially having to mock a module import. That said, I'm not sure it was entirely worth it at the time I wrote it and it became completely pointless once the functionality in question became a method on the `DocumentContent`, which was already being passed to the `ToolShelf`, so that prop is gone now.

v3/src/models/document/document-content.ts
```diff
@@ -1,217 +1,133 @@
...
-        })
-      }
-    }
-  }))
-export interface IDocumentContentModel extends Instance<typeof DocumentContentModel> {}
-export interface IDocumentContentSnapshotIn extends SnapshotIn<typeof DocumentContentModel> {}
+// export function cloneContentWithUniqueIds(content?: DocumentContentModelType,
+//                                           asTemplate?: boolean): DocumentContentModelType | undefined {
+//   return content && DocumentContentModel.create(content.snapshotWithUniqueIds(asTemplate))
+// }
```
emcelroy, line 133: Should these commented-out lines be removed?

emcelroy (approved):
Looks good 👍 The only little issue I noticed was some commented-out lines that look like they could be removed.

PR #842, opened 2023-07-19
fix: Remove toast pop-ups for implemented adornments (PT-185639719)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/185639719

Updates the `GraphMeasurePalette` component's `handleSetting` function so it only shows toast pop-up messages for adornments that haven't been implemented yet.


kswenson (approved):
👍 LGTM

PR #843, opened 2023-07-19
fix: Removing axis attributes does not remove line (PT-185639677)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/185639677

Changes the `Adornments` component so it will skip adding adornments in the Graph model that don't support the current plot type. This also fixes a separate bug identified in this [PT story](https://www.pivotaltracker.com/story/show/185639448). In addition, there is some clean up of the Movable Point component that removes unnecessary code and fixes an issue where the point moved incorrectly when axis attributes were removed or changed.

### DIFF
--- a/v3/src/components/graph/adornments/movable-point/movable-point.tsx
+++ b/v3/src/components/graph/adornments/movable-point/movable-point.tsx
@@ -1,6 +1,7 @@
 import React, { useCallback, useEffect, useRef, useState } from "react"
 import {drag, select, Selection} from "d3"
 import {tip as d3tip} from "d3-v6-tip"
+import { autorun } from "mobx"
 import { observer } from "mobx-react-lite"
 import { IMovablePointModel } from "../adornment-models"
 import { INumericAxisModel } from "../../../axis/models/axis-model"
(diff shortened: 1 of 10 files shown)


v3/src/components/graph/adornments/movable-point/movable-point.tsx
```diff
@@ -51,17 +51,19 @@ export const MovablePoint = observer(function MovablePoint(props: IProps) {
...
     xAttr = allAttributes?.find(attr => attr.id === xAttrId),
     yAttr = allAttributes?.find(attr => attr.id === yAttrId),
     xAttrName = xAttr?.name ?? '',
-    yAttrName = yAttr?.name ?? ''
+    yAttrName = yAttr?.name ?? '',
+    xAttrNameRef = useRef(xAttrName),
+    yAttrNameRef = useRef(yAttrName)
 
   const showCoordinates = useCallback((event: MouseEvent) => {
-    if (!model.points) return
```
emcelroy, line 64: These `if (!model.points) return` lines were introduced while troubleshooting a problem I ran into in the Add Movable Point PR. They aren't necessary. I should have removed them after solving the problem, but forgot.

kswenson (changes requested):
👍 Generally looks good -- I added a commit which:
- moves the initial creation of the points from the React component to the model
- uses the `DataConfigurationModel` for determining categories rather than the `GraphLayout`
- fixes the D3 types

Serialization in the split case doesn't work perfectly, but I can't tell whether that's related to what I did or a preexisting issue. It may be related to the fact that split axes don't restore properly even without adornments, which seems to be a preexisting condition. In any case, I'm turning it over to you to take it from here.

emcelroy:
As discussed elsewhere, there is a remaining issue where the movable point will not be correctly placed when an attribute is removed from an axis, and then a new attribute is added to that axis. There are changes required outside the realm of adornments, however, to fix that problem. This code should work correctly for placing the movable point correctly in that case once those changes are made.

kswenson (approved):
👍 Looks good!

PR #844, opened 2023-07-21
fix: Movable line not draggable beyond plot boundaries (PT-185640617)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/185640617

These changes prevent a movable line from being dragged beyond a plot's boundaries which renders it inaccessible. Following the behavior in V2, the movable line will be reset to its initial state if it is dragged outside of the plot boundaries while in the perfectly vertical or perfectly horizontal state.

Included is small fix for [Movable line shows wrong equation when the line is vertical](https://www.pivotaltracker.com/story/show/185638862). This was fixed with the refactoring of the `equationString` function in graph-utils.ts.


kswenson (approved):
👍 LGTM

PR #846, opened 2023-07-24
fix: Changing attributes doesn't update Movable Point coordinate info box (PT-185648727)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/185648727

Moves the setting of mouse over/out data tip handlers to the `addBehaviors` `useEffect`. That way the `showCoordinates` function will use the latest attribute names when setting the data tip content instead of the names that existed when the line was initialized in the `createElements` `useEffect` (which only runs once).


kswenson (approved):
👍 LGTM

PR #847, opened 2023-07-24
fix: Movable line custom mouse cursors should change (PT-185631489)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/185631489

In addition to making the custom cursors change appropriately when the line gets dragged/rotated, these changes consolidate the `continueRotation1` and `continueRotation2` functions into a single function.

### DIFF
--- a/v3/src/components/graph/adornments/movable-line/movable-line.scss
+++ b/v3/src/components/graph/adornments/movable-line/movable-line.scss
@@ -27,6 +27,10 @@ svg[class^="line-"] {
 
 .movable-line-lower-cover {
   cursor: url("../assets/pivot-bot-left-cursor.png") 8 8, grab;
+
+  &.negative-slope {
+    cursor: url("../assets/pivot-top-left-cursor.png") 8 8, grab;
+  }
 }
 
 .movable-line-middle-cover {
@@ -35,6 +39,10 @@ svg[class^="line-"] {
 
 .movable-line-upper-cover {
   cursor: url("../assets/pivot-top-right-cursor.png") 8 8, grab;
+
+  &.negative-slope {
+    cursor: url("../assets/pivot-bot-right-cursor.png") 8 8, grab;
+  }
 }
 
 .movable-line-equation-container {
--- a/v3/src/components/graph/adornments/movable-line/movable-line.tsx
+++ b/v3/src/components/graph/adornments/movable-line/movable-line.tsx
@@ -182,59 +182,56 @@ export const MovableLine = observer(function MovableLine(props: IProps) {
       model.setLine({slope, intercept: newIntercept, equationCoords}, instanceKey)
     }, [instanceKey, model, xAxis, xScaleCopy, yAxis, yScaleCopy]),
 
-    // Lower cover drag handler
-    continueRotation1 = useCallback((event: { x: number, y: number, dx: number, dy: number }) => {
-      if (!pointsOnAxes.current) return
-      const lineParams = model.lines?.get(instanceKey),
-        currentPivot2 = lineParams?.pivot2,
-        equationCoords = lineParams?.equationCoords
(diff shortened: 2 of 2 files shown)


kswenson (approved):
👍 I can't say I'm thoroughly following/reviewing all of the line/pivot logic, but consolidating the `continueRotation` functions seems like a good thing and from what I can tell the code makes sense.

emcelroy:
> 👍 I can't say I'm thoroughly following/reviewing all of the line/pivot logic, but consolidating the `continueRotation` functions seems like a good thing and from what I can tell the code makes sense.

@kswwenson Do you think modifications like the below would be an improvement in terms of making it easier to follow?

```
  continueRotation = useCallback((
      event: { x: number, y: number, dx: number, dy: number },
      lineSection: string
    ) => {
      if (!pointsOnAxes.current) return
      const lineParams = model.lines?.get(instanceKey)
      // The current pivot is the pivot point on the line section not being dragged.
      // lineParams.pivot1 is the pivot point on the lower section, lineParams.pivot2 is the
      // pivot point on the upper section
      const currentPivot = lineSection === "lower" ? lineParams?.pivot2 : lineParams?.pivot1
      const equationCoords = lineParams?.equationCoords
    
      if (event.dx !== 0 || event.dy !== 0) {
        let isVertical = false
        // The new pivot will be the point on the line section where it is being dragged.
        const newPivot = { x: xScaleCopy.invert(event.x), y: yScaleCopy.invert(event.y) }
        // If the current pivot isn't valid, use the point where the other line section intersects the
        // axes as the pivot point.
        const pivot = currentPivot?.isValid()
                        ? currentPivot
                        : lineSection === "lower"
                          ? pointsOnAxes.current.pt2
                          : pointsOnAxes.current.pt1
    
        // If the line is perfectly vertical, set the new pivot's x coordinate to the x coordinate of the
        // original pivot. If the line is perfectly horizontal, set the new pivot's y coordinate to the y
        // coordinate of the original pivot.
        if (Math.abs(xScaleCopy(newPivot.x) - xScaleCopy(pivot.x)) < kTolerance) {
          newPivot.x = pivot.x
          isVertical = true
        } else if (Math.abs(yScaleCopy(newPivot.y) - yScaleCopy(pivot.y)) < kTolerance) {
          newPivot.y = pivot.y
        }

        let newSlope, newIntercept
        if (isVertical) {
          newSlope = Number.POSITIVE_INFINITY
          newIntercept = pivot.x
        } else {
          newSlope = lineSection === "lower"
            ? (pivot.y - newPivot.y) / (pivot.x - newPivot.x)
            : (newPivot.y - pivot.y) / (newPivot.x - pivot.x)
          newIntercept = newPivot.y - newSlope * newPivot.x
        }

        lineObject.lower.classed('negative-slope', newSlope < 0)
        lineObject.upper.classed('negative-slope', newSlope < 0)
    
        model.setLine(
          {
            slope: newSlope,
            intercept: newIntercept,
            pivot1: lineSection === "lower" ? newPivot : pivot,
            pivot2: lineSection === "lower" ? pivot : newPivot,
            equationCoords,
          },
          instanceKey
        )
      }
    }, [instanceKey, lineObject.lower, lineObject.upper, model, xScaleCopy, yScaleCopy]),

```

bfinzer (approved):
👍🏻LGTM

kswenson:
> > 👍 I can't say I'm thoroughly following/reviewing all of the line/pivot logic, but consolidating the `continueRotation` functions seems like a good thing and from what I can tell the code makes sense.
> 
> @kswenson Do you think modifications like the below would be an improvement in terms of making it easier to follow? I've added some comments and tried to make some parts a bit easier to read, like the setting of `newSlope` and `newIntercept`.

My comment was mainly meant to imply that I wasn't thoroughly reviewing the calculations per se, but your suggested improvements seem like a step in the right direction and the comments definitely help.

PR #849, opened 2023-07-25
chore: Restructure adornment models files.

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/181914454

As part of the work for adding the Count and Percent adornments, we're restructuring the adornment model files -- changing from a single model file for all adornment models to a separate file for each adornment's model, plus the base model. We've split this part out to get the changes into `main` quicker since we'll be porting the updated Graph code to CLUE soon and the Count and Percent adornments probably won't be done in time.


v3/src/components/graph/adornments/movable-line/movable-line-model.ts
```diff
@@ -0,0 +1,69 @@
+import { Instance, types } from "mobx-state-tree"
+import { AdornmentModel, IAdornmentModel, IUpdateCategoriesOptions, PointModel,
+         kInfinitePoint } from "../adornment-models"
+import { Point } from "../../graphing-types"
+import { IAxisModel } from "../../../axis/models/axis-model"
+import { computeSlopeAndIntercept } from "../../utilities/graph-utils"
+import { kMovableLineType } from "./movable-line-types"
+
+export const MovableLineParams = types.model("MovableLineParams", {
```
kswenson, line 9: I'm thinking perhaps this would be better named something like `MovableLineInstance` -- what do you think?

v3/src/components/graph/adornments/movable-line/movable-line-model.ts
```diff
@@ -0,0 +1,69 @@
...
+      const instanceKey = self.instanceKey(xCategories, yCategories, i)
+      if (!self.lines.get(instanceKey) || resetPoints) {
+        self.setInitialLine(xAxis, yAxis, instanceKey)
+      }
+    }
+  }
+}))
+export interface IMovableLineModel extends Instance<typeof MovableLineModel> {}
+export function isMovableLine(adornment: IAdornmentModel): adornment is IMovableLineModel {
+return adornment.type === kMovableLineType
```
kswenson, line 68:
```suggestion
  return adornment.type === kMovableLineType
```

kswenson (approved):
👍 LGTM -- just a couple of suggestions to consider

v3/src/components/graph/adornments/movable-line/movable-line-model.ts
emcelroy: Yeah, that makes sense 👍

PR #850, opened 2023-07-26
feat: tool shelf right buttons

kswenson:
[PT-185700500](https://www.pivotaltracker.com/story/show/185700500)

In preparation for enabling undo/redo, this PR adds the following buttons to the right side of the tool shelf:
- Undo
- Redo
- Tiles
- Options
- Help
- Guide

Icons were extracted from the custom `icomoon` font used by CODAP v2 using an online extractor: https://iconly.io/tools/font-to-icons-converter.

For this initial implementation, but buttons just put up a toast indicating that the button has been clicked.


v3/src/components/tool-shelf/tool-shelf.tsx
```diff
@@ -1,68 +1,121 @@
...
-      <>
-        {Icon && <Icon />}
-        <Tag className='tool-shelf-tool-label'>{t(label)}</Tag>
-      </>
-    </Box>
+      <Spacer/>
+      <Flex className="tool-shelf-right-buttons">
+        {rightButtons.map(entry => {
+          const { className, icon, label, hint } = entry
+          return <ToolShelfButton key={entry.label} className={className} icon={icon} label={label} hint={hint}
```
emcelroy, line 115: This very likely doesn't matter much, but is there a reason you used `entry.label` for the `key` attribute, and `label` for the `label` attribute? If I understand correctly, they're the same thing.

emcelroy (approved):
Looks good 👍

v3/src/components/tool-shelf/tool-shelf.tsx
kswenson: Nope -- will fix

bfinzer (approved):
👍🏻LGTM

PR #852, opened 2023-07-28
Fix for graph transparency option bug

bfinzer:
[#185707586] Bug fix: Transparent background for graph not working

* The user specification that the graph's background be transparent applies only to the interior of the graph; i.e. the plot. The background elsewhere should remain opaque.
* A background color was being set for the entire div containing the graph. This was preventing the transparency from working. Presumably this got instituted after the initial implementation of the transparency option.
* Simply eliminating the background color didn't work because then the axes, legends, and a few other miscellaneous rectangles became transparent as well.
  * The solution is to separately place a background rectangle behind axes and the legend, being careful with the top and bottom axes to make this rectangle as wide as the graph, not just as wide as the axis.


emcelroy (approved):
Looks good 👍

PR #853, opened 2023-07-29
Legend selection of value in grouped attribute

bfinzer:
[#185325793] Feature: Clicking on legend category key should, if appropriate, select case(s) at the level of the legend attribute.

* In DataConfigurationModel:selectCasesForLegendValue we detect when the legend attribute belongs to a group and, if so, select the cases from this group that have the appropriate value. If there is no such group, we, as before, iterate through all the childmost cases.


kswenson (approved):
👍 LGTM

PR #855, opened 2023-08-03
fix: fix and simplify Edit Attribute modal closing [PT-185757174]

pjanik:
https://www.pivotaltracker.com/story/show/185757174

Modal closing didn't seem to be implemented correctly and the attribute tooltip was never restored after the user closed the Edit Attribute modal using "X" button.  I've simplified the code, so the edit attribute props dialog needs to call just `onClose` instead of `onClose` + `onModalOpen(false)`.


kswenson (approved):
👍 Looks good!

PR #856, opened 2023-08-03
feat: add basic formula editor modal [PT-185745389]

pjanik:
https://www.pivotaltracker.com/story/show/185745389

1. The new modal code is based on `EditAttributePropertiesModal` that was already existing.
2. I've moved attribute menu-related code to a new dir, as I guess there'll be more dialogs, utils, etc. related to this context menu. But I don't really feel the codebase yet, so if it doesn't feel right, I'm happy to revert, or use another dir name.
3. I was surprised I had to add `onKeyDown={(e) => e.stopPropagation()}` to Textarea, as otherwise, some unexpected things were happening to the case table. This is also done in `EditAttributePropertiesModal`. Do we understand the issue well? Maybe some more generic approach could be found.
4. I haven't looked at Charka UI docs yet, but generally it looks nice. I was slightly surprised to see all the styles defined via inline attributes (like `ml` to set margin-left, etc.). Is this a pattern you agreed on and thought it'd be the best? Or something I'm randomly copying from `EditAttributePropertiesModal`? :)
5. I was also not sure about translation files. I'm assuming all the translation JSONs are coming from CODAP v2. Should we try to reuse all the strings as much as possible? Or it's fine to add new entries and tweak existing ones? In this case I've added the missing "Edit formula" header and updated the formula prompt to match the style of the `EditAttributePropertiesModal` labels (lowercase). But I don't know if this kind of update is worth it, as it'll obviously require us to re-translate these labels.

### DIFF
--- a/v3/src/components/case-table/attribute-menu/attribute-menu-list.tsx
+++ b/v3/src/components/case-table/attribute-menu/attribute-menu-list.tsx
@@ -1,11 +1,12 @@
 import React, { forwardRef } from "react"
 import { MenuItem, MenuList, useDisclosure, useToast } from "@chakra-ui/react"
 import { CalculatedColumn } from "react-data-grid"
-import { useCaseMetadata } from "../../hooks/use-case-metadata"
-import { useDataSetContext } from "../../hooks/use-data-set-context"
-import { TRow } from "./case-table-types"
-import { EditAttributePropertiesModal } from "./edit-attribute-properties"
-import t from "../../utilities/translation/translate"
+import { useCaseMetadata } from "../../../hooks/use-case-metadata"
+import { useDataSetContext } from "../../../hooks/use-data-set-context"
+import { TRow } from "../case-table-types"
+import { EditAttributePropertiesModal } from "./edit-attribute-properties-modal"
+import t from "../../../utilities/translation/translate"
+import { EditFormulaModal } from "./edit-formula-modal"
 
 interface IProps {
   column: CalculatedColumn<TRow, unknown>
@@ -18,34 +19,50 @@ export const AttributeMenuList = forwardRef<HTMLDivElement, IProps>(
   const toast = useToast()
   const data = useDataSetContext()
   const caseMetadata = useCaseMetadata()
-  const { isOpen, onOpen, onClose } = useDisclosure()
+  const {
+    isOpen: isEditAttributePropsOpen, onOpen: onEditAttributePropsOpen, onClose: onEditAttributePropsClose
+  } = useDisclosure()
+  const {
+    isOpen: isEditFormulaOpen, onOpen: onEditFormulaOpen, onClose: onEditFormulaClose
+  } = useDisclosure()
+  const columnName = column.name as string
 
   const handleMenuItemClick = (menuItem: string) => {
     toast({
       title: 'Menu item clicked',
-      description: `You clicked on ${menuItem} on ${column.name}`,
+      description: `You clicked on ${menuItem} on ${columnName}`,
       status: 'success',
       duration: 5000,
       isClosable: true,
     })
   }
   const handleHideAttribute = () => {
-    const attrId = data?.attrIDFromName(column.name as string)
+    const attrId = data?.attrIDFromName(columnName)
     attrId && caseMetadata?.setIsHidden(attrId, true)
   }
 
   const handleDeleteAttribute = () => {
-    const attrId = data?.attrIDFromName(column.name as string)
+    const attrId = data?.attrIDFromName(columnName)
     attrId && data?.removeAttribute(attrId)
   }
 
-  const handleEditAttributeProps = () => {
-    onOpen()
+  const handleEditAttributePropsOpen = () => {
+    onEditAttributePropsOpen()
     onModalOpen(true)
   }
 
   const handleEditAttributePropsClose = () => {
-    onClose()
+    onEditAttributePropsClose()
+    onModalOpen(false)
+  }
+
+  const handleEditFormulaOpen = () => {
+    onEditFormulaOpen()
+    onModalOpen(true)
+  }
+
+  const handleEditFormulaClose = () => {
+    onEditFormulaClose()
     onModalOpen(false)
   }
 
@@ -62,10 +79,10 @@ export const AttributeMenuList = forwardRef<HTMLDivElement, IProps>(
         <MenuItem onClick={() => handleMenuItemClick("Fit width")}>
           {t("DG.TableController.headerMenuItems.resizeColumn")}
         </MenuItem>
-        <MenuItem onClick={handleEditAttributeProps}>
+        <MenuItem onClick={handleEditAttributePropsOpen}>
           {t("DG.TableController.headerMenuItems.editAttribute")}
         </MenuItem>
-        <MenuItem onClick={() => handleMenuItemClick("Edit Formula")}>
+        <MenuItem onClick={handleEditFormulaOpen}>
           {t("DG.TableController.headerMenuItems.editFormula")}
         </MenuItem>
         <MenuItem onClick={() => handleMenuItemClick("Delete Formula")}>
@@ -90,9 +107,11 @@ export const AttributeMenuList = forwardRef<HTMLDivElement, IProps>(
           {t("DG.TableController.headerMenuItems.deleteAttribute")}
         </MenuItem>
       </MenuList>
-      <EditAttributePropertiesModal columnName={column.name as string} isOpen={isOpen}
+      <EditAttributePropertiesModal columnName={columnName} isOpen={isEditAttributePropsOpen}
         onClose={handleEditAttributePropsClose} />
+      <EditFormulaModal columnName={columnName} isOpen={isEditFormulaOpen} onClose={handleEditFormulaClose} />
     </>
   )
-  })
+})
+
 AttributeMenuList.displayName = "AttributeMenuList"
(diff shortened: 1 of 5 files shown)


kswenson:
> 1. The new modal code is based on `EditAttributePropertiesModal` that was already existing.

Seems like a reasonable place to start.

> 2. I've moved attribute menu-related code to a new dir, as I guess there'll be more dialogs, utils, etc. related to this context menu. But I don't really feel the codebase yet, so if it doesn't feel right, I'm happy to revert, or use another dir name.

The `case-table` directory is overdue for some cleanup, so the new `attribute-menu` subdirectory seems fine.

> 3. I was surprised I had to add `onKeyDown={(e) => e.stopPropagation()}` to Textarea, as otherwise, some unexpected things were happening to the case table. This is also done in `EditAttributePropertiesModal`. Do we understand the issue well? Maybe some more generic approach could be found.

We understand it insofar as Chakra menus, dialogs, etc. don't seem to automatically stop propagation which leads to keyboard events propagating to the tile that had focus when they were opened. It would be nice if there were a more generic approach that didn't require dealing with it on a dialog-by-dialog basis, but I don't think anyone's spent time looking for one.

> 4. I haven't looked at Charka UI docs yet, but generally it looks nice. I was slightly surprised to see all the styles defined via inline attributes (like `ml` to set margin-left, etc.). Is this a pattern you agreed on and thought it'd be the best? Or something I'm randomly copying from `EditAttributePropertiesModal`? :)

Chakra supports those inline [style props](https://chakra-ui.com/docs/styled-system/style-props) as well as some broader [theming](https://chakra-ui.com/docs/styled-system/theme) and of course one can always use straight CSS. Evangeline wrote most of the early Chakra-related code, but I don't think we've used it enough to really settle on optimal usage patterns so feel free to weigh in if you find yourself working in this area.

> 5. I was also not sure about translation files. I'm assuming all the translation JSONs are coming from CODAP v2. Should we try to reuse all the strings as much as possible? Or it's fine to add new entries and tweak existing ones? In this case I've added the missing "Edit formula" header and updated the formula prompt to match the style of the `EditAttributePropertiesModal` labels (lowercase). But I don't know if this kind of update is worth it, as it'll obviously require us to re-translate these labels.

Yes, the translation files are currently being shared with v2, so we should try to reuse existing strings as much as possible. I see what you mean about the inconsistency in prompts between the edit attributes dialog (title bar, `name`, `description`, etc.) and the formula editor dialog (no title bar, `Attribute Name:`, `Formula:`). Note that the v2 formula editor dialog shows different prompts in different contexts, e.g. for a plotted value the prompts are `Plotted Value` (no colon), `value =`, so there's inconsistency there as well. For the purposes of this PR I'd recommend sticking with the v2 strings and perhaps remove the title bar or leave the title blank, but we should definitely raise it as a design issue either as a PT story and/or a topic for discussion in notion.

Rest of review to follow shortly.

v3/src/utilities/translation/lang/en-US.json5
```diff
@@ -706,8 +706,9 @@
     "DG.AttributeFormat.DatePrecisionShort.millisecond": "M/D/YY HH:mm:ss.SSS",
 
     // DG.AttributeFormulaView
-    "DG.AttrFormView.attrNamePrompt": "Attribute Name:",
-    "DG.AttrFormView.formulaPrompt": "Formula:",
+    "DG.AttrFormView.title": "Edit Formula",
+    "DG.AttrFormView.attrNamePrompt": "name",
+    "DG.AttrFormView.formulaPrompt": "formula",
```
kswenson, line 711: I see your point about the inconsistency in the dialogs, but for this PR I think we should stick with the existing strings. I raised the design issue in slack and if there is a decision to change the UI we can address the new/changed strings at that time.

v3/src/components/case-table/attribute-menu/attribute-menu-list.tsx
```diff
@@ -18,34 +19,50 @@ export const AttributeMenuList = forwardRef<HTMLDivElement, IProps>(
   const toast = useToast()
   const data = useDataSetContext()
   const caseMetadata = useCaseMetadata()
-  const { isOpen, onOpen, onClose } = useDisclosure()
+  const {
+    isOpen: isEditAttributePropsOpen, onOpen: onEditAttributePropsOpen, onClose: onEditAttributePropsClose
+  } = useDisclosure()
+  const {
+    isOpen: isEditFormulaOpen, onOpen: onEditFormulaOpen, onClose: onEditFormulaClose
+  } = useDisclosure()
```
kswenson, line 27: It took me a few minutes to figure out why `useDisclosure()` can be used multiple times this way. Perhaps a comment like the following would be useful:
```suggestion
  // each use of useDisclosure() maintains its own state and callbacks so they can be used for independent dialogs
  const {
    isOpen: isEditAttributePropsOpen, onOpen: onEditAttributePropsOpen, onClose: onEditAttributePropsClose
  } = useDisclosure()
  const {
    isOpen: isEditFormulaOpen, onOpen: onEditFormulaOpen, onClose: onEditFormulaClose
  } = useDisclosure()
```

kswenson (changes requested):
👍 Looks good -- several suggested tweaks but none of them very substantial.

v3/src/components/case-table/attribute-menu/attribute-menu-list.tsx
```diff
@@ -1,11 +1,12 @@
...
-import t from "../../utilities/translation/translate"
+import { useCaseMetadata } from "../../../hooks/use-case-metadata"
+import { useDataSetContext } from "../../../hooks/use-data-set-context"
+import { TRow } from "../case-table-types"
+import { EditAttributePropertiesModal } from "./edit-attribute-properties-modal"
+import t from "../../../utilities/translation/translate"
+import { EditFormulaModal } from "./edit-formula-modal"
 
 interface IProps {
   column: CalculatedColumn<TRow, unknown>
```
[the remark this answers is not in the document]
pjanik: Fixed in the last commit.

v3/src/components/case-table/attribute-menu/edit-attribute-properties-modal.tsx
```diff
@@ -1,19 +1,19 @@
...
+import t from "../../../utilities/translation/translate"
 
 interface IProps {
   columnName: string
   isOpen: boolean
   onClose: () => void
 }
 
-export const EditAttributePropertiesModal = ({columnName, isOpen, onClose}: IProps) => {
+export const EditAttributePropertiesModal: React.FC<IProps> = ({columnName, isOpen, onClose}) => {
```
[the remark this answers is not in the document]
pjanik: Interesting, I've never thought about it. My main reason for using React.FC was to ensure that the function returns a correct type (string or JSX) + for "visual clarity" - it's clearly saying it's a component. :) But it's true the return type will be caught anyway. And it's not difficult to guess that we're dealing with components either. I've removed it and I'll avoid it in the future. 👍

v3/src/utilities/translation/lang/en-US.json5
pjanik: Fixed in the last commit.

pjanik:
Thanks for the review. I think I've addressed all the comments.

> Chakra supports those inline [style props](https://chakra-ui.com/docs/styled-system/style-props) as well as some broader [theming](https://chakra-ui.com/docs/styled-system/theme) and of course one can always use straight CSS. Evangeline wrote most of the early Chakra-related code, but I don't think we've used it enough to really settle on optimal usage patterns so feel free to weigh in if you find yourself working in this area.

I see. I'm assuming that CODAPv3 is not planning to use Michael's help to design every aspect of its UI. So, if we don't have this pixel-perfect requirement, I feel that CODAP could be a good use case for the more generalized theme and shared components. I could imagine Michael providing some basic guidelines for typography, spacing, some general layout of forms and dialogs, etc., and we could translate them into themes and reusable components. In a relatively fresh rewrite like CODAP v3, I'd be happy to write code closer to that:

```javascript
      <ModalFooter>
        {
          buttons.map((b, idx) => (
            <ModalButton onClick={b.onClick} tooltip={b.tooltip}>{b.label}</ModalButton>
          ))
        }
      </ModalFooter>
```

rather than the current:

```javascript
      <ModalFooter mt="-5">
        {
          buttons.map((b, idx) => (
            <Tooltip key={idx} label={b.tooltip} h="20px" fontSize="12px" color="white" openDelay={1000}
              placement="bottom" bottom="15px" left="15px" data-testid="modal-tooltip"
            >
              <Button size="xs" variant="ghost" ml="5" onClick={b.onClick} data-testid={`${b.label}-button`}>
                {b.label}
              </Button>
            </Tooltip>
          ))
        }
      </ModalFooter>
```

Also, as a lifelong fan of CSS modules 😉  I could see their benefits here since it's a relatively large codebase for SPA, and class name conflicts are likely to happen unless everything is namespaced very carefully (but that's manual work that can be avoided). Obviously, I don't think these things are the highest priority and it's relatively easy to fix that at any time. But I'd probably try to think about it sooner than later, as otherwise, we'll probably never revisit components that "just work fine". :)

v3/src/components/case-table/attribute-menu/attribute-menu-list.tsx
pjanik: I see. I've committed your comment + replaced destructuring assignment with `const formulaModal` and `const attributePropsModal` as it might be more readable and easier to use than separate variables.

kswenson (approved):
👍 Changes look good

PR #857, opened 2023-08-03
feat: Add count percent graph adornments (PT-181914454)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/181914454

- Adds the Count adornment for the Graph. Checking the Count option in the Measure menu results in a number appearing in the upper right corner of each subplot in the Graph. The number is equal to the number of cases within that subplot.
- Improves/extends the construction of the grid used for placing adornments over the Graph's subplots.


v3/src/components/graph/models/data-configuration-model.ts
```diff
@@ -144,6 +144,14 @@ export const DataConfigurationModel = types
     placeShouldShowClickHereCue(place: GraphPlace, tileHasFocus: boolean) {
       return this.placeAlwaysShowsClickHereCue(place) ||
         (this.placeCanShowClickHereCue(place) && tileHasFocus)
+    },
+    isCaseInSubPlot(subPlotKey: Record<string, string>, caseData: Record<string, any>) {
+      const numOfKeys = Object.keys(subPlotKey).length
+      let matchedValCount = 0
+      Object.keys(subPlotKey).forEach(key => {
+        if (!subPlotKey[key] || subPlotKey[key] === caseData[key]) matchedValCount++
```
kswenson, line 152:
```suggestion
        if (subPlotKey[key] === caseData[key]) matchedValCount++
```
Do we still need the `!subPlotKey[key]` part of the test? I wouldn't think that it would occur at this point.

kswenson (changes requested):
👍 Looks good! Just a few relatively minor suggestions.

v3/src/components/graph/models/data-configuration-model.ts
emcelroy: @kswenson I think we do need this check, at least with the way things are right now. If I understand correctly, `!subPlotKey[key]` would be true when `!subPlotKey[key]` is equal to the empty string. In the latest commit I've changed it to `subPlotKey[key] === ""` to make it a bit more clear.

A key's value will be equal to the empty string when the associated attribute is numeric. In [adornments.tsx](https://github.com/concord-consortium/codap/blob/2e30125416e623b948c0645ccc64708c52d55ddb/v3/src/components/graph/adornments/adornments.tsx#L44-L54), when we set up the arrays used to iterate over when building the adornment grid, if an attribute is not categorical, we set the associated array to `[""]`. That way there's a single item to iterate over. I'm wondering now if we should revisit that idea since it's not self-evident a single empty string means the data isn't categorical. I think it also may be odd to do that mainly just to get the associated `for` loop to run once. Would it be worth trying to modify that approach so the arrays would be empty or undefined instead? Or is it reasonably straightforward as is?

v3/src/components/graph/adornments/adornments.tsx
```diff
@@ -9,75 +9,136 @@ import { getAdornmentContentInfo } from "./adornment-content-info"
...
+  // have the same attribute on two axes or splits, we need to make sure the subPlotKey is unique.
+  // So if an attribute is on more than one axis or split, we set the value of that attribute's ID
+  // to "__IMPOSSIBLE__" instead of overwriting the key's value because it's impossible for a
+  // single case to have two different values for the same attribute.
+  const updateSubPlotKey = (subPlotKey: Record<string, string>, attrId: string, cat: string) => {
+    const newSubPlotKey = { ...subPlotKey }
+    const propertyAlreadyPresent = Object.keys(newSubPlotKey).includes(attrId)
+    newSubPlotKey[attrId] = propertyAlreadyPresent && newSubPlotKey[attrId] !== cat
+      ? "__IMPOSSIBLE__"
+      : cat
```
kswenson, line 37:
```suggestion
    if (cat) {
      const propertyAlreadyPresent = Object.keys(newSubPlotKey).includes(attrId)
      newSubPlotKey[attrId] = propertyAlreadyPresent && newSubPlotKey[attrId] !== cat
        ? "__IMPOSSIBLE__"
        : cat
    }
```
Vis-a-vis the discussion about handling empty strings for numeric attributes, I think if we just don't include empty strings in the `subPlotKey`, then we wouldn't need to deal with them down the road. Empty strings as categories should only occur when there are no other categories.

v3/src/components/graph/models/data-configuration-model.ts
```diff
@@ -144,6 +144,14 @@ export const DataConfigurationModel = types
     placeShouldShowClickHereCue(place: GraphPlace, tileHasFocus: boolean) {
       return this.placeAlwaysShowsClickHereCue(place) ||
         (this.placeCanShowClickHereCue(place) && tileHasFocus)
+    },
+    isCaseInSubPlot(subPlotKey: Record<string, string>, caseData: Record<string, any>) {
+      const numOfKeys = Object.keys(subPlotKey).length
+      let matchedValCount = 0
+      Object.keys(subPlotKey).forEach(key => {
+        if (subPlotKey[key] === "" || subPlotKey[key] === caseData[key]) matchedValCount++
```
kswenson, line 152:
```suggestion
        if (subPlotKey[key] === caseData[key]) matchedValCount++
```
I think if we make the change I suggested above about not including empty strings in subplot keys _then_ we wouldn't need this test here.

kswenson (changes requested):
👍 Changes all look good -- I have one more suggestion about how to deal with the problem of empty strings in subplot keys. Give it a try and let me know how it goes.

v3/src/components/graph/adornments/adornments.tsx
emcelroy: @kswenson Yeah, that works -- thanks! I also had to make a small change to `setSubPlotKey` in adornment-models.ts in [the latest commit](https://github.com/concord-consortium/codap/pull/857/commits/d682bf2879c6ea37f0858d7dae0a721506aea5a6). I think the checks on 76 and 77 weren't quite right before. When I made the suggested changes in adornments.tsx and data-configuration-model.ts, the Point and Line wouldn't render correctly until I made that change to `setSubPlotKey`.

kswenson (approved):
👍 Changes look good

PR #858, opened 2023-08-05
Correctly plot parent level cases

bfinzer:
[#185447912] Feature: Number of points in graph is at most the number of cases in the child-most collection for the given attributes

* Added collectionID as property of FilteredCases so that, if present, the raw cases could be taken from a collection rather than the dataset as a whole.
* In DataConfigurationModel improved the logic for deciding on the 'caption' attribute to take into account the dataset hierarchy. Buried `defaultCaptionAttribute` inside `attributeID` method to avoid circularity. 
* Added idOfChildmostCollectionForAttributes to data-set-utils.ts
* Modified data-configuration-model.test.ts to accommodate loss of `defaultCaptionAttribute`
* Include 'invalidateCollectionGroups' in actions to be responded to in use-plot.ts
* Fixed latent bug whereby case plot was getting its cases from the dataset instead of the data configuration for initialization of random positions
* Fixed latent bug whereby if no attributes were assigned, the caption attribute would be assigned to the dataset's 0th attribute instead of its 0th ungrouped attribute
* Made `Dataset:setCaseValues` and `Dataset:setCaseValue` smarter about when to call invalidateCollectionGroups


v3/src/components/data-display/models/data-configuration-model.ts
```diff
@@ -112,9 +108,30 @@ export const DataConfigurationModel = types
           : this.attributeDescriptions[role]
     },
     attributeID(role: GraphAttrRole) {
+
+      const defaultCaptionAttributeID = () => {
+        // We find the childmost collection and return the first attribute in that collection. If there is no
+        // childmost collection, we return the first attribute in the dataset.
+        const attrIDs = ['x', 'y', 'rightNumeric', 'topSplit', 'rightSplit']
```
kswenson, line 115: Not `legend` or `caption`? If so, you could use `TipAttrRoles`.

kswenson (approved):
👍 Looks good -- I added a commit which tweaked a few things and then rebased to `main`. I left two questions to consider.

v3/src/components/data-display/models/data-configuration-model.ts
bfinzer: Nope. Caption is not a candidate. But 'legend', so I'm adding that.

v3/src/components/graph/components/casedots.tsx
```diff
@@ -108,24 +105,24 @@ export const CaseDots = function CaseDots(props: {
...
       })
     }
 
     initCases(cases)
-    const disposer = dataset && onAnyAction(dataset, action => {
-      if (isAddCasesAction(action)) {
-        initCases(action.args[0])
+    const disposer = dataConfiguration?.onAction(action => {
+      if (['addCases', 'removeCases'].includes(action.name)) {
+        initCases(dataConfiguration?.caseDataArray)
```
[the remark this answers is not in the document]
bfinzer: Right you are. I improved initCases to leave existing point coordinates intact.

PR #859, opened 2023-08-07
feat: Add percent graph adornment (PT-185758373)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/185758373

Adds support for the Percent adornment which indicates the percent of cases contained within a portion of the graph. The Percent adornment is very similar to the Count adornment, and is essentially an add-on to the Count adornment in the code, although it can appear independently from the Count depending on the options the user selects from the Measure menu.


v3/src/components/graph/adornments/count/count-model.ts
```diff
@@ -1,12 +1,26 @@
...
   .props({
-    type: 'Count'
+    type: 'Count',
+    showCount: true,
+    showPercent: false,
+    percentType: types.optional(types.enumeration(["cell", "column", "row"]), "row"),
   })
+  .actions(self => ({
+    setShowCount(showCount: boolean) {
+      self.showCount = showCount
```
kswenson, line 15:
```suggestion
      self.showCount = showCount
      self.isVisible = self.showCount || self.showPercent
```

v3/src/components/graph/adornments/count/count-model.ts
```diff
@@ -1,12 +1,26 @@
...
+    },
+    setShowPercent(showPercent: boolean) {
+      self.showPercent = showPercent
```
kswenson, line 18:
```suggestion
      self.showPercent = showPercent
      self.isVisible = self.showCount || self.showPercent
```

kswenson (commented):
👍 This is looking good -- to move more of the logic out of the `GraphMeasurePanel` and into the adornments themselves I think we could change the `IAdornmentComponentInfo` to something like (note that this is all untested pseudo-code):
```typescript
export interface IAdornmentComponentInfo {
  adornmentEltClass: string
  Component: React.ComponentType<IAdornmentComponentProps>
  Controls: React.ComponentType<IAdornmentControlsProps>
  order: number
  type: string
}
```
Then the adornment checkbox code can be made into a reusable component that looks something like:
```typescript
export function AdornmentCheckbox(adornmentModel: IAdornmentModel, labelKey: string) {
  const titleSlug = adornmentModel.type.replace(/\s/g, "-").toLowerCase()
  return (
    <FormControl>
      <Checkbox
        data-testid={`adornment-checkbox-${titleSlug}`}
        defaultChecked={!adornmentModel.isHidden}
        onChange={e => adornmentModel.setVisible(e.target.checked)}
      >
        {t(labelKey)}
      </Checkbox>
    </FormControl>
  )
}
```
so that the `Controls` property for simple adornments would look something like:
```typescript
{
  // ...
  Controls: (adornmentModel: IAdornmentModel) => <AdornmentCheckbox adornmentModel={adornmentModel} labelKey={labelKey} />
  // ...
}
```
The count/percent adornment could then have its own unique Controls component that looks something like this:
```typescript
export function CountAdornmentControls(adornmentModel: IAdornmentModel) {
  const dataConfiguration = useDataConfigurationContext()
  const countAdornment = isCountAdornment(adornmentModel) ? adornmentModel : undefined
  const shouldShowPercentOptions = true // ... some function of the data configuration
  const titleSlug = adornmentModel.type.replace(/\s/g, "-").toLowerCase()
  return (
    <>
      <FormControl>
        <Checkbox
          data-testid={`adornment-checkbox-${titleSlug}-count`}
          defaultChecked={countAdornment.showCount}
          onChange={e => adornmentModel.setShowCount(e.target.checked)}
        >
          {t(countLabelKey)}
        </Checkbox>
      </FormControl>
      <FormControl>
        <Checkbox
          data-testid={`adornment-checkbox-${titleSlug}-percent`}
          defaultChecked={countAdornment.showPercent}
          onChange={e => adornmentModel.setShowPercent(e.target.checked)}
        >
          {t(percentLabelKey)}
        </Checkbox>
        {shouldShowPercentOptions &&
          // should the radio group be in its own `<FormControl>`?
          <div className="sub-options percent-type" data-test-id="adornment-percent-type-options">
            <RadioGroup
              name="percent-type"
              size="md"
              value={percentType}
              onChange={e => adornmentModel.setPercentType(e.target.value)}
            >
              <Radio value="row" size="md">
                {t("DG.Inspector.graphRow")}
              </Radio>
              <Radio value="column" size="md">
                {t("DG.Inspector.graphColumn")}
              </Radio>
              <Radio value="cell" size="md">
                {t("DG.Inspector.graphCell")}
              </Radio>
            </RadioGroup>
          </div>
        }
      </FormControl>
    </>
  )}
```
Then the `GraphMeasurePanel` just needs to render the `Controls` component for each adornment in the appropriate `order` but the adornments themselves handle all of the interactions between the `Controls` and the model. See if this makes sense and we can touch base on it tomorrow.

v3/src/components/data-display/models/data-configuration-model.ts
```diff
@@ -354,8 +334,120 @@ export const DataConfigurationModel = types
...
+      let matchedValCount = 0
+      Object.keys(subPlotKey).forEach(key => {
+        if (subPlotKey[key] === caseData[key]) matchedValCount++
+      })
+      return matchedValCount === numOfKeys
+    }
+  }))
+  .views(self => ({
+    subPlotCases(subPlotKey: Record<string, string>) {
+      const casesInPlot = [] as ICase[]
```
bfinzer, line 375: In the first `forEach` below, when there are multiple numeric y-attributes, cases will (usually) be counted more than once. I've written a PT story (#185854897) about this, a fix of which should probably be in a separate PR. This comment applies to each of the following three methods as well.

v3/src/components/graph/utilities/adornment-utils.ts
```diff
@@ -0,0 +1,11 @@
+export function shouldShowPercentOption(attrTypes: Record<string, string | undefined>) {
+  const categoricalAttrCount = Object.values(attrTypes).filter(a => a === "categorical").length
+  return categoricalAttrCount > 0
+}
+
```
bfinzer, line 5: Doesn't the function below replicate logic in the DataConfigurationModel?

bfinzer (approved):
👍🏻Looks very good. Just a few comments, mostly about naming.

v3/src/components/graph/adornments/count/count-model.ts
```diff
@@ -1,12 +1,43 @@
...
   .named('CountModel')
   .props({
-    type: 'Count'
+    type: 'Count',
+    showCount: false,
+    showPercent: false,
+    percentType: types.optional(types.enumeration(["cell", "column", "row"]), "row"),
   })
+  .views(self => ({
+    percentValue(dataConfig: any, casesInPlot: number, cellKey: Record<string, string>) {
```
kswenson, line 14: Wouldn't this be:
```suggestion
    percentValue(dataConfig: IDataConfigurationModel, casesInPlot: number, cellKey: Record<string, string>) {
```

v3/src/components/graph/adornments/count/count-model.ts
```diff
@@ -1,12 +1,43 @@
...
+      const casesInRow = dataConfig?.rowCases(cellKey)?.length ?? 0
+      const casesInColumn = dataConfig?.columnCases(cellKey)?.length ?? 0
+      const casesInCell = dataConfig?.cellCases(cellKey)?.length ?? 0
+      const divisor = self.percentType === "row"
+        ? casesInRow
+        : self.percentType === "column"
+          ? casesInColumn
+          : casesInCell
```
kswenson, line 22: This is going to compute all three even though we only need one. How about:
```suggestion
      const divisor = self.percentType === "row"
        ? dataConfig?.rowCases(cellKey)?.length ?? 0
        : self.percentType === "column"
          ? dataConfig?.columnCases(cellKey)?.length ?? 0
          : dataConfig?.cellCases(cellKey)?.length ?? 0
```

v3/src/components/graph/adornments/count/count-model.ts
```diff
@@ -1,12 +1,43 @@
...
+      const percentValue = ((casesInPlot / divisor) * 100).toFixed(2)
+      return isFinite(Number(percentValue)) ? Math.round(Number(percentValue) * 100)/100 : 0
```
kswenson, line 24: Some unnecessary conversions going on here. How about something like:
```suggestion
      const percentValue = casesInPlot / divisor
      return isFinite(percentValue) ? percentValue : 0
```
and then leave it to the view to do any necessary rounding for display purposes. The model should return the actual percent value.

v3/src/components/graph/adornments/count/count-registration.tsx
```diff
@@ -0,0 +1,116 @@
...
+  const [percentTypeValue, setPercentTypeValue] = useState(
+    existingAdornment && isCount(existingAdornment) ? existingAdornment.percentType : "row"
+  )
+
+  const handleSetting = (checked: boolean, checkBoxType: string) => {
+    const existingCountAdornment = graphModel.adornments.find(a => a.type === kCountType) as ICountModel
+    const componentContentInfo = getAdornmentContentInfo(kCountType)
+    const adornment = existingCountAdornment
+      ? existingCountAdornment
+      : componentContentInfo.modelClass.create() as ICountModel
```
kswenson, line 28:
```suggestion
    const adornment = existingCountAdornment ?? componentContentInfo.modelClass.create() as ICountModel
```

v3/src/components/graph/adornments/count/count.tsx
```diff
@@ -7,17 +7,34 @@ import "./count.scss"
...
-export const Count = observer(function Count({model, subPlotKey}: IProps) {
+export const Count = observer(function Count({model, cellKey}: IProps) {
+  const classFromKey = model.classNameFromKey(cellKey)
   const dataConfig = useDataConfigurationContext()
-  const casesInPlot = dataConfig?.subPlotCases(subPlotKey)?.length ?? 0
-  const classFromKey = model.classNameFromKey(subPlotKey)
+  const casesInPlot = dataConfig?.subPlotCases(cellKey)?.length ?? 0
+  const percent = model.percentValue(dataConfig, casesInPlot, cellKey)
+  const displayPercent = model.showCount ? ` (${percent}%)` : `${percent}%`
+  const shouldShowPercentOption = dataConfig?.categoricalAttrCount ? dataConfig?.categoricalAttrCount() > 0 : false
```
kswenson, line 19:
```suggestion
  const shouldShowPercentOption = dataConfig ? dataConfig.categoricalAttrCount() > 0 : false
```

kswenson (changes requested):
👍 Looks good -- a handful of suggested changes but they're all pretty minor.

v3/src/components/graph/adornments/count/count-model.ts
emcelroy: @kswenson Understood about leaving it to the view to do the necessary rounding. That makes a lot of sense.

To get the correct display value, though, I think at least some of the conversion was actually necessary. The way I was doing it definitely wasn't optimal or clear, though. I made the changes you suggested here, but also added a `percentString` utility function that handles the formatting in a clearer way. It's used by the Count component to set the value that gets displayed. Does that look good? Or would it be better to make `percentString` a view on the count model?

kswenson (approved):
👍 LGTM -- just one potential undo/redo issue that you'll need to address at some point but it need not be now.

PR #860, opened 2023-08-08
Fix problems introduced in merge of PRs

bfinzer:
* The merges to main by Ethan and Bill conflicted a bit introducing problems in adornments.tsx and graph-content-model.ts
* Also fixed a lint error in data-display-layer-union.tx


v3/src/components/graph/models/graph-content-model.ts
```diff
@@ -93,25 +93,25 @@ export const GraphContentModel = DataDisplayContentModel
       return self.plotType === 'scatterPlot' && ['left', 'bottom'].includes(place)
     },
     getUpdateCategoriesOptions(resetPoints=false): IUpdateCategoriesOptions {
-      const xAttrId = self.getAttributeID("x"),
-        xAttrType = self.config.attributeType("x"),
+      const xAttrId = this.getAttributeID("x"),
```
kswenson, line 96: Rather than using `this` here it would be preferable to put `getUpdateCategoriesOptions()` into its own `views()` block because `self` gives better typing than `this`. If you don't want to deal with it here I can deal with it on rebase of the next PR.

kswenson (approved):
👍 Looks good -- just one suggested tweak.

v3/src/components/graph/models/graph-content-model.ts
bfinzer: Good idea

PR #864, opened 2023-08-11
Refactor formula-related code, add initial handling of lookup functions, add aggregate function with filter expression

pjanik:
https://www.pivotaltracker.com/story/show/185745574

~~This code is in a very draft state and most of it should be cleaned up~~, but it can give you an impression of how some things have been approached:
- custom MathJs functions
- lookup functions evaluation
- lookup dependency observing
- custom aggregate function (`mean`) with filtering implemented

~~I'm definitely parsing the same expressions and dealing with math nodes too often, so probably I'll try to cache parsed formulas and/or reuse logic related to that. Also, there's no caching, linear search for lookupByKey, etc.~~

~~It might look pretty scary that we had to add so much code related to observing just to handle these two new lookup functions.~~ But I'm hoping these functions are very specific cases and we won't have many of these (~~probably something for semi-aggregates or for filters and that's it?). And the way we observe updates could be definitely cleaned up too.~~

Demo:
https://codap3.concord.org/branch/lookup-func/index.html

UPDATE: 
I think that this PR is ready for review now. I've posted updates in comments below, so some of the description above is out of date.


kswenson (commented):
I took a quick look at this, but a more detailed review will have to wait until I have more time.

v3/src/models/data/formula-utils.ts
```diff
@@ -29,31 +48,48 @@ export const canonicalizeExpression = (displayExpression: string, displayNameMap
...
-          newNode.name += AGGREGATE_SYMBOL_SUFFIX
+          symbolNode.name += AGGREGATE_SYMBOL_SUFFIX
+        }
+      }
+    }
+    if (node.type === "FunctionNode" && isLookupFunction((node as FunctionNode).fn.name)) {
+      const functionNode = node as FunctionNode
+      if (functionNode.args[0].type === "ConstantNode") {
+        const dataSetName = getValueOrName(functionNode.args[0])?.toString() || ""
+        ;(functionNode.args[0] as ConstantNode<string>).value = displayNameMap.dataSet[dataSetName]?.id
```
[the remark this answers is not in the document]
pjanik: Done, thanks.

pjanik:
Aug 15th update:
Lots of the code has been refactored, and I hope that `formula-fn-registry.ts` starts to look reasonable.
There's an example of the custom operator being overwritten (`equal`/`==` so we can compare strings) and the first actual implementation of the aggregate function that supports filters - `mean`. Caching is still not there.

pjanik:
Aug 16th update:
I've fixed a bug in lookupByIndex implementation, simplified lookup dependencies handling, and completely rewrote MathJS custom scope, so it's now cleaner. It should be faster (plus it's no longer necessary to recreate it constantly and use weird typings).

At this point, I think this PR is ready to review.

v3/src/models/data/formula-fn-registry.ts
```diff
@@ -0,0 +1,157 @@
...
+        const val = dataSet.getValue(c.__id__, keyAttrId)
+        if (val === keyAttrValue) {
+          return dataSet.getValue(c.__id__, attrId) || ""
+        }
+      }
+      return ""
+    },
+  },
+
+  // mean(expression, filterExpression)
```
bfinzer, line 126: Note that aggregate functions work on numeric values (like `mean`) ignore not just empty values but non-numeric values as well. Also, if an attribute is typed as numeric, values are coerced to numeric where possible; e.g. the value '1a' is coerced to 1 and included in the calculation. You can see this in V2.

bfinzer (approved):
👍🏻LGTM
Very few comments. I didn't attempt to dive very deeply into this.

v3/src/models/data/formula-fn-registry.ts
```diff
@@ -0,0 +1,157 @@
...
+    isAggregate: false,
+    validateArguments: (args: MathNode[]): [ConstantNode<string>, ConstantNode<string>, MathNode] => {
+      if (args.length !== 3) {
+        throw new Error(`lookupByIndex function expects exactly 3 arguments, but it received ${args.length}`)
+      }
+      if (!isConstantStringNode(args[0]) || !isConstantStringNode(args[1])) {
+        throw new Error("lookupByIndex function expects first two arguments to be strings " +
+          "and the third one to be a number")
+      }
+      return [args[0], args[1], args[2]]
```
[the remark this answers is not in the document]
pjanik: > Should there be a test to check that args[2] is a number?

Initially, I thought so (hence the error message above that I forgot to update), but later I realized it's not true.
The third argument can be any expression that finally resolves to a number. But we cannot check its validity until the expression is actually calculated. Eg. user can say `lookupByIndex("Mammals", "Height", 2 + 2)` or even `lookupByIndex("Mammals", "Height", round(LocalAttribute))`. This also works in V2. 

> And is the constructed array different than args?

Yes, there's subtle difference related to types. Note that `args`  is `MathNode[]`. And the main point of `validateArguments` function is to narrow down types to the expected ones, so we don't have to do it over and over again. I've defined that in line 39 as a fn return type:
`[ConstantNode<string>, ConstantNode<string>, MathNode]` -> it means that the first two arguments need to be string constants, and the last one is just an expression.

If we tried to simply `return args` from this function, TS would complain:
`Type 'MathNode[]' is not assignable to type '[ConstantNode<string>, ConstantNode<string>, MathNode]'.`

It all works fine as in line 43, I use `isConstantStringNode()` function that narrows down the type to `ConstantNode<string>`.

v3/src/models/data/formula-fn-registry.ts
```diff
@@ -0,0 +1,157 @@
...
+      const attrId = evaluateNode(args[1], scope)
+      const keyAttrId = evaluateNode(args[2], scope)
+      const keyAttrValue = evaluateNode(args[3], scope)
+
+      const dataSet: IDataSet | undefined = scope.getDataSet(dataSetId)
+      if (!dataSet) {
+        return ""
+      }
+      // TODO: is that the right way to access all the dataset cases?
+      // TODO: Optimize? Sort and use binary search?
```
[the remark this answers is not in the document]
pjanik: In later PRs I've added a very simple (but pretty flexible) approach to caching. 

I guess I could sort the attribute values when this is executed the first time, cache sorted values, and then restore sorted values from the cache and use binary search for each lookup. It'll reduce O(n^2) to O(nlogn).

v3/src/models/data/formula-fn-registry.ts
pjanik: Thanks, good to know. Yes, all my PRs totally miss handling of user errors, edge cases, and tests. I didn't try to get into that yet.

v3/src/models/data/formula-fn-registry.ts
```diff
@@ -0,0 +1,157 @@
...
+  // lookupByKey("dataSetName", "attributeName", "keyAttributeName", "keyAttributeValue" | localKeyAttribute)
+  lookupByKey: {
+    rawArgs: true,
+    isAggregate: false,
+    validateArguments: (args: MathNode[]):
+      [ConstantNode<string>, ConstantNode<string>, ConstantNode<string>, MathNode] => {
+      if (args.length !== 4) {
+        throw new Error("lookupByKey function expects exactly 4 arguments")
+      }
+      if (!isConstantStringNode(args[0]) || !isConstantStringNode(args[1]) || !isConstantStringNode(args[2])) {
```
kswenson, line 81: Could also use:
```suggestion
      if (!args.slice(0, 3).every(arg => isConstantStringNode(arg)) {
```

v3/src/models/data/formula-fn-registry.ts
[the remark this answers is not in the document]
kswenson: Caching is a good idea as long as we invalidate the cache when values in the referenced data set change.

v3/src/models/data/formula-mathjs-scope.ts
```diff
@@ -0,0 +1,90 @@
...
+    throw new Error("It's not allowed to delete local values in the formula scope.")
+  }
+
+  clear () {
+    throw new Error("It's not allowed to clear values in the formula scope.")
+  }
+
+  // In regular MathJS use case, a subscope is used to create a new scope for a function call. It needs to ensure
+  // that variables from subscope cannot overwrite variables from the parent scope. However, since we don't allow
+  // any variables to be set in the formula scope, we can just return the same scope for simplicyt and
```
kswenson, line 76:
```suggestion

  // In regular MathJS use case, a subscope is used to create a new scope for a function call. It needs to ensure
  // that variables from subscope cannot overwrite variables from the parent scope. However, since we don't allow
  // any variables to be set in the formula scope, we can just return the same scope for simplicity and
```

v3/src/models/data/formula-types.ts
```diff
@@ -15,4 +27,23 @@ export interface IGlobalValueDependency {
...
+
+export interface IFormulaMathjsFunction {
+  rawArgs: boolean
+  isAggregate: boolean
+  evaluate: (args: MathNode[], mathjs: any, scope: FormulaMathJsScope) => EvalValue | EvalValue[]
+  canonicalize?: (args: MathNode[], displayNameMap: DisplayNameMap) => void
+  getDependency?: (args: MathNode[]) => IFormulaDependency
+}
+
+export type ICODAPMathjsFunctionRegistry = Record<string, IFormulaMathjsFunction>
```
kswenson, line 49:
```suggestion
export type CODAPMathjsFunctionRegistry = Record<string, IFormulaMathjsFunction>
```
For simple types (rather than interfaces) I'd leave off the `I`, but that's a personal preference.

kswenson (approved):
👍 Looks good -- just a smattering of minor suggestions

v3/src/models/data/formula-fn-registry.ts
pjanik: It seems that it won't work in this particular case. These simple `if` guards allow TS to assume that in the following lines args[0], args[1], and args[1] are instances of `ConstantNode<string>`. Your expression is nicer and more concise, but TS is not smart enough to parse it and infer correct types from it. It'll generate an error in the next line:
` Type 'MathNode' is not assignable to type 'ConstantNode<string>'.`

v3/src/models/data/formula-types.ts
pjanik: I also usually follow this naming. Fixed in the last comment.

PR #866, opened 2023-08-17
feat: add random() and randomPick() functions, implement basic rerandomization [PT-185745433]

pjanik:
https://www.pivotaltracker.com/story/show/185745433

This PR adds two random functions: `random` and `randomPick`. Also, it implements "rerandomize" feature and makes this button disabled when there's no random function in the formula. 

The implementation is very basic, so I likely missed some quirks and challenges related to the random functions, but I still thought I could just push it and open PR.

Note that there are some differences in behavior compared to V2, but I could almost call them improvements unless there are some specific reasons for the previous behavior. An example:

CODAP V2: Open Mammals and add a new formula like `random() * Height`. Simple, non-aggregate function. Now, when I edit just one "Height" cell, all the cases get recomputed and since we use random(), they all get a new value. However, my implementation in CODAP V3 updates just the case that was edited by the user. Is that fine? From my naive perspective, it makes even more sense. 

I can find other similar scenarios - e.g., when I add a new case in V2, all the values get recomputed. V3 recomputes only the new ones. Would we consider such behavior okay and/or even an improvement? Or is there some reason to recalculate everything?

Note that differences are caused by the property observing. I guess V3 is more specific about what and when to recompute a case. Although this will depend on type of the dependency - e.g., lookup dependency probably works similarly to V2.

### DIFF
--- a/v3/src/models/data/formula-fn-registry.ts
+++ b/v3/src/models/data/formula-fn-registry.ts
@@ -1,4 +1,4 @@
-import { create, all, mean, MathNode, ConstantNode } from 'mathjs'
+import { create, all, mean, random, pickRandom, MathNode, ConstantNode } from 'mathjs'
 import { FormulaMathJsScope } from './formula-mathjs-scope'
 import {
   DisplayNameMap, FValue, ICODAPMathjsFunctionRegistry, ILookupDependency, isConstantStringNode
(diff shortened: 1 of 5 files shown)


bfinzer (approved):
👍🏻LGTM

v3/src/models/data/formula-fn-registry.ts
kswenson, line 1: As this registry continues to grow as more functions are added, it would be good to have a registration system so that sets of functions can be implemented in separate files and added to the registry dynamically.

v3/src/models/data/formula.ts
```diff
@@ -25,6 +28,12 @@ export const Formula = types.model("Formula", {
   },
   setCanonical(canonical: string) {
     self.canonical = canonical
+  },
+  rerandomize() {
+    if (!self.formulaManager) {
+      return
+    }
+    self.formulaManager.recalculateFormula(self.id)
```
kswenson, line 36:
```suggestion
    self.formulaManager?.recalculateFormula(self.id)
```

kswenson (approved):
👍 Looks good

v3/src/models/data/formula-fn-registry.ts
pjanik: Agree. It won't be a problem to export the register function or move implementations to separate files and import them statically here. Let's see how it goes when we keep adding more functions.

PR #867, opened 2023-08-18
Initial approach/spike to handling collections in formulas

pjanik:
https://www.pivotaltracker.com/story/show/185745420

This PR adds an initial implementation that handles resolving the aggregate symbols when the tables are hierarchical.
I doubt it's complete, and many edge cases are probably missing, but I still thought it might be worth sharing before I leave for vacation.

There are two main issues so far:
- I had to disable previous caching, as obviously, we can no longer cache aggregate functions as simply as before. I have some ideas on how to deal with it, but first, I'd like to wrap up and validate the whole approach.
- There's a significant difference in how I handle aggregate functions in the child collections. [Please see my message on Slack](https://concord-consortium.slack.com/archives/C035J6RDAK0/p1692357421148309), where I included two videos clearly showing the difference. It mainly happened because I didn't realize V2 was working that way. I can adjust this behavior if necessary (I bet it is, as it's a significant difference).
- I'll need to handle a scenario when an attribute with a formula is dragged to another collection.
- I didn't even try to think what to do when the formula references an attribute from the parent collection. But I've noticed that V2 gets broken or it just doesn't render anything.
- I've also noticed a few bugs in CODAP v3 related to hierarchal parents. Initial dragging of attributes usually works, but often after a few moves back and forth, things get a bit broken. I bet you're aware of that. But I'm mentioning that as it might affect testing/demoing.

### DIFF
--- a/v3/src/models/data/formula-mathjs-scope.ts
+++ b/v3/src/models/data/formula-mathjs-scope.ts
@@ -43,7 +43,7 @@ export class FormulaMathJsScope {
           // - the table is flat and aggregate function is referencing another attribute
           // - the table is hierarchical and aggregate function is referencing an attribute from the same collection
           // In both cases it's enough to compare collection ids. When the table is flat, they might be equal to
-          // undefined but equality check it's gonna work anyway.
+          // undefined but equality check works anyway.
           const attrCollectionId = context.localDataSet.getCollectionForAttribute(attr.id)?.id
           const useSameLevelCases = context.formulaAttributeCollectionId === attrCollectionId
           // Note that mapping childCaseIds might look like potential performance issue / O(n^2), but it's not.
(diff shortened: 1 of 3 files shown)


v3/src/models/data/formula-mathjs-scope.ts
```diff
@@ -31,9 +34,27 @@ export class FormulaMathJsScope {
...
+        context.localDataSet.getValue(caseId, attr.id)
+      )
       Object.defineProperty(this.dataStorage, `${LOCAL_ATTR}${attr.id}${AGGREGATE_SYMBOL_SUFFIX}`, {
         get: () => {
-          return attr.strValues
+          // Note that we have to return all the same level cases in two cases:
+          // - the table is flat and aggregate function is referencing another attribute
+          // - the table is hierarchical and aggregate function is referencing an attribute from the same collection
+          // In both cases it's enough to compare collection ids. When the table is flat, they might be equal to
+          // undefined but equality check it's gonna work anyway.
```
kswenson, line 46:
```suggestion
          // undefined but equality check works anyway.
```

kswenson (approved):
👍 Seems like a decent start

pjanik:
Thanks. I'm going to merge this PR as it's been reviewed. I'm aware that the behavior is currently different than in V2 and it's not desired. I'll open new PRs to change that and to re-enable caching.

PR #868, opened 2023-08-30
feat: implement handling of same-level grouping in aggregate formulas [PT-185745420]

pjanik:
https://www.pivotaltracker.com/story/show/185745420

This PR makes CODAP V3 handle same-level grouping in the same way as V2 does. I've described the previous differences in detail here:
https://concord-consortium.slack.com/archives/C035J6RDAK0/p1692357421148309

### DIFF
--- a/v3/src/models/data/formula-manager.ts
+++ b/v3/src/models/data/formula-manager.ts
@@ -5,7 +5,7 @@ import { onAnyAction } from "../../utilities/mst-utils"
 import { getFormulaDependencies } from "./formula-utils"
 import {
   DisplayNameMap, IFormulaDependency, GLOBAL_VALUE, LOCAL_ATTR, ILocalAttributeDependency, IGlobalValueDependency,
-  ILookupDependency
+  ILookupDependency, NO_PARENT_KEY
 } from "./formula-types"
 import { math } from "./formula-fn-registry"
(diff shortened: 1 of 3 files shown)


pjanik:
@kswenson, I believe this PR is now ready for review. New version should handle complex hierarchical tables correctly and calculate aggregate functions on all hierarchy levels.

I hope it's more or less clear, but this kind of code might always require some deeper understanding of the desired behavior of the aggregate formulas and data set hierarchies. The same-level grouping looks like an edge case (as it is IMO), and it brings the most confusion. If you have suggestions for new comments or different naming, I'd happily apply them.

kswenson (approved):
👍 Looks good!

pjanik:
@kswenson, I can see you've left an inline comment referring to outdated/nonexistent piece of code. I'm wondering if you've reviewed the very last version or the one before (maybe you started the review, and then GH kept the previous revision). The last commit completely changes the implementation.

kswenson:
> @kswenson, I can see you've left an inline comment referring to outdated/nonexistent piece of code. I'm wondering if you've reviewed the very last version or the one before (maybe you started the review, and then GH kept the previous revision). The last commit completely changes the implementation.

I think I had added that comment to the PR before your latest changes, but I did go back and review everything after the changes.

PR #870, opened 2023-09-01
feat: restore caching in aggregate functions that works with hierarchical tables [PT-185947550]

pjanik:
https://www.pivotaltracker.com/story/show/185947550

Hierarchical tables complicated everything quite a lot.
This is an approach that seems to support two cases when the aggregate function needs to be cached:
- aggregate function referencing another attribute in a flat table
- aggregate function referencing another attribute from the same collection in a hierarchical tables

When the aggregate function references an attribute from a child collection, caching is impossible and unnecessary.
Does it cover all the use cases for caching or am I missing something?

Semi-aggregate functions work only within the flat tables or the same collection, so they should be easier to handle.

### DIFF
--- a/v3/src/models/data/formula-fn-registry.ts
+++ b/v3/src/models/data/formula-fn-registry.ts
@@ -15,13 +15,20 @@ const evaluateNode = (node: MathNode, scope?: FormulaMathJsScope) => {
 const cachedAggregateFnFactory =
 (fnName: string, fn: (args: MathNode[], mathjs: any, scope: FormulaMathJsScope) => FValue | FValue[]) => {
   return (args: MathNode[], mathjs: any, scope: FormulaMathJsScope) => {
-    const cacheKey = `${fnName}(${args.toString()})`
+    const cacheKey = `${fnName}(${args.toString()})-${scope.getCaseGroupId()}`
     const cachedValue = scope.getCached(cacheKey)
     if (cachedValue !== undefined) {
       return cachedValue
     }
     const result = fn(args, mathjs, scope)
-    scope.setCached(cacheKey, result)
+    // In practice, the cacheability is defined by the fact whether the function accesses attributes from the same
+    // collection only or not. It can be checked while the expression is calculated for the first time. `scope` object
+    // keeps track of that. Each scope is cacheable by default, but once it accesses an attribute from a different
+    // collection, it becomes not cacheable (and propagates this update to the parent scopes). Note that a new scope
+    // is created for each function call (and each instance keeps reference to its parent).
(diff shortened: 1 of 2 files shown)


v3/src/models/data/formula-fn-registry.ts
```diff
@@ -15,13 +15,20 @@ const evaluateNode = (node: MathNode, scope?: FormulaMathJsScope) => {
...
-    const cacheKey = `${fnName}(${args.toString()})`
+    const cacheKey = `${fnName}(${args.toString()})-${scope.getCaseGroupId()}`
     const cachedValue = scope.getCached(cacheKey)
     if (cachedValue !== undefined) {
       return cachedValue
     }
     const result = fn(args, mathjs, scope)
-    scope.setCached(cacheKey, result)
+    // In practice, the cacheability is defined by the fact whether the function accesses attributes from the same
+    // collection only or not. It can be checked while the expression is calculated for the first time. `scope` object
```
kswenson, line 25:
```suggestion
    // In practice, the cacheability is defined by whether or not the function accesses attributes from the same
    // collection only. It can be checked while the expression is calculated for the first time. `scope` object
```

v3/src/models/data/formula-mathjs-scope.ts
```diff
@@ -16,20 +16,45 @@ export interface IFormulaMathjsScopeContext {
...
+    // performance reasons.
+    if (parent) {
+      this.parent = parent
+      this.caseId = parent.caseId
+      this.cache = parent.cache
+    }
+    // Cacheability does not depend on the parent. It's actually defined the other way - parent is not cacheable if
+    // any of its children is not cacheable (see setNotCacheable() method).
+    this.cacheable = true
+    // Note that dataStorage cannot be reused, as it relies on closures and bounding to correct `this` instance.
```
kswenson, line 43:
```suggestion
    // Note that dataStorage cannot be reused, as it relies on closures and binding to correct `this` instance.
```

v3/src/models/data/formula-mathjs-scope.ts
```diff
@@ -16,20 +16,45 @@ export interface IFormulaMathjsScopeContext {
...
+    // It needs to be recreated, but it is not be a costly operation. Make sure that all the caching and
```
kswenson, line 44:
```suggestion
    // It needs to be recreated, but it is not a costly operation. Make sure that all the caching and
```

kswenson (approved):
👍 Looks good -- at some point it would be good to go through some examples but for now this looks like a step in the right direction.

PR #871, opened 2023-09-04
Evaluate aggregate function against the child-most collection in hierarchical tables (following V2 approach)

pjanik:
https://www.pivotaltracker.com/story/show/185967142

In V2, formulas with aggregate functions were evaluated against the child-most collection in hierarchical tables. This pull request introduces the same behavior in V3.

[Discussion on Slack #1](https://concord-consortium.slack.com/archives/C035J6RDAK0/p1693584036196519)
[Discussion on Slack #2](https://concord-consortium.slack.com/archives/C035J6RDAK0/p1693833564638489)

Details:
I've managed to streamline the update using an approach that still relies on arrays. Now, when encountering a formula like `mean(LifeSpan + newAttr)` (where newAttr is in a parent collection and LifeSpan is in a child one), it is resolved as follows:

- `mean([70, 70, 40, 25, ...] + [1, 1, 1, ...])`
- `mean([19, 25, 14, ...] + [2, 2, 2, ...])`
- `mean([40, 9, 3, ...] + [3, 3, 3, ...])`

We discussed this as one of the options on Friday.

I find this approach appealing because it allows us to implement aggregate functions in a self-contained and elegant manner, without diving into the details of how formulas are evaluated. Therefore, the code in the formula-fn-registry (or whatever set of files it evolves into in the future) should remain easy to read, and adding new functions, even aggregate ones, should be relatively straightforward. Let's see if it holds up as we encounter more edge cases in the future.

Additionally, this new approach has significantly simplified caching for aggregate functions. We now use a single grouping type per formula, making many aspects of our implementation much smoother.

### DIFF
--- a/v3/src/models/data/formula-mathjs-scope.ts
+++ b/v3/src/models/data/formula-mathjs-scope.ts
@@ -2,15 +2,17 @@ import { AGGREGATE_SYMBOL_SUFFIX, GLOBAL_VALUE, LOCAL_ATTR, NO_PARENT_KEY } from
 import type { IGlobalValueManager } from "../global/global-value-manager"
 import type { IDataSet } from "./data-set"
 import type { IValueType } from "./attribute"
+import type { ICase } from "./data-set-types"
 
 const CACHE_ENABLED = true
 
 export interface IFormulaMathjsScopeContext {
   localDataSet: IDataSet
   dataSets: Map<string, IDataSet>
-  globalValueManager?: IGlobalValueManager
-  formulaAttributeCollectionId?: string
+  childMostCollectionCases: ICase[]
+  useSameLevelGrouping: boolean
   caseGroupId: Record<string, string>
+  globalValueManager?: IGlobalValueManager
 }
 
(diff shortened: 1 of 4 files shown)


v3/src/models/data/formula-mathjs-scope.ts
```diff
@@ -53,32 +36,20 @@ export class FormulaMathJsScope {
...
       Object.defineProperty(this.dataStorage, `${LOCAL_ATTR}${attr.id}`, {
         get: () => {
-          // Any expression that depends on case-dependent attribute is not cacheable.
-          this.setNotCacheable()
           return context.localDataSet.getValue(this.caseId, attr.id)
         }
       })
 
-      const allCasesForAttr = context.localDataSet.getCasesForAttributes([attr.id])
+      // Make sure that all the caching and  case processing is done lazily, only for attributes that are actually
```
kswenson, line 43:
```suggestion
      // Make sure that all the caching and case processing is done lazily, only for attributes that are actually
```

kswenson (approved):
👍 Looks good -- nice simplification

PR #872, opened 2023-09-05
Slightly refactor DataSet collection-related methods (-> ungrouped col consistency)

pjanik:
I've been working a bit with DataSet's collection-related helpers and noticed some inconsistencies. Some collection methods treat "ungrouped" fake-collection as a collection and return it, while others do not.

Eg. you can receive "ungrouped" when you ask for an attribute collection, but when you ask explicitly for a collection or collection index, `ungrouped` is no longer treated as one.  

These assumptions are invisible from the outside and require some special handling in the client code (in my case, it's a formula engine). So, I thought it'd be nice to make it more consistent and always include or always exclude `ungrouped` collection.

My current idea is that unless the client explicitly asks for a "real" collection (naming suggestions welcome), `ungrouped` is included in all the logic and results.

### DIFF
--- a/v3/src/models/data/data-set.ts
+++ b/v3/src/models/data/data-set.ts
@@ -209,12 +209,19 @@ export const DataSet = types.model("DataSet", {
   let _childCases: IGroupedCase[] = []
   let isValidCollectionGroups = false
 
-  function getCollection(collectionId: string): ICollectionModel | undefined {
+  function getRealCollection(collectionId: string): ICollectionModel | undefined {
     return self.collections.find(coll => coll.id === collectionId)
   }
 
+  function getCollection(collectionId: string): ICollectionPropsModel | undefined {
+    return collectionId === self.ungrouped.id ? self.ungrouped : getRealCollection(collectionId)
+  }
+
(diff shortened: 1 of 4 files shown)


v3/src/models/data/data-set.ts
```diff
@@ -209,12 +209,19 @@ export const DataSet = types.model("DataSet", {
   let _childCases: IGroupedCase[] = []
   let isValidCollectionGroups = false
 
-  function getCollection(collectionId: string): ICollectionModel | undefined {
+  function getRealCollection(collectionId: string): ICollectionModel | undefined {
```
kswenson, line 212: How about `getGroupedCollection()`?

kswenson (approved):
👍 Looks good -- I've encountered that same inconsistency but hadn't gotten around to figuring out a solution. This seems like a reasonable one, with the proviso that I suggested a name change below.

v3/src/models/data/data-set.ts
pjanik: Sounds good. I didn't like the "real" naming, I just couldn't come up with anything better. I just pushed a commit that changes this name.

PR #874, opened 2023-09-05
Detect formulas incorrectly referencing parent or child attributes in hierarchical tables

pjanik:
https://www.pivotaltracker.com/story/show/185972525

Also, related Slack discussion:
https://concord-consortium.slack.com/archives/C035J6RDAK0/p1693414962764689

This PR is a first pass on displaying formula errors to users. Two mistakes that are now handled explicitly:
1. Aggregate functions should not be able to reference attributes from parent collections. 
2. Formula in parent collection should not be able to reference child attributes without using the aggregate function

One of these errors (1.) was not handled by CODAP V2, so I had to add a new error message to the translation strings.

Also, if mathjs throws an error while evaluating a formula, it'll be displayed to the user. We won't be able to translate it, so long-term we'll probably to catch these error ourselves even before letting mathjs do it. But I think it's still a reasonable fallback mechanism for scenarios that we won't think of or forget to handle.

### DIFF
--- a/v3/src/models/data/formula-manager.ts
+++ b/v3/src/models/data/formula-manager.ts
@@ -1,11 +1,15 @@
 import { makeObservable, observable, reaction } from "mobx"
+import { EvalFunction } from "mathjs"
 import { FormulaMathJsScope } from "./formula-mathjs-scope"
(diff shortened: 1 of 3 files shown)


kswenson (approved):
👍 Looks good! Things seem to be coming together nicely.

PR #875, opened 2023-09-05
feat: Movable values can be placed on a dot plot (PT-185758373)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/185758373

- Enables the Movable Value option in the Graph component's Measure panel.
- Clicking the button makes the Add and Remove options appear. Clicking Add will add a new movable value to the graph. Clicking Remove will remove the most-recently-added movable value from the graph.
- Any number of movable values can be added to a graph. Each new value appears 1/3 of the way from the lower bound of the largest gap formed by the axis' min, axis' max, and any other existing movable values between the min and max.
- The gaps between the axis' min, axis' max, and movable values alternate between being shaded blue and being transparent -- blue between the first and second, transparent between the second and third, blue between the third and fourth, and so on.
- Movable values can be dragged to change their value.


v3/src/components/graph/adornments/movable-value/movable-value.tsx
```diff
@@ -1,116 +1,270 @@
...
+      const selection = select(valueRef.current)
+      selection.html(null)
+      select(`#${containerId}`).selectAll("div").remove()
 
-  // Add the behavior to the line cover
-  useEffect(function addBehaviors() {
-    valueObject.cover?.call(drag().on('drag', dragValue))
-  }, [valueObject, dragValue])
+      for (let i = 0; i < valueSet.length; i++) {
+        const newValueObject: any = {}
```
kswenson, line 215: Shouldn't the type be `IValueObject`?

v3/src/components/graph/adornments/movable-value/movable-value.tsx
```diff
@@ -1,116 +1,270 @@
...
-        .transition()
-        .duration(1000)
-        .style('opacity', 0)
-        .end().then(() => {
-        newValueObject.valueLabel.remove()
-      })
+      valueObjects.current = []
     }
-  }, [transform])
+  }, [])
```
kswenson, line 255: This effect seems unnecessary since it's only going to run on unmount at which point the ref will be destroyed automatically.

kswenson (changes requested):
👍 Looks good! A bunch of suggestions, most of them minor, but the undo/redo-related ones may require a bit of work.

kswenson:
@bfinzer I've reviewed this but I think you should take a look at it, specifically the graph and D3-specific areas.

bfinzer:
Bug: Error in console by adding to movable values and then removing one.
Error: <rect> attribute x: Expected length, "NaN".

bfinzer:
Bug: When I have a shaded region, it seems the shaded rectangle is in front of the line so that the cursor only shows correctly if I'm slightly outside.

bfinzer:
Error when adding three movable values, dragging them around across each other and then removing one.
[mobx] Encountered an uncaught exception that was thrown by a reaction or observer component, in: 'Reaction[Autorun@1633]'
TypeError: Cannot read properties of undefined (reading 'line') at eval (webpack-internal:///./src/components/graph/adornments/movable-value/movable-value.tsx:149:131) at eval (webpack-internal:///./src/components/graph/adornments/movable-value/movable-value.tsx:177:13) at reactionRunner (webpack-internal:///./node_modules/mobx/dist/mobx.esm.js:2537:5) at trackDerivedFunction (webpack-internal:///./node_modules/mobx/dist/mobx.esm.js:1746:18) at Reaction.track (webpack-internal:///./node_modules/mobx/dist/mobx.esm.js:2263:18) at Reaction.eval [as onInvalidate_] (webpack-internal:///./node_modules/mobx/dist/mobx.esm.js:2518:12) at Reaction.runReaction_ (webpack-internal:///./node_modules/mobx/dist/mobx.esm.js:2228:16) at runReactionsHelper (webpack-internal:///./node_modules/mobx/dist/mobx.esm.js:2376:29) at reactionScheduler (webpack-internal:///./node_modules/mobx/dist/mobx.esm.js:2352:10) at eval (webpack-internal:///./node_modules/mobx/dist/mobx.esm.js:2386:14) {stack: "TypeError: Cannot read properties of undefined (re…:///./node_modules/mobx/dist/mobx.esm.js:2386:14)", message: "Cannot read properties of undefined (reading 'line')"}

bfinzer:
Not able to reproduce it, but after removing one and then adding back in, the shaded rectangle came up with the wrong edge until I moved the value, as though it was coming up where the value used to be?

bfinzer:
Maybe a separate story, but in V2 if I add three movable values and then split with categorical on right, each of the subplots gets three movable values. I'm only see one in each subplot here.

bfinzer:
Bug: Add movable value. Split by categorical on right. Remove split attribute. Result is one movable values as it should be but it does have the full extent of the plot even after dragging.

bfinzer:
Bug: With numeric attribute on vertical axis, movable value is not draggable by its handle. But is still draggable by the line.

bfinzer:
Bug: Movable value does not follow the axis when the axis is dragged. (Only able to do this with numeric vertical axis currently.)

kswenson (approved):
👍 Changes look good -- I'll let you and Bill decide whether to address the bugs he found as part of this PR or leave them for a subsequent PR.

bfinzer:
For some reason I can't leave a review separate from this conversation. I leave it to you to decide which of the bugs I found to fix as part of this PR and which to log as PT stories.

emcelroy:
> For some reason I can't leave a review separate from this conversation. I leave it to you to decide which of the bugs I found to fix as part of this PR and which to log as PT stories.

@bfinzer Thanks for catching those bugs. Most of them were relatively straightforward to fix. The two that are a bit tricky are:

1) **Movable value does not follow the axis when the axis is dragged.**
I definitely had this working at one point but yeah it's broken again. It's not yet clear to me what's wrong now, but I'll spend some more time on it this afternoon. Hopefully it's something I can figure out today.

2) **In V2 if I add three movable values and then split with categorical on right, each of the subplots gets three movable values. I'm only see one in each subplot here.** I suspect this may warrant a separate story. I spent a few minutes working out a fix, but it didn't quite work right. I'll try to spend a little more time on that after hopefully fixing #1. Will report back later.

kswenson (changes requested):
👍 Looks good -- I made a bunch of suggestions, then I ended up trying some of them out locally to make sure I wasn't giving you a bum steer, and in the end I pushed a couple of commits to the PR. Take a look and see what you think. Try it out and make sure I didn't break anything -- I only gave it some light testing. Also, there are several suggestions _not_ addressed by my commits so take a look at those as well.

v3/src/components/graph/adornments/movable-value/movable-value.tsx
```diff
@@ -1,116 +1,279 @@
...
+    yCats = yAttrType === "categorical" && yCatSet ? Array.from(yCatSet.values) : [""],
+    xCellCount = xCats.length * xSubAxesCount,
+    yCellCount = yCats.length * ySubAxesCount,
+    values = model.values.get(instanceKey),
+    valueRef = useRef<SVGSVGElement>(null),
+    valueObjects = useRef<IValueObject[]>([]),
+    draggingIndex = useRef<number>(-1),
+    preDragValue = useRef<number | undefined>(undefined),
+    isVertical = useRef(!!(xAttrType && xAttrType === "numeric")),
+    dragValue = useRef<number | undefined>(undefined)
```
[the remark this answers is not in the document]
emcelroy: I did, but ran into bugs when there was more than one movable value instance. With your changes and a couple small tweaks, though, those issues went away. So I do think that's the way to go. Being able to drop the separate `sortedValuesWithDragValue()` is great. Thanks 👍

kswenson (approved):
👍 Looks good

PR #877, opened 2023-09-12
Update dependencies, fix lint errors

pjanik:
While I was struggling with and debugging `reaction` in FormulaManager, I decided to update mobx and react-mobx-lite out of desperation. Although it didn't resolve the issue, it also didn't cause any harm. So, I proceeded to update most of the other dependencies that appeared to be safe to update.

Also, a few lint fixes - possibly related to newer versions of eslint plugins.


pjanik:
Well, I've noticed that Cypress tests started to fail. But it seems it might be actually what we want. I might try fixing the problematic code or ask Ethan to help me. I've commented on the problematic lines:
https://github.com/concord-consortium/codap/pull/839/files#r1323111819

kswenson (approved):
👍 LGTM -- updating dependencies was on my list of things to get to soon but you beat me to it! Thx

pjanik:
So, I spent some time yesterday trying to narrow down the problem. The Cypress test starts to fail after updating `react-mobx-lite` from v3 to v4. However, I don't think it is the fault of `react-mobx-lite`. It is more about the related code that I linked in the previous comment. I think it relies too much on refs that affect rendering, so it is somewhat non-deterministic, and the update of `react-mobx-lite` is just a random trigger.

For now, I will downgrade `react-mobx-lite` to v3 so we don't break the `main` branch.

I've added a PT story to track the issue:
https://www.pivotaltracker.com/story/show/186031756

PR #879, opened 2023-09-12
Support importing of V2 docs with formulas

pjanik:
This PR adds support for importing V2 docs with formulas.
It required a couple of fixes, but the most important one was to change the formula's source of truth to display string.
The canonical variant is computed as a derived value, so in most cases, it should be cached (that's why I've added `observer` to a few components). I bet over time, we might decide that it's better to store canonical string too, but for now, it simplifies quite a few things (no need to worry about synchronization between these two strings).

### DIFF
--- a/v3/src/components/case-table/attribute-menu/attribute-menu-list.tsx
+++ b/v3/src/components/case-table/attribute-menu/attribute-menu-list.tsx
@@ -1,5 +1,5 @@
 import React, { forwardRef } from "react"
-
+import { observer } from "mobx-react-lite"
 import { MenuItem, MenuList, useDisclosure, useToast } from "@chakra-ui/react"
 import { useCaseMetadata } from "../../../hooks/use-case-metadata"
 import { useDataSetContext } from "../../../hooks/use-data-set-context"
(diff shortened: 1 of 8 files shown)


v3/src/models/data/formula.ts
```diff
@@ -5,29 +5,29 @@ import { getFormulaManager } from "../tiles/tile-environment"
 
 export const Formula = types.model("Formula", {
   id: types.optional(types.identifier, () => typedId("FORMULA")),
```
kswenson, line 7: Conventionally, the `typedId` prefix is four characters, so this would be `FORM`.

kswenson (approved):
👍 Looks good

v3/src/components/case-table/attribute-menu/attribute-menu-list.tsx
```diff
@@ -1,19 +1,21 @@
 import React, { forwardRef } from "react"
+
 import { MenuItem, MenuList, useDisclosure, useToast } from "@chakra-ui/react"
 import { useCaseMetadata } from "../../../hooks/use-case-metadata"
 import { useDataSetContext } from "../../../hooks/use-data-set-context"
 import { TCalculatedColumn } from "../case-table-types"
 import { EditAttributePropertiesModal } from "./edit-attribute-properties-modal"
 import t from "../../../utilities/translation/translate"
 import { EditFormulaModal } from "./edit-formula-modal"
+import { observer } from "mobx-react-lite"
```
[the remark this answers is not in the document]
pjanik: Doh, forgotten VSCode auto-import again. I need to know what was the result of @scytacki's spike 😉 There are two ESlint packages that can order imports for us automatically, and they have lots of rules - I'd hope we could replicate our patterns.

v3/src/models/data/formula.ts
pjanik: Thanks, fixed.

PR #880, opened 2023-09-12
feat: Add plotted value (PT-181940090)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/181940090


v3/src/components/graph/adornments/plotted-value/edit-formula-modal.tsx
kswenson, line 1: Ideally, we'll eventually be able to share a single formula editor implementation, but as a stopgap measure this is fine for now.

v3/src/components/graph/adornments/plotted-value/plotted-value-model.ts
```diff
@@ -0,0 +1,34 @@
+import { Instance, types } from "mobx-state-tree"
+import { mean } from "mathjs"
+import { AdornmentModel, IAdornmentModel, IUpdateCategoriesOptions } from "../adornment-models"
+import { kPlottedValueType } from "./plotted-value-types"
+
+export const PlottedValueModel = AdornmentModel
+  .named("PlottedValueModel")
+  .props({
+    type: "Plotted Value",
+    value: types.maybe(types.union(types.number, types.string))
```
kswenson, line 10: For the plotted value, the formula should be serialized, not the value. It may be that all you need is an instance of `FormulaModel`. Probably okay to leave this for @pjanik to work out as well.

v3/src/components/graph/adornments/plotted-value/plotted-value.tsx
```diff
@@ -0,0 +1,162 @@
...
+  const yScale = layout.getAxisScale("left") as ScaleNumericBaseType
+  const yAttrId = dataConfig?.attributeID("y")
+  const yAttrType = dataConfig?.attributeType("y")
+  const classFromKey = model.classNameFromKey(cellKey)
+  const valueRef = useRef<SVGGElement>(null)
+  const isVertical = useRef(!!(xAttrType && xAttrType === "numeric"))
+  const previousAttrTypes = usePrevious({ xAttrType, yAttrType})
+  // TODO: The offset value below shouldn't be needed once there's a method available for adding
+  // space to the top of the graph to accommodate the Plotted Value's UI elements.
+  const offsetTop = 20
```
kswenson, line 50: This is obviously a bit of a hack, but I'm okay with it for now until we get a better feel for what the requirements should be. For instance, should we support the possibility of multiple banners at the top of the plot?

v3/src/components/graph/adornments/plotted-value/plotted-value.tsx
```diff
@@ -0,0 +1,162 @@
...
+        : []
+    return cases
+  }, [dataset, isVertical, xAttrId, yAttrId])
+
+  const getCaseValues = useCallback(() => {
+    const cases = getCases()
+    const caseValues = cases && xAttrId
+      ? cases.map(c => dataset?.getNumeric(c.__id__, xAttrId)).filter(x => x) as number[]
+      : cases && yAttrId
+        ? cases.map(c => dataset?.getNumeric(c.__id__, yAttrId)).filter(x => x) as number[]
```
kswenson, line 66: Seems like this would strip out zeros. I think you want something like:
```suggestion
      ? cases.map(c => dataset?.getNumeric(c.__id__, xAttrId)).filter(x => isFinite(x)) as number[]
      : cases && yAttrId
        ? cases.map(c => dataset?.getNumeric(c.__id__, yAttrId)).filter(x => isFinite(x)) as number[]
```

v3/src/components/graph/adornments/plotted-value/plotted-value.tsx
```diff
@@ -0,0 +1,162 @@
...
+        : [] as number[]
+    return caseValues
+  }, [getCases, dataset, xAttrId, yAttrId])
+
+  // Updates the coordinates of the line, its cover, and text label
+  const refreshValue = useCallback(() => {
+    const disposer = autorun(() => {
```
kswenson, line 73:
```suggestion
```
I don't think you want the `autorun` here. `refreshValue()` is just the function that updates the value.

v3/src/components/graph/adornments/plotted-value/plotted-value.tsx
```diff
@@ -0,0 +1,162 @@
...
+    return () => disposer()
+  }, [getCaseValues, classFromKey, containerId, model, plotHeight, plotWidth, xScale, yScale])
+
+  // Refresh the value when it changes
+  useEffect(function refreshValueChange() {
+    const disposer = autorun(() => {
+      refreshValue()
+    }, { name: "PlottedValue.refreshValueChange" })
+    return () => disposer()
+  }, [refreshValue, model.value])
```
kswenson, line 121: You shouldn't need to dispose of and reinstall the `autorun` on value changes. The `refreshValue()` function already accesses the model value, so any change should trigger the `autorun` which should call `refreshValue()`.
```suggestion
  useEffect(function refreshValueChange() {
    const disposer = autorun(() => {
      refreshValue()
    }, { name: "PlottedValue.refreshValueChange" })
    return () => disposer()
  }, [refreshValue])
```

kswenson (changes requested):
👍 Looks good I made some suggestions on some of the autoruns that should be fixed. As for the notion of helpers, the hardest part about it is naming. How about calling them banners, i.e. each adornments has the opportunity to specify a banner which will be rendered at/across the top of the plot?

emcelroy:
> As for the notion of helpers, the hardest part about it is naming. How about calling them banners, i.e. each adornments has the opportunity to specify a banner which will be rendered at/across the top of the plot?

Banners makes sense to me. Thanks. I went ahead and used that in place of helpers.

v3/src/components/graph/adornments/plotted-value/plotted-value-model.ts
emcelroy: I spent a little time on this but it wasn't clear to me how to do it. So unless @pjanik feels strongly otherwise, I'm going to leave it for him.

kswenson (approved):
👍 Changes look good

PR #881, opened 2023-09-13
feat: Support symbol names enclosed in backticks (``) within formulas

pjanik:
https://www.pivotaltracker.com/story/show/186033178

CODAP v2 handles special attribute names if they're enclosed in backticks. The most common use case can be attribute name with whitespace. This PR adds this feature to V3.


kswenson (approved):
👍 LGTM

PR #882, opened 2023-09-13
fix: fade in / out effects in graph adornment (PT-186031756)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/186031756

Simplifies the application of classes associated with the CSS animations that make adornments fade in and out when they're added and removed from a graph. Also updates mobx-react-lite to version 4.0.4.


v3/cypress/support/commands.ts
```diff
@@ -11,7 +11,7 @@ Cypress.Commands.add("dragAttributeToTarget", (source, attribute, target, num =
     caseCardHeaderDropZone: ".react-data-card .data-cell-lower",
     caseCardCollectionDropZone: ".react-data-card .collection-header-row",
     graphTile: ".graph-plot svg",
-    legend: ".graph-plot .droppable-legend",
+    legend: ".graph-plot .legend .chakra-menu__menu-button",
```
emcelroy, line 14: This change was needed to get the legend tests that started to fail to pass. I'm not 100% sure why, to be honest. There _is_ a DIV element with the class `.droppable-legend` inside another DIV element with the class `.graph-plot` but for whatever reason, the Cypress test was not able to drop an attribute there. When it tried to, the attribute would end up on the y axis instead of on the legend.

I ran this change by @tejal-shah who thought it was acceptable. She pointed out we really should be using data-testid values instead of class names, but that change should be part of a separate story.

kswenson (approved):
👍 LGTM

PR #883, opened 2023-09-13
feat(formula): add `count` function [PT-186033205]

pjanik:
https://www.pivotaltracker.com/story/show/186033205

### DIFF
--- a/v3/src/models/data/formula-fn-registry.ts
+++ b/v3/src/models/data/formula-fn-registry.ts
@@ -148,6 +148,24 @@ export const fnRegistry = {
     }
   },
 
+  // count(expression, filterExpression)
+  count: {
+    isAggregate: true,
+    cachedEvaluateFactory: cachedAggregateFnFactory,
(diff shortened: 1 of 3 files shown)


v3/src/models/data/formula-fn-registry.ts
```diff
@@ -148,6 +148,24 @@ export const fnRegistry = {
     }
   },
 
+  // count(expression, filterExpression)
+  count: {
+    isAggregate: true,
+    cachedEvaluateFactory: cachedAggregateFnFactory,
+    evaluateRaw: (args: MathNode[], mathjs: any, scope: FormulaMathJsScope) => {
+      const expression = args[0]
+      const filter = args[1]
```
kswenson, line 157: You can also use:
```suggestion
      const [expression, filter] = args
```

kswenson (approved):
👍 Looks good

v3/src/models/data/formula-fn-registry.ts
pjanik: True, I'll update all existing functions when I add a new one.

PR #885, opened 2023-09-14
Update formula fn registry

pjanik:
A couple of new functions (aggregate + `if`), fixed == operator, and fixed `prev` and `next` functions (they didn't support the scenario without the filter arg).


kswenson (approved):
👍 Looks good -- I don't see anything that would affect the `prev` and `next` functions as described in the PR description, however.

PR #886, opened 2023-09-15
fix(formula): update prev() and next() to handle calls without filter arg [PT-185745491]

pjanik:
This PR fixes prev and next functions, as they didn't support the scenario without the filter arg.


kswenson (approved):
👍 LGTM

PR #887, opened 2023-09-15
Handle and render formula syntax errors

pjanik:
https://www.pivotaltracker.com/story/show/186052466

This PR cleans up the handling of syntax errors. CODAP will not crash in such a case, and the syntax error will be rendered as formula output (similarly to V2).


kswenson (approved):
👍 Looks good

PR #889, opened 2023-09-18
Eevaluate `prev()` and `next()` respecting grouping of hierarchical tables

pjanik:
https://www.pivotaltracker.com/story/show/186061842

During my work on the recursive `prev()` method, I realized that neither `prev()` nor `next()` properly take into account hierarchical grouping in tables. With this pull request, I have addressed this issue and improved the implementation of these methods to be more streamlined.

### DIFF
--- a/v3/src/models/data/formula-fn-registry.ts
+++ b/v3/src/models/data/formula-fn-registry.ts
@@ -30,14 +30,13 @@ const cachedAggregateFnFactory =
 // The only difference is the final math operation applies to the expression results.
 const aggregateFnWithFilterFactory = (fn: (values: number[]) => number) => {
   return (args: MathNode[], mathjs: any, scope: FormulaMathJsScope) => {
-    const expression = args[0]
-      const filter = args[1]
-      let expressionValues = evaluateNode(expression, scope)
-      if (filter) {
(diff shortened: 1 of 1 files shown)


v3/src/models/data/formula-fn-registry.ts
```diff
@@ -30,17 +30,19 @@ const cachedAggregateFnFactory =
...
+    let expressionValues = evaluateNode(expression, scope)
+    if (filter) {
+      const filterValues = evaluateNode(filter, scope)
+      expressionValues = expressionValues.filter((v: any, i: number) => !!filterValues[i])
+    }
+    return fn(expressionValues)
   }
 }
 
+// CODAP formulas assume that 0 is a truthy value, which is different from default JS behavior.
```
kswenson, line 43: One might add something like "so that, for instance, `count(attribute)` will return a count of valid data values, since 0 is a valid numeric value." to provide some context for the decision.

kswenson (approved):
👍 Looks good

v3/src/models/data/formula-fn-registry.ts
pjanik: Thanks, I've just added it to my next PR.

PR #890, opened 2023-09-18
feat: axis drags undoable as a single action; also fix bugs in axis dragging

kswenson:
- feat: axis drags undoable as a single action
- fix: graph x-axis dragging
- fix: slider axis dragging


kswenson:
Reviewed synchronously with @bfinzer.

PR #891, opened 2023-09-19
Support formulas with `prev()` that references its own attribute

pjanik:
https://www.pivotaltracker.com/story/show/186050666

It's been a lot of work for me to develop a solution covering all the scenarios. There's a basic running sum like  `OtherAtrr + prev(FormulaAttr)`, but also other forms that caused issues in my previous approaches: `prev(round(ceil(FormulaAttr)), 0) + 1`, `prev(FormulaAttr + OtherAttr, 0) + 1`, etc. While they might not seem useful, we should not have a semi-correct implementation that works only for presumably useful scenarios. 

Finally, I had to rely on an iterative approach that is "simulated" using caching within the `prev` implementation. This differs from other aggregate or semi-aggregate functions that depend on array operations. I'm not happy we need to support that, but after talking to Dan, it seems it might be important for him and used in quite a few documents - mostly to calculate running sum like `OtherAtrr + prev(FormulaAttr)`. Here's the whole Slack thread that might be worth reading:
https://concord-consortium.slack.com/archives/C035J6RDAK0/p1695129101652899

Note that CODAP V2 also supports self-reference in `next()` - probably by reversing the order of processing cases. But my suggestion is not to implement that in V3 for now. I think the idea of not treating self-reference as circular dependency is a bit fragile - it enforces a strong assumption about the order in which the formula is evaluated. For `prev()` to work, we must iterate from 0 to N. From `next`, it needs to be from N to 0. That sounds fine at first glance, we could swap interaction order if `next` is detected, but it complicates the code even more. Plus, what happens when you combine `prev` and `next` in one formula?

Also, while I see use cases for recursive `prev`, I cannot think of any reasonable use case for self-reference in `next`. So, I'd suggest not allowing it there and justifying it by saying that formulas are always calculated in order of cases, from 0 to N (and that's why `prev` is an exception that works with self-reference).

### DIFF
--- a/v3/src/models/data/formula-fn-registry.ts
+++ b/v3/src/models/data/formula-fn-registry.ts
@@ -288,6 +288,9 @@ export const fnRegistry = {
 
   // prev(expression, defaultValue, filter)
   prev: {
+    // Circular reference might be used to define a formula that calculates the cumulative value, e.g.:
+    // `CumulativeValue` attribute formula: `Value + prev(CumulativeValue, 0)`
+    selfReferenceAllowed: true,
     // expression and filter are evaluated as aggregate symbols, defaultValue is not - it depends on case index
     isSemiAggregate: [true, false, true],
     evaluateRaw: (args: MathNode[], mathjs: any, scope: FormulaMathJsScope) => {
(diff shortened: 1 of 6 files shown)


kswenson (approved):
👍 Looks good! Changes were less significant than I expected.

PR #892, opened 2023-09-20
Fix misc issues related to formulas showing while importing V2 document

pjanik:
1. Before checking for syntax errors, we need to apply formula transformations (e.g. \`\` replacement)
2. All formulas need to be registered (=> metadata with parent attribute, dataset, etc.) before we can check for errors and dependency cycles. The previous implementation was working with import only if formulas were registered in the "right" order, so it was pretty random. Now, it should always work - even when the formula is referencing another formula that is registered later.


kswenson (approved):
👍 LGTM

PR #893, opened 2023-09-21
Recalculate formulas when attr is moved between collections

pjanik:
https://www.pivotaltracker.com/story/show/186040941

"setCollectionForAttribute" seemed like the only action always being called for various use cases and called only once (moving between existing collections or while creating a new one). While I see some benefits of `onAnyAction`, I have a feeling that it might be quite fragile long-term. It feels like relying on the implementation details of various models, also making them harder to refactor if necessary. That makes me think that some simple event dispatcher (like basic event-emitte2) with more controlled, explicit events could be useful. But that's not a topic for this PR anyway. :wink:

### DIFF
--- a/v3/src/models/data/formula-manager.ts
+++ b/v3/src/models/data/formula-manager.ts
@@ -22,11 +22,17 @@ type IFormulaMetadata = {
   registeredDisplay: string
   attributeId: string
   dataSetId: string
+  active: boolean
   dispose?: () => void
+}
+
+type IDataSetMetadata = {
+  dispose: () => void
 }
 
 export class FormulaManager {
   @observable dataSets = new Map<string, IDataSet>()
+  dataSetMetadata = new Map<string, IDataSetMetadata>()
   formulaMetadata = new Map<string, IFormulaMetadata>()
   globalValueManager?: IGlobalValueManager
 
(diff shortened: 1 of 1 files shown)


v3/src/models/data/formula-manager.ts
```diff
@@ -347,6 +376,20 @@ export class FormulaManager {
     return displayNameMap
   }
 
+  observeDatasetChanges(dataSet: IDataSet) {
+    const dispose = onAnyAction(dataSet, mstAction => {
+      switch (mstAction.name) {
+        // When attribute is moved to a new collection, it usually affects grouping that is respected by formulas.
+        // It'd be possible to optimize this by checking formula dependencies and limit number of updates, but
+        // for now let's keep it simple and see if we encounter
```
kswenson, line 384:
```suggestion
        // for now let's keep it simple and see if we encounter any problems.
```
Or end the sentence however you were originally intending to. 😉

v3/src/models/data/formula-manager.ts
```diff
@@ -22,11 +22,17 @@ type IFormulaMetadata = {
   registeredDisplay: string
   attributeId: string
   dataSetId: string
+  active: boolean
```
kswenson, line 25: The meaning of this flag is not entirely clear. It seems that it's false when initially registered but then made true in the `registerAllFormulas` reaction, which suggests that it's just tracking whether it's fully initialized. Perhaps a name change (isInitialized?) and/or a comment would help.

kswenson (approved):
👍 Looks good -- just a couple potential clarifications.

>While I see some benefits of `onAnyAction`, I have a feeling that it might be quite fragile long-term. It feels like relying on the implementation details of various models, also making them harder to refactor if necessary. That makes me think that some simple event dispatcher (like basic event-emitter2) with more controlled, explicit events could be useful.

On the other hand, MST provides a notification mechanism which is tied directly to the actions that we care about. Adding a parallel notification mechanism is more code to write/maintain which duplicates functionality provided by MST, so until we encounter a situation in which the MST notifications don't meet our needs I'm inclined to make use of it.

v3/src/models/data/formula-manager.ts
pjanik: Thanks, I agree it's a much better name. I just renamed active -> isInitialized in the last commit.

PR #894, opened 2023-09-21
feat: create new table undoable; close component undoable

kswenson:


kswenson:
Sync review with @bfinzer.

PR #897, opened 2023-09-22
feat: delete attribute from table is undoable

kswenson:


v3/src/models/data/attribute.ts
```diff
@@ -117,12 +118,14 @@ export const Attribute = types.model("Attribute", {
   // should be called before retrieving snapshot (i.e. before serialization)
   prepareSnapshot() {
     if (isDevelopment()) {
```
bfinzer, line 120: For my information: Why is this a no-op in production?

v3/src/models/data/data-set.ts
```diff
@@ -180,6 +180,23 @@ export const DataSet = types.model("DataSet", {
   }
 })
 .views(self => ({
```
bfinzer, line 182: I like how these two new views simplify things by removing need to maintain these maps. 👍🏻

bfinzer (approved):
👍🏻LGTM
I don't pretend to understand running action contexts and calls. Hopefully I won't need to.

v3/src/models/data/attribute.ts
kswenson: See the comment in line 111 above, but the short answer is that MST freezes properties at the JS level in development but not in production (since there's a performance cost), so the implementation is slightly different in the two cases.

v3/src/models/data/data-set.ts
kswenson: I was originally maintaining all of the maps manually out of concern for performance, but adding/removing attributes doesn't happen often enough and the numbers of attributes aren't great enough to warrant the concern. For now, I'm leaving the case IDs map maintained manually because rebuilding it from scratch on every change could be expensive with a large number of cases, whereas we can handle the most common case -- adding a new case to the end -- much more efficiently when responding manually. Will probably have to address that issue down the road as well.

kswenson:
> I don't pretend to understand running action contexts and calls. Hopefully I won't need to.

There are subtleties you probably won't need to understand, but the basics are pretty straightforward. When you call an action like `setAttributeID()`, that creates an action context. When an outer action like `applyUndoableAction()` calls an inner action like `setAttributeID()`, then there are nested action contexts. The history/undo system has to track those calls and make sure it tracks all of the changes that occur within the outermost action context. For the purposes of this PR, there were a couple places where we need to navigate the nested action contexts to find the right one.

PR #898, opened 2023-09-25
Evaluate `caseIndex` in formulas

pjanik:
https://www.pivotaltracker.com/story/show/185896734

Initially, I hoped that `caseIndex` could be handled only by the mathjs scope. But it's not so simple - we can use aggregate or semi-aggregate functions with `caseIndex`, like `mean(caseIndex)` or `prev(caseIndex)`. So, `caseIndex` became a "fake local attribute" of a case. It essentially behaves like other local attributes, so aggregate or semi-aggregate functions are handled by the same logic and code.

### DIFF
--- a/v3/src/models/data/formula-utils.ts
+++ b/v3/src/models/data/formula-utils.ts
@@ -1,6 +1,6 @@
 import { parse, MathNode, isFunctionNode, isSymbolNode } from "mathjs"
 import {
-  AGGREGATE_SYMBOL_SUFFIX, LOCAL_ATTR, GLOBAL_VALUE, DisplayNameMap, IFormulaDependency, ILocalAttributeDependency,
+  AGGREGATE_SYMBOL_SUFFIX, LOCAL_ATTR, GLOBAL_VALUE, DisplayNameMap, IFormulaDependency, ILocalAttributeDependency
 } from "./formula-types"
 import { typedFnRegistry } from "./formula-fn-registry"
 import type { IDataSet } from "./data-set"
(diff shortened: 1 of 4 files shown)


v3/src/models/data/formula-utils.ts
```diff
@@ -180,15 +182,16 @@ export const getExtremeCollectionDependency =
 export const getFormulaChildMostAggregateCollectionIndex = (formulaCanonical: string, dataSet: IDataSet) => {
   const attrId = getExtremeCollectionDependency(formulaCanonical, dataSet, { order: "max", aggregate: true })
   const collectionId = dataSet.getCollectionForAttribute(attrId || "")?.id
-  return dataSet.getCollectionIndex(collectionId || "") ?? null
+  const collectionIndex = dataSet.getCollectionIndex(collectionId || "")
+  return collectionIndex !== -1 ? dataSet.getCollectionIndex(collectionId || "") : null
```
kswenson, line 186: Personal preference, but I generally use `>= 0` in situations like this because it reads more like _is a valid index_ than _isn't the magic sentinel value used to indicate invalid indices, which means it's valid_.

kswenson (approved):
👍 Looks good -- just one stylistic suggestion

v3/src/models/data/formula-utils.ts
pjanik: My brain reads `-1` as an invalid index (kinda like checking against null or undefined), but I get your point. I'll update it soon.

pjanik: Fixed in https://github.com/concord-consortium/codap/commit/e71c56ac75dc1583af2252a09d1cb76f9f902f26 that will be part of another PR (minimizing chances of merge conflicts).

PR #899, opened 2023-09-25
feat: Dot plots can show mean and median (PT-181914405)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/181914405


v3/src/components/graph/adornments/adornments-store.ts
```diff
@@ -0,0 +1,102 @@
...
+      let addedShowMeasureLabels = false
+      measures[plotType].map((measure: IMeasure) => {
+        const { title, type } = measure
+        const checked = !!self.adornments.find(a => a.type === type)?.isVisible
+        const registeredAdornment = registeredAdornments.find(a => a.type === title)
+        const isUnivariateMeasure = registeredAdornment?.subTypeOf === "Univariate Measure"
+        if (isUnivariateMeasure && !addedShowMeasureLabels) {
+          measureMenuItems.push({
+            checked: self.showMeasureLabels,
+            title: t("DG.Inspector.showLabels"),
```
kswenson, line 68: My inclination is that the model should deal in keys and the component should be responsible for performing the translation at render time.

v3/src/components/graph/adornments/adornments-store.ts
```diff
@@ -0,0 +1,102 @@
...
+  }))
+  .views(self => ({
+    getAdornmentsMenuItems(plotType: PlotTypes) {
+      const registeredAdornments = getAdornmentTypes()
+      const measureMenuItems: IMeasureMenuItem[] = []
+      let addedShowMeasureLabels = false
+      measures[plotType].map((measure: IMeasure) => {
+        const { title, type } = measure
+        const checked = !!self.adornments.find(a => a.type === type)?.isVisible
+        const registeredAdornment = registeredAdornments.find(a => a.type === title)
```
kswenson, line 63: You can't rely on the fact that the translated title of the menu item will match the model type of the adornment we're interested in -- that only works in English (and isn't necessarily guaranteed there). You either need to add the actual adornment type to the `measures` table (why is the existing `type` for all of the entries currently `adornment`?) or perhaps this is the time to get rid of the `measures` table entirely and build it from the registered adornments.

v3/src/components/graph/adornments/count/count-registration.tsx
```diff
@@ -12,7 +12,8 @@ import { useDataConfigurationContext } from "../../hooks/use-data-configuration-
 const Controls = () => {
   const graphModel = useGraphContentModelContext()
   const dataConfig = useDataConfigurationContext()
-  const existingAdornment = graphModel.adornments.find(a => a.type === kCountType) as ICountModel
+  const adornmentsStore = graphModel.adornmentsStore
+  const existingAdornment = adornmentsStore.adornments.find(a => a.type === kCountType) as ICountModel
```
kswenson, line 16: There are a number of this kind of cast throughout the code. I suggest adding a `view` like this to the `AdornmentsStore`:
```typescript
    findAdornmentOfType<T extends IAdornmentModel = IAdornmentModel>(type: string): T | undefined {
      return self.adornments.find(adornment => adornment.type === type) as T
    }
```
and then clients can use:
```typescript
  const countAdornment = adornmentsStore.findAdornmentOfType<ICountModel>(kCountType)
```

v3/src/components/graph/adornments/univariate-measures/univariate-measure-adornment-component.tsx
```diff
@@ -0,0 +1,245 @@
...
+        .attr("y", y / yCellCount)
+    }
+
+  }, [adornmentsStore?.activeUnivariateMeasures, classFromKey, containerId, handleEndMoveLabel,
+      handleMoveLabel, highlightCover, highlightLabel, layout, measureType, model.type, plotHeight,
+      plotWidth, showLabel, xAttrType, xScale, yAttrType, yScale])
+
+  // Add the lines and their associated covers and labels
+  const refreshValues = useCallback(() => {
+    const disposer = autorun(() => {
```
kswenson, line 187: I don't think you want the `autorun` here. `refreshValues` should just be a function that runs when called and then any necessary `autorun`s or `reaction`s would be installed in `useEffect`s like `refreshAxisChange` below.

v3/src/components/graph/adornments/univariate-measures/univariate-measure-adornment-component.tsx
```diff
@@ -0,0 +1,245 @@
...
+    return autorun(() => {
+      isVertical.current = dataConfig?.attributeType("x") === "numeric"
+      refreshValues()
+    }, { name: "UnivariateMeasureAdornmentComponent.refreshAxisChange" })
+  }, [dataConfig, refreshValues, xAxis?.max, xAxis?.min, yAxis?.max, yAxis?.min])
+
+  // On unmount, remove measure
+  useEffect(function removeMean() {
+    return () => {
+      model?.removeMeasure(instanceKey)
```
kswenson, line 219: Why are we removing measures from the model on unmount? In general, model changes should be triggered by user actions or responding directly to other model changes.

kswenson (changes requested):
👍 This is looking good but there are still some things I think can be improved. Let me know if you have any questions about the specific suggestions.

v3/src/components/graph/adornments/adornments-store.ts
emcelroy: Sorry I missed it, but that `type` property in `measures` should've been removed. I added it as part of an abandoned idea for identifying the "Show Measure Labels" item as something other than an adornment when it was still listed in the `measures` table (i.e. its type would've been something other than `adornment`, like `control`). I dropped that idea after removing "Show Measure Labels" from the list, but forgot to remove the new `type` property. 

I can easily enough change that `type` property to be used for the individual adornment types, though. I did that and changed the line to `const registeredAdornment = registeredAdornments.find(a => a.type === type)` which makes more sense. Thanks.

I had the impression we wanted to keep using the `measures` table for now. It'd be nice to drop it, but that'd mean there won't be any placeholder listings in the menu for the adornments that don't exist yet. Should we wait on that? Or do you think at this point that's not a concern?

v3/src/components/graph/adornments/univariate-measures/univariate-measure-adornment-component.tsx
emcelroy: I think I was in a mixed up state while trying to troubleshoot a problem when I added this. I've removed it and things work fine.

v3/src/components/graph/adornments/adornments-store.ts
```diff
@@ -0,0 +1,102 @@
...
+      self.activeUnivariateMeasures = activeUnivariateMeasures
+    },
+    removeActiveUnivariateMeasure(measure: string) {
+      const activeUnivariateMeasures = [...self.activeUnivariateMeasures]
+      const index = activeUnivariateMeasures.indexOf(measure)
+      if (index > -1) {
+        activeUnivariateMeasures.splice(index, 1)
+      }
+      self.activeUnivariateMeasures = activeUnivariateMeasures
+    },
```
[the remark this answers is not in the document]
emcelroy: Yeah, that seems to work just fine. Thanks. I've removed the `activeUnivariateMeasures` property and its associated actions.

v3/src/components/graph/adornments/adornment-content-info.ts
```diff
@@ -5,6 +5,7 @@ export interface IAdornmentContentInfo {
   modelClass: typeof AdornmentModel
   plots: PlotType[]
   prefix: string
+  subTypeOf?: string
```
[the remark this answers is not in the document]
emcelroy: As discussed elsewhere, for now at least, this information does need to be provided in the adornment content info so it's available before an instance of each adornment's model is created. I changed the property name to `parentType`, though, as you suggested. I also added that `isUnivariateMeasure()` view method to the models since it's useful elsewhere.

v3/src/components/graph/adornments/adornment-content-info.ts
```diff
@@ -23,5 +25,10 @@ export function getAdornmentContentModels() {
 }
 
 export function getAdornmentTypes() {
-  return Object.values(gAdornmentContentInfoMap).map(info => info.type)
+  return Object.values(gAdornmentContentInfoMap).map((info) => {
+    return {
+      parentType: info.parentType,
+      type: info.type,
+    }
+  })
```
kswenson, line 33: FYI, you should be able to replace this with:
```suggestion
  return Object.values(gAdornmentContentInfoMap).map(({ parentType, type }) => ({ parentType, type }))
```

v3/src/components/graph/adornments/adornment-models.ts
```diff
@@ -38,6 +38,7 @@ export interface IUpdateCategoriesOptions {
   rightCats: string[]
   rightAttrId: string
   resetPoints?: boolean
+  dataConfig?: any
```
kswenson, line 41: Is there a reason not to type this as `IDataConfigurationModel`?

v3/src/components/graph/adornments/adornment-types.ts
```diff
@@ -21,6 +25,52 @@ const adornmentTypeDispatcher = (adornmentSnap: IAdornmentModel) => {
...
+    { title: "DG.Inspector.graphPlottedStDev", type: "Standard Deviation" },
+    { title: "DG.Inspector.graphPlottedMeanAbsDev", type: "Mean Absolute Deviation" },
+    { title: "DG.Inspector.graphPlottedBoxPlot", type: "Box Plot" },
+    { title: "DG.Inspector.graphBoxPlotShowOutliers", type: "Show Outliers" },
+    { title: "DG.Inspector.graphPlottedValue", type: "Plotted Value" },
+    { title: "DG.Inspector.graphMovableValue", type: "Movable Value" }
+  ],
+  "scatterPlot": [
+    { title: "DG.Inspector.graphCount", type: "Count" },
+    { title: "DG.Inspector.graphConnectingLine", type: "adornment" },
```
kswenson, line 67:
```suggestion
    { title: "DG.Inspector.graphConnectingLine", type: "Connecting Lines" },
```

v3/src/components/graph/adornments/univariate-measures/mean/mean-adornment-model.ts
```diff
@@ -0,0 +1,25 @@
...
+  .views(self => ({
+    getMeasureValue(attrId: string, cellKey: Record<string, string>, dataConfig: IDataConfigurationModel) {
+      const casesInPlot = dataConfig.subPlotCases(cellKey)
+      const caseValues = self.getCaseValues(attrId, casesInPlot, dataConfig)
+      return mean(caseValues)
+    }
+  }))
+
+export interface IMeanAdornmentModel extends Instance<typeof MeanAdornmentModel> {}
+export function isMeanAdornment(adornment: IUnivariateMeasureAdornmentModel): adornment is IMeanAdornmentModel {
```
kswenson, line 23:
```suggestion
export function isMeanAdornment(adornment: IAdornmentModel): adornment is IMeanAdornmentModel {
```

v3/src/components/graph/adornments/univariate-measures/median/median-adornment-model.ts
```diff
@@ -0,0 +1,25 @@
...
+  .views(self => ({
+    getMeasureValue(attrId: string, cellKey: Record<string, string>, dataConfig: IDataConfigurationModel) {
+      const casesInPlot = dataConfig.subPlotCases(cellKey)
+      const caseValues = self.getCaseValues(attrId, casesInPlot, dataConfig)
+      return median(caseValues)
+    }
+  }))
+
+export interface IMedianAdornmentModel extends Instance<typeof MedianAdornmentModel> {}
+export function isMedianAdornment(adornment: IUnivariateMeasureAdornmentModel): adornment is IMedianAdornmentModel {
```
kswenson, line 23:
```suggestion
export function isMedianAdornment(adornment: IAdornmentModel): adornment is IMedianAdornmentModel {
```

v3/src/components/graph/adornments/univariate-measures/univariate-measure-adornment-component.tsx
```diff
@@ -0,0 +1,233 @@
...
+  cellKey: Record<string, string>
+  containerId?: string
+  model: IMeanAdornmentModel | IMedianAdornmentModel
+  plotHeight: number
+  plotWidth: number
+  xAxis?: INumericAxisModel
+  yAxis?: INumericAxisModel
+}
+
+export const UnivariateMeasureAdornmentComponent = observer(function UnivariateMeasureAdornmentComponent (props: IProps) {
```
kswenson, line 39: Shorten this line somehow to avoid lint warning.

v3/src/components/graph/adornments/univariate-measures/median/median-adornment-model.ts
```diff
@@ -0,0 +1,25 @@
...
+import {
+  UnivariateMeasureAdornmentModel, IUnivariateMeasureAdornmentModel
+  } from "../univariate-measure-adornment-model"
+import { kMedianType } from "./median-adornment-types"
+import { IDataConfigurationModel } from "../../../../data-display/models/data-configuration-model"
+
+export const MedianAdornmentModel = UnivariateMeasureAdornmentModel
+  .named("MedianAdornmentModel")
+  .props({
+    type: "Median"
```
kswenson, line 12: I should have pointed this out earlier, but this should be
```suggestion
    type: types.optional(types.literal(kMedianType), kMedianType)
```
for all adornments (and any other models where types are used to discriminate between models). As written, the TypeScript type is `string` with a default value of `"Median"`, whereas `types.literal()` indicates that there is only one valid value, and `types.optional()` means that it will be provided automatically if not provided by the caller. Once all adornments are using `types.literal()` we should test whether the explicit dispatcher is still necessary.

v3/src/components/graph/adornments/univariate-measures/univariate-measure-adornment-model.ts
```diff
@@ -0,0 +1,92 @@
...
+      throw "type must be overridden"
+    }),
+  })
+  .views(self => ({
+    getCaseValues(attrId: string, casesInPlot: ICase[], dataConfig: IDataConfigurationModel) {
+      const dataset = dataConfig?.dataset
+      const caseValues: number[] = []
+      casesInPlot.forEach((c: ICase) => {
+        const caseValue = Number(dataset?.getValue(c.__id__, attrId))
+        if (Number.isInteger(caseValue)) {
```
kswenson, line 37:
```suggestion
        if (Number.isFinite(caseValue)) {
```

v3/src/components/graph/adornments/univariate-measures/univariate-measure-adornment-model.ts
```diff
@@ -0,0 +1,92 @@
...
+    }
+  }))
+
+export interface IMeasureInstance extends Instance<typeof MeasureInstance> {}
+export interface IUnivariateMeasureAdornmentModel extends Instance<typeof UnivariateMeasureAdornmentModel> {}
+export function isUnivariateMeasureAdornment(adornment: IAdornmentModel):
+  adornment is IUnivariateMeasureAdornmentModel {
+    return adornment.type === "UnivariateMeasureAdornmentModel" ||
+           adornment.type === "Mean" ||
+           adornment.type === "Median"
```
kswenson, line 91:
```suggestion
    return adornment.isUnivariateMeasure
```

v3/src/components/graph/adornments/adornments-store.ts
kswenson: > I had the impression we wanted to keep using the `measures` table for now. It'd be nice to drop it, but that'd mean there won't be any placeholder listings in the menu for the adornments that don't exist yet. Should we wait on that? Or do you think at this point that's not a concern?

At this point I'd do whatever involves less work, which I think is leaving it in for now.

kswenson (changes requested):
👍 Changes look good -- just a few more relatively minor changes.

kswenson (approved):
👍 Changes looks good!

PR #900, opened 2023-09-25
Formulas can evaluate nested `prev` calls (e.g. prev(fib, 1) + prev(prev(fib, 1), 1))

pjanik:
https://www.pivotaltracker.com/story/show/186090743

This PR adds support for nested `prev` calls. It seems to work well with the Fibonacci example, or even with a filter, e.g. `prev(prev(LifeSpan, -1, Habitat = "water"), -1, Habitat = "water")`.


kswenson (approved):
👍 Looks good

PR #901, opened 2023-09-25
fix: auto-scroll bugs

kswenson:
- improve document _does_ auto-scroll when trying to drag attribute to graph x-axis
- fix case table _doesn't_ auto-scroll when dragging attributes to create hierarchical table

The problem of the document auto-scrolling when trying to drag to the x-axis of the graph is improved (although not fixed entirely) by using existing `dnd-kit` configuration to reduce the size of the auto-scroll area. If we want to fine-tune the auto-scroll behavior based on the element being dragged over, for instance, we'll need to make more substantial modification to `dnd-kit` itself.

The case table bug is fixed by patching `@dnd-kit/core`. The dnd-kit auto-scroll code maintains a mapping between scrollable elements and their bounding rects. In the `useRects()` hook, this test was used to determine when to recompute the rectangles associated with the scrollable elements:

```typescript
  if (elements.length > 0 && rects === defaultValue$2) {
    measureRects();
  }
```

where `defaultValue$2` is the default empty array. This doesn't catch the situation in which the length of the elements array changes, which is the situation we were encountering. Changing the test to this fixes the bug:

```typescript
  if (elements.length !== rects.length) { // [CC]
    measureRects();
  }
```

For now, we're just using `patch-package` to deploy the fix, but I plan to submit a PR to the `dnd-kit` repo to contribute the fix back.

### DIFF
--- a/v3/patches/@dnd-kit+core+6.0.8.patch
+++ b/v3/patches/@dnd-kit+core+6.0.8.patch
@@ -1,0 +1,13 @@
+diff --git a/node_modules/@dnd-kit/core/dist/core.esm.js b/node_modules/@dnd-kit/core/dist/core.esm.js
+index 91671e9..2deb096 100644
+--- a/node_modules/@dnd-kit/core/dist/core.esm.js
++++ b/node_modules/@dnd-kit/core/dist/core.esm.js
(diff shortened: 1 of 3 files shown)


emcelroy (approved):
Looks good 👍

PR #902, opened 2023-09-26
Let `next` be nested within other `next` or `prev` calls

pjanik:
https://www.pivotaltracker.com/story/show/186115784

Semi-aggregate functions have made a return. Frankly, I'm struggling to reason about them and fit them into the rest of the formula's code. The problem was that `next(next(LifeSpan))` didn't work previously. Similar issues arose with other combinations, such as `prev(next(LifeSpan))` or more complex ones like `prev(next(LifeSpan, -1, Habitat = "water"), -2, Habitat = "both")`.

So, the solution I came up with is to treat `next` as the reverse of `prev`. However, since it requires a reverse iteration order, it iterates over all cases and stores results when the very first case is evaluated. Then, for each subsequent case, the cached result is returned.

There are also changes to the MathJS Scope class, allowing the casePointer to be movable (I used the term 'pointer' to avoid confusion with 'caseIndex,' which has a very specific meaning in formulas).

The drawback of this solution is that when next switches the order of iteration, nested prev with complex filters might not work as expected. But I believe this is a very rare use case, if not non-existent. I wonder how all of this worked in V2. I looked at the V2 code, but recreating the entire iteration process just from reading it was challenging.

I'm opening this PR anyway, as it still improves how `next` and `prev` are evaluated, and I don't want to get bogged down in this area for more days. Perhaps taking a break and focusing on other areas will help me come up with a better solution.

### DIFF
--- a/v3/src/models/data/formula-fn-registry.ts
+++ b/v3/src/models/data/formula-fn-registry.ts
@@ -51,10 +51,7 @@ export const isValueTruthy = (value: any) => value !== "" && value !== false && 
 export const prevOrNextFactory = (variant: "next" | "prev") =>
   (args: MathNode[], mathjs: any, scope: FormulaMathJsScope) => {
   interface ICachedData {
-    currentIndex: number
-    resultIndex: number
-    expressionValues: FValue[]
-    filterValues: FValue[]
+    result?: FValue
   }
 
   const cacheKey = `${variant}(${args.toString()})-${scope.getSemiAggregateGroupId()}`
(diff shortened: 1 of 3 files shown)


v3/src/models/data/formula-fn-registry.ts
```diff
@@ -44,6 +44,50 @@ const aggregateFnWithFilterFactory = (fn: (values: number[]) => number) => {
...
+  const cacheKey = `${variant}(${args.toString()})-${scope.getSemiAggregateGroupId()}`
+  const [ expression, defaultValue, filter ] = args
+  const cachedData = scope.getCached(cacheKey) as ICachedData | undefined
+  let result
+
+  if (cachedData !== undefined) {
+    // In case we don't find a new result, we need to reuse the old one.
+    result = cachedData.result
+    let newExpressionValue, newFilterValue
+    // This block will resolve attribute names to previous case values.
```
kswenson, line 66:
```suggestion
    // This block will resolve attribute names to previous/next case values.
```

pjanik:
Well, this is most likely obsolete. I believe I have finally found a proper approach, but I'll open a new PR tomorrow after I clean up the code a bit. It might still be worth a brief look, as some of the changes from this PR are still present in the new one.

kswenson (approved):
👍 LGTM

v3/src/models/data/formula-fn-registry.ts
pjanik: This code is gone in the newest version, so I'm not committing this change to avoid merge conflicts.

PR #904, opened 2023-09-26
fix: Ensure adornment appears after attribute changes

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/181940090

This fixes a bug where starting from a configuration where the x attribute type is numeric and the y attribute type is categorical, to a configuration where the x attribute is numeric and the y attribute is numeric, and then switching back to the previous configuration, the adornment's line would not appear in the plot even if its checkbox was checked and a value provided.

Also included are fixes to correct some styling issues, and ensure that a formula value will use the correct case data when there are multiple subplots.

### DIFF
--- a/v3/src/components/graph/adornments/plotted-value/plotted-value-model.ts
+++ b/v3/src/components/graph/adornments/plotted-value/plotted-value-model.ts
@@ -23,7 +23,7 @@ export const PlottedValueModel = AdornmentModel
       const caseValues: number[] = []
       casesInPlot.forEach((c: ICase) => {
         const caseValue = Number(dataset?.getValue(c.__id__, attrId))
-        if (Number.isInteger(caseValue)) {
+        if (Number.isFinite(caseValue)) {
           caseValues.push(caseValue)
         }
       })
--- a/v3/src/components/graph/adornments/plotted-value/plotted-value-model.ts
+++ b/v3/src/components/graph/adornments/plotted-value/plotted-value-model.ts
@@ -1,7 +1,9 @@
 import { Instance, types } from "mobx-state-tree"
 import { mean } from "mathjs"
-import { AdornmentModel, IAdornmentModel, IUpdateCategoriesOptions } from "../adornment-models"
+import { AdornmentModel, IAdornmentModel } from "../adornment-models"
 import { kPlottedValueType } from "./plotted-value-types"
+import { ICase } from "../../../../models/data/data-set-types"
+import { IDataConfigurationModel } from "../../../data-display/models/data-configuration-model"
 
 export const PlottedValueModel = AdornmentModel
   .named("PlottedValueModel")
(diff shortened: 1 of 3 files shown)


v3/src/components/graph/adornments/plotted-value/plotted-value-model.ts
```diff
@@ -14,19 +16,25 @@ export const PlottedValueModel = AdornmentModel
       // This is just a proof-of-concept placeholder for function strings. It will always return
       // the mean of caseValues no matter what the string value is.
       return mean(caseValues)
+    },
+    getCaseValues(attrId: string, cellKey: Record<string, string>, dataConfig: IDataConfigurationModel) {
```
emcelroy, line 20: Just to note: I added a [very similar view to `UnivariateMeasureAdornmentModel`](https://github.com/concord-consortium/codap/blob/5afdf315fc5db55178261fff23c125f4b02b7a40/v3/src/components/graph/adornments/univariate-measures/univariate-measure-adornment-model.ts#L32-L42) in the the Mean and Median adornments PR. I _think_ it'd make sense to move this to the main `AdornmentModel` and just have a single instance of the view there, but I'm going to try doing that as part of the work on the Standard Deviation adornment.

v3/src/components/graph/adornments/plotted-value/plotted-value-model.ts
```diff
@@ -14,19 +16,25 @@ export const PlottedValueModel = AdornmentModel
...
+      const dataset = dataConfig?.dataset
+      const casesInPlot = dataConfig.subPlotCases(cellKey)
+      const caseValues: number[] = []
+      casesInPlot.forEach((c: ICase) => {
+        const caseValue = Number(dataset?.getValue(c.__id__, attrId))
+        if (Number.isInteger(caseValue)) {
```
kswenson, line 26: Why does it have to be integer? Seems like `isFinite()` would be preferable.

v3/src/components/graph/adornments/plotted-value/plotted-value.tsx
```diff
@@ -145,7 +137,7 @@ export const PlottedValue = observer(function PlottedValue (props: IProps) {
   return (
     <>
       <div className="plotted-value-container" id={`plotted-value-${containerId}`}>
-        { model.value &&
+        { value && model.isVisible &&
```
kswenson, line 140:
```suggestion
        { value != null && model.isVisible &&
```
Otherwise, zero values won't plot.

kswenson (approved):
👍 Looks good -- just a couple of suggestions that shouldn't require further review once addressed.

v3/src/components/graph/adornments/plotted-value/plotted-value.tsx
emcelroy: Right. I'll have to do a better job of remembering that.

PR #905, opened 2023-09-27
Reimplement semi-aggregate functions + refactor Formula MathJS scope and manager

pjanik:
https://www.pivotaltracker.com/story/show/186115784

Yet another approach to semi-aggregates. I hope I've found a proper approach (which is also much simpler) that maintains an O(n) time complexity without the need for reverse iteration for `next`. So, we can avoid the potential O(n^2) issues that could happen in V2 (if I've correctly interpreted its code). The only trade-off is the lack of self-reference in `next()` function that we discussed before. But it works seamlessly for `prev()`, allowing us to implement the Fibonacci sequence or running sum.

This PR also includes some refactoring and simplification of the formula-related code:
- I've added or cleaned up `within...` methods in the mathjs custom scope and explained in comments why I use this approach instead of the more elegant sub-scopes. 
- I've removed the aggregate suffix (`_ALL`) from symbol names in the canonical form and replaced that with `withAggregateContext` scope method. Since I had to add `within...` method(s) anyway, this was making more sense now and simplified the scope code too.
- I've removed `casePointerModifier` and simply let the functions move the pointer as they want. This makes the scope simpler.
- Other smaller changes like changing names or moving some logic to places that are probably better suited for it.

I've tested `prev` and `next` functions using the [mammals dataset](https://codap.concord.org/app/static/dg/en/cert/index.html#file=examples:Mammals) in V2 and V3, and for each formula, I've checked both the flat table and the hierarchical one (with the "Diet" attribute moved to a new top-level collection). Results were identical for each of the formulas listed:

```
next(LifeSpan, -1)
prev(LifeSpan, -1)
next(LifeSpan, -1, Habitat = "water") 
prev(LifeSpan, -1, Habitat = "water")
prev(Height, 1) + prev(prev(Height, 1), 1) + prev(prev(prev(Height, 1), 1), 1)
prev(next(LifeSpan))
next(next(LifeSpan))
next(next(LifeSpan, -2, Habitat = "water"), -1, Habitat = "water")
next(prev(LifeSpan, -2, Habitat = "water"), -1, Habitat = "water")
prev(next(LifeSpan, -2, Habitat = "water"), -1, Habitat = "water")
next(prev(next(LifeSpan, -3, Habitat = "water"), -2, Habitat = "water"), -1, Habitat = "both")
prev(caseIndex)
next(caseIndex)
prev(prev(caseIndex))
next(next(caseIndex))
next(prev(caseIndex))
prev(next(caseIndex))
prev(LifeSpan) + mean(LifeSpan)
prev(newAttr2) + mean(LifeSpan) (in the top-level collection with Diet and newAttr2)
```

There are some scenarios where V2 and V3 would produce different results. But I think they either don’t make much sense, or v3 behavior seems to be better or logical:

1. Aggregate within next or prev: prev(mean(LifeSpan))
a. V2 returns 70, 70, 70…
b. V3 returns the actual mean of LifeSpan: 24.8, 24.8, 24.8…
2. prev/next referencing child attribute: prev(LifeSpan)  in the top-level collection with Diet
a. V2 fails quietly / renders blank fields
b. V3 returns the first value from the previous group: 70, 19
3. prev/next referencing child attribute within the aggregate function: prev(mean(LifeSpan)) in the top-level collection with Diet
a. V2 fails quietly / renders blank fields
b. V3 returns the mean of the previous group, which seems to make sense (same as combining prev and mean calls in two separate columns): 35, 21.8

### DIFF
--- a/v3/src/models/data/formula-fn-registry.ts
+++ b/v3/src/models/data/formula-fn-registry.ts
@@ -16,7 +16,7 @@ const evaluateNode = (node: MathNode, scope?: FormulaMathJsScope) => {
 const cachedAggregateFnFactory =
 (fnName: string, fn: (args: MathNode[], mathjs: any, scope: FormulaMathJsScope) => FValue | FValue[]) => {
   return (args: MathNode[], mathjs: any, scope: FormulaMathJsScope) => {
-    const cacheKey = `${fnName}(${args.toString()})-${scope.getCaseGroupId()}`
+    const cacheKey = `${fnName}(${args.toString()})-${scope.getCaseAggregateGroupId()}`
     const cachedValue = scope.getCached(cacheKey)
     if (cachedValue !== undefined) {
       return cachedValue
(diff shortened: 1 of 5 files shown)


kswenson (approved):
👍 Looks good -- cleaned up nicely!

PR #907, opened 2023-09-27
Fix count() caching

pjanik:
https://www.pivotaltracker.com/story/show/186033205

This pull request fixes formulas that included a `count()` call without an argument. It was incorrectly cached and resulted in the wrong result in most cases. This is an atypical scenario, as `count()` doesn't have any arguments from the child collection, yet it should use grouping logic / behavior / IDs, as it did. Manual caching seemed to be the easiest solution.


kswenson (approved):
👍 Looks good

PR #908, opened 2023-09-28
feat: more case table actions support undo/redo -- hide/show and attribute move/hierarchy changes

kswenson:
Should be undoable as a single action with appropriate undo/redo strings:
- hiding attribute(s)
- showing attribute(s)
- moving attributes within a collection
- moving attributes to create a new collection
- moving attributes to remove a collection

### DIFF
--- a/v3/src/components/case-table/case-table.tsx
+++ b/v3/src/components/case-table/case-table.tsx
@@ -2,7 +2,7 @@ import { useDndContext } from "@dnd-kit/core"
 import { observer } from "mobx-react-lite"
 import React, { CSSProperties, useCallback, useEffect, useRef } from "react"
 import { AttributeDragOverlay } from "../drag-drop/attribute-drag-overlay"
-import { kChildMostTableCollectionId, kIndexColumnKey } from "./case-table-types"
+import { kIndexColumnKey } from "./case-table-types"
 import { CollectionTable } from "./collection-table"
 import { useCaseTableModel } from "./use-case-table-model"
 import { useSyncScrolling } from "./use-sync-scrolling"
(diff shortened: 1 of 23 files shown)


v3/src/components/case-table/case-table.tsx
```diff
@@ -85,11 +87,11 @@ export const CaseTable = observer(function CaseTable({ setNodeRef }: IProps) {
...
           <div className="case-table-content" onScroll={handleHorizontalScroll}>
             {collections.map((collection, i) => {
-              const key = collection?.id || kChildMostTableCollectionId
+              const key = collection.id
               const parent = i > 0 ? collections[i - 1] : undefined
               return (
-                <ParentCollectionContext.Provider key={key} value={parent}>
-                  <CollectionContext.Provider key={key} value={collection}>
+                <ParentCollectionContext.Provider key={key} value={parent?.id}>
+                  <CollectionContext.Provider key={key} value={collection.id}>
```
emcelroy, line 94: Is it right that both contexts have the same key?

v3/src/test/setupTests.ts
```diff
@@ -17,3 +18,50 @@ declare global {
...
 global.assertIsDefined = assertIsDefined
 global.jestSpyConsole = jestSpyConsole
+
+let message = () => ""
+
+function removeIdsReplacer(key: string, value: any) {
+  return key === "id" ? undefined : value
+}
+
+// for use with lodash's isEqualWith function to implement toEqualWithUniqueIds custom jest matcher
```
emcelroy, line 28: Should `toEqualWithUniqueIds` in this comment be `toEqualWithoutIds`?

v3/src/test/setupTests.ts
```diff
@@ -17,3 +18,50 @@ declare global {
...
+}
+
+// custom jest matcher for testing that copied objects are identical to the originals
+// except for the object ids which should all be different.
+expect.extend({
+  toEqualWithoutIds(received: any, expected: any) {
+    message = () => ""
+    const pass = isEqualWith(received, expected, withoutIdsCustomizer)
+    if (!pass && !message()) {
+      message = () => "values differed outside of geometry objects"
```
emcelroy, line 55: What does "geometry objects" mean here? I think I understand the general meaning of the message -- that there was a mismatch of something outside of the checks in withoutIdsCustomizer -- but not the "geometry objects" part. I'm probably just ignorant of the term, sorry.

emcelroy (approved):
Looks good 👍 I asked questions about a few details, but I'm pretty sure that's mostly because I'm unaware of some things and need to learn.

One issue I noticed while trying the changes out in a browser is that after moving an attribute within a collection, you can undo that move, but you can't seem to redo it. Is that the expected behavior? Redo after undo seems to work for all the other actions. FWIW, I see an `/undoLastAction` being added to the undo history when I undo an attribute move. I don't see anything like that happening when I undo the other actions.

v3/src/test/setupTests.ts
kswenson: Holdover from CLUE -- I'll adjust the message.

v3/src/components/case-table/case-table.tsx
kswenson: It seems that the outer `Provider` needs a key but the inner one does not, so we can just eliminate the second and then we don't have to answer that question. 😉

PR #909, opened 2023-09-29
fix: Dot plots can show mean and median (PT-181914405)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/181914405

Various bug fixes:

1.  Prevent attempts to add adornments to subplots that don't contain case data
2. Ensure adornments adjust their height correctly when the number of subplots change
3. Make adornments respond to rescaling of axes while axes are being rescaled, not just after
4. Make adornments react to case data changes

NOTE: Fix number 2 was also applicable to the Plotted Value adornment, and fix number 3 was applicable to Plotted Value and Movable Value. Since they were relatively simple, I included the changes for those adornments, too.

### DIFF
--- a/v3/src/components/graph/adornments/movable-value/movable-value.tsx
+++ b/v3/src/components/graph/adornments/movable-value/movable-value.tsx
@@ -202,7 +202,7 @@ export const MovableValue = observer(function MovableValue (props: IProps) {
       adjustAllValues()
       renderFills()
     }, { name: "MovableValue.refreshAxisChange" })
-  }, [adjustAllValues, dataConfig, renderFills, xAxis.max, xAxis.min, yAxis.max, yAxis.min])
+  }, [adjustAllValues, dataConfig, renderFills, xAxis.domain, yAxis.domain])
 
   // Make the movable values and their cover segments
   useEffect(function createElements() {
--- a/v3/src/components/graph/adornments/movable-value/movable-value.tsx
+++ b/v3/src/components/graph/adornments/movable-value/movable-value.tsx
@@ -201,10 +201,8 @@ export const MovableValue = observer(function MovableValue (props: IProps) {
       // We observe changes to the axis domains within the autorun by extracting them from the axes below.
       // We do this instead of including domains in the useEffect dependency array to prevent domain changes
       // from triggering a reinstall of the autorun.
-      // eslint-disable-next-line @typescript-eslint/no-unused-vars
-      const { domain: xDomain } = xAxis
-      // eslint-disable-next-line @typescript-eslint/no-unused-vars
-      const { domain: yDomain } = yAxis
+      const { domain: xDomain } = xAxis // eslint-disable-line @typescript-eslint/no-unused-vars
+      const { domain: yDomain } = yAxis // eslint-disable-line @typescript-eslint/no-unused-vars
       isVertical.current = dataConfig?.attributeType("x") === "numeric"
       adjustAllValues()
       renderFills()
--- a/v3/src/components/graph/adornments/movable-value/movable-value.tsx
+++ b/v3/src/components/graph/adornments/movable-value/movable-value.tsx
@@ -198,7 +198,12 @@ export const MovableValue = observer(function MovableValue (props: IProps) {
   // Refresh the value when the axis changes
   useEffect(function refreshAxisChange() {
     return autorun(() => {
+      // We observe changes to the axis domains within the autorun by extracting them from the axes below.
+      // We do this instead of including domains in the useEffect dependency array to prevent domain changes
+      // from triggering a reinstall of the autorun.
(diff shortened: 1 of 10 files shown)


v3/src/components/graph/adornments/univariate-measures/univariate-measure-adornment-model.ts
```diff
@@ -74,10 +81,18 @@ export const UnivariateMeasureAdornmentModel = AdornmentModel
...
       for (let i = 0; i < totalCount; ++i) {
         const cellKey = self.setCellKey(options, i)
-        const instanceKey = self.instanceKey(cellKey)
+        // If there are no cases in the cell, do not add an adornment
+        if (dataConfig.subPlotCases(cellKey).length === 0) continue
+
+        const instanceKey = self.instanceKey(cellKey) 
         const value = Number(self.getMeasureValue(attrId, cellKey, dataConfig))
+        // If the value is not a number, do not add an adornment
+        if (!Number.isFinite(value)) continue
```
kswenson, line 90: I suspect it will work better to go ahead and add the measure with a value of `NaN` and then deal with that possibility at render time (e.g. by not rendering). Otherwise, you have to listen for every possible circumstance that can change the value and be prepared to add/remove measures dynamically as the value changes. For instance, imagine if there were an animating slider controlling a filter that affected the number of cases -- then potentially you'd have to add/remove measures with each tick of the slider. If you always add the measures, however, then you just need to change the (volatile) values as the slider animates. This will also have the effect of preserving user-positioned labels as values appear/disappear.

kswenson (approved):
👍 Generally looks good -- the changes to `updateAfterSharedModelChanges()` and the introduction of `updateMeasureValue()` you asked about on slack both seem fine. I left a couple of suggestions:
- support measures with `NaN` values rather than trying to add/remove them
- access axis domains inside `autorun`s so they trigger the `autorun` rather than reinstalling it via `useEffect` dependencies.

v3/src/components/graph/adornments/movable-value/movable-value.tsx
```diff
@@ -198,11 +198,18 @@ export const MovableValue = observer(function MovableValue (props: IProps) {
   // Refresh the value when the axis changes
   useEffect(function refreshAxisChange() {
     return autorun(() => {
+      // We observe changes to the axis domains within the autorun by extracting them from the axes below.
+      // We do this instead of including domains in the useEffect dependency array to prevent domain changes
+      // from triggering a reinstall of the autorun.
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      const { domain: xDomain } = xAxis
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      const { domain: yDomain } = yAxis
```
kswenson, line 207: FYI, you can also use:
```suggestion
      const { domain: xDomain } = xAxis // eslint-disable-line @typescript-eslint/no-unused-vars
      const { domain: yDomain } = yAxis // eslint-disable-line @typescript-eslint/no-unused-vars
```

v3/src/components/graph/adornments/univariate-measures/univariate-measure-adornment-model.ts
```diff
@@ -71,13 +78,18 @@ export const UnivariateMeasureAdornmentModel = AdornmentModel
       const columnCount = topCatCount * xCatCount
       const rowCount = rightCatCount * yCatCount
       const totalCount = rowCount * columnCount
-      const attrId = xAttrId ? xAttrId : yAttrId
+      const attrId = xAttrId && dataConfig.attributeType("x") ? xAttrId : yAttrId
       for (let i = 0; i < totalCount; ++i) {
         const cellKey = self.setCellKey(options, i)
-        const instanceKey = self.instanceKey(cellKey)
+        // If there are no cases in the cell, do not add an adornment
+        if (dataConfig.subPlotCases(cellKey).length === 0) continue
```
kswenson, line 85: As mentioned previously, it may be better to add the adornment in this case and just not render it if there are no cases, but I'm also fine with leaving this as is and crossing that bridge if/when we come to it.

kswenson (approved):
👍 Looks good

v3/src/components/graph/adornments/movable-value/movable-value.tsx
emcelroy: That's a bit cleaner. Thanks 👍

v3/src/components/graph/adornments/univariate-measures/univariate-measure-adornment-model.ts
emcelroy: Sorry -- I removed the `if (!Number.isFinite(value)) continue` line previously, but for some reason didn't think to remove this one, too. I've removed it in the latest commit.

PR #910, opened 2023-09-29
Implement `canonicalToDisplay` helper + unit tests

pjanik:
https://www.pivotaltracker.com/story/show/186144182

This PR adds the `canonicalToDisplay` helper along with quite a few unit tests to cover it and demonstrate how it works. This function will be quite important, so a separate PR might help identify any potential issues.

I've given some thought to how to maintain the whitespace, and this is the simplest and cleanest idea I've come up with so far. I couldn't find a V2 algorithm that performs the same task, and the CLUE version is not very useful, as it only deals with the 'x' symbol.


kswenson (approved):
👍 Looks good

PR #911, opened 2023-09-29
Fix handling of scenario when attribute name is the same as function name (e.g. "mean(mean)")

pjanik:
https://www.pivotaltracker.com/story/show/186152786

`mean(mean)` is a valid formula, but it used to crash in V3. I realized this would be a problem when I added the `isNonFunctionSymbolNode` helper as part of the previous PR. This is all due to how MathJS parses expressions (function node + separate function name node).

I also renamed `canonicalizeExpression` to `displayToCanonical`, so it's the inverse of `canonicalToDisplay`.

Additionally, I've added quite a few unit tests that also cover the scenario that caused a real-life bug discussed above and in the PT story.

### DIFF
--- a/v3/src/models/data/formula-manager.ts
+++ b/v3/src/models/data/formula-manager.ts
@@ -332,7 +332,7 @@ export class FormulaManager {
     })
   }
 
-  getDisplayNameMapForFormula(formulaId: string) {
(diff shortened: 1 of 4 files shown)


kswenson (approved):
👍 Looks good

PR #913, opened 2023-10-02
Update formula's display form when attribute name is changed

pjanik:
https://www.pivotaltracker.com/story/show/186144182

This PR is a continuation of https://github.com/concord-consortium/codap/pull/910. The formula model and a few other helpers are updated, and a new reaction acts as an observer of the formula attribute name change.


kswenson (approved):
👍 Looks good!

PR #915, opened 2023-10-03
fix: Improve adornments' undo/redo behavior (PT-186135050)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/186135050

Still a draft. There are a couple small issues to work out.


v3/src/components/graph/adornments/count/count-adornment-model.ts
```diff
@@ -26,17 +27,23 @@ export const CountModel = AdornmentModel
...
     },
     setPercentType(percentType: string) {
       self.percentType = percentType
     }
   }))
 
-export interface ICountModel extends Instance<typeof CountModel> {}
-export function isCount(adornment: IAdornmentModel): adornment is ICountModel {
+export interface ICountAdornmentModel extends Instance<typeof CountAdornmentModel> {}
+export function isCount(adornment: IAdornmentModel): adornment is ICountAdornmentModel {
```
kswenson, line 47:
```suggestion
export function isCountAdornment(adornment: IAdornmentModel): adornment is ICountAdornmentModel {
```

v3/src/components/graph/adornments/univariate-measures/univariate-measure-adornment-model.ts
```diff
@@ -5,11 +5,9 @@ import { IDataConfigurationModel } from "../../../data-display/models/data-confi
 import { ICase } from "../../../../models/data/data-set-types"
 
 export const MeasureInstance = types.model("MeasureInstance", {
-  labelCoords: types.maybe(PointModel)
+  labelCoords: types.maybe(PointModel),
+  value: types.optional(types.number, 0)
```
kswenson, line 9: We've been here before, but do we need to store the value persistently? I would have thought that it could be computed when needed but apparently there's something here I don't understand.

kswenson (approved):
👍 Looks good -- aside from some minor stuff, the one big question is about whether we should store adornment values persistently.

v3/src/components/graph/adornments/univariate-measures/univariate-measure-adornment-model.ts
emcelroy: @kswenson With the previous configuration where value is volatile and set to 0 by default, if you add the Mean or Median adornment, then undo adding it, then redo adding it, the adornment returns with a value of 0.  It seems when the action is redone, `adornmentStore`'s`addAdornment` action isn't triggered, which means the adornment model's `updateCategories` isn't being called. And so the instance of the adornment doesn't get a real value set.

There's probably a way to have the volatile value properly set when the adornment returns, but it's not obvious to me. I'll see if I can work something out, though.

v3/src/components/graph/adornments/movable-value/movable-value-adornment-model.ts
```diff
@@ -59,7 +59,7 @@ export const MovableValueModel = AdornmentModel
     }
   }))
   .actions(self => ({
-    addValue(aValue?: number) {
+    addValue(aValue?: number, init=false) {
```
kswenson, line 62: Unused argument `init`?

v3/src/components/graph/adornments/univariate-measures/univariate-measure-adornment-component.tsx
```diff
@@ -93,9 +93,12 @@ export const UnivariateMeasureAdornmentComponent = observer(
     }, [instanceKey, model])
 
     const addLineCoverAndLabel = useCallback((
-      valueObj: IValueObject, labelObj: ILabelObject, adornment: IMeasureInstance
+      valueObj: IValueObject, labelObj: ILabelObject, measure: IMeasureInstance
     ) => {
-      const value = adornment.value
+      const attrId = dataConfig?.primaryAttributeID
+      const value = attrId ? model.measureValue(attrId, cellKey, dataConfig) : undefined
+      if (!value && value !== 0) return
```
kswenson, line 100: I'm not sure what we're testing for here -- what's the significance of a falsy value other than `0`?

kswenson (approved):
👍 looks good -- just a couple code clarity questions

v3/src/components/graph/adornments/movable-value/movable-value-adornment-model.ts
emcelroy: Yeah, no longer needed. I've removed it.

v3/src/components/graph/adornments/univariate-measures/univariate-measure-adornment-component.tsx
emcelroy: Since `value` could possibly be `undefined` (see line 99), we need to check for that, and if it is undefined and not 0, return before continuing on to render the line. Does that make sense? It does to me but I could be missing something.

emcelroy: Yeah, those are better. Thanks 👍

PR #917, opened 2023-10-05
fix: custom undo/redo implementation

kswenson:
[Re-]fixes the undo/redo-ability of moving attributes in the case table.

I'm pretty sure this was working properly when I first implemented it, but something must have changed since then. In any case, we need to make sure that changes that occur during undo/redo are not treated as newly undoable actions.


bfinzer (approved):
👍🏻LGTM

PR #919, opened 2023-10-05
Fix double quote string escaping, refactor char escaping related code, add/modify unit tests to include escaped chars

pjanik:
https://www.pivotaltracker.com/story/show/186145125

PR changes:
- Fix double quote string escaping.
- Refactor the code for character escaping. I've renamed variables and reorganized function calls to make it easier to understand why certain things are being called in specific places. It will still be challenging to grasp without delving deep into the context, but at least now, it makes more sense to me. Previously, I would occasionally get lost (and I'm still deeply immersed in this context :wink:).
- Add/modify unit tests of formula utils to include escaped characters.

Also, enclosing a doc with all the manual test cases: [CODAP V3 Formulas: Special Character Test Cases](https://docs.google.com/document/d/1w6-kSAMD-84EEo7N8McN-7EdA8y7jsiH2RhH5_Q3NMU/edit?usp=sharing)

### DIFF
--- a/v3/src/models/data/formula-manager.ts
+++ b/v3/src/models/data/formula-manager.ts
@@ -347,23 +347,24 @@ export class FormulaManager {
       dataSet: {}
     }
 
-    const mapAttributeNames = (dataSet: IDataSet, prefix: string) => {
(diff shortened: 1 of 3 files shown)


kswenson (approved):
👍 Looks good

PR #921, opened 2023-10-06
Fix import of V2 doc with formulas

pjanik:
This PR fixes a regression caught by Tejal:
https://www.pivotaltracker.com/story/show/186002016/comments/238700252

Now, the manager updates the canonical form when it detects the difference between the old and a new display form. This will handle both manual updates and updates via import / snapshots, etc.


kswenson (approved):
👍 Looks good -- I worry a little about updating _all_ formulas when _any_ formula changes, but I guess we'll find out if it creates any problems.

pjanik:
@kswenson,

>  I worry a little about updating all formulas when any formula changes, but I guess we'll find out if it creates any problems.

I don't think it currently happens. We generally recalculate only formulas that have been changed.

Sometimes we recalculate display expressions (canonical -> display) of all the formulas, e.g. when the attribute is renamed, but I doubt we need to worry about performance there.

PR #922, opened 2023-10-06
Add VSCode launch.json config

pjanik:
This is similar to what Boris showed during the last dev day.

There's no checked-in launch.json, and we don't ignore it either. I don't think this would conflict with anyone's preferences or customizations, so it's probably nice to have it checked in and keep improving it as a group.

I'm also happy to keep it in my own user settings.json if we think it's better, as this particular configuration is very generic anyway. Not sure if there's any conclusion on what to do about launch.json.

### DIFF
--- a/v3/.vscode/launch.json
+++ b/v3/.vscode/launch.json
@@ -1,0 +1,14 @@
+{
+  "version": "1.0.0",
+  "configurations": [
+    {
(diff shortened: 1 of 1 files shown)


kswenson (approved):
👍 I'm fine with committing it.

PR #923, opened 2023-10-06
chore: Group Plotted Value with Univariate Measures (PT-186184994)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/186184994

Groups the Plotted Value graph adornment with Mean and Median so that it uses the same base component and model used by the latter.

Note that the following changes to the Plotted Value's appearance and behavior have been approved:

- The placement of the Plotted Value's tool tip that appears on hover is at the same level that the Mean and Median's tool tips appear. (This differs from V2 where the Plotted Value's tool tip appears closer to the top of the line.)

- The Plotted Value now supports the Show Measure Labels feature that's supported by Mean and Median.

- The text that appears before the numeric value in the Plotted Value's tool tip and label will be `value=` (Similar to how the Mean's value is preceded by `mean=`, for example).


kswenson (commented):
👍 Mostly this looks good. I'm not sure how to handle the strings changes. How was this handled in v2?

emcelroy:
> 👍 Mostly this looks good. I'm not sure how to handle the strings changes. How was this handled in v2?

@kswenson In v2 the Plotted Value's text label value is just the numeric value, there is no `value=` in front of it. I don't think you were present at the meeting, but I discussed the differences between the Plotted Value and Mean and Median with @bfinzer, which ones were important to keep, which ones might be open to change. One of the changes Bill thought should be made was to have `value=` before the number in the label.

kswenson (approved):
>discussed the differences between the Plotted Value and Mean and Median with @bfinzer, which ones were important to keep, which ones might be open to change. One of the changes Bill thought should be made was to have `value=` before the number in the label.

👍 In that case, LGTM

PR #924, opened 2023-10-06
Fix count function

pjanik:
https://www.pivotaltracker.com/story/show/186185032

This PR fixes `count()` function so it checks if the expression value is truthy instead of just checking if it's not empty.


kswenson (approved):
👍 Looks good

PR #925, opened 2023-10-07
Don't consider blank cells in aggregate functions, fix equal function for arrays

pjanik:
https://www.pivotaltracker.com/story/show/186145218

The aggregate function should filter out empty strings as MathJS resolves them to 0, while CODAP v2 ignores them.

Also, I've fixed the equal operator when arrays are compared (or when an array is compared with a single value). This happens when we deal with aggregate functions. I noticed it when I tested `mean(Mass = 80)` both in V2 and V3. Now it should work as expected.

### DIFF
--- a/v3/src/models/data/formula-fn-registry.ts
+++ b/v3/src/models/data/formula-fn-registry.ts
@@ -43,25 +43,35 @@ const aggregateFnWithFilterFactory = (fn: (values: number[]) => number) => {
     let expressionValues = evaluateNode(expression, scope)
     if (filter) {
       const filterValues = evaluateNode(filter, scope)
-      expressionValues = expressionValues.filter((v: any, i: number) => !!filterValues[i])
+      expressionValues = expressionValues.filter((v: any, i: number) =>
+        // Non-truthy expression values should not be part of the aggregate function result following V2 behavior.
+        // E.g. empty cells should not be counted in mean() function.
+        isValueNonEmpty(v) && isValueTruthy(filterValues[i])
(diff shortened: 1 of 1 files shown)


v3/src/models/data/formula-fn-registry.ts
```diff
@@ -43,25 +43,35 @@ const aggregateFnWithFilterFactory = (fn: (values: number[]) => number) => {
...
       const filterValues = evaluateNode(filter, scope)
-      expressionValues = expressionValues.filter((v: any, i: number) => !!filterValues[i])
+      expressionValues = expressionValues.filter((v: any, i: number) =>
+        // Non-truthy expression values should not be part of the aggregate function result following V2 behavior.
+        // E.g. empty cells should not be counted in mean() function.
+        isValueNonEmpty(v) && isValueTruthy(filterValues[i])
+      )
+    } else {
+      // Non-truthy expression values should not be part of the aggregate function result following V2 behavior.
+      // E.g. empty cells should not be counted in mean() function.
```
kswenson, line 53: Seems like this comment could be reduced to:
```suggestion
      // Empty cells should not be included in aggregate functions.
```

kswenson (approved):
👍 Looks good -- one suggested tweak to a comment.

v3/src/models/data/formula-fn-registry.ts
pjanik: Thanks, I've committed this one and replaced another copy of this comment.

PR #927, opened 2023-10-09
Fix a bug that was causing an error when attr with formula was deleted

pjanik:
https://www.pivotaltracker.com/story/show/186201130

This PR fixes a bug described in the linked PT story. Also, it fixes MST warnings when an attribute with an empty formula is deleted.

Kirk, I remember that during one of the recent arch meetings, you mentioned some issues with `isAlive`, but I can't remember exactly what that was. It'd be enough to use `isAlive` in this PR, but I've added it as a last resort to be more explicit (and potentially avoid the issues you described?).

### DIFF
--- a/v3/src/models/data/formula-manager.ts
+++ b/v3/src/models/data/formula-manager.ts
@@ -290,6 +290,7 @@ export class FormulaManager {
       // In 99% of cases, the two `if` statements above will be sufficient to detect deleted formulas. However, this
       // could serve as the ultimate check in case a formula instance was deleted in a different manner.
       if (!isAlive(metadata.formula)) {
+        console.warn("FormulaManager.unregisterDeletedFormulas unregistering a defunct formula that was not detected by the usual means.")
         this.unregisterFormula(formulaId)
       }
     })
--- a/v3/src/models/data/formula-manager.ts
+++ b/v3/src/models/data/formula-manager.ts
@@ -283,11 +283,7 @@ export class FormulaManager {
     this.formulaMetadata.forEach((metadata, formulaId) => {
       const { dataSetId, attributeId } = this.getFormulaMetadata(formulaId)
       const dataSet = this.dataSets.get(dataSetId)
-      if (!dataSet) {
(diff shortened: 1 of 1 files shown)


v3/src/models/data/formula-manager.ts
```diff
@@ -270,7 +272,31 @@ export class FormulaManager {
...
+      const { dataSetId, attributeId } = this.getFormulaMetadata(formulaId)
+      const dataSet = this.dataSets.get(dataSetId)
+      if (!dataSet) {
+        this.unregisterFormula(formulaId)
+        return
+      }
+      if (!dataSet.attrFromID(attributeId)) {
+        this.unregisterFormula(formulaId)
+        return
+      }
```
kswenson, line 293: Seems like this could be combined to:
```suggestion
      if (!dataSet?.attrFromID(attributeId)) {
        this.unregisterFormula(formulaId)
        return
      }
```

v3/src/models/data/formula-manager.ts
```diff
@@ -270,7 +272,31 @@ export class FormulaManager {
...
+      // In 99% of cases, the two `if` statements above will be sufficient to detect deleted formulas. However, this
+      // could serve as the ultimate check in case a formula instance was deleted in a different manner.
+      if (!isAlive(metadata.formula)) {
+        this.unregisterFormula(formulaId)
+      }
```
kswenson, line 298: >I remember that during one of the recent arch meetings, you mentioned some issues with `isAlive`, but I can't remember exactly what that was.

It was just a recommendation to not use it as a primary control flow mechanism but rather as a means of detecting that something isn't being handled optimally, as you are doing here. Toward that end, I'd suggest adding a `console.warn`, something like:
```suggestion
      if (!isAlive(metadata.formula)) {
        console.warn("FormulaManager.unregisterDeletedFormulas unregistering a defunct formula that was not detected by the usual means.")
        this.unregisterFormula(formulaId)
      }
```

kswenson (approved):
👍 Looks good

PR #928, opened 2023-10-10
Recalculate formula when dependant cell is set to "" 

pjanik:
https://www.pivotaltracker.com/story/show/186200846

This PR fixes a bug described in the PT story linked above. `if (c[attrId])` was not executed when `c[attrId]` was equal to `""`. I've refactored some helpers a bit and added unit tests to cover this use case.

### DIFF
--- a/v3/src/models/data/formula-utils.test.ts
+++ b/v3/src/models/data/formula-utils.test.ts
@@ -24,7 +24,7 @@ const displayNameMapExample: DisplayNameMap = {
     "Roller Coaster": {
       "id": "DATA_ROLLER_COASTER",
       "attribute": {
-        // \, ", and " are added to test characters escaping
+        // \, ', and " are added to test characters escaping
         "Park\"": "ATTR_PARK",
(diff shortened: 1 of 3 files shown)


v3/src/models/data/formula-utils.test.ts
```diff
@@ -22,7 +24,7 @@ const displayNameMapExample: DisplayNameMap = {
     "Roller Coaster": {
       "id": "DATA_ROLLER_COASTER",
       "attribute": {
-        // \, ', and " are added to test characters escaping
+        // \, ", and " are added to test characters escaping
```
kswenson, line 27: Seems like this makes the comment redundant -- was this change intended?

kswenson (approved):
👍 Looks good

v3/src/models/data/formula-utils.test.ts
pjanik: Thanks, it wasn't, probably some search-and-replace mistake. Fixed in the last commit.

PR #932, opened 2023-10-10
Improve error handling when an attribute used by formula is deleted

pjanik:
https://www.pivotaltracker.com/story/show/186201910

This PR fixes the following bug:

**Test scenario**

1. Open https://codap3.concord.org/branch/main/index.html?sample=mammals
2. Add formula: `Height + Mass`
3. Remove the attribute Mass

**Current behavior**

1. The formula displays an error (as expected), but the error mentions the canonical symbol name (instead of the display one). 
2. When you open the formula dialog, you'll see: `Height + LOCAL_ATTR_ATTRHKTbQ9BF_f2w`
3. When you rename another attribute to Mass, the formula doesn't get fixed.

**Expected behavior**

1. The formula displays an error about undefined `Mass`, 
2. The formula dialog still shows `Height + Mass`.
3. When the user renames another attribute to Mass, the formula should be fixed and recalculated.

---

It took me a bit more time than expected because, initially, I wanted to avoid prefixing all the canonical names with an additional prefix. We could manage regular symbols, but things became tricky with lookup function arguments. Since they're normal strings, at some point, it becomes difficult to guess if they're canonical names or valid user-defined string constants. This new prefix makes things more explicit.

### DIFF
--- a/v3/src/models/data/formula-fn-registry.ts
+++ b/v3/src/models/data/formula-fn-registry.ts
@@ -1,7 +1,7 @@
 import { create, all, mean, median, mad, max, min, sum, random, pickRandom, MathNode, ConstantNode } from 'mathjs'
 import { FormulaMathJsScope } from './formula-mathjs-scope'
 import {
-  DisplayNameMap, FValue, CODAPMathjsFunctionRegistry, ILookupDependency, isConstantStringNode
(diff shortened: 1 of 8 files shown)


kswenson (approved):
👍 Looks good!

PR #933, opened 2023-10-10
fix: childCases after adding/removing cases [#186182255]

kswenson:
[[#186182255]](https://www.pivotaltracker.com/story/show/186182255)

Invalidate/refresh the `childCases` cache when adding/removing cases to fix formula evaluation of newly added cases.


pjanik (approved):
Thanks, looks good 👍

PR #934, opened 2023-10-11
fix: formula evaluation during slider dragging

kswenson:
[[#186182162]](https://www.pivotaltracker.com/story/show/186182162)

Move `dynamicValue` implementation from `SliderModel` to `GlobalValueModel`.

### DIFF
--- a/v3/src/components/slider/slider-model.test.ts
+++ b/v3/src/components/slider/slider-model.test.ts
@@ -100,14 +100,15 @@ describe("SliderModel", () => {
     slider.setDynamicValue(dynamicValue1)
     expect(slider.isUpdatingDynamically).toBe(true)
     expect(slider.value).toBe(dynamicValue1)
-    expect(slider.globalValue.value).toBe(initialValue)
+    expect(slider.globalValue._value).toBe(initialValue)
     slider.setDynamicValueIfDynamic(dynamicValue2)
     expect(slider.value).toBe(dynamicValue2)
     slider.applyUndoableAction(() => slider.setValue(finalValue), "Undo slider change", "Redo slider change")
     expect(slider.isUpdatingDynamically).toBe(false)
     expect(slider.value).toBe(finalValue)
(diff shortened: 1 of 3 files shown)


pjanik (approved):
Nice, it's more straightforward than I thought it'd be. 👍

I definitely don't have that much experience with undo/redo, but actually, I did some implementation in the drawing tool back in the day, and I had to deal with similar problems (ofc way smaller scale and generally simpler). I think the way I handled these prolonged user actions was by providing a method or closure like `suppressUndoHistoryUpdates(callback)`.

I'm wondering if there's currently any possibility of suppressing the undo history from being updated? Or if it wouldn't be a nice approach to similar problems in the future (at least from the client code perspective)? Or it wouldn't work, as we'd always have some changes to MST tree that we cannot ignore?

I could imagine that we would have two handlers in the slider, both using `.setValue()` only:

```javascript
    const handlePointerMove = (e: PointerEvent) => {
      // ...
      sliderModel.suppressUndoHistoryUpdates(
        () => sliderModel.setValue(sliderValue),
      )
      // ...

    const handlePointerUp = (e: PointerEvent) => {
      // ...
      sliderModel.applyUndoableAction(
        () => sliderModel.setValue(sliderValue),
        "DG.Undo.slider.change", "DG.Redo.slider.change"
      )
      // ...
```

kswenson:
> I'm wondering if there's currently any possibility of suppressing the undo history from being updated? Or if it wouldn't be a nice approach to similar problems in the future (at least from the client code perspective)? Or it wouldn't work, as we'd always have some changes to MST tree that we cannot ignore?

There is a mechanism for suppressing undo, `withoutUndo()`, but it can be a bit finicky (e.g. doesn't handle nested actions, etc.) and so my approach has been to make use of volatile state when it's "reasonably straightforward" to do so to avoid invoking the undo machinery at all rather than rely on invoking it and having it suppressed correctly. That said, this may well be a situation in which it would work.

PR #935, opened 2023-10-11
fix(formula): fix lookup functions [PT-186201910]

pjanik:
This PR fixes a bug that I introduced as a part of https://github.com/concord-consortium/codap/pull/932.

`.getDependency` functions should remove the canonical name prefix from IDs. Also, I've reused this function in `evaluateRaw` to reuse this logic there.


kswenson (approved):
👍 Looks good

PR #936, opened 2023-10-11
Handle escaping of single quote strings in formulas

pjanik:
https://www.pivotaltracker.com/story/show/186179035

This PR finally fixes the last missing piece of special character handling: single-quote strings.

This became possible because of the MathJS v11.11.2 fix described here: https://github.com/josdejong/mathjs/issues/3073

However, some additional handling was still required on our side. MathJS doesn't store the type of string constant, whether it's a single-quote or double-quote string. Consequently, we don't know how to escape the given string when converting it from canonical form back to the display form.

That's why the formulaIndexOf helper was added. We take advantage of the fact that there are only two string delimiters available, allowing us to search for both variants, select the one that appears first in the string, and perform the remaining processing based on this result (=> escaping).

### DIFF
--- a/v3/package-lock.json
+++ b/v3/package-lock.json
@@ -19,7 +19,7 @@
         "escape-string-regexp": "^5.0.0",
         "leaflet": "^1.9.4",
         "lodash": "^4.17.21",
-        "mathjs": "^11.11.0",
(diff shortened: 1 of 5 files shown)


kswenson (approved):
👍 Looks good!

PR #937, opened 2023-10-11
feat: add standard deviation and mean absolute deviation (PT-181914408)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/181914408

Adds the Standard Deviation and Mean Absolute Deviation graph adornments. The two new adornments are grouped with the Univariate Measures adornments (Mean, Median, Plotted Value).

Changes have been made to the common `UnivariateMeasureAdornmentComponent` to account for features unique to the Standard Deviation and Mean Absolute Deviation. These features are:

1) a shaded rectangle indicating the range
2) two additional lines at the range min and max
3) hover behaviors for the min and max lines (instead of hovering over the mean line in the center of the range to reveal the text tip or label, hovering over the min and max lines reveals the tip or label)


v3/src/components/graph/adornments/univariate-measures/standard-deviation/standard-deviation-adornment-model.ts
```diff
@@ -0,0 +1,31 @@
...
+export const StandardDeviationAdornmentModel = UnivariateMeasureAdornmentModel
+  .named("StandardDeviationAdornmentModel")
+  .props({
+    type: "Standard Deviation",
+    labelTitle: types.optional(types.literal(kStandardDeviationKey), kStandardDeviationKey)
+  })
+  .views(self => ({
+    computeMeasureRange(attrId: string, cellKey: Record<string, string>, dataConfig: IDataConfigurationModel) {
+      const caseValues = self.getCaseValues(attrId, cellKey, dataConfig)
+      return std(caseValues)
```
kswenson, line 18: Have you verified that the MathJS implementation handled edge cases like no cases or only one case the way we would like? That's always the potential downside of delegating to another library.

v3/src/components/graph/adornments/univariate-measures/univariate-measure-adornment-component.tsx
```diff
@@ -87,119 +120,191 @@ export const UnivariateMeasureAdornmentComponent = observer(
...
+        .attr("x1", x1)
+        .attr("x2", x2)
+        .attr("y1", y1)
+        .attr("y2", y2)
+    }, [])
+
+    const addRange = useCallback((
+      range: number, coverClass: string, lineClass: string, valueObj: IValue, coords: ILineCoords
+    ) => {
+      if (!dataConfig || (!isStandardDeviationAdornment(model) && !isMeanAbsoluteDeviationAdornment(model))) return
```
kswenson, line 143: How about adding a property like `hasRange` to the registration or having a view on the model that returns true for `hasRange` so that we don't have to hard-code specific types here?

v3/src/components/graph/adornments/univariate-measures/univariate-measure-adornment-component.tsx
```diff
@@ -87,119 +120,191 @@ export const UnivariateMeasureAdornmentComponent = observer(
...
+
+    const addLabels = useCallback((
+      labelObj: ILabel, measure: IMeasureInstance, textContent: string, valueObj: IValue,
+      plotValue: number, range?: number
     ) => {
+      const labelSelection = select(labelRef.current)
+      const labelCoords = measure.labelCoords
+      const activeUnivariateMeasures = adornmentsStore?.activeUnivariateMeasures
+      const adornmentIndex = activeUnivariateMeasures?.indexOf(model) ?? null
+      const topOffset = activeUnivariateMeasures.length > 1 ? adornmentIndex * 20 : 0
```
kswenson, line 201: Can we have a named constant for the magic number 20?

v3/src/components/graph/adornments/univariate-measures/univariate-measure-adornment-component.tsx
```diff
@@ -87,119 +120,191 @@ export const UnivariateMeasureAdornmentComponent = observer(
...
+    }, [adornmentsStore?.activeUnivariateMeasures, classFromKey, containerId, handleEndMoveLabel, handleMoveLabel,
+        highlightCovers, highlightLabel, measureSlug, model, xCellCount, xScale])
+
+    const addTextTip = useCallback((plotValue: number, textContent: string, valueObj: IValue, range?: number) => {
+      const selection = select(valueRef.current)
+      const textId = generateIdString("tip")
+      const textClass = clsx("measure-tip", `${measureSlug}-tip`)
+      let x = isVertical.current ? xScale(plotValue) / xCellCount + 5 : (plotWidth - plotWidth/2) / xCellCount
+      if (range && isVertical.current) x = x + range
+      let y = isVertical.current ? 50 : yScale(plotValue) / yCellCount - 5
```
kswenson, line 242: Named constants for some of these magic numbers?

kswenson (changes requested):
👍 Looks good -- just a few suggestions

v3/src/components/graph/adornments/univariate-measures/univariate-measure-adornment-component.tsx
emcelroy: @kswenson I tried exactly that first, but there was a problem. I think for this particular instance, we can actually drop the adornment type check since it's also done in `addLineCoverAndLabel` before that function calls `addRange`. Still, if I add a `hasRange` property to the `UnivariateMeasureAdornmentModel`, set it to false by default, and set it to true on the SD and MAD adornment models, the code below will return the error `Property 'computeMeasureRange' does not exist on type 'IUnivariateMeasureAdornmentModel'.`

```
const measureRange = attrId && model.hasRange
  ? model.computeMeasureRange(attrId, cellKey, dataConfig)
  : undefined
```
I couldn't figure out another way to make it work other than to use `isStandardDeviationAdornment` and `isMeanAbsoluteDeviationAdornment` instead. Are you aware of a straightforward way to make the `hasRange` check work the way we'd want it to? Maybe adding a default `computeMeasure` view to `UnivariateMeasureAdornmentModel` that returns `NaN` or `undefined` is a reasonable option?

v3/src/components/graph/adornments/univariate-measures/standard-deviation/standard-deviation-adornment-model.ts
emcelroy: @kswenson I believe so. It seems to match V2's behavior. With no cases, no univariate adornments will appear (except for Plotted Value) when their options are checked. With one case, the Mean, Median, SD, and MAD will appear but they are all a single line. SD and MAD do not show a range.

v3/src/components/graph/adornments/univariate-measures/univariate-measure-adornment-component.tsx
kswenson: Add a `hasRange` view to `UnivariateMeasureAdornmentModel` that returns `false`.

v3/src/components/graph/adornments/univariate-measures/univariate-measure-adornment-model.ts
```diff
@@ -48,12 +49,16 @@ export const UnivariateMeasureAdornmentModel = AdornmentModel
     },
     get isUnivariateMeasure() {
       return true
+    },
+    computeMeasureRange(attrId: string, cellKey: Record<string, string>, dataConfig: IDataConfigurationModel) {
+      // derived models should override if they have a range
+      return undefined
+    },
+    computeMeasureValue(attrId: string, cellKey: Record<string, string>, dataConfig: IDataConfigurationModel) {
```
emcelroy, line 57: Note that I made this a view instead of an action. Even in the individual "sub adornment" models, `computeMeasureValue` doesn't set or change any model properties, and is already defined as a view, not an action.

v3/src/components/graph/adornments/univariate-measures/mean-absolute-deviation/mean-absolute-deviation-adornment-model.ts
```diff
@@ -10,6 +10,7 @@ export const MeanAbsoluteDeviationAdornmentModel = UnivariateMeasureAdornmentMod
   .named("MeanAbsoluteDeviationAdornmentModel")
   .props({
     type: "Mean Absolute Deviation",
+    hasRange: true,
```
kswenson, line 14: Make this a view that returns `true` rather than a property -- there's no reason to store it in the model/document given that it never changes.

kswenson (approved):
👍 Looks good -- just one suggestion to make `hasRange` be views rather than properties since there's no reason to store them.

v3/src/components/graph/adornments/univariate-measures/mean-absolute-deviation/mean-absolute-deviation-adornment-model.ts
emcelroy: Ah, OK 👍

PR #938, opened 2023-10-11
fix: undo slider creation

kswenson:
Guarantees that creating a slider can be undone/redone in a single step and that no MST warnings are generated.

- Adds `MobXAutorun` class for managing a MobX `autorun` that is automatically uninstalled when the underlying MST model it is monitoring is destroyed.
- Adds `isAliveSafe()` function which is like MST's `isAlive()` but accepts `undefined`, which precludes the need to type `if (someModel && isAlive(someModel))`.
- I/we have said that `isAlive()` (or `isAliveSafe()`) should not be used as a standard control flow mechanism, but rather as a means to detect situations in which we haven't handled model destruction appropriately and log a warning, for instance. It turns out, however, that due to the vagaries of the `observer` implementation, there doesn't seem to be a mechanism for detecting model destruction and preventing invalid model access from a React component wrapped in `observer`, so that appears to be a case in which `isAlive()` or `isAliveSafe()` is required.

### DIFF
--- a/v3/src/components/axis/hooks/use-sub-axis.ts
+++ b/v3/src/components/axis/hooks/use-sub-axis.ts
@@ -1,6 +1,5 @@
 import {BaseType, drag, format, ScaleLinear, select, Selection} from "d3"
 import {reaction} from "mobx"
-import {isAlive} from "mobx-state-tree"
 import {MutableRefObject, useCallback, useEffect, useMemo, useRef} from "react"
 import {transitionDuration} from "../../data-display/data-display-types"
 import {AxisBounds, AxisPlace, axisPlaceToAxisFn, AxisScaleType, otherPlace} from "../axis-types"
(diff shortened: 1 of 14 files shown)


v3/src/components/axis/hooks/use-sub-axis.ts
```diff
@@ -350,8 +351,8 @@ export const useSubAxis = ({
 
   // update d3 scale and axis when axis domain changes
   useEffect(function installDomainSync() {
-    const disposer = autorun(() => {
-      const _axisModel = axisProvider.getAxis?.(axisPlace)
+    const mobXAutorun = new MobXAutorun(() => {
+      const _axisModel = axisProvider?.getAxis?.(axisPlace)
       if (_axisModel && isAlive(_axisModel)) {
```
emcelroy, line 356: Probably not important, but could you use `isAliveSafe` here? Change the line to `if (isAliveSafe(_axisModel)) {`?

emcelroy (approved):
Looks good 👍 I just left on question about a possible change in a comment. I don't think it's critical.

Once these changes are merged into `main` should we generally favor using `MobXAutorun` over `autorun`?

v3/src/components/axis/hooks/use-sub-axis.ts
kswenson: Good catch -- I wrote this part before I added `isAliveSafe()` and never revisited. I'll make the change.

kswenson:
> Once these changes are merged into `main` should we generally favor using `MobXAutorun` over `autorun`?

`reaction`s/`autorun`s that access MST models probably should default to `MobXAutorun` and `MobXReaction`. I've been converting them as I encounter them while debugging undo/redo issues but you should be able to tell which ones are likely to be problems by inspection.

PR #939, opened 2023-10-12
chore: refactor autorun/reaction wrappers

kswenson:
I originally implemented these wrappers as classes because I thought I needed the MobX `disposer` property to be stored in an instance member of the class so that I could prevent it from being called twice, once by the internal MST model disposer and once by the external client, by setting it to `undefined` once it had been called. @pjanik was right, however, that the same thing can be accomplished with a function that makes use of an internal closure variable and the resulting API seems simpler, so I'm making the change.


pjanik (approved):
Looks good 👍

PR #940, opened 2023-10-12
fix: undo create graph

kswenson:
Use `mstReaction` to prevent access to defunct models from `reaction`s in the graph code when undoing/redoing the creation of a graph tile.


bfinzer (approved):
👍🏻LGTM

PR #941, opened 2023-10-12
fix: Adornments not always rendered for all subplots (PT-186164083)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/186164083

This corrects an issue where adornments sometimes weren't appearing in all relevant subplots. The problem seemed to only arise when there was a single numeric attribute on either the x or y axis, and then categorical attributes on the top split and right split. It stemmed from the `setCellKey` action's index calculation for associating the correct right-split category with the cell. It didn't account for all the possible configurations that could affect that calculation.

While making these changes, I also made `setCellKey` a view instead of an action since it doesn't actually set anything, and then changed its name to `generateCellKey` to clarify that it doesn't set anything.

### DIFF
--- a/v3/src/components/graph/adornments/adornment-models.ts
+++ b/v3/src/components/graph/adornments/adornment-models.ts
@@ -63,16 +63,8 @@ export const AdornmentModel = types.model("AdornmentModel", {
     },
     get isUnivariateMeasure() {
       return false
-    }
-  }))
-  .actions(self => ({
-    setVisibility(isVisible: boolean) {
-      self.isVisible = isVisible
     },
-    updateCategories(options: IUpdateCategoriesOptions) {
-      // derived models should override to update their models when categories change
-    },
-    setCellKey(options: IUpdateCategoriesOptions, index: number) {
+    generateCellKey(options: IUpdateCategoriesOptions, index: number) {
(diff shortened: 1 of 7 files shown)


v3/src/components/graph/adornments/adornment-models.ts
```diff
@@ -63,45 +63,53 @@ export const AdornmentModel = types.model("AdornmentModel", {
...
-  }))
-  .actions(self => ({
-    setVisibility(isVisible: boolean) {
-      self.isVisible = isVisible
     },
-    updateCategories(options: IUpdateCategoriesOptions) {
-      // derived models should override to update their models when categories change
-    },
-    setCellKey(options: IUpdateCategoriesOptions, index: number) {
+    generateCellKey(options: IUpdateCategoriesOptions, index: number) {
```
kswenson, line 67: What about `getCellKey` or even just `cellKey`? 🤷

kswenson (approved):
👍 Looks good

v3/src/components/graph/adornments/adornment-models.ts
emcelroy: I'm going to go with `cellKey`. I think that makes sense. Thanks.

PR #943, opened 2023-10-13
Add tests of numericFnFactory, numericMultiArgsFnFactory, and all the arithmetic functions

pjanik:
[Develop testing pattern for non-aggregate formula functions #186245948](https://www.pivotaltracker.com/story/show/186245948)

This PR is related to https://github.com/concord-consortium/codap/pull/942, but I'm opening it separately because it's the first time I've added unit tests for the actual functions.

There are a couple of options on how to approach this, but the pattern with `math.compile` and `math.evaluate` with a basic object literal scope seems to be quite clean. These tests are close to real unit tests, and there's no need to involve a custom MathJS scope (which can be tested separately). Each of the arithmetic functions tests slightly overlaps with the tests of the helper functions (handling arrays and non-numeric values). However, it's easy to do and requires only a few lines of code, so I believe it's still worth keeping this pattern (even to test whether the function actually uses the helpers or equivalent logic). If someone needs to add new functions, testing them should be straightforward as well.

Aggregate and semi-aggregate functions will require more work since they depend on our custom MathJS features. So, I'll either need to initialize it in these tests or provide some mocks. I'll leave that for later.

I also considered semi-integration tests (still written in Jest) where we use `DataSet`, `MathJSScope`, and `FormulaManager` to perform complete tests. However, these tests might be limited to just one function from each group (e.g., one arithmetic, one aggregate, etc.). Writing such tests for each new function could be tedious, and the tests would be much slower.

Also, the formula-fn-registry should be divided into smaller parts, and some helpers are spread across various files and could probably be regrouped. I'm planning to refactor all the formula files and perform some cleanup, but I'll address this cleanup work separately.

### DIFF
--- a/v3/src/models/data/formula-fn-registry.test.ts
+++ b/v3/src/models/data/formula-fn-registry.test.ts
@@ -24,7 +24,7 @@ describe("numericMultiArgsFnFactory", () => {
   const fn = (valueA: number, valueB = 0) => valueA + valueB
   const numericFn = numericMultiArgsFnFactory(fn, { numOfRequiredArgs: 1 })
 
-  it("returns a function that applies the specified function to a multiple numeric values", () => {
+  it("returns a function that applies the specified function to multiple numeric values", () => {
     const result = numericFn(10, 20)
     expect(result).toBe(30)
   })
--- a/v3/src/models/data/formula-fn-registry.test.ts
+++ b/v3/src/models/data/formula-fn-registry.test.ts
@@ -1,0 +1,210 @@
+import { UNDEF_RESULT, numericFnFactory, numericMultiArgsFnFactory, math } from "./formula-fn-registry"
+
+describe("numericFnFactory", () => {
+  const fn = (value: number) => value * 2
+  const numericFn = numericFnFactory(fn)
+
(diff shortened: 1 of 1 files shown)


v3/src/models/data/formula-fn-registry.test.ts
```diff
@@ -0,0 +1,210 @@
...
+    const result = numericFn([1, 2, "3", "4", "five", "", UNDEF_RESULT])
+    expect(result).toEqual([2, 4, 6, 8, UNDEF_RESULT, UNDEF_RESULT, UNDEF_RESULT])
+  })
+})
+
+describe("numericMultiArgsFnFactory", () => {
+  const fn = (valueA: number, valueB = 0) => valueA + valueB
+  const numericFn = numericMultiArgsFnFactory(fn, { numOfRequiredArgs: 1 })
+
+  it("returns a function that applies the specified function to a multiple numeric values", () => {
```
kswenson, line 27:
```suggestion
  it("returns a function that applies the specified function to multiple numeric values", () => {
```

kswenson (approved):
👍 Nice!

PR #944, opened 2023-10-13
Map component displays

bfinzer:
It has taken some work to get back into the map component. Hopefully the architecture will hold up for awhile.

* Fixing merge conflict problems
* GraphDataConfigurationModel is now based on DataConfigurationModel
* Extensive refactoring
  * Moved some functions from graph-utils to data-display-utils for use by maps
  * Created PointDescription to be used by GraphContentModel and by MapPointLayerModel
  * Points (circles) now have the data configuration ID as a class for separation of point layers in map
  * Defined MapAttrRoles
  * Replaced kGraphFont with kDataDisplayFont
  * Factored computePointRadius for use by GraphContentModel and MapPointLayerModel
* Additional work on map point layers
* Points are showing but not in the right place


v3/src/components/map/map-types.ts
```diff
@@ -1,10 +1,18 @@
...
   kMapUrl = "https://services.arcgisonline.com/ArcGIS/rest/services/World_Street_Map/MapServer/tile/{z}/{y}/{x}",
 
   kDefaultMapLocation = [45.4408, 12.3155],
   kDefaultMapZoom = 1,
-  kDefaultMapZoomForGeoLocation = 8
+  kDefaultMapZoomForGeoLocation = 8,
+
+  kLatNames = ['latitude', 'lat', 'latitud', 'breitengrad', '緯度', 'عرض جغرافیایی'],
+  kLongNames = ['longitude', 'long', 'lng', 'lon', 'longitud', 'längengrad', '経度', 'طول جغرافیایی'],
+  kPolygonNames = ['boundary', 'boundaries', 'polygon', 'polygons', 'grenze', '境界', 'مرز']
```
kswenson, line 17: We're not using our normal localized strings for these?

v3/src/components/map/models/map-content-model.ts
```diff
@@ -20,63 +28,73 @@ export const MapContentModel = DataDisplayContentModel
...
+              } else {
+                // Add a new layer for this dataset
+                this.addPointLayer(sharedDataSet.dataSet)
+              }
+            }
+          })
+          // Remove any layers that are no longer in the shared model
+          layersToCheck.forEach(layer => {
+            self.layers.splice(self.layers.indexOf(layer), 1)
+          })
```
kswenson, line 75: This whole block could use a comment since its intent is not obvious.

v3/src/components/map/models/map-content-model.ts
```diff
@@ -20,63 +28,73 @@ export const MapContentModel = DataDisplayContentModel
...
-          // auto-link to DataSet if it contains lat/long and/or polygons
-          else if (!tileDataSet && sharedDataSets.length === 1) {
-            linkTileToDataSet(self, sharedDataSets[0].dataSet)
-          }
+          const layersToCheck = self.layers.splice(0)
+          sharedDataSets.forEach(sharedDataSet => {
+            if (datasetHasPointData(sharedDataSet.dataSet)) {
+              const layer = layersToCheck.find(aLayer => aLayer.data === sharedDataSet.dataSet)
+              if (layer) {
+                layersToCheck.splice(layersToCheck.indexOf(layer), 1)
```
kswenson, line 65:
```suggestion
              const foundIndex = layersToCheck.findIndex(aLayer => aLayer.data === sharedDataSet.dataSet)
              if (foundIndex >= 0) {
                layersToCheck.splice(foundIndex, 1)
```

v3/src/components/map/utilities/map-utils.ts
```diff
@@ -1,8 +1,98 @@
...
+import {latLngBounds} from 'leaflet'
+import {isFiniteNonEmpty} from "../../../utilities/math-utils"
+import {IDataSet} from "../../../models/data/data-set"
+import {kLatNames, kLongNames} from "../map-types"
+import {IDataConfigurationModel} from "../../data-display/models/data-configuration-model"
 
 
-export const syncMapModelWithDatasets = (mapModel: IMapContentModel) => {
-  // const datasets = getSharedDataSets()
+export const datasetHasPointData = (dataset: IDataSet) => {
```
kswenson, line 8:
```suggestion
export const datasetHasLatLongData = (dataset: IDataSet) => {
```
and also add a comment about what the criteria are.

v3/src/components/map/utilities/map-utils.ts
```diff
@@ -1,8 +1,98 @@
...
+        hasLatAttribute = true
+      } else if (kLongNames.includes(attrName)) {
+        hasLngAttribute = true
+      }
+    }
+  }
+  return hasLatAttribute && hasLngAttribute
+}
+
+export const pointAttributesFromDataSet = (dataSet: IDataSet) => {
```
kswenson, line 25:
```suggestion
export const latLongAttributesFromDataSet = (dataSet: IDataSet) => {
```

v3/src/components/map/utilities/map-utils.ts
kswenson, line 1: Comments would help for each of these functions.

v3/src/utilities/math-utils.ts
```diff
@@ -96,3 +96,7 @@ export function neededSigDigitsArrayForQuantiles(quantiles: number[], values: nu
     return Math.max(leftDigits, rightDigits)
   })
 }
+
+export function isFiniteNonEmpty(val:any) {
```
kswenson, line 100: It's not actually performing a non-empty test -- perhaps `isFiniteNumber()`?

v3/src/components/codap-component.scss
```diff
@@ -52,6 +53,7 @@ $corner-drag-size: calc($border-drag-width * 2);
   position: absolute;
   margin: 0;
   user-select: none;
+  z-index: 1000;
```
kswenson, line 56: I assume this is more `z-index` machinations to work around leaflet. We should set up a [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_positioned_layout/Understanding_z-index/Stacking_context) so that leaflet's `z-index` values don't affect the surrounding content. The simplest map-specific solution would seem to be adding a `z-index` value to the `.map-container` `div`, given that one of the situations that creates a new stacking context is "Element with a [`position`](https://developer.mozilla.org/en-US/docs/Web/CSS/position) value `absolute` or `relative` and [`z-index`](https://developer.mozilla.org/en-US/docs/Web/CSS/z-index) value other than auto." Arguably, we could/should also add a `z-index` to the `.codap-component` `div` to create a stacking context for the entire tile, but that probably warrants a larger conversation.

v3/src/components/codap-component.scss
```diff
@@ -23,6 +23,7 @@ $corner-drag-size: calc($border-drag-width * 2);
   position: absolute;
   margin: 0;
   user-select: none;
+  z-index: 11;
```
kswenson, line 26: See longer `z-index` comment elsewhere in this file.

kswenson (changes requested):
👍 Looking good -- a lot of comments/suggestions. Feel free to table some of them for later if it seems like they would conflict with ongoing work, but a number of the suggestions shouldn't present such difficulties.

kswenson:
👍 Assuming you're planning to take care of the remaining items later, e.g. z-indices, role definitions, localized strings for lat/lng attribute names, etc.

PR #945, opened 2023-10-16
fix: Wrong attribute passed to formula (PT-186266057)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/186266057

@pjanik pointed out that the wrong attribute ID can sometimes be passed to the `computeMeasureValue` view in Plotted Value model. (See [PT story](https://www.pivotaltracker.com/story/show/186266057) for an example.)

In `UnivariateMeasureAdornmentModel` and `UnivariateMeasureAdornmentComponent` I had used `dataConfig.primaryAttributeID` to determine the attribute ID to use. In the case of the Plotted Value -- which unlike other univariate measures can be present when there are numeric attributes on both the x and y axes -- that won't work. According to [the original Plotted value story](https://www.pivotaltracker.com/story/show/181940090) description, "When functions (e.g. mean) are specified, if there is no argument, the x-attribute is assumed as the attribute over which computation should take place." So I believe the thing to do is always pass the x attribute's ID if there is a numeric attribute on x.


pjanik (approved):
Thank you. The assumption that we should always use the X-axis attribute looks good to me, but I'll wait for Kirk's opinion too. :)

kswenson (approved):
👍 Seems like that should work

PR #946, opened 2023-10-17
fix: add standard deviation and mean absolute deviation (PT-181914408)

emcelroy:
PT Story: https://www.pivotaltracker.com/n/projects/2556922/stories/181914408

Fixes a few bugs uncovered by QA...

1. At least in some instances, MathJS's `mad` function wasn't returning the same value that CODAP v2's method for determining the MAD does. I'm not still not sure why that is, but copying over CODAP v2's method to v3 so it uses the expected values was straightforward enough.
2. The dimensions of the shaded area on the graph were not being properly set. These changes properly calculate the dimensions of the range in the context of the plot. The dimensions will now match what appears in v2.
3. The values shown in text tips and labels for SD and MAD could be wrong in the case that the SD or MAD values were 0. 
4. The text tips that appear on hover were sometimes not visible because they were placed outside the visible bounds of the plot. These changes check to make sure that doesn't happen by adjusting the text tip's `x` value to ensure it's at least mostly visible (if a subplot is exceptionally narrow, the tip could then still be partially hidden).

### DIFF
--- a/v3/src/components/graph/adornments/univariate-measures/univariate-measure-adornment-component.tsx
+++ b/v3/src/components/graph/adornments/univariate-measures/univariate-measure-adornment-component.tsx
@@ -13,6 +13,7 @@ import { valueLabelString } from "../../utilities/graph-utils"
 import { Point } from "../../../data-display/data-display-types"
 import { useGraphContentModelContext } from "../../hooks/use-graph-content-model-context"
 import { isPlottedValueAdornment } from "./plotted-value/plotted-value-adornment-model"
+import { measureText } from "../../../../hooks/use-measure-text"
 
 import "./univariate-measure-adornment-component.scss"
 
@@ -246,7 +247,7 @@ export const UnivariateMeasureAdornmentComponent = observer(
 
       // If x plus the approximate width of the text tip would extend beyond the right boundary of the subplot, set x to
       // plotWidth minus the text tip width or zero, whichever is greater.
-      const textTipWidth = textContent.length * 8
+      const textTipWidth = measureText(textContent)
       if (x + textTipWidth > plotWidth) x = Math.max(plotWidth - textTipWidth, 0)
 
       valueObj.text = selection.append("text")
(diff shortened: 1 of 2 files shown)


v3/src/components/graph/adornments/univariate-measures/univariate-measure-adornment-component.tsx
```diff
@@ -239,9 +240,14 @@ export const UnivariateMeasureAdornmentComponent = observer(
...
       let x = isVertical.current ? xScale(plotValue) / xCellCount + lineOffset : (plotWidth - plotWidth/2) / xCellCount
-      if (range && isVertical.current) x = x + range
+      if (range && isVertical.current) x = xScale(plotValue + range) / xCellCount + lineOffset
       let y = isVertical.current ? topOffset : yScale(plotValue) / yCellCount - lineOffset
-      if (range && !isVertical.current) y = y - range
+      if (range && !isVertical.current) y = yScale(plotValue + range) / yCellCount - lineOffset
+
+      // If x plus the approximate width of the text tip would extend beyond the right boundary of the subplot, set x to
+      // plotWidth minus the text tip width or zero, whichever is greater.
+      const textTipWidth = textContent.length * 8
```
emcelroy, line 249: This method for determining the width of the tool tip was suggested by Copilot. It won't be 100% accurate, but it seems reasonable to me as an approximation that will work most of the time. The only alternative I'm aware of would depend on adding a hidden element to the DOM, inserting the text into that element, and then checking that element's width. There's very possibly some other option, but it would probably depend on revamping how the text tips are constructed. If this seems reasonable enough, I'd rather not investigate that further right now.

kswenson: Did you look into the available `measureText` or `useMeasureText` utilities already in use?

kswenson (approved):
👍 Looks good -- one question about the mechanism for estimating the width of the tooltip.

emcelroy:
> ...one question about the mechanism for estimating the width of the tooltip.

I wasn't aware of the available `measureText` option. Thanks, that works well. I think it's similar to what I was thinking of in adding an element to the DOM, except I wasn't thinking of `canvas` specifically and, more importantly, it looks like it never adds the canvas to the DOM? I didn't realize you could do that and still get an accurate measure of the rendered width. Good to know 👍

PR #947, opened 2023-10-18
Cypress tests for formula engine

tejal-shah:


pjanik:
Great work on the coverage! This will be extremely beneficial for my ongoing work that involves a lot of refactoring.

I noticed that this is the first time we're using JSON-based tests in this repository, correct? I assume this pattern was adopted from another project, possibly CLUE? I'm guessing this part doesn't need a review, right? I can see how this approach significantly reduces the amount of typing required when creating these tests. As for whether it simplifies working with them later on, you're likely more knowledgeable on that than I am. :)

I have two minor suggestions:

1. For `arithmetic-formula`, could we include two additional cells: one empty and one with a random string (e.g., "abc", "foo", etc.)? It would be useful to test empty strings and non-numeric strings, as it was part of the task. This would likely apply to any other group of functions that we add over time.
2. When I ran the test live, I noticed numerous 500ms `wait` calls. It seems that we're waiting 500ms before each comparison when testing the values of multiple attribute cells. While I can see the benefit of waiting before the first check, I believe we could eliminate the subsequent ones. This would significantly speed up the test run. I understand this might be based on your experience, but is there a possibility we could remove these wait calls altogether? Or perhaps just use one before the first call (and maybe for less than 500ms?).

pjanik (approved):

tejal-shah:
@pjanik , thank you for your inputs.
1. I've now included a "foo" cell in the arithmetic-formula test. An empty cell was already present in my original commit.
2. I had added the extra waits in my first draft and had since forgotten to remove them. I've now gotten rid of them, thank you for catching that.
3. I'm also fixing a minor bug I found in the visitURL helper when fixing the first two points above.

As for the JSON-based tests, I originally started writing the tests in the traditional way that we do but hald way through, I realized that the code was getting extremely repetitive and writing them this way might make more sense, for that reason and it might also make the test more readable. It might also have the added advantage of making it easier to ads more tests in the future, but we'll see. I don't think we have JSON-based (aka keyword driven testing framework here at CC, but I believe when we have such extensive repetitive tests, having this might be beneficial.

pjanik:
> As for the JSON-based tests, I originally started writing the tests in the traditional way that we do but hald way through, I realized that the code was getting extremely repetitive and writing them this way might make more sense, for that reason and it might also make the test more readable.

@tejal-shah, I wonder if calling functions (these step executors) directly wouldn't be even more concise.  E.g., instead of:
```javascript
"testName": "Check arithmetic functions",
"steps": [
{
  "name": "addFormula",
  "attributeName": "pow",
  "formula": "pow(num, 2)"
},
```
we could say:
```javascript
addFormula({ attributeName: "pow", formula: "pow(num, 2") };
```
It seems it would be even more compact. And the same amount of work, as either way, we need these step executor helpers. Another point with JSON is that it's a bit harder to customize. Let's say that in one of the tests you need a few custom steps. Then, you're forced to add step executor for each of them. If it's part of a regular test, you can just write normal JS code.

pjanik:
> Regarding the JSON-based tests, I initially began writing them in the traditional manner that we are accustomed to. However, halfway through the process, I realized that the code was becoming excessively repetitive. I started to consider that this alternative approach might not only make more sense but also enhance the readability of the tests.

@tejal-shah, I'm curious if directly calling functions (the step executors) would result in even more concise code. For instance, instead of:

```javascript
"testName": "Check arithmetic functions",
"steps": [
  {
    "name": "addFormula",
    "attributeName": "pow",
    "formula": "pow(num, 2)"
  },
]
```

we could opt for:

```javascript
it("Check arithmetic functions", () => {
  addFormula({ attributeName: "pow", formula: "pow(num, 2") };
})
```

This approach seems to offer a more compact representation and would require the similar amount of effort, as we'd still need to implement these step executor helpers in either case. Another challenge with using JSON is that it can be somewhat less flexible when it comes to customization. For instance, if a few custom steps are required for one of the tests, you'd be forced to add a step executor for each of them, even if they are only used once. In a standard test, you can simply write regular JavaScript code to handle such situations.

Having said all of this, I'm not proposing any changes since it's presentation rather than the functionality. Let's test it out and observe how it works for you. These are just my two cents. Thanks again for all the tests and coverage.

PR #949, opened 2023-10-18
fix: tile ordering on undo/redo close component

kswenson:
[[#186075314]](https://www.pivotaltracker.com/story/show/186075314)

@tejal-shah noted that sometimes redoing the removal of a tile would remove two tiles rather than one. This could occur when the `tiles` map and the `order` array got out of sync. The fix here is to move the `order` array to volatile storage so that the undo history doesn't attempt to apply any patches to it (we don't want simple tile ordering changes to be undoable anyway). We do want the tile order to be saved with the document, however, so that when users return to a saved document the tile order is the way they left it. To accomplish this we make use of the same `prepareSnapshot()`/`completeSnapshot()` mechanism we've used elsewhere, but now we apply it document-wide rather than just to `DataSet`s.

### DIFF
--- a/v3/src/models/document/free-tile-row.ts
+++ b/v3/src/models/document/free-tile-row.ts
@@ -57,7 +57,7 @@ export interface IFreeTileInRowOptions extends ITileInRowOptions {
   height?: number
 }
 export const isFreeTileInRowOptions = (options?: ITileInRowOptions): options is IFreeTileInRowOptions =>
-              (options as any)?.x != null && (options as any).y != null
+              !!options && ("x" in options && options.x != null) && ("y" in options && options.y != null)
 
 /*
   Tiles are represented as a map of layouts and an array of tile ids representing the order.
--- a/v3/src/models/document/free-tile-row.ts
+++ b/v3/src/models/document/free-tile-row.ts
@@ -1,4 +1,5 @@
-import { Instance, SnapshotIn, types } from "mobx-state-tree"
+import { observable } from "mobx"
+import { Instance, SnapshotIn, addDisposer, onPatch, types } from "mobx-state-tree"
 import { ITileInRowOptions, ITileRowModel, TileRowModel } from "./tile-row"
 import { withoutUndo } from "../history/without-undo"
 import { withUndoRedoStrings } from "../history/codap-undo-types"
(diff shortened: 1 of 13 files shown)


v3/src/models/history/tree-types.ts
```diff
@@ -50,6 +50,10 @@ export function isChildOfUndoRedo(actionCall?: IActionContext) {
   return false
 }
 
+export function isUndoingOrRedoing() {
+  return isChildOfUndoRedo(getRunningActionContext())
+}
```
kswenson, line 55: This is currently unused. I added it thinking it would be necessary but ended up not needing it, but it seems like it could be useful so I'm leaving it in for now.

v3/src/models/shared/shared-data-utils.ts
```diff
@@ -77,18 +77,3 @@ export function linkTileToDataSet(tile: ITileContentModel, dataSet: IDataSet) {
...
-    sharedDataSets.forEach(model => model.dataSet.prepareSnapshot())
-
-    // perform the serialization with the prepared data sets
-    return serializeFn()
-  }
-  finally {
-    // complete the serialization process for each data set
-    sharedDataSets.forEach(model => model.dataSet.completeSnapshot())
-  }
-}
```
kswenson, line 94: This functionality moved to `serializeDocument`.

v3/src/models/document/free-tile-row.ts
```diff
@@ -56,7 +57,7 @@ export interface IFreeTileInRowOptions extends ITileInRowOptions {
   height?: number
 }
 export const isFreeTileInRowOptions = (options?: ITileInRowOptions): options is IFreeTileInRowOptions =>
-              (options as any)?.x != null && (options as any)?.y != null
+              (options as any)?.x != null && (options as any).y != null
```
emcelroy, line 60: Could you drop the `?` from `(options as any)?.x` like you dropped it from the check for `y`? And could you use `IFreeTileInRowOptions` in place of `any`?

emcelroy (approved):
Looks good! I left one non-blocking comment with a couple questions. I'm not sure they matter much, and it's possible I'm missing a reason why the code is the way it is.

v3/src/models/document/free-tile-row.ts
kswenson: I tweaked the test to eliminate the cast altogether.

PR #955, opened 2023-10-20
chore: Update adornment type properties (PT-186297171)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/186297171


pjanik (approved):
Looks good, thanks! 👍

PR #956, opened 2023-10-20
fix: redo .csv import

kswenson:
The issue was that the data values weren't being captured in the redo patch because of our use of volatile storage for data values. The fix is to make sure `prepareSnapshot()` is called before the patch is created and `completeSnapshot()` is called once the action is complete. I also added a jest test to increase the patch coverage. The project coverage reports a drop in coverage of some unrelated files that shouldn't be affected at all, so I don't know what's happening there.


emcelroy (approved):
Looks good 👍

PR #957, opened 2023-10-21
Formula observers cleanup/refactor, bug fixes, and unit tests

pjanik:
https://www.pivotaltracker.com/story/show/186267057

1. The most significant change in this PR is the extraction of all formula-related observers into a separate file and converting them into regular functions, making them closer to pure functions. The primary goal was to make the FormulaManager slimmer, make these functions easier to test, and allow for composing these observers when necessary. For example, attribute formulas observe changes to the dataset hierarchy, while plotted value formulas do not. I have also added quite a few unit tests that have already helped me catch a few subtle bugs, such as the need to recalculate all cases when new cases are added and the formula includes an aggregate function.

2. Additionally, I have used a new constant, `kPlottedValueType`, added by @emcelroy.

3. Finally, I updated some code to allows us to support symbols with empty names. While it may not seem very useful, it can occur. One of Tejal's tests modifies the name of the slider global value, but first, it clears its value. The GlobalName is updated immediately, and it was causing crashes in display formula updates. This bug was previously hidden, as global value name updates were slightly buggy and incomplete. Both of these issues have now been resolved.

### DIFF
--- a/v3/src/models/data/attribute-formula-adapter.ts
+++ b/v3/src/models/data/attribute-formula-adapter.ts
@@ -5,13 +5,14 @@ import {
   formulaError, getFormulaChildMostAggregateCollectionIndex, getFormulaDependencies, getIncorrectChildAttrReference,
   getIncorrectParentAttrReference
 } from "./formula-utils"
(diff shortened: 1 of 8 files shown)


kswenson (approved):
👍 Looks good!

PR #958, opened 2023-10-21
Move formula code to one dir, split function definitions into smaller files

pjanik:
This pull request moves all the formula-related source files to a `src/models/formula` directory. Additionally, the large `formula-fn-registry.ts` file has been split into multiple smaller ones, all residing in `src/models/formula/functions`. Consequently, arithmetic functions are now defined in the `arithmetic-functions.ts` file, aggregate functions in the `aggregate-functions.ts` file, and so on. I've tried to use file names that mostly correspond to the V2 grouping of functions, although semi-aggregate functions seemed to deserve their own file. This change should make it easier to work with and add new functions.


v3/src/models/formula/functions/other-functions.ts
```diff
@@ -0,0 +1,24 @@
...
+    // just the name is different.
+    evaluate: (...args: FValue[]) => pickRandom(args)
+  },
+
+  // random(min, max)
+  random: {
+    isRandomFunction: true,
+    // Nothing to do here, mathjs.random() has exactly the same signature as CODAP V2 random() function.
+    evaluate: (...args: FValue[]) => random(...args as number[])
+  }
```
kswenson, line 23: We may end up wanting to move the `random...` functions to their own file as we add more of them, but this seems fine for now.

kswenson (approved):
👍 Looks good

PR #961, opened 2023-10-24
fix: graph attribute drop in presence of multiple data sets

kswenson:
Alternative to #959 as a fix for [[#186316077]](https://www.pivotaltracker.com/story/show/186316077).

Ultimately, the issue is that we have global mechanisms (e.g. the shared model manager) and local mechanisms (e.g. `DataConfigurationModel.dataset`) for tracking which tiles are connected to which data sets and the two weren't being kept in sync. #959 proposed fixing it by eliminating all references to the global mechanisms and relying entirely on the local mechanisms, which allows the graph to work despite the two being out of sync. In this PR we propose to actually keep the two in sync.


bfinzer (approved):
👍🏻LGTM

PR #962, opened 2023-10-25
[#185781518] Feature: A map displays points for datasets that have lat/long attributes

bfinzer:
* Properly restore point layer
* Display points in correct positions
* Respond to leaflet notifications
* Show case selection
* Show data tip
* Allow selection
* Properly show selection for points on map
* Click on point stops event propagation so that the map won't receive the click and deselect all
* Ensure that the map center and zoom are stored in the map model
* Expand the containing bounds slightly after computing it so that the data fit better
* Show data tips on each point
* Fine tuning of leaflet event handling
* Creating a stack context for the leaflet MapContainer by giving the enclosing div its own z-index
* Moved definitions of AttrRoles into data-display-types.ts


kswenson (changes requested):
👍 Looks good -- a couple of minor changes, including changes to one of the `useEffect`s.

kswenson (approved):
👍 Changes LGTM

PR #964, opened 2023-10-30
chore: update dependencies

kswenson:
Notes:
- @pjanik :` mathjs` remains pinned at  v11.11.2 because v11.12.0 introduced an escaping-related change which breaks the cypress tests. There's a comment which indicates the change is responding to one of your comments, so hopefully it's an improvement.
- @tejal-shah: I tried updating to cypress 13 and it worked locally but failed on the CI build, so I backed down to v12.17.


pjanik:
Thank you. I had planned to check v11.11.2, but I didn't get to it. I will do it soon.

PR #966, opened 2023-10-31
Improve lookupByKey and lookupByIndex errors handling

pjanik:
So, this PR came about because of a bug reported by @bfinzer: [lookupByIndex was causing an uncaught exception #186293413](https://www.pivotaltracker.com/story/show/186293413).

The main reason for the bug was that I never really set up proper error handling for the lookup functions. So, I've added another story that goes beyond just fixing this one exception: [Dealing with user errors in lookupByIndex and lookupByKey #186354628](https://www.pivotaltracker.com/story/show/186354628).

You might have noticed that I got rid of the `validateArguments` function that I added a while back. Originally, I thought it could be used for handling errors in these functions, as some errors can be spotted before the function actually runs, like the need for string constants as arguments. I tried to stick to this idea, but it required a completely different approach compared to other functions, which usually only flag errors during evaluation. For example, "mean(Height)" only realizes that "Height" is giving unsupported values when it evaluates "Height."

So, I decided to put error handling inside the evaluation process to keep things consistent with how other functions handle incorrect arguments. This way, I don't need a separate code path or the `validateArguments` helper.

To make this work, I also had to tweak some other code to be more forgiving and not crash until the formula is actually evaluated.

Error messages attempt to reuse existing V2 strings, but I ended up adding a new one. It could look something like: "The lookupByIndex function expects one argument to be a string constant," so it doesn't include ordinal numbers. However, I skipped those on purpose to keep things simple. After all, it's just for one error message that probably won't pop up too often or for too long.

Unit tests are coming in a separate PR, as they led to some more extensive work (https://github.com/concord-consortium/codap/pull/967)

### DIFF
--- a/v3/src/models/formula/functions/lookup-functions.ts
+++ b/v3/src/models/formula/functions/lookup-functions.ts
@@ -5,21 +5,21 @@ import { UNDEF_RESULT, equal, evaluateNode } from "./function-utils"
 import type { IDataSet } from "../../data/data-set"
 import t from "../../../utilities/translation/translate"
 
-type LookupStringContantArg = ConstantNode<string> | undefined
+type LookupStringConstantArg = ConstantNode<string> | undefined
 
 export const lookupFunctions = {
   // lookupByIndex("dataSetName", "attributeName", index)
   lookupByIndex: {
     getDependency: (args: MathNode[]): ILookupDependency => {
+      const [dataSetNameArg, attrNameArg] = args as LookupStringConstantArg[]
       return {
         type: "lookup",
-        dataSetId: rmCanonicalPrefix((args[0] as LookupStringContantArg)?.value),
-        attrId: rmCanonicalPrefix((args[1] as LookupStringContantArg)?.value),
+        dataSetId: rmCanonicalPrefix(dataSetNameArg?.value),
+        attrId: rmCanonicalPrefix(attrNameArg?.value),
       }
     },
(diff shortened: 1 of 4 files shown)


v3/src/models/formula/functions/lookup-functions.ts
```diff
@@ -3,78 +3,103 @@ import { FormulaMathJsScope } from "../formula-mathjs-scope"
...
       }
     },
     canonicalize: (args: MathNode[], displayNameMap: DisplayNameMap) => {
-      const validArgs = lookupFunctions.lookupByIndex.validateArguments(args)
-      const dataSetName = validArgs[0].value
-      const attrName = validArgs[1].value
-      validArgs[0].value = displayNameMap.dataSet[dataSetName]?.id
-      validArgs[1].value = displayNameMap.dataSet[dataSetName]?.attribute[attrName]
+      const dataSetNameArg = args[0] as LookupStringContantArg
+      const attrNameArg = args[1] as LookupStringContantArg
```
kswenson, line 22: Could also use:
```suggestion
      const [dataSetNameArg, attrNameArg] = args as LookupStringConstantArg[]
```
And same could be done in other similar functions. Note type fix as well.

kswenson (approved):
👍 Looks good

v3/src/models/formula/functions/lookup-functions.ts
pjanik: Thanks, fixed in the last commit.

PR #967, opened 2023-10-31
 Formula testing utils and unit tests for lookup and aggregate functions

pjanik:
This PR is a follow-up to https://github.com/concord-consortium/codap/pull/966. It extends tests to new areas that weren't tested before.

So, I added a new PT story:
[Develop testing patterns, helpers, and unit tests for more advanced formulas like aggregate and lookup functions #186374843](https://www.pivotaltracker.com/story/show/186374843)

Before, I only had a pattern for basic arithmetic functions that don't rely on the custom MathJS scope. However, lookup functions clearly depend on datasets, and aggregate functions use custom scope features for caching.

I considered a few options and decided to use a predefined document with some datasets and global values, along with some helpers to streamline the initialization process for the unit tests.

Mocking DataSet would be too cumbersome, and relying too much on the internal details of the formula manager and scope doesn't seem ideal.

Creating datasets manually is possible, but I found it to be quite cumbersome. Additionally, as more tests are added, it often turns out that the initial small dataset is insufficient, and new cases or attributes need to be added. I also found the tests hard to read, as the formulas and their results don't always make much sense if you mostly work with foo and bars. :wink:

Working with the existing Mammals (or Mammals and Cats) example works much better for me. I can jump right into writing the tests without worrying about setting up the dataset each time. Since I use it for development regularly, it's easier to read these tests and understand the formulas and their expected results.

I've added tests for the lookups function, aggregate functions, `getDisplayNameMap`, and `getCanonicalNameMap` helpers, which were refactored, as well as some simple tests for the test helpers themselves.


v3/src/components/slider/slider-registration.ts
```diff
@@ -15,14 +13,10 @@ import { SliderModel } from "./slider-model"
 import { SliderTitleBar } from "./slider-title-bar"
 import { AnimationDirections, AnimationModes, kDefaultAnimationDirection, kDefaultAnimationMode } from "./slider-types"
 import SliderIcon from '../../assets/icons/icon-slider.svg'
+import { getGlobalValueManager } from "../../models/tiles/tile-environment"
```
kswenson, line 16: Move import ⬆️ with the other `../../models/tiles/...` imports.

kswenson (approved):
👍 Nice!

v3/src/components/slider/slider-registration.ts
pjanik: Fixed in the last commit.

PR #968, opened 2023-10-31
Upgrade to Cypress 13

tejal-shah:


kswenson (changes requested):
Note the failing tests in the CI build. That's what I ran into when I tried to update to v13.

tejal-shah:
Oh okay. Looking into that now.

tejal-shah:
I finally got these to work @kswenson . It turns out we were running our cypress tests on Electron browser and there are many `click` issues for Cypress 13 on Electron. When I made the switch to Chrome, they all magically disappeared, except for one failure, which I fixed in the adornment tests.

kswenson (approved):
👍 Looks good!

PR #969, opened 2023-10-31
Update MathJS to v12, update formula code to work with fixed special chars in MathJS, add unit tests

pjanik:
@kswenson mentioned that MathJS v11.12.0 is causing Cypress tests to break: https://github.com/concord-consortium/codap/pull/964

This issue happened due to a recent MathJS bug fix:
- https://github.com/josdejong/mathjs/issues/3073#issuecomment-1768326862
- https://github.com/josdejong/mathjs/pull/3082

This PR removes some lines from formula-utils.ts that are no longer necessary. Additionally, I've added unit tests that would fail with MathJS < 11.12.0 and the new code, providing us with double coverage (not just relying on the Cypress tests that saved me before, thanks to @tejal-shah :wink:).

### DIFF
--- a/v3/package-lock.json
+++ b/v3/package-lock.json
@@ -19,7 +19,7 @@
         "escape-string-regexp": "^5.0.0",
         "leaflet": "^1.9.4",
         "lodash": "^4.17.21",
-        "mathjs": "~11.11.2",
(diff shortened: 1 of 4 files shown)


kswenson (approved):
👍 LGTM

PR #973, opened 2023-11-02
Show plotted value formula errors in UI

pjanik:
https://www.pivotaltracker.com/story/show/186358491

This PR adds a new `<div>` below the plotted value "value=..." box to show potential errors. These errors are mostly the same as those in the case table context, but there's one that's specific to the plotted value context, which was actually introduced as part of another PR:
Screenshot 2023-11-02 at 16 55 59


pjanik:
The failing Cypress test doesn't seem to be related to this PR.

I've restarted the last build in the main branch, and it... failed:
https://github.com/concord-consortium/codap/actions/runs/6724645673
So, it seems that the v13 Cypress issue fix that worked yesterday no longer works today (?).

emcelroy (approved):
Looks good 👍 There's a failed Cypress run here, but maybe that was a fluke? I checked out this branch locally and ran the tests and they passed there.

pjanik:
@emcelroy, I commented on the failing test above. Probably we both posted around the same time. :)

emcelroy:
> @emcelroy, I commented on the failing test above. Probably we both posted around the same time. :)

Yeah, you left that comment while I was doing my review, but I didn't see it till right after I left my approval. Sorry for any confusion. Anyway, it seems like we can definitely ignore that failure here.

v3/src/components/graph/adornments/univariate-measures/plotted-value/plotted-value-adornment-model.ts
```diff
@@ -9,7 +9,8 @@ export const PlottedValueAdornmentModel = UnivariateMeasureAdornmentModel
   .props({
     type: types.optional(types.literal(kPlottedValueType), kPlottedValueType),
     formula: types.optional(Formula, () => Formula.create()),
-    labelTitle: types.optional(types.literal(kPlottedValueValueTitleKey), kPlottedValueValueTitleKey)
+    labelTitle: types.optional(types.literal(kPlottedValueValueTitleKey), kPlottedValueValueTitleKey),
+    error: types.optional(types.string, "")
```
kswenson, line 13: In this case, the simpler form should suffice:
```suggestion
    error: ""
```

kswenson (approved):
👍 LGTM

PR #974, opened 2023-11-03
Fix adornment test

tejal-shah:
@kswenson , @emcelroy , There are a few (6) lines in the adornment cypress tests that are flaky and I tried to fix them but the coupel things that I tried didn't seem to get rid of the flakiness of that part of the code. I'm commenting this part of the tests for now and am creating a new story for adding it back and getting rid of the flakiness since we don't want to block/delay deployments because of this.


emcelroy (approved):
Looks good -- thanks 👍

PR #975, opened 2023-11-03
Add support of the default argument to the formula system and use it in Plotted Value formulas

pjanik:
https://www.pivotaltracker.com/story/show/186358510.

This PR introduces the concept of default arguments to the formula system and integrates them into the Plotted Value formulas.

There are two key areas where significant changes have occurred:

1. **Observers**: The default argument is now treated as a new dependency, even when it's not explicitly defined. For instance, if a user simply uses `mean()` and the default argument is set to `Height`, the system will now recalculate the formula whenever any of the Height values change.

2. **Evaluation**: This aspect is more straightforward. The default argument has become a part of the FormulaMathJSScope. Functions do not require manual handling of default arguments. This handling is automatic. The only thing each function needs to specify is the number of required arguments. Without this information, it becomes impossible to determine whether the default argument should be used. In theory, we could always push the default argument into the arguments array, which would work for evaluation, as each function could just ignore this last argument. However, it does not align well with observing. In this scenario, we wouldn't be able to determine whether the default argument needs to be observed or not.

I've also included a few new helpers, did some refactoring, and added unit tests.

However, there is one outstanding issue that can be triggered by dragging a new numeric attribute to the X-axis, essentially updating the default argument. In this situation, calculations do not proceed correctly. I'm working on a solution and will open a separate PR to address this issue.

### DIFF
--- a/v3/src/models/formula/formula-manager.ts
+++ b/v3/src/models/formula/formula-manager.ts
@@ -2,7 +2,7 @@ import { comparer, makeObservable, observable, reaction, action } from "mobx"
 import { isAlive } from "mobx-state-tree"
 import { ICase } from "../data/data-set-types"
 import {
-  getFormulaDependencies, formulaError, getDisplayNameMap, getCanonicalNameMap
+  getFormulaDependencies, formulaError, getDisplayNameMap, getCanonicalNameMap, displayToCanonical
 } from "./formula-utils"
 import { CaseList } from "./formula-types"
 import { IDataSet } from "../data/data-set"
@@ -28,6 +28,7 @@ export interface IFormulaMetadata {
 export interface IFormulaExtraMetadata {
   dataSetId: string
   attributeId?: string
+  defaultArgument?: string
 }
 
 export interface IFormulaContext {
@@ -45,11 +46,10 @@ export interface IFormulaAdapterApi {
 export interface IFormulaManagerAdapter {
   type: string
   getAllFormulas: () => ({ formula: IFormula, extraMetadata?: any })[]
-  recalculateFormula: (formulaContext: IFormulaContext, extraMetadata: any,
-    casesToRecalculateDesc?: CaseList) => void
-  setupFormulaObservers: (formulaContext: IFormulaContext, extraMetadata: any) => () => void
-  getFormulaError: (formulaContext: IFormulaContext, extraMetadata: any) => undefined | string
+  recalculateFormula: (formulaContext: IFormulaContext, extraMetadata: any, casesToRecalculateDesc?: CaseList) => void
   setFormulaError: (formulaContext: IFormulaContext, extraMetadata: any, errorMsg: string) => void
+  getFormulaError: (formulaContext: IFormulaContext, extraMetadata: any) => undefined | string
+  setupFormulaObservers?: (formulaContext: IFormulaContext, extraMetadata: any) => () => void
 }
 
 export class FormulaManager {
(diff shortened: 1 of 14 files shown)


v3/src/models/formula/formula-utils.ts
```diff
@@ -88,36 +93,37 @@ export const getDisplayNameMap = (options: IDisplayNameMapOptions, useSafeSymbol
...
-    ...mapAttributeNames(localDataSet, LOCAL_ATTR, useSafeSymbolNames),
-    // caseIndex is a special name supported by formulas. It essentially behaves like a local data set attribute
-    // that returns the current, 1-based index of the case in its collection group.
-    caseIndex: `${CANONICAL_NAME}${LOCAL_ATTR}${CASE_INDEX_FAKE_ATTR_ID}`
-  }
+  const key = (name: string, _useSafeSymbolNames = useSafeSymbolNames) =>
+    nonEmptyName(_useSafeSymbolNames ? safeSymbolName(name) : name)
 
+  // When localNames are generated, the order of processing various sources of names is important.
+  // The last last source would provide the final canonical name for the symbol. So, currently the global values
```
kswenson, line 100:
```suggestion
  // The last source would provide the final canonical name for the symbol. So, currently the global values
```

v3/src/models/formula/formula-utils.ts
```diff
@@ -317,12 +324,21 @@ export const getFormulaDependencies = (formulaCanonical: string, formulaAttribut
...
+    if (functionInfo) {
+      // Some functions have special kind of dependencies that need to be calculated in a custom way
+      // (eg. lookupByIndex, lookupByKey).
+      if (functionInfo.getDependency) {
+        const dependency = functionInfo.getDependency(node.args)
+        if (dependency) {
+          result.push(dependency)
+        }
+      }
+      // When default argument is provided and the function has less arguments than required, we need to visit the
```
kswenson, line 338:
```suggestion
      // When default argument is provided and the function has fewer arguments than required, we need to visit the
```

v3/src/models/formula/formula-utils.test.ts
```diff
@@ -353,6 +358,54 @@ describe("getDisplayNameMap", () => {
       }
     })
   })
+
+  describe("when there are local attributes or globals with the name 'caseIndex'", () => {
+    it("resolves 'caseIndex' to a special value (special value take precedence over anything else)", () => {
```
kswenson, line 363:
```suggestion
    it("resolves 'caseIndex' to a special value (special value takes precedence over anything else)", () => {
```

v3/src/models/formula/formula-utils.test.ts
```diff
@@ -400,3 +453,21 @@ describe("getCanonicalNameMap", () => {
     })
   })
 })
+
+describe("localAttrIdToCanonical", () => {
+  it("returns a string that is recognized parseBasicCanonicalName as a local attribute dependency", () => {
```
kswenson, line 458:
```suggestion
  it("returns a string that is recognized by parseBasicCanonicalName as a local attribute dependency", () => {
```

v3/src/models/formula/formula-utils.test.ts
```diff
@@ -400,3 +453,21 @@ describe("getCanonicalNameMap", () => {
...
+    expect(parseBasicCanonicalName(localAttrIdToCanonical("foo"))).toEqual({ type: "localAttribute", attrId: "foo" })
+  })
+})
+
+describe("globalValueIdToCanonical", () => {
+  it("returns a string that is recognized parseBasicCanonicalName as a global value dependency", () => {
```
kswenson, line 464:
```suggestion
  it("returns a string that is recognized by parseBasicCanonicalName as a global value dependency", () => {
```

v3/src/models/formula/formula-manager.ts
```diff
@@ -45,11 +44,10 @@ export interface IFormulaAdapterApi {
...
   type: string
   getAllFormulas: () => ({ formula: IFormula, extraMetadata?: any })[]
-  recalculateFormula: (formulaContext: IFormulaContext, extraMetadata: any,
-    casesToRecalculateDesc?: CaseList) => void
-  setupFormulaObservers: (formulaContext: IFormulaContext, extraMetadata: any) => () => void
-  getFormulaError: (formulaContext: IFormulaContext, extraMetadata: any) => undefined | string
+  recalculateFormula: (formulaContext: IFormulaContext, extraMetadata: any, casesToRecalculateDesc?: CaseList) => void
   setFormulaError: (formulaContext: IFormulaContext, extraMetadata: any, errorMsg: string) => void
+  getFormulaError: (formulaContext: IFormulaContext, extraMetadata: any) => undefined | string
+  setupFormulaObservers?: (formulaContext: IFormulaContext, extraMetadata: any) => () => void
```
kswenson, line 50: Can we do better than `any` for the `extraMetadata` types here?

kswenson (approved):
👍 Looks good -- nice work. Just a smattering of suggestions.

PR #976, opened 2023-11-03
fix: graph response to delete case table along with undo/redo

kswenson:
Fix: [[#186383038]](https://www.pivotaltracker.com/story/show/186383038)

There were several aspects to getting this to work:
1. The `DataConfigurationModel` needs to update its internal state, e.g. the `filteredCases` array, when its `dataset` is changed. Previously, this was accomplished by having calls to `setDataset()` internally call `_setDataset()`, which had code to update the necessary internal state. But the `dataset` can be changed by other means than directly calling `setDataset()`, e.g. by deleting it from the shared model manager. Instead, we install a MobX `reaction` in `afterCreate()` which observes the `dataset` property and performs the necessary updates whenever it changes, thus eliminating the need to trap calls to `setDataset()` directly. @bgoldowsky This change will probably benefit the CLUE code as well.
2. The `Graph` component installs a MobX `reaction` that listens specifically to the number of elements in the `filteredCases` array. This guarantees that it responds _after_ the `DataConfigurationModel` has responded to the change in `dataset`. This `reaction` calls the new `GraphController.clearGraph()` method in the specific case that the `filteredCases` array becomes empty (and we're not in the middle of an undo/redo operation). In all other cases it calls the existing `GraphController.callMatchCirclesToData()` method, which guarantees that this critical method gets called whenever the `filteredCases` array changes.
3. The new `GraphController` method `clearGraph()` handles the case of returning a graph to its default/unlinked state. It complements the existing `initializeGraph()` (for initializing a graph whose models are up to date, e.g. after deserialization) and `handleAttributeAssignment()` (for adding/removing an attribute to/from a graph).

Aside from the obvious benefit of having graphs handle deletion of their case table/data set correctly, this has the added benefit of allowing this deletion to be undoable, which is an improvement over v2.

Along the way, this PR also fixes a smattering of other issues encountered along the way:
1. enhances a `console.warn` message in the `useSubAxis` hook.
2. adds a missing dependency array to a `useEffect` in the `Graph` component
3. removes an unnecessary dependency from another `useEffect` in the `Graph` component
4. reduces the size of the dependency array from a `useEffect` in the `useGraphModel` hook
5. fixes a bug in the `isChildOfUndoRedo()` utility function

### DIFF
--- a/.github/workflows/v3.yml
+++ b/.github/workflows/v3.yml
@@ -47,7 +47,8 @@ jobs:
       fail-fast: false
       matrix:
         # run multiple copies of the current job in parallel
-        containers: [1, 2, 3]
+        # containers: [1, 2, 3]
+        containers: [1]
     steps:
       - name: Checkout
         uses: actions/checkout@v3
--- a/v3/src/components/axis/hooks/use-sub-axis.ts
+++ b/v3/src/components/axis/hooks/use-sub-axis.ts
@@ -57,7 +57,7 @@ export const useSubAxis = ({
     renderSubAxis = useCallback(() => {
       const _axisModel = axisProvider.getAxis?.(axisPlace)
       if (!isAliveSafe(_axisModel)) {
-        console.warn("useSubAxis.renderSubAxis skipping rendering of defunct axis model")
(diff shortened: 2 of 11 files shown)


bfinzer (approved):
👍🏻LGTM

PR #977, opened 2023-11-06
Fix evaluation of custom operators

pjanik:
I didn't notice that https://github.com/concord-consortium/codap/pull/975 broke some functionality.
I thought I could completely replace `evaluate` with `evaluateRaw` but apparently MathJS is calling operator functions in a slightly different way. So, I still need to have basic evaluate function available and don't transform it into the evaluateRaw variant. To keep it more descriptive, I've called it `evaluateOperator` and added `isOperator` flag.


kswenson (approved):
👍 LGTM

PR #978, opened 2023-11-06
Recalculate plotted value formula on graph attribute updates

pjanik:
https://www.pivotaltracker.com/story/show/186358510

This PR ensures that plotted value formulas are recalculated when the user drags a new attribute to one of the attribute slots. Now, `extraMetadata` is also used to trigger the re-registration and recalculation of a formula. This change was necessary for default argument handling.

To handle other graph axes (e.g. user dragging a category to the top or right axis), I could add something like that:
```javascript
  setupFormulaObservers(formulaContext: IFormulaContext, extraMetadata: IPlottedValueFormulaExtraMetadata) {
    const graphContentModel = this.getGraphContentModel(extraMetadata)
    return reaction(() => getGraphCellKeys(graphContentModel),
    () => {
      this.recalculateFormula(formulaContext, extraMetadata)
    })
  }
```

But we would recalculate the formula twice when the default argument is updated - once because of this reaction and once because of the extra metadata update. Of course, I could calculate the defaultArgument in the reaction and not trigger recalculation, but I've added graphCellKeys to the extra metadata just to keep things simpler and together.

The last commit is not really related to plotted value, but also related to observers, so I just added it to this PR.

### DIFF
--- a/v3/src/models/formula/attribute-formula-adapter.ts
+++ b/v3/src/models/formula/attribute-formula-adapter.ts
@@ -216,7 +216,8 @@ export class AttributeFormulaAdapter implements IFormulaManagerAdapter {
   }
 
   setupFormulaObservers(formulaContext: IFormulaContext, extraMetadata: IAttrFormulaExtraMetadata) {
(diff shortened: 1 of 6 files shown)


kswenson (approved):
👍 LGTM

PR #980, opened 2023-11-07
fix: Users can plot a function on a scatterplot (PT-181940062)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/181940062

These changes eliminate problems causing duplicate cellKeys to be generated for graphs with multiple subplots. The duplicate cellKeys were caused by a few separate issues: 1) the adornment model's `cellKey` view wasn't always using the correct index values for getting category values, 2) the adornment model's `cellKey` view wasn't using the duplicate-preventing `__IMPOSSIBLE__` key used by the adornment component's `updateCellKey` function, and 3) the latter wasn't correctly using `__IMPOSSIBLE__` as a _key_. They also move `updateCellKey` to a new `adornment-utils.ts` file so it can be used by both the component and model.

### DIFF
--- a/v3/src/components/graph/adornments/adornment-utils.ts
+++ b/v3/src/components/graph/adornments/adornment-utils.ts
@@ -1,7 +1,7 @@
 export const updateCellKey = (cellKey: Record<string, string>, attrId: string, cat: string) => {
   const newCellKey = { ...cellKey }
   if (cat) {
-    const propertyAlreadyPresent = Object.keys(newCellKey).includes(attrId)
+    const propertyAlreadyPresent = Object.prototype.hasOwnProperty.call(newCellKey, attrId)
     if (propertyAlreadyPresent && newCellKey[attrId] !== cat) {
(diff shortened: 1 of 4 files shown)


v3/src/components/graph/adornments/adornment-utils.ts
```diff
@@ -0,0 +1,15 @@
+export const updateCellKey = (cellKey: Record<string, string>, attrId: string, cat: string) => {
+  const newCellKey = { ...cellKey }
+  if (cat) {
+    const propertyAlreadyPresent = Object.keys(newCellKey).includes(attrId)
```
kswenson, line 4:
```suggestion
    const propertyAlreadyPresent = Object.prototype.hasOwnProperty.call(newCellKey, attrId)
```

kswenson (approved):
👍 Looks good -- just one suggestion to use `hasOwnProperty` rather than iterating through all properties when checking for duplicates.

PR #981, opened 2023-11-07
Add unit tests to check removal and addition of symbol name

pjanik:
https://www.pivotaltracker.com/story/show/186252555

I've fixed the actual bug while working on another task, but I'm just adding a few more unit tests to cover the scenario that used to cause some issues for Tejal.


kswenson (approved):
👍 LGTM

PR #982, opened 2023-11-07
Do not serialize values computed by formulas, make formula attrs non-editable

pjanik:
https://www.pivotaltracker.com/story/show/186424226

This PR ensures that values computed by formulas will not be serialized in both development and production modes. I've also modified the top-level attribute file comment because, unfortunately, prepareSnapshot() and completeSnapshot() are no longer no-ops in production.

Additionally, I've added a semi-related feature (mostly due to it being overdue for a long time) - formula cells are now non-editable.

### DIFF
--- a/v3/src/models/data/attribute.ts
+++ b/v3/src/models/data/attribute.ts
@@ -167,6 +167,9 @@ export const Attribute = types.model("Attribute", {
   get format() {
     return self.precision != null ? `.${self.precision}~f` : kDefaultFormatStr
   },
+  get isEditable() {
+    return self.editable && self.formula.empty
+  },
   value(index: number) {
     return self.strValues[index]
   },
@@ -211,7 +214,6 @@ export const Attribute = types.model("Attribute", {
   },
   setDisplayFormula(displayFormula: string) {
     self.formula.setDisplayFormula(displayFormula)
-    this.setEditable(displayFormula === "")
   },
   addValue(value: IValueType = "", beforeIndex?: number) {
     const strValue = self.importValue(value)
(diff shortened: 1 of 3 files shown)


v3/src/models/data/attribute.ts
```diff
@@ -192,10 +207,11 @@ export const Attribute = types.model("Attribute", {
     self.editable = editable
   },
   clearFormula() {
-    self.formula.setDisplayFormula("")
+    this.setDisplayFormula("")
   },
   setDisplayFormula(displayFormula: string) {
     self.formula.setDisplayFormula(displayFormula)
+    this.setEditable(displayFormula === "")
```
kswenson, line 214: The `editable` property is meant to correspond to the user's ability to specify editability. I think for this purpose a better solution would be to add a view like:
```
  get isEditable() {
    return self.editable && !self.displayFormula
  }
```

kswenson (changes requested):
👍 Looks good -- I hadn't thought about the implications of formulas for `preapreSnapshot`/`completeSnapshot`, but I'm glad that supporting them wasn't too difficult. The one requested change is to the handling of editability -- see comment elsewhere.

pjanik:
Thank you, @kswenson. It definitely makes sense. I've just pushed a commit that adds this new view. I'm requesting a re-review just to ask if you can think of any other part of the app, apart from the user cells, that should now use "isEditable" instead of "editable"?

kswenson (approved):
>I'm requesting a re-review just to ask if you can think of any other part of the app, apart from the use-cells, that should now use "isEditable" instead of "editable"?

That's the only one that comes to mind.

PR #983, opened 2023-11-07
fix: select tile on drop [#186382997]

kswenson:
[[#186382997]](https://www.pivotaltracker.com/story/show/186382997)


bfinzer (approved):
👍🏻LGTM
Nice!

PR #985, opened 2023-11-08
Maps display boundaries + drop legend attributes

bfinzer:
* This is another large PR with some refactoring. In order to test, download this codap3 file: https://drive.google.com/file/d/17_ncDq0saVBoeLws2zdkpsW-xS60EAW0/view?usp=drive_link

Things to try:
* Hover to get data tips
* Selection of points and polygons
* Drop legend attributes from each of the datasets

[#186304968] Feature: In the presence of datasets with a boundary attribute, maps include a boundary layer showing polygons

* Search for boundary attributes in available datasets and construct a MapPolygonLayerModel for each one found
* Implement MapPolygonLayer component
* Store features as volatile property of MapPolygonLayer
* Include MapPolygonLayers in MapInterior component
* Some renaming of point-specific routines to more obviously apply to polygons as well
* Include overrides to Leaflet css to conform to CODAP styling of data tips
* Implement responses to user hover and click on polygons
* Include boundaries in fitting map to data
* Detect attribute drop over map as intention to provide legend attribute
* In MapContentModel implement setLegendAttributeID and canAcceptAttributeIDDrop
* In MapPolygonLayer respond to changes in dataconfiguration's legend id
* Refactoring in preparation for adding Legends to maps
* Bring 'map-dot-area' above map using z-index
* Fix to map point layer's reaction to changes in displayChangeCount
* A map with both a polygon layer and a point layer functions properly wrt hover and click
* Responds to drop of legend attribute for both point and polygon layers


v3/src/components/graph/adornments/movable-line/movable-line.tsx
kswenson, line 1: Empty file should be removed.

kswenson (changes requested):
👍 Looks good -- sorry it took so long to turn this around. I added a few commits which fix some remaining dependency issues. The remaining comments are relatively minor and unrelated to the refactoring so I left them for you.

kswenson (approved):
👍 Code review changes look good

PR #986, opened 2023-11-08
Reset case values evaluated by formula before moving attr to a different collection

pjanik:
https://www.pivotaltracker.com/story/show/186323665

This PR resets formula-evaluated values to blank right before the attribute is moved to another collection. This prevents the grouping logic from considering cases that have been evaluated using the formula.

Note that all I wanted to do in this PR is to add literally 4 lines to `setCollectionForAttribute`:
```javascript
setCollectionForAttribute(attributeId: string, options?: IMoveAttributeCollectionOptions) {
  // (...)
  if (!attribute.formula.empty) {
    // If the attribute has a formula, we need to reset all the calculated values to blank values so that they
    // are not taken into account while calculating case grouping. After the grouping is done, the formula will
    // be re-evaluated, and the values will be updated to the correct values again.
    const cases = getCasesForCollection(oldCollection?.id).map(aCase => ({ ...aCase, [attributeId]: "" }))
    this.setCaseValues(cases)
  }
}
```

However, this resulted in moving around 175 lines, involving the switching from `self` to `this`, or vice versa, and direct use of "private" variables. And all this re-shuffling of things actually took me an amount of time I'm ashamed to disclose. :wink: Also, I did what I consider to be the minimal reshuffling, so probably the new layers/blocks may not necessarily be logically grouped anymore. I assumed that if my "primitive" action calls a more "intermediate" action or view, I need to move these more "intermediate" actions up in the file and make them more "primitive." Not the other way around (assuming this would result in less reshuffling, but it's somewhat of a blind guess).

While I believe the concept of using layers or blocks can work well for smaller files, in this case, it gets a bit challenging. As the file size grows, it becomes difficult to keep track of the number of blocks, groupings, and overall structure, and to understand the intended purpose of these layers or blocks. Of course, one separate suggestion could be to divide this file, but let's assume for now that this is not always possible, and sometimes we have to deal with larger files.

I think having a pattern that doesn't require any thinking about all that would be great. At some point, all the reshuffling and the uncertainty of whether I did it correctly or not (fortunately, this file is well-tested, but that might not always be the case) might outweigh the benefits of file/text-level grouping. Also, we can achieve similar grouping simply by keeping methods close to each other and using comments (if we have the freedom to lay them out as we want).

### DIFF
--- a/v3/src/models/data/data-set.ts
+++ b/v3/src/models/data/data-set.ts
@@ -256,7 +256,7 @@ export const DataSet = types.model("DataSet", {
   get ungroupedAttributes(): IAttribute[] {
     const grouped = new Set(self.groupedAttributes.map(attr => attr.id))
     return self.attributes.filter(attr => attr && !grouped.has(attr.id))
-  },
+  }
 }))
 .extend(self => {
   // we do our own caching because MST's auto-caching wasn't working as expected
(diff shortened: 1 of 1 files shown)


v3/src/models/data/data-set.ts
```diff
@@ -409,13 +376,13 @@ export const DataSet = types.model("DataSet", {
...
-                                      ? newCollectionGroups[newCollectionGroups.length - 1]
-                                      : undefined
+          ? newCollectionGroups[newCollectionGroups.length - 1]
+          : undefined
         if (lastCollectionGroup) {
           // if there are collections, then child cases are determined by the parents
           lastCollectionGroup?.groups.forEach(group => {
             _childCases.push(...group.childCaseIds.map((caseId, index) =>
-              ({__id__: caseId, [symParent]: group.pseudoCase.__id__, [symIndex]: index })))
+              ({ __id__: caseId, [symParent]: group.pseudoCase.__id__, [symIndex]: index })))
```
pjanik, line 385: Ops, these whitespace changes were done by VSC auto-formatter when I used it to format the code in the new block. I can revert the old one if it's a more common style in this repo.

v3/src/models/data/data-set.ts
```diff
@@ -464,6 +592,13 @@ export const DataSet = types.model("DataSet", {
         const newCollection = options?.collection ? getGroupedCollection(options.collection) : undefined
         const oldCollection = getCollectionForAttribute(attributeId)
         if (attribute && oldCollection !== newCollection) {
+          if (!attribute.formula.empty) {
+            // If the attribute has a formula, we need to reset all the calculated values to blank values so that they
+            // are not taken into account while calculating case grouping. After the grouping is done, the formula will
+            // be re-evaluated, and the values will be updated to the correct values again.
+            const cases = getCasesForCollection(oldCollection?.id).map(aCase => ({ ...aCase, [attributeId]: "" }))
+            this.setCaseValues(cases)
```
kswenson, line 600: Do we need to use `DataSet` API for this? What if there were a `clearValues()` method on the `Attribute` model that accomplished the same thing? I wouldn't think we want/need external clients of the `DataSet` responding to this call to `setCaseValues()` anyway.

v3/src/models/data/data-set.ts
kswenson: Preferences differ on whether there should be spaces inside braces so we allow both, but it should be consistent between the beginning and ending braces of the same construct, so I think this change is fine.

v3/src/models/data/data-set.ts
pjanik: It's a good point. I had concerns about using some custom methods because I wasn't sure about their safety regarding dataset caching, validations, and other potential helper structures that I might not be aware of yet. I also had doubts about whether the grouping would be calculated correctly. That's why I had used the existing API just to be safe.

However, if you are suggesting this approach, I agree that it sounds like a better option. It involves fewer changes to the DataSet, and it might be faster since there won't be any unnecessary client reactions to the extra 'setCaseValues' call.

I'll go ahead and make that change. Thanks!

pjanik:
@kswenson, I'm going to close this PR and open a new one instead:
https://github.com/concord-consortium/codap/pull/988

PR #987, opened 2023-11-08
feat: User can add Least Squares Line to scatterplot (PT-181939979)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/181939979

### DIFF
--- a/v3/src/components/graph/adornments/lsrl/lsrl-adornment-component.tsx
+++ b/v3/src/components/graph/adornments/lsrl/lsrl-adornment-component.tsx
@@ -73,6 +73,10 @@ export const LSRLAdornment = observer(function LSRLAdornment(props: IProps) {
   const lineObjectsRef = useRef<ILineObject[]>([])
   const pointsOnAxes = useRef<IAxisIntercepts>({pt1: {x: 0, y: 0}, pt2: {x: 0, y: 0}})
 
+  const getLines = useCallback(() => {
+    return dataConfig && model.getLines(xAttrId, yAttrId, cellKey, dataConfig)
+  }, [cellKey, dataConfig, model, xAttrId, yAttrId])
+
   const fixEndPoints = useCallback((iLine: Selection<SVGLineElement, unknown, null, undefined>) => {
     if (
       !Number.isFinite(pointsOnAxes.current.pt1.x) || !Number.isFinite(pointsOnAxes.current.pt2.x) ||
@@ -122,7 +126,7 @@ export const LSRLAdornment = observer(function LSRLAdornment(props: IProps) {
         .style("top", `${top}px`)
 
       if (isFinished) {
-        const lines = dataConfig && model.getLines(xAttrId, yAttrId, cellKey, dataConfig)
+        const lines = getLines()
         if (!lines) return
         // Get the percentage of plot width and height of the equation box's coordinates
         // for a more accurate placement of the equation box.
@@ -134,10 +138,10 @@ export const LSRLAdornment = observer(function LSRLAdornment(props: IProps) {
         )
       }
     }
-  }, [cellKey, dataConfig, equationContainerSelector, graphModel, model, plotHeight, plotWidth, xAttrId, yAttrId])
+  }, [equationContainerSelector, getLines, graphModel, plotHeight, plotWidth])
 
   const updateEquations = useCallback(() => {
-    const lines = dataConfig && model.getLines(xAttrId, yAttrId, cellKey, dataConfig)
+    const lines = getLines()
     if (!lines) return
     const equationDiv = select(equationContainerSelector)
     equationDiv.style("width", `${plotWidth}px`)
@@ -170,7 +174,7 @@ export const LSRLAdornment = observer(function LSRLAdornment(props: IProps) {
           .style("top", `${top}%`)
       }
     }
-  }, [cellKey, dataConfig, equationContainerSelector, model, plotHeight, plotWidth, showConfidenceBands,
+  }, [cellKey, dataConfig, equationContainerSelector, getLines, model, plotHeight, plotWidth, showConfidenceBands,
       xAttrId, xAttrName, xScale, xSubAxesCount, yAttrId, yAttrName, yScale, ySubAxesCount])
 
   const confidenceBandPaths = useCallback((caseValues: Point[], lineIndex: number) => {
@@ -211,7 +215,7 @@ export const LSRLAdornment = observer(function LSRLAdornment(props: IProps) {
   }, [cellKey, confidenceBandPaths, dataConfig, model, showConfidenceBands, toggleConfidenceBandTip, xAttrId, yAttrId])
 
   const updateLines = useCallback(() => {
-    const lines = dataConfig && model.getLines(xAttrId, yAttrId, cellKey, dataConfig)
+    const lines = getLines()
     if (!lines) return
 
     for (let lineIndex = 0; lineIndex < lineObjectsRef.current.length; lineIndex++) {
@@ -227,7 +231,7 @@ export const LSRLAdornment = observer(function LSRLAdornment(props: IProps) {
       lineObj.cover && fixEndPoints(lineObj.cover)
       updateConfidenceBands(lineIndex, line)
     }
-  }, [dataConfig, model, xAttrId, yAttrId, cellKey, xAxis, yAxis, fixEndPoints, updateConfidenceBands])
+  }, [getLines, xAxis, yAxis, fixEndPoints, updateConfidenceBands])
 
   const updateLSRL = useCallback(() => {
     updateLines()
@@ -235,7 +239,7 @@ export const LSRLAdornment = observer(function LSRLAdornment(props: IProps) {
   }, [updateEquations, updateLines])
 
   useEffect(function createElements() {
-    const lines = dataConfig && model.getLines(xAttrId, yAttrId, cellKey, dataConfig)
+    const lines = getLines()
     if (!lines) return
 
     // Clear any previously added elements
@@ -324,9 +328,8 @@ export const LSRLAdornment = observer(function LSRLAdornment(props: IProps) {
     }
 
   }, [cellKey, classFromKey, confidenceBandPaths, containerId, dataConfig, equationContainerClass, fixEndPoints,
-      handleHighlightLineAndEquation, handleMoveEquation, instanceKey, model, plotHeight, plotWidth,
-      showConfidenceBands, xAttrId, xAxis, xCellCount, xScale, xSubAxesCount, yAttrId, yAxis, yCellCount, yScale,
-      ySubAxesCount])
+      getLines, handleHighlightLineAndEquation, handleMoveEquation, instanceKey, model, plotHeight, plotWidth,
+      showConfidenceBands, xAxis, xCellCount, xScale, xSubAxesCount, yAxis, yCellCount, yScale, ySubAxesCount])
 
   // Refresh values on axis changes
   useEffect(function refreshAxisChange() {
--- a/v3/src/components/graph/adornments/lsrl/lsrl-adornment-component.tsx
+++ b/v3/src/components/graph/adornments/lsrl/lsrl-adornment-component.tsx
@@ -9,7 +9,7 @@ import { Point } from "../../../data-display/data-display-types"
 import { INumericAxisModel } from "../../../axis/models/axis-model"
 import { IAxisIntercepts, curveBasis, lineToAxisIntercepts, lsrlEquationString } from "../../utilities/graph-utils"
 import { useGraphDataConfigurationContext } from "../../hooks/use-data-configuration-context"
-import { ILSRLAdornmentModel } from "./lsrl-adornment-model"
+import { ILSRLAdornmentModel, ILSRLInstance } from "./lsrl-adornment-model"
 import { useGraphContentModelContext } from "../../hooks/use-graph-content-model-context"
 
 import "./lsrl-adornment-component.scss"
@@ -122,7 +122,7 @@ export const LSRLAdornment = observer(function LSRLAdornment(props: IProps) {
         .style("top", `${top}px`)
 
       if (isFinished) {
-        const lines = model.lines.get(instanceKey)
+        const lines = dataConfig && model.getLines(xAttrId, yAttrId, cellKey, dataConfig)
         if (!lines) return
         // Get the percentage of plot width and height of the equation box's coordinates
         // for a more accurate placement of the equation box.
@@ -134,25 +134,26 @@ export const LSRLAdornment = observer(function LSRLAdornment(props: IProps) {
         )
       }
     }
-  }, [equationContainerSelector, graphModel, instanceKey, model.lines, plotHeight, plotWidth])
+  }, [cellKey, dataConfig, equationContainerSelector, graphModel, model, plotHeight, plotWidth, xAttrId, yAttrId])
 
   const updateEquations = useCallback(() => {
-    const lines = model.lines.get(instanceKey)
-    if (!lines || !dataConfig) return
+    const lines = dataConfig && model.getLines(xAttrId, yAttrId, cellKey, dataConfig)
+    if (!lines) return
     const equationDiv = select(equationContainerSelector)
     equationDiv.style("width", `${plotWidth}px`)
       .style("height", `${plotHeight}px`)
     for (let linesIndex = 0; linesIndex < lines.length; linesIndex++) {
-      const legendCat = dataConfig.categoryArrayForAttrRole("legend")?.[linesIndex]
-      const caseValues = model.getCaseValues(xAttrId, yAttrId, cellKey, dataConfig, legendCat)
-      // If there is one or fewer cases there is no equation to update
-      if (caseValues.length <= 1) continue
+      const category = lines[linesIndex].category
+      const caseValues = model.getCaseValues(xAttrId, yAttrId, cellKey, dataConfig, category)
+      const catColor = category && category !== "__main__" ? dataConfig?.getLegendColorForCategory(category) : undefined
       const { slope, intercept, rSquared } = lines[linesIndex]
+      if (!slope || !intercept || !rSquared) continue
       const screenX = xScale((pointsOnAxes.current.pt1.x + pointsOnAxes.current.pt2.x) / 2) / xSubAxesCount
       const screenY = yScale((pointsOnAxes.current.pt1.y + pointsOnAxes.current.pt2.y) / 2) / ySubAxesCount
       const attrNames = {x: xAttrName, y: yAttrName}
-      const string = lsrlEquationString(slope, intercept, rSquared, attrNames, caseValues, showConfidenceBands)
-      const equation = equationDiv.selectAll("p").filter(`:nth-child(${linesIndex + 1})`)
+      const string =
+        lsrlEquationString(slope, intercept, rSquared, attrNames, caseValues, showConfidenceBands, catColor)
+      const equation = equationDiv.select(`#lsrl-equation-${model.classNameFromKey(cellKey)}-${linesIndex}`)
       equation.html(string)
 
       // The equation may have been unpinned from its associated line if the user dragged it away from the line.
@@ -169,7 +170,7 @@ export const LSRLAdornment = observer(function LSRLAdornment(props: IProps) {
           .style("top", `${top}%`)
       }
     }
-  }, [cellKey, dataConfig, equationContainerSelector, instanceKey, model, plotHeight, plotWidth, showConfidenceBands,
+  }, [cellKey, dataConfig, equationContainerSelector, model, plotHeight, plotWidth, showConfidenceBands,
       xAttrId, xAttrName, xScale, xSubAxesCount, yAttrId, yAttrName, yScale, ySubAxesCount])
 
   const confidenceBandPaths = useCallback((caseValues: Point[], lineIndex: number) => {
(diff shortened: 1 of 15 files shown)


v3/src/components/graph/adornments/lsrl/lsrl-adornment-component.tsx
```diff
@@ -0,0 +1,359 @@
...
+        equationP.style("left", `${left}%`)
+          .style("top", `${top}%`)
+      }
+
+      lineObj.equation = equationDiv
+      lineObjectsRef.current = [...lineObjectsRef.current, lineObj]
+
+      const equation = equationDiv.selectAll("p").filter((d, i) => i === lineIndex)
+      equation.call(
+        drag<any, any>().on("drag", (e) => handleMoveEquation(e, false, lineIndex))
```
emcelroy, line 325: I did try to find more specific types to use here, but had trouble and decided not to spend anymore time on it for now.

v3/src/components/graph/adornments/lsrl/lsrl-adornment-model.ts
```diff
@@ -0,0 +1,157 @@
...
+    cat?: string
+  ) {
+    if (!dataConfig) return []
+    const dataset = dataConfig?.dataset
+    const legendAttrId = dataConfig?.attributeID("legend")
+    const casesInPlot = dataConfig.subPlotCases(cellKey)
+    const caseValues: Point[] = []
+    casesInPlot.forEach((c: ICase) => {
+      const caseValueX = Number(dataset?.getValue(c.__id__, xAttrId))
+      const caseValueY = Number(dataset?.getValue(c.__id__, yAttrId))
```
kswenson, line 54:
```suggestion
      const caseValueX = dataset?.getNumeric(c.__id__, xAttrId)
      const caseValueY = dataset?.getNumeric(c.__id__, yAttrId)
```
You don't need to perform the numeric conversion yourself.

v3/src/components/graph/adornments/lsrl/lsrl-adornment-model.ts
```diff
@@ -0,0 +1,157 @@
...
+      }
+    })
+    return caseValues
+  }
+}))
+.views(self => ({
+  confidenceValues(iX: number, caseValues: Point[], cellKey: Record<string, string>, lineIndex: number) {
+    const lines = self.lines.get(self.instanceKey(cellKey))
+    const line = lines?.[lineIndex]
+    const interceptIsLocked = false
```
kswenson, line 68: Should there be a TODO comment here about supporting the locked case?

v3/src/components/graph/adornments/lsrl/lsrl-adornment-model.ts
```diff
@@ -0,0 +1,157 @@
...
+    const slopeIntercept = leastSquaresLinearRegression(caseValues, interceptIsLocked)
+    const { count, mse, xSumSquaredDeviations, xMean } = slopeIntercept
```
kswenson, line 70: Seems like `slopeIntercept` is the wrong name here, and maybe you don't need an intermediate variable at all:
```suggestion
    const { count, mse, xSumSquaredDeviations, xMean } = leastSquaresLinearRegression(caseValues, interceptIsLocked)
```

v3/src/components/graph/utilities/graph-utils.ts
```diff
@@ -275,3 +293,220 @@ export function computeSlopeAndIntercept(xAxis?: IAxisModel, yAxis?: IAxisModel)
...
+  sdResiduals: number | null
+  slope: number | null
+  sse: number // sum of squared errors
+  sumSquaresResiduals?: number | null
+  xMean: number | null
+  xSumSquaredDeviations: number | null
+  yMean?: number | null
+}
+
+export const leastSquaresLinearRegression = (iValues: any, iInterceptLocked: boolean) => {
```
kswenson, line 418: Type?

v3/src/components/graph/utilities/graph-utils.ts
```diff
@@ -275,3 +293,220 @@ export function computeSlopeAndIntercept(xAxis?: IAxisModel, yAxis?: IAxisModel)
...
+  [2000, 1.96115],
+  [10000, 1.9602],
+  [100000, 1.95999]
+]
+
+ export const tAt0975ForDf = (iDf: number) => {
+  const foundIndex = t_quantile_at_0975_for_df.findIndex((iPair: any) => iPair[0] > iDf)
+  return foundIndex <= 0 ? 1.96 : t_quantile_at_0975_for_df[foundIndex - 1][1]
+}
+export interface ISlopeIntercept {
```
kswenson, line 404: Perhaps `IRegression` or `IRegressionResults`?

v3/src/components/graph/utilities/graph-utils.ts
```diff
@@ -275,3 +293,220 @@ export function computeSlopeAndIntercept(xAxis?: IAxisModel, yAxis?: IAxisModel)
...
+  intercept: number | null
+  mse: number | null // mean squared error
+  rSquared: number | null
+  sdResiduals: number | null
+  slope: number | null
+  sse: number // sum of squared errors
+  sumSquaresResiduals?: number | null
+  xMean: number | null
+  xSumSquaredDeviations: number | null
+  yMean?: number | null
```
kswenson, line 415: Is there a reason to use `null` here? My inclination would have been to make them all optional properties so they get initialized to `undefined` automatically.

v3/src/components/graph/utilities/graph-utils.ts
```diff
@@ -275,3 +293,220 @@ export function computeSlopeAndIntercept(xAxis?: IAxisModel, yAxis?: IAxisModel)
...
+}
+
+export const leastSquaresLinearRegression = (iValues: any, iInterceptLocked: boolean) => {
+  const tSlopeIntercept: ISlopeIntercept = {
```
kswenson, line 419: Seems like there are better names.

v3/src/components/graph/adornments/lsrl/lsrl-adornment-model.ts
```diff
@@ -0,0 +1,157 @@
...
+    if (
+      !line ||
+      (!count && count !== 0) ||
+      (!mse && mse !== 0) ||
+      (!xMean && xMean !== 0) ||
+      (!xSumSquaredDeviations && xSumSquaredDeviations !== 0)
```
kswenson, line 76: I don't understand these test pairs -- if it's meant to be a test for `null`/`undefined` then I'd just use `count != null`, etc.

v3/src/components/graph/adornments/lsrl/lsrl-adornment-model.ts
```diff
@@ -0,0 +1,157 @@
...
+        // TODO: Once the Intercept Locked feature is implemented, we will need to pass in something like 
+        // isInterceptLocked instead of false in the call to self.computeValues.
+        const { intercept, rSquared, slope, sdResiduals } = self.computeValues(
+          xAttrId, yAttrId, cellKey, dataConfig, instanceKey, false, legendCats[j]
+        )
+        if (
+          (intercept === null && intercept !== 0) ||
+          (rSquared === null && rSquared !== 0) ||
+          (slope === null && slope !== 0) ||
+          (sdResiduals === null && sdResiduals !== 0)
```
kswenson, line 145: A value that `=== null` is by definition `!== 0`.

v3/src/components/graph/adornments/lsrl/lsrl-adornment-registration.tsx
```diff
@@ -0,0 +1,96 @@
...
+import { LSRLAdornment } from "./lsrl-adornment-component"
+import { useGraphContentModelContext } from "../../hooks/use-graph-content-model-context"
+
+const Controls = observer(() => {
+  const graphModel = useGraphContentModelContext()
+  const adornmentsStore = graphModel.adornmentsStore
+  const existingAdornment = adornmentsStore.findAdornmentOfType<ILSRLAdornmentModel>(kLSRLType)
+
+  const handleBoxPlotSetting = (checked: boolean) => {
+    const existingBoxPlotAdornment = adornmentsStore.findAdornmentOfType<ILSRLAdornmentModel>(kLSRLType)
```
kswenson, line 20: Naming?

v3/src/components/graph/utilities/graph-utils.ts
```diff
@@ -275,3 +293,220 @@ export function computeSlopeAndIntercept(xAxis?: IAxisModel, yAxis?: IAxisModel)
...
   return {slope, intercept}
 }
+
+  /**
+   * Returns an object that has the slope and intercept
+   * @param iCoordPairs [{x: {Number}, y: {Number}}]
+   * @returns {{count: {Number}, xSum: {Number}, xSumOfSquares: {Number}, xSumSquaredDeviations: { Number},
+  *          ySum: {Number}, ySumOfSquares: {Number}, ySumSquaredDeviations: {Number}, sumOfProductDiffs: {Number} }
+  */
+ const computeBivariateStats = (iCoordPairs: any) => {
```
kswenson, line 303: Type?

v3/src/components/graph/utilities/graph-utils.ts
```diff
@@ -275,3 +293,220 @@ export function computeSlopeAndIntercept(xAxis?: IAxisModel, yAxis?: IAxisModel)
...
+      sumOfProductDiffs: 0
+    }
+    let tSumDiffsX = 0
+    let tSumDiffsY = 0
+
+    // Under certain circumstances (adding new case) an empty value can sneak in here. Filter out.
+    iCoordPairs = iCoordPairs.filter((iPair: any) => {
+      return !(!iPair.x || !iPair.y)
+    })
+   iCoordPairs.forEach(function (iPair: any) {
```
kswenson, line 321: Types?

v3/src/utilities/translation/lang/en-US.json5
kswenson, line 1: Generally, we've been adding new v3 strings at the top of the file. Are these meant to be v2 strings as well? Has the working been vetted by the appropriate people? @bfinzer?

kswenson (changes requested):
👍 Impressive work -- I made a bunch of suggestions. Some of them I realize in retrospect probably relate to code that was copied from v2, but I still think it's worth a pass to clean up some of the naming and typing issues because otherwise they'll probably never be addressed.

bfinzer:
Found bugs by checking out an running the branch. These could be logged in a separate story or stories if you prefer: 
* When the intercept is negative the "+" in the equation should not be present
* The slope should only be parenthesized if there are units present. Since we're not dealing with units yet, the parentheses will come in later.
* I rename the x-axis attribute to be a single letter; e.g. *x*. In that case we don't parenthesize it in the equation.
* The equation color should match the line color when there is a legend
* When dragging the equation, if the mouse goes down near one end, the equation "jumps" so the mouse is centered. It shouldn't jump.
* I put Top_Speed on x and Max_Height on y and showed LSRL. Then I added Type as legend. Finally, I added Inversions to the right axis. The lines did not show in the top subplot.

bfinzer (approved):
👍🏻LGTM
Kirk has commented on the code. My contribution at this stage is noting bugs. I notice that some of the bugs I found apply equally to movable line equation.

pjanik:
@emcelroy, in case you're surprised by my last commit - I just merged main into your PR, as we both touched the same files (plotted function component and model). But it shouldn't affect your work at all. I just adjusted the component to use graph-utils helpers that you added in this PR (and removed copies of the same code from the component file itself - my previous approach).

emcelroy:
> @emcelroy, in case you're surprised by my last commit - I just merged main into your PR, as we both touched the same files (plotted function component and model). But it shouldn't affect your work at all. I just adjusted the component to use graph-utils helpers that you added in this PR (and removed copies of the same code from the component file itself - my previous approach).

@pjanik Thanks for doing that and letting me know. Makes sense 👍

v3/src/components/graph/adornments/lsrl/lsrl-adornment-component.tsx
```diff
@@ -0,0 +1,359 @@
...
+  const yCellCount = yCats.length * ySubAxesCount
+  const allAttributes = dataConfig?.dataset?.attributes
+  const xAttrId = dataConfig?.attributeID("x") || ""
+  const yAttrId = dataConfig?.attributeID("y") || ""
+  const xAttr = allAttributes?.find(attr => attr.id === xAttrId)
+  const yAttr = allAttributes?.find(attr => attr.id === yAttrId)
+  const xAttrName = xAttr?.name ?? ""
+  const yAttrName = yAttr?.name ?? ""
+  const instanceKey = model.instanceKey(cellKey)
+  const classFromKey = model.classNameFromKey(cellKey)
```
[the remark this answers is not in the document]
emcelroy: That makes sense, but is something I'd like to attack in a separate piece of work.

v3/src/components/graph/adornments/lsrl/lsrl-adornment-component.tsx
```diff
@@ -0,0 +1,359 @@
...
+  // Refresh values on axis changes
+  useEffect(function refreshAxisChange() {
+    return mstAutorun(() => {
+      // We observe changes to the axis domains within the autorun by extracting them from the axes below.
+      // We do this instead of including domains in the useEffect dependency array to prevent domain changes
+      // from triggering a reinstall of the autorun.
+      const { domain: xDomain } = xAxis // eslint-disable-line @typescript-eslint/no-unused-vars
+      const { domain: yDomain } = yAxis // eslint-disable-line @typescript-eslint/no-unused-vars
+      updateLSRL()
+    }, { name: "LSRLAdornmentComponent.refreshAxisChange" }, model)
```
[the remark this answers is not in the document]
emcelroy: @kswenson I tried this but the line would then not dynamically update when you did things like drag a case dot around the plot. I'm not sure why exactly. I left it as is, for now at least. If changing it is important, though, I can try again.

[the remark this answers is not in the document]
kswenson: I was thinking that the`mstAutorun` was specifically reacting to domain changes. Apparently, it's also reacting to stuff that's being accessed in `updateLSRL()`. Leave it for now and we can come back to it later if it becomes a problem, e.g. if it's calling `updateLSRL()` too frequently.

v3/src/components/graph/adornments/lsrl/lsrl-adornment-component.tsx
```diff
@@ -0,0 +1,355 @@
...
+
+      lineObj.equation = equationDiv
+      lineObjectsRef.current = [...lineObjectsRef.current, lineObj]
+
+      const equation = equationDiv.select<HTMLElement>(`#lsrl-equation-${model.classNameFromKey(cellKey)}-${lineIndex}`)
+      equation?.call(
+        drag<HTMLElement, unknown>().on("drag", (e) => handleMoveEquation(e, false, lineIndex))
+          .on("end", (e) => handleMoveEquation(e, true, lineIndex))
+      )
+    }
```
emcelroy, line 324: In case it helps someone in the future, the key to being able to use `drag<HTMLElement, unknown>` on line 321 instead of resorting to `drag<any, any>` was adding `<HTMLElement>` to the select in line 319. Without the latter, there was an error thrown by line 321 `Type 'Selection<BaseType, unknown, HTMLElement, any>' is not assignable to type 'Selection<HTMLElement, unknown, any, any>'` `BaseType` could potentially equal `null` which isn't assignable to `HTMLElement`.

v3/src/components/graph/adornments/lsrl/lsrl-adornment-model.ts
emcelroy: Thanks. I still get confused sometimes by `null`, `0`, and `undefined` and which is considered equal to which. I need to get that squared away in my head.

v3/src/utilities/translation/lang/en-US.json5
emcelroy: @kswenson I lifted the new strings from [this V2 translation file](https://github.com/concord-consortium/codap/blob/master/lang/strings/en-US.json) in `master`. I believe they were added in V2 relatively recently which is why they weren't already in the V3 file. In that case, is it OK that they're located where they are in the V3 file? Or should they still be moved to the top?

emcelroy:
> Found bugs by checking out an running the branch. These could be logged in a separate story or stories if you prefer:
> 
> * When the intercept is negative the "+" in the equation should not be present
> * The slope should only be parenthesized if there are units present. Since we're not dealing with units yet, the parentheses will come in later.
> * I rename the x-axis attribute to be a single letter; e.g. _x_. In that case we don't parenthesize it in the equation.
> * The equation color should match the line color when there is a legend
> * When dragging the equation, if the mouse goes down near one end, the equation "jumps" so the mouse is centered. It shouldn't jump.
> * I put Top_Speed on x and Max_Height on y and showed LSRL. Then I added Type as legend. Finally, I added Inversions to the right axis. The lines did not show in the top subplot.

@bfinzer Thanks. I believe the latest commit fixes all the above except for the equation jump on drag. I'd like to work on that separately since there's a lot going on in this PR already and I believe that problem exists in at least one other adornment  (Movable Line) as well.

Could you please confirm that the other bugs are fixed now?

v3/src/utilities/translation/lang/en-US.json5
kswenson: No, it suggests that we should do a better job of keeping the v3 file up-to-date with changes in v2, so what you did was exactly correct.

v3/src/components/graph/adornments/lsrl/lsrl-adornment-model.ts
```diff
@@ -0,0 +1,198 @@
...
+.volatile(() => ({
+  category: undefined as string | undefined,
+  intercept: undefined as number | undefined,
+  rSquared: undefined as number | undefined,
+  slope: undefined as number | undefined,
+  sdResiduals: undefined as number | undefined
+}))
+.views(self => ({
+  get isValid() {
+    return self.intercept && self.rSquared && self.slope && self.sdResiduals
```
kswenson, line 22: You probably want `isFinite()` here because this will fail when any of the values `=== 0`.

v3/src/components/graph/adornments/lsrl/lsrl-adornment-component.tsx
```diff
@@ -122,7 +122,7 @@ export const LSRLAdornment = observer(function LSRLAdornment(props: IProps) {
         .style("top", `${top}px`)
 
       if (isFinished) {
-        const lines = model.lines.get(instanceKey)
+        const lines = dataConfig && model.getLines(xAttrId, yAttrId, cellKey, dataConfig)
```
kswenson, line 125: Given that this occurs multiple times in multiple other callbacks, you may want to pull it out into its own callback:
```
  const getLines = useCallback(() => {
    return dataConfig && model.getLines(xAttrId, yAttrId, cellKey, dataConfig)
  }, [cellKey, dataConfig, xAttrId, yAttrId])
```
which then shortens the dependency list for the other effects.

v3/src/components/graph/adornments/lsrl/lsrl-adornment-model.ts
```diff
@@ -50,30 +61,29 @@ export const LSRLAdornmentModel = AdornmentModel
...
-    const slopeIntercept = leastSquaresLinearRegression(caseValues, interceptIsLocked)
-    const { count, mse, xSumSquaredDeviations, xMean } = slopeIntercept
+    const { count, mse, xSumSquaredDeviations, xMean } = leastSquaresLinearRegression(caseValues, isInterceptLocked)
     if (
-      !line ||
-      (!count && count !== 0) ||
-      (!mse && mse !== 0) ||
-      (!xMean && xMean !== 0) ||
-      (!xSumSquaredDeviations && xSumSquaredDeviations !== 0)
+      !line || !line.intercept || !line.slope || count == null || mse == null || xMean == null ||
```
kswenson, line 85: `line.intercept` and `line.slope` can validly be 0, so the test should be `== null` for them as well.

v3/src/components/graph/utilities/graph-utils.ts
```diff
@@ -297,28 +299,30 @@ export function computeSlopeAndIntercept(xAxis?: IAxisModel, yAxis?: IAxisModel)
...
+     ySumOfSquares: 0,
+     ySumSquaredDeviations: 0,
+     sumOfProductDiffs: 0
+   }
+   let tSumDiffsX = 0
+   let tSumDiffsY = 0
+
+   // Under certain circumstances (adding new case) an empty value can sneak in here. Filter out.
+   iCoordPairs = iCoordPairs.filter((iPair: Point) => {
+     return !(!iPair.x || !iPair.y)
```
kswenson, line 323: These should probably be `isFinite()` tests as well to handle 0 values.

kswenson (changes requested):
👍 Changes generally look good -- there's a few places where tests should be changed to properly handle 0 values.

bfinzer (approved):
👍🏻LGTM
I checked the bugs that I previously identified and they are all fixed except, as you noted, the equation jumping bug.
However, please note that the bug regarding the negative sign for the intercept also still occurs for a movable line equation.

emcelroy:
> However, please note that the bug regarding the negative sign for the intercept also still occurs for a movable line equation.

@bfinzer Thanks. I'll fix that in the locked intercept work.

kswenson (approved):
👍 Changes look good

PR #988, opened 2023-11-09
Clear case values evaluated by formula before moving attr to a different collection

pjanik:
https://www.pivotaltracker.com/story/show/186323665

This PR resets formula-evaluated values to blank right before the attribute is moved to another collection. This prevents the grouping logic from considering cases that have been evaluated using the formula.

A simpler version of https://github.com/concord-consortium/codap/pull/986 that addresses @kswenson suggestion.


kswenson (approved):
👍 Looks good! Nice that we didn't have to make all of those `DataSet` changes and thx for the helpful jest tests!

PR #989, opened 2023-11-09
Small refactoring of plotted-function model and component

pjanik:
https://www.pivotaltracker.com/story/show/186437311

This PR doesn't introduce any new functionality or formula-related code. It's preparatory work for upcoming changes. It mainly involves modifications to existing code, so I believe it makes sense to have a separate PR to validate whether these changes align with your understanding.

Previously, in the PlottedFunctionAdornmentModel, `self.computePoints(...)` was called, then the resulting array was converted to a number (resulting in NaN), which was set as a MeasureInstance value. I was a bit confused about how the points were plotted. It turns out that the PlottedFunctionAdornmentComponent was calling `model.computePoints(...)` again. To streamline this process:
- I changed `MeasureInstance.value` to be an instance of `(x: number) => number` (`FormulaFn` type).
- Moved `computePoints` and all related helpers (`pathBasic`, `curveBasic`) out of the model file and defined them as pure functions in the components. They don't require access to any model state, so pure functions are always preferable to me (clarity, testability). Additionally, I believe they are more fitting for the view (component) than the actual MST model, as the actual points/coordinates are more of a view concern.
- Removed `.isValid` from `MeasureInstance` as it was never used (should we double-check if that's not necessary, as I remember seeing that in other adornments?).
- Removed `IValue` and its instances from the component file, as they didn't serve any purpose.
- Kept the naming convention like `set/addValue` even though now it's a function, assuming it might be intentional to maintain similarity with other adornments. However, we can certainly change that if nothing else depends on this particular `MeasureInstance`.

### DIFF
--- a/v3/src/components/graph/adornments/plotted-function/plotted-function-adornment-model.ts
+++ b/v3/src/components/graph/adornments/plotted-function/plotted-function-adornment-model.ts
@@ -1,20 +1,16 @@
 import { Instance, types } from "mobx-state-tree"
 import { AdornmentModel, IAdornmentModel, IUpdateCategoriesOptions } from "../adornment-models"
-import { kPlottedFunctionType, kPlottedFunctionValueTitleKey } from "./plotted-function-adornment-types"
-import { ScaleNumericBaseType } from "../../../axis/axis-types"
-import { Point } from "../../../data-display/data-display-types"
+import { kPlottedFunctionType, kPlottedFunctionValueTitleKey, FormulaFn } from "./plotted-function-adornment-types"
 import { ICase } from "../../../../models/data/data-set-types"
 import { IGraphDataConfigurationModel } from "../../models/graph-data-configuration-model"
 
 export const MeasureInstance = types.model("MeasureInstance", {})
 .volatile(self => ({
-  value: NaN,
-  isValid: false
+  formulaFunction: (x: number) => NaN,
 }))
 .actions(self => ({
-  setValue(value: number) {
-    self.value = value
-    self.isValid = true
+  setValue(formulaFunction: FormulaFn) {
+    self.formulaFunction = formulaFunction
   }
 }))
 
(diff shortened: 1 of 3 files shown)


emcelroy (approved):
Generally these adjustments look good and make a lot of sense -- thanks 👍 There are just two points I should make, the first of which we just now discussed on Slack, Piotr, but I'll include here for Kirk's benefit.

1) In the [Least Squares Line PR](https://github.com/concord-consortium/codap/pull/987/), I moved `curveBasis` and `pathBasis` to the shared graph-utils.ts. The Least Squares Line adornment uses those functions, too, so it seems like it should go in a shared location. I'm not sure graph-utils.ts is exactly the right place, but maybe. I'm waiting to hear Kirk and Bill's thoughts on that.

2) I think it's OK to drop `isValid` from `MeasureInstance`. For some history on why in this case I may have mistakenly thought it was needed, see [this comment](https://github.com/concord-consortium/codap/blob/6736cee41faa88c8ff06a2c27d4f87fa45a625fb/v3/src/components/graph/adornments/univariate-measures/univariate-measure-adornment-model.ts#L84-L87) about its use in `UnivariateMeasureAdornmentModel`. It doesn't seem like we reference the `MeasureInstance`'s `value` property anywhere, so that issue doesn't seem like it'll be a factor for the Plotted Function. If we do end up referencing the `value` property in the future, it may need to come back, though.

pjanik:
Thanks, @emcelroy. I'll wait for https://github.com/concord-consortium/codap/pull/987/ to be merged. After that, I'll rebase my PR and use the graphing helpers from the location you used in your PR.

v3/src/components/graph/adornments/plotted-function/plotted-function-adornment-model.ts
```diff
@@ -1,22 +1,16 @@
 import { Instance, types } from "mobx-state-tree"
 import { AdornmentModel, IAdornmentModel, IUpdateCategoriesOptions } from "../adornment-models"
-import { kPlottedFunctionType, kPlottedFunctionValueTitleKey } from "./plotted-function-adornment-types"
-import { ScaleNumericBaseType } from "../../../axis/axis-types"
-import { Point } from "../../../data-display/data-display-types"
-import { ICase } from "../../../../models/data/data-set-types"
-import { IGraphDataConfigurationModel } from "../../models/graph-data-configuration-model"
+import { kPlottedFunctionType, kPlottedFunctionValueTitleKey, FormulaFn } from "./plotted-function-adornment-types"
 
 export const MeasureInstance = types.model("MeasureInstance", {})
```
kswenson, line 5: Not directly part of this PR, but this `MeasureInstance` and the one in `univariate-measure-adornment-model.ts` should have unique names.

kswenson (approved):
👍 LGTM -- one suggestion on naming that needn't be part of this PR.

v3/src/components/graph/adornments/plotted-function/plotted-function-adornment-model.ts
pjanik: Maybe it'd leave it up to @emcelroy, as he knows the differences better and probably can come up with some better naming that I would.

BTW, what's the purpose of this name property? The docs don't explain anything about it. Is it just for debugging?

pjanik:
> After that, I'll rebase my PR and use the graphing helpers from the location you used in your PR.

@emcelroy, I actually did something else. I merged this PR so I could merge all my following PRs into main. And then I merged main into your PR https://github.com/concord-consortium/codap/pull/987/ so you don't have to deal with merge conflicts yourself. I removed helpers from the component file and used the code that you placed in graph-utils.

v3/src/components/graph/adornments/plotted-function/plotted-function-adornment-model.ts
kswenson: > BTW, what's the purpose of this name property? The docs don't explain anything about it. Is it just for debugging?

It's _mostly_ just for debugging/human readability, but the name is retained at runtime and so it can be used for lookup purposes. In `mst-utils.ts` we have a `getParentWithTypeName()` function which can be used, for instance, to walk up the tree to find the `DocumentContent` model.

PR #991, opened 2023-11-09
fix: Standard deviation + mean absolute deviation (PT-181914408)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/181914408

As noted in the comments of the PT story, the hover tip for SD and MAD were not appearing correctly on hover. These changes adjust the calculation for the x and y coordinates of the hover tip element so they appear in the correct spot.

In addition, these changes fix a separate bug in calculating the width of the range fills when the graph has a numeric attribute on the bottom and a categorical attribute on the top split.


kswenson (approved):
👍 LGTM

PR #992, opened 2023-11-10
Don't recalculate hidden adornment formulas, refactor related code

pjanik:
https://www.pivotaltracker.com/story/show/186452589

This PR fixes a bug that was causing hidden adornment formulas to still recalculate when the user modified one of the dependent attributes.

### DIFF
--- a/v3/src/models/formula/formula-manager.ts
+++ b/v3/src/models/formula/formula-manager.ts
@@ -147,7 +147,7 @@ export class FormulaManager {
         { id: formula.id, formula: formula.display, extraMetadata }
       ))
     }, () => {
-      const activeFormulas: Set<string> = new Set()
+      const activeFormulas = new Set<string>()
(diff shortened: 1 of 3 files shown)


v3/src/models/formula/formula-manager.ts
```diff
@@ -133,24 +136,25 @@ export class FormulaManager {
...
+  registerActiveFormulas() {
     reaction(() => {
       // Observe all the formulas
-      return this.getAllFormulas().map(({ formula, extraMetadata }) => (
+      return this.getActiveFormulas().map(({ formula, extraMetadata }) => (
         { id: formula.id, formula: formula.display, extraMetadata }
       ))
     }, () => {
-      this.unregisterDeletedFormulas()
+      const activeFormulas: Set<string> = new Set()
```
kswenson, line 150: FYI, I think you can just use:
```suggestion
      const activeFormulas = new Set<string>()
```

kswenson (approved):
👍 LGTM

PR #993, opened 2023-11-10
Refactor formula utilities

pjanik:
https://www.pivotaltracker.com/story/show/186452591

This PR splits formula-utils into multiple files, now located under formula/utils/... The file had grown quite large and was mixing various helper categories, so I hope it's cleaner now. Tests have also been divided. Additionally, non-type elements (such as string constants and functions) have been removed from formula-types.ts. These have either been relocated to more relevant categories or placed in new files (e.g., mathjs-utils + tests).


kswenson (approved):
👍 Looks good

PR #994, opened 2023-11-11
Graph rescale button implementation

bfinzer:
[#186438328] Feature: Graph inspector has a rescale button that scales the axes to show all the data
[#186448224] Feature: For a case plot the graph's rescale button causes the points to rerandomize their positions
[#186448263] Feature: The graph's rescale button is disabled if the plot is not a case plot and has no numeric axes.
* Added `handleGraphRescale` in `GraphInspector`
* Implemented the rescaling in `GraphContentModel`
* CaseDots pays attention to graphModel.changeCount and rerandomizes points in response
* The rescale button gets tooltips that depend on graph state
* The rescale button is disabled when appropriate

Rescale transitions should be animated but that awaits some refactoring of enableAnimation

### DIFF
--- a/v3/src/components/graph/components/casedots.tsx
+++ b/v3/src/components/graph/components/casedots.tsx
@@ -29,7 +29,7 @@ export const CaseDots = function CaseDots(props: {
     currPos = useRef({x: 0, y: 0}),
     target = useRef<any>()
 
-  const randomlyDistributePoints = useCallback((cases?: CaseData[] | undefined) => {
+  const randomlyDistributePoints = useCallback((cases?: CaseData[]) => {
     const uniform = randomUniform()
       const points = randomPointsRef.current
       cases?.forEach(({caseID}) => {
--- a/v3/src/components/graph/components/casedots.tsx
+++ b/v3/src/components/graph/components/casedots.tsx
@@ -1,8 +1,9 @@
 import {randomUniform, select} from "d3"
+import {mstReaction} from "../../../utilities/mst-reaction"
 import React, {useCallback, useEffect, useRef, useState} from "react"
 import {CaseData} from "../../data-display/d3-types"
 import {IDotsRef} from "../../data-display/data-display-types"
-import {handleClickOnDot, setPointSelection} from "../../data-display/data-display-utils"
+import {handleClickOnDot, setPointSelection, startAnimation} from "../../data-display/data-display-utils"
 import {useDragHandlers, usePlotResponders} from "../hooks/use-plot"
 import {useGraphDataConfigurationContext} from "../hooks/use-data-configuration-context"
 import {useDataSetContext} from "../../../hooks/use-data-set-context"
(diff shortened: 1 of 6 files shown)


v3/src/components/graph/components/casedots.tsx
```diff
@@ -28,18 +29,28 @@ export const CaseDots = function CaseDots(props: {
...
-      const aCaseData: CaseData = target.current.node().__data__
-      if (aCaseData && target.current.node()?.nodeName === 'circle') {
-        target.current.transition()
-          .attr('r', dragPointRadius)
-        setDragID(aCaseData.caseID)
-        currPos.current = {x: event.clientX, y: event.clientY}
-        handleClickOnDot(event, aCaseData.caseID, dataset)
-      }
-    }, [dragPointRadius, dataset, enableAnimation]),
+  const randomlyDistributePoints = useCallback((cases?: CaseData[] | undefined) => {
```
emcelroy, line 32: The `| undefined` is unnecessary. (In case it's not clear, the `?` indicates `cases` can be undefined.)

v3/src/components/graph/components/casedots.tsx
```diff
@@ -28,18 +29,28 @@ export const CaseDots = function CaseDots(props: {
...
+    const uniform = randomUniform()
```
emcelroy, line 33: This is probably fine, and I see it was carried over from the previous state of the code, but why don't we just use `randomUniform()` on line 37 instead of creating the new `uniform()` here and using that instead?

v3/src/components/graph/components/graph-inspector.tsx
```diff
@@ -42,12 +44,36 @@ export const GraphInspector = ({ tile, show }: ITileInspectorPanelProps) => {
...
+
+    const getRescaleTooltip = () => {
+      if (graphModel?.noPossibleRescales) {
+        return "V3.Inspector.rescale.noRescale.toolTip"
+      }
+      else if (graphModel?.plotType === 'casePlot') {
+        return "V3.Inspector.rescale.casePlot.toolTip"
+      }
+      return "DG.Inspector.rescale.toolTip"
+    }
```
emcelroy, line 57: I'm not sure how much this matters really, but it doesn't seem like a function is necessary for this. Wouldn't something like the below work just as well?

```
const getRescaleTooltip = graphModel?.noPossibleRescales
  ? "V3.Inspector.rescale.noRescale.toolTip"
  : graphModel?.plotType === "casePlot"
    ? "V3.Inspector.rescale.casePlot.toolTip"
    : "DG.Inspector.rescale.toolTip"
```

emcelroy (changes requested):
Looks good 👍 I left some comments on the code, mostly things you can probably ignore if you prefer the way things are for specific reasons. There are a couple unrelated problems I noticed though:

1) There's a failing Cypress test. I think that can be fixed by modifying the value for `graphResizeButton` on line 10 of v3/cypress/support/elements/component-elements.ts. (It's currently `"Resize all columns to fit data"` but I think it should be `"Move the points to new random positions"`?)

2) I'm not 100% sure this is a problem, but if I add categorical attributes to both x and y, then change either the attribute on y or the attribute on x to a numeric attribute, the Rescale button remains disabled. I think it should become enabled at that point. Is that right? (If I change _both_ attributes to numeric, then the button becomes enabled.)

emcelroy (approved):
Looks good 👍

PR #995, opened 2023-11-12
Inspector palette for maps

bfinzer:
This commit just puts up the inspector palette but doesn't implement any of its functionality.

[#186459298] Feature: Maps have an inspector palette
* Register Map component as having a MapInspector
* Only display ruler icon if there is at least one point layer
* Show checkboxes in ruler icon menu
* Show a layers icon
* Show menu for export image icon

Strangely, two strings were missing from en-US.json5: 
"DG.DataDisplayMenu.exportSvgImage"
"DG.DataDisplayMenu.exportImage"
even though they were present in individual language json's

### DIFF
--- a/v3/src/components/map/components/inspector-panel/hide-show-menu-list.tsx
+++ b/v3/src/components/map/components/inspector-panel/hide-show-menu-list.tsx
@@ -1,0 +1,43 @@
+import { MenuItem, MenuList, useToast } from "@chakra-ui/react"
+import React from "react"
+import { useDataSetContext } from "../../../../hooks/use-data-set-context"
+import { ITileModel } from "../../../../models/tiles/tile-model"
+import {isMapContentModel} from "../../models/map-content-model"
+import t from "../../../../utilities/translation/translate"
+
+interface IProps {
+  tile?: ITileModel
+}
+
+export const HideShowMenuList = ({tile}: IProps) => {
+  const data = useDataSetContext()
+  const toast = useToast()
+  const mapModel = isMapContentModel(tile?.content) ? tile?.content : undefined
+  const handleMenuItemClick = (menuItem: string) => {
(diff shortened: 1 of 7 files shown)


v3/src/components/map/components/inspector-panel/hide-show-menu-list.tsx
```diff
@@ -0,0 +1,43 @@
...
+import t from "../../../../utilities/translation/translate"
+
+interface IProps {
+  tile?: ITileModel
+}
+
+export const HideShowMenuList = ({tile}: IProps) => {
+  const data = useDataSetContext()
+  const toast = useToast()
+  const mapModel = isMapContentModel(tile?.content) ? tile?.content : undefined
```
emcelroy, line 15: `mapModel` is defined here but doesn't seem to be used anywhere. Will it be utilized in the future?

emcelroy (approved):
Looks good 👍 I left some comments on very minor code issues. Generally, there's a lot of commented-out code that I wondered about, too, but I'm guessing that code will be useful for later work?

There's a failing Cypress test, but it looks like it's the same one that's failing in [this PR](https://github.com/concord-consortium/codap/pull/994). It'd probably be best to fix that before merging this in, though.

Also, when I tried to view the inspector palette in the browser using this branch so I could get a better idea of how it appears, should work, etc., I couldn't seem to get it to show up. I wasn't sure if that was because it's not actually supposed to appear just yet? Or maybe there's potentially something wrong with my set up?

v3/src/components/map/components/inspector-panel/hide-show-menu-list.tsx
bfinzer: Yes, it will be used once we start implementing functionality of the buttons.

PR #996, opened 2023-11-13
Add FormulaManager tests, refactor Formula model

pjanik:
This PR focuses on enhancing the testability of FormulaManager and introducing Jest tests. To achieve this, I've modified the initialization process of adapters to enable the use of FakeAdapter in the tests. Additionally, I've transferred certain logic from the Formula MST model to FormulaManager. The Formula model primarily serves as a simple storage, it can't do much without access to all the context (datasets, attributes, global values, etc.).


pjanik:
I'm not sure how CodeCov calcualted -11.77% of test coverage, but I think it's impossible given the fact that I was essentially just adding tests in this PR without adding any new functionality. FormulaManager is now covered in 95% according to jest coverage:

Screenshot 2023-11-14 at 13 56 00

kswenson (approved):
👍 Looks good

PR #997, opened 2023-11-13
Fix for graph rescale button state

bfinzer:
[#186468144] Bug fix: Graph Rescale button gets in wrong state

* In graph-inspector.tsx respond to change in graph's plotType
* Also fixed a lint complaint in graph-content-model.ts

### DIFF
--- a/v3/src/components/graph/components/graph-inspector.tsx
+++ b/v3/src/components/graph/components/graph-inspector.tsx
@@ -1,5 +1,5 @@
 import React, {useRef, useEffect, useState} from "react"
-import {mstAutorun} from "../../../utilities/mst-autorun"
+import {observer} from "mobx-react-lite"
 import {InspectorButton, InspectorMenu, InspectorPanel} from "../../inspector-panel"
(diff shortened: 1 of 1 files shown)


v3/src/components/graph/components/graph-inspector.tsx
```diff
@@ -1,33 +1,41 @@
...
-  useEffect(()=>{
+  useEffect(() => {
     !show && setShowPalette(undefined)
   }, [active, show])
 
+  useEffect(function respondToChangeInPlotType() {
+    return mstAutorun(() => {
+      setPlotType(graphModel?.plotType)
+    }, { name: "respondToChangeInPlotType" }, graphModel)
+  }, [graphModel])
```
kswenson, line 37: Rather than setting up an `autorun` here to synchronize graph model state to local React state, I think you could just make the `GraphInspector` an `observer` and access `graphModel.plotType` directly.

emcelroy (approved):
Looks good 👍

PR #999, opened 2023-11-14
Implementation of graph inspector hide/show commands

bfinzer:
[#186472161] Feature: Graph inspector palette allows user to hide selected/unselected cases and show all hidden

* As a side improvement, pass only the dotsRef.current svg element to useDragHandlers. (Previously was passing **window**. Clearly a bad idea!)
* Added `applyUndoableAction`, `hiddenCases` and `unselectedCases` to DataConfigurationModel
* Fleshed out the appropriate menu items in hide-show-menu-list.tsx
* Added `respondToHiddenCasesChange` to `usePlotResponders`


emcelroy (approved):
Looks good 👍 I think once the conflict with `main` in v3/src/components/data-display/models/data-configuration-model.ts is fixed, it's good to merge.

v3/src/components/data-display/models/data-configuration-model.ts
```diff
@@ -135,6 +138,8 @@ export const DataConfigurationModel = types
   }))
   .views(self => ({
     filterCase(data: IDataSet, caseID: string, caseArrayNumber?: number) {
+      // If the case is hidden we don't plot it
+      if (self.hiddenCases.includes(caseID)) return false
```
kswenson, line 142: This is an O(n^2) lookup which could become a problem with large data sets. Could have a view method which returned a  set to get O(n) lookups, for instance.

PR #1001, opened 2023-11-14
Add formula manager adapters tests (attribute, plotted value and plotted function)

pjanik:
This pull request includes tests for the formula adapters. Setting this up became quite messy due to obvious dependencies on various external models. The Graph formulas adapters were particularly tricky to set up because of the GraphContentModel, which I couldn't instantiate in the required manner. However, I believe that some basic concepts are covered, and it increases coverage to 80-90%.


kswenson (approved):
👍 LGTM

PR #1002, opened 2023-11-14
Add tests of semi-aggregate functions

pjanik:
This pull request adds a test for semi-aggregate functions, increasing coverage to 95%. I have verified that the complexity is indeed O(n) by considering the most pessimistic scenarios for both functions: using `prev` with a filter that works only for the first case and using `next` with a filter that works only for the last case.


kswenson (approved):
👍 LGTM

PR #1003, opened 2023-11-14
Fix `equal` helper, add tests of the basic function helpers 

pjanik:
This file seemed important; however, it wasn't tested. In fact, I found a bug in the `equal` function. The previous version would return true for both `true` and `1`. Now it's fixed (simply by reordering the if statements) and has been tested.


kswenson (approved):
👍 LGTM

PR #1005, opened 2023-11-15
Fix O(n^2) performance bottleneck in DataConfigurationModel

bfinzer:
[#186487944] Chore: Fix potential performance bottleneck having to do with testing for hidden cases

* Introduced hiddenCasesSet as a view in DataConfigurationModel and refer to it in filterCase (rather than testing for includes of hiddenCases array).


kswenson (approved):
👍 Looks good

PR #1006, opened 2023-11-15
Fix inefficiency in subPlotCases method

pjanik:
https://docs.google.com/document/d/1RS5kWAGC8qN-0kxafvb1zES6i8HLZuNpA7fkMg4pM-M/edit?usp=sharing:

Screenshot 2023-11-15 at 19 12 58

### DIFF
--- a/v3/src/components/graph/models/graph-data-configuration-model.ts
+++ b/v3/src/components/graph/models/graph-data-configuration-model.ts
@@ -258,12 +258,14 @@ export const GraphDataConfigurationModel = DataConfigurationModel
   .views(self => ({
     subPlotCases(cellKey: Record<string, string>) {
       const casesInPlot: ICase[] = []
+      const casesInPlotSet = new Set<string>()
       self.filteredCases?.forEach(aFilteredCases => {
         aFilteredCases.caseIds.forEach((id) => {
           const caseData = self.dataset?.getCase(id)
-          const caseAlreadyMatched = casesInPlot.find(aCase => aCase.__id__ === id)
-          if (caseData && !caseAlreadyMatched) {
-            self.isCaseInCell(cellKey, caseData) && casesInPlot.push(caseData)
+          const caseAlreadyMatched = casesInPlotSet.has(id)
+          if (caseData && !caseAlreadyMatched && self.isCaseInCell(cellKey, caseData)) {
+            casesInPlot.push(caseData)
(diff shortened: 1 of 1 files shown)


v3/src/components/graph/models/graph-data-configuration-model.ts
```diff
@@ -258,12 +258,14 @@ export const GraphDataConfigurationModel = DataConfigurationModel
...
       self.filteredCases?.forEach(aFilteredCases => {
         aFilteredCases.caseIds.forEach((id) => {
           const caseData = self.dataset?.getCase(id)
-          const caseAlreadyMatched = casesInPlot.find(aCase => aCase.__id__ === id)
-          if (caseData && !caseAlreadyMatched) {
-            self.isCaseInCell(cellKey, caseData) && casesInPlot.push(caseData)
+          const caseAlreadyMatched = casesInPlotSet.has(id)
+          if (caseData && !caseAlreadyMatched && self.isCaseInCell(cellKey, caseData)) {
+            casesInPlot.push(caseData)
+            casesInPlotSet.add(caseData.__id__)
```
kswenson, line 268: Rather than maintaining both the array and the set on every iteration, my inclination would have been to just maintain the set and then `return Array.from(casesInPlot)` at the end. Not sure what impact that would have on performance.

kswenson (approved):
👍 Looks good -- I added a comment about a slight tweak to the implementation for your consideration.

bfinzer (approved):
👍 LGTM

v3/src/components/graph/models/graph-data-configuration-model.ts
pjanik: In theory, it shouldn't have any impact. It's N * O(1) versus O(N). In practice, there might be some, but probably well below levels that are important for us. Another minor concern could be the fact that a Set might not maintain the order of the cases (even if nothing depends on that, it can be helpful while debugging or writing unit tests), but looking at the docs, it seems it does. I'll make the change. 👍

pjanik: @kswenson, well, I ended up having to use a Map. Set wouldn't work too well, I think, as I was using it just to store IDs. I could store the full case object, but since references would be compared in this case, I bet it might not be safe (e.g., if case data is generated dynamically or something like that - ID comparison seems safer).

PR #1008, opened 2023-11-16
Animation during graph rescaling

bfinzer:
There are a **lot** of affected files, but not so many substantive changes.

[#186448221] Bug fix: When user rescales graph axes, the rescaling is animated

* This involved a large refactoring to move the state of animation being enabled into the DataDisplayContentModel where it is a volatile boolean property and no longer a React ref held onto by the GraphController
* Getting the rescale triggered in the graph inspector to animate was then trivial


kswenson (commented):
Added a commit to tweak the implementation and then sync reviewed the changes with @bfinzer.

PR #1011, opened 2023-11-17
Maps respond to new datasets with lat/long data

bfinzer:
[#186435138] Feature: Importing a csv with lat/long attributes should display points for the cases on existing map without further user intervention

* Move the setTimeout for fitting bounds into fitMapBoundsToData so clients don't have to deal with it
* In the reaction in MapContentModel that reacts to changes in the number of available datasets with points or boundaries, when layers have been changed, fit the bounds.


kswenson (approved):
👍 Looks good

PR #1012, opened 2023-11-18
maps respond to component resize

bfinzer:
Previous to this PR maps would not properly bring in new portions of the base map when the map component was resized. Now they do. In a slight touchup from V2, the map shrinks and grows around the center of the map rather than the upper left corner.

[#186408583] Feature: Maps should expand when a user expands the component using the drag handle

* Set up a useEffect in use-map-model.ts to inform leafletMap that its size has changed
* Considerable cleanup on getting rid of MapController

### DIFF
--- a/v3/src/components/map/hooks/use-map-model.ts
+++ b/v3/src/components/map/hooks/use-map-model.ts
@@ -1,18 +1,16 @@
+import {reaction} from "mobx"
 import {useEffect} from "react"
 import {latLng} from 'leaflet'
 import {useMap} from "react-leaflet"
+import {useMapModelContext} from "./use-map-model-context"
+import {useMapLayoutContext} from "../models/map-layout"
 import {kDefaultMapZoomForGeoLocation} from "../map-types"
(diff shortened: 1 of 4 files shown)


v3/src/components/map/hooks/use-map-model.ts
```diff
@@ -68,4 +66,13 @@ export function useMapModel(props: IProps) {
     mapModel.setHasBeenInitialized()
   }, [leafletMap, mapModel, mapModel.layers])
 
+  // Respond to content width and height changes
+  useEffect(function updateMapSize() {
+    const disposer = reaction(
+      () => [layout.mapWidth, layout.mapHeight],
+      () => leafletMap.invalidateSize(),
+      {name: "updateMapSize"}
```
kswenson, line 74:
```suggestion
      {name: "MapContentModel.updateMapSize"}
```
so it's easier to know where to find it.

kswenson (approved):
👍 LGTM

PR #1013, opened 2023-11-19
feat: Movable Lines + Least Squares Lines can have their intercepts locked (PT-181939990)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/181939990

Adds support for the Intercept Locked option in the measures menu. When activated, the Moving Line and LSRL adornments are forced to go through the origin.

These changes include a fix for a bug related to the drag behavior of the Movable Line: [Dragging movable line isn't following the mouse correctly](https://www.pivotaltracker.com/story/show/186448628). And there is also a fix for a bug related to the drag behavior of the Movable Line and LSRL's equation box: [Equation box for Movable Line and LSRL "jumps" when dragged](https://www.pivotaltracker.com/story/show/186485315).

### DIFF
--- a/v3/src/components/graph/adornments/lsrl/lsrl-adornment-component.tsx
+++ b/v3/src/components/graph/adornments/lsrl/lsrl-adornment-component.tsx
@@ -360,9 +360,10 @@ export const LSRLAdornment = observer(function LSRLAdornment(props: IProps) {
         // from triggering a reinstall of the autorun.
         const { domain: xDomain } = xAxis // eslint-disable-line @typescript-eslint/no-unused-vars
         const { domain: yDomain } = yAxis // eslint-disable-line @typescript-eslint/no-unused-vars
+        graphModel.getUpdateCategoriesOptions()
         buildElements()
       }, { name: "LSRLAdornmentComponent.refreshAxisChange" }, model)
-  }, [buildElements, dataConfig, model, xAxis, yAxis, updateLSRL])
+  }, [buildElements, graphModel, model, xAxis, yAxis])
 
   return (
     <svg
--- a/v3/src/components/graph/adornments/lsrl/lsrl-adornment-component.tsx
+++ b/v3/src/components/graph/adornments/lsrl/lsrl-adornment-component.tsx
@@ -46,6 +46,7 @@ export const LSRLAdornment = observer(function LSRLAdornment(props: IProps) {
   const graphModel = useGraphContentModelContext()
   const dataConfig = useGraphDataConfigurationContext()
   const layout = useAxisLayoutContext()
+  const adornmentsStore = graphModel?.adornmentsStore
   const xScale = layout.getAxisScale("bottom") as ScaleNumericBaseType
   const yScale = layout.getAxisScale("left") as ScaleNumericBaseType
   const xAttrType = dataConfig?.attributeType("x")
@@ -65,17 +66,17 @@ export const LSRLAdornment = observer(function LSRLAdornment(props: IProps) {
   const yAttr = allAttributes?.find(attr => attr.id === yAttrId)
   const xAttrName = xAttr?.name ?? ""
   const yAttrName = yAttr?.name ?? ""
-  const instanceKey = model.instanceKey(cellKey)
   const classFromKey = model.classNameFromKey(cellKey)
   const showConfidenceBands = model.showConfidenceBands
+  const interceptLocked = adornmentsStore?.interceptLocked
   const { equationContainerClass, equationContainerSelector } = equationContainer(model, cellKey, containerId)
   const lineRef = useRef() as React.RefObject<SVGSVGElement>
   const lineObjectsRef = useRef<ILineObject[]>([])
   const pointsOnAxes = useRef<IAxisIntercepts>({pt1: {x: 0, y: 0}, pt2: {x: 0, y: 0}})
 
   const getLines = useCallback(() => {
-    return dataConfig && model.getLines(xAttrId, yAttrId, cellKey, dataConfig)
-  }, [cellKey, dataConfig, model, xAttrId, yAttrId])
+    return dataConfig && model.getLines(xAttrId, yAttrId, cellKey, dataConfig, interceptLocked)
+  }, [cellKey, dataConfig, interceptLocked, model, xAttrId, yAttrId])
 
   const fixEndPoints = useCallback((iLine: Selection<SVGLineElement, unknown, null, undefined>) => {
     if (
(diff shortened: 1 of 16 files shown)


v3/src/components/graph/adornments/movable-line/movable-line-adornment-component.tsx
```diff
@@ -90,83 +110,93 @@ export const MovableLineAdornment = observer(function MovableLineAdornment(props
...
+  }, [equationContainerSelector, instanceKey, model.lines, plotHeight, plotWidth, xAttrName, xSubAxesCount,
+      yAttrName, ySubAxesCount])
+
+  const breakPointCoords = useCallback((pixelPtsOnAxes: IPointsOnAxes, breakPointNum: number) => {
+    if (interceptLocked) return {x: xScaleRef.current(0), y: yScaleRef.current(0)}
+    const weight = breakPointNum === 1 ? 3/8 : 5/8
+    const x = pixelPtsOnAxes.pt1.x + weight * (pixelPtsOnAxes.pt2.x - pixelPtsOnAxes.pt1.x)
+    const y = pixelPtsOnAxes.pt1.y + weight * (pixelPtsOnAxes.pt2.y - pixelPtsOnAxes.pt1.y)
+    return {x, y}
+  }, [interceptLocked])
```
kswenson, line 127: If `interceptLocked` were passed in as an argument then this could be a utility function outside the component that didn't require `useCallback` at all.

kswenson (changes requested):
👍 Looks good overall -- a few minor suggestions that should be easy to address. I asked a few questions about callback and effect dependencies that warrant some additional thought, but there may be a can of worms there that we don't want to open too far at this point. If nothing else, it's something to think about moving forward.

v3/src/components/graph/adornments/movable-line/movable-line-adornment-component.tsx
emcelroy: I moved this to graph-utils.ts. Doing so required also passing `xScaleRef.current` and `yScaleRef.current`. Does that seem reasonable? Or am I maybe missing a way that  wouldn't be necessary? I also had to move the `IPointsOnAxes` interface (to graph-utils.ts), although I'm wondering now if there's a better place for that. Maybe graphing-types.ts?

v3/src/components/graph/adornments/lsrl/lsrl-adornment-component.tsx
```diff
@@ -65,17 +66,17 @@ export const LSRLAdornment = observer(function LSRLAdornment(props: IProps) {
...
   const { equationContainerClass, equationContainerSelector } = equationContainer(model, cellKey, containerId)
   const lineRef = useRef() as React.RefObject<SVGSVGElement>
   const lineObjectsRef = useRef<ILineObject[]>([])
   const pointsOnAxes = useRef<IAxisIntercepts>({pt1: {x: 0, y: 0}, pt2: {x: 0, y: 0}})
 
   const getLines = useCallback(() => {
-    return dataConfig && model.getLines(xAttrId, yAttrId, cellKey, dataConfig)
-  }, [cellKey, dataConfig, model, xAttrId, yAttrId])
+    return dataConfig && model.getLines(xAttrId, yAttrId, cellKey, dataConfig, interceptLocked)
+  }, [cellKey, dataConfig, interceptLocked, model, xAttrId, yAttrId])
```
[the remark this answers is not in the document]
emcelroy: If I understand correctly, I do think with things as they are extracting `interceptLocked` outside of the callbacks is the right thing to do. I did experiment a bit with extracting it inside the callbacks, but I kept running into problems with the feature not working correctly. It's very possible there's a way to fix those problems by making more significant changes, but I'm not sure and think I should move on for now.

v3/src/components/graph/adornments/lsrl/lsrl-adornment-component.tsx
```diff
@@ -168,14 +166,14 @@ export const LSRLAdornment = observer(function LSRLAdornment(props: IProps) {
...
+        const left = equationCoords.x
+        const top = equationCoords.y
+        equation.style("left", `${left}px`)
+          .style("top", `${top}px`)
       }
     }
-  }, [cellKey, dataConfig, equationContainerSelector, getLines, model, plotHeight, plotWidth, showConfidenceBands,
-      xAttrId, xAttrName, xScale, xSubAxesCount, yAttrId, yAttrName, yScale, ySubAxesCount])
+  }, [cellKey, dataConfig, equationContainerSelector, getLines, interceptLocked, model, plotHeight, plotWidth,
+      showConfidenceBands, xAttrId, xAttrName, xScale, xSubAxesCount, yAttrId, yAttrName, yScale, ySubAxesCount])
```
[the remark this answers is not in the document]
emcelroy: I do think we need to invalidate this when the specified dependencies change since changes to them have an effect on how the adornment's elements are rendered. Similar to my other comment specifically about `interceptLocked`, I expect it'd be possible to change that by making some significant changes to the component as whole and how it renders things, but I'm not sure and tend to think I should move on for now. If you think this is a critical area that needs improvement, though, I could probably spend some more time investigating that.

v3/src/components/graph/adornments/lsrl/lsrl-adornment-component.tsx
```diff
@@ -265,7 +271,7 @@ export const LSRLAdornment = observer(function LSRLAdornment(props: IProps) {
       const { slope, intercept } = lines[lineIndex]
       const { domain: xDomain } = xAxis
       const { domain: yDomain } = yAxis
-      if (!slope || !intercept) continue
+      if (!slope || (!intercept && intercept !== 0)) continue
```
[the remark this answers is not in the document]
emcelroy: Thanks for calling that out. I believe I've fixed this part.

v3/src/components/graph/adornments/lsrl/lsrl-adornment-component.tsx
```diff
@@ -327,21 +335,34 @@ export const LSRLAdornment = observer(function LSRLAdornment(props: IProps) {
...
-    }, { name: "LSRLAdornmentComponent.refreshAxisChange" }, model)
-  }, [dataConfig, model, xAxis, yAxis, updateLSRL])
+    return mstAutorun(
+      () => {
+        // We observe changes to the axis domains within the autorun by extracting them from the axes below.
+        // We do this instead of including domains in the useEffect dependency array to prevent domain changes
+        // from triggering a reinstall of the autorun.
+        const { domain: xDomain } = xAxis // eslint-disable-line @typescript-eslint/no-unused-vars
+        const { domain: yDomain } = yAxis // eslint-disable-line @typescript-eslint/no-unused-vars
+        buildElements()
```
kswenson, line 363: Based on the lengthy discussion above:
```suggestion
        graphModel.getUpdateCategoriesOptions()
        buildElements()
```

v3/src/components/graph/adornments/movable-line/movable-line-adornment-component.tsx
kswenson: I hadn't considered the scale refs. Given that the function is only used in the movable line, I'd be inclined to leave it where it was as a callback but to pass `interceptLocked` as an argument so that the callback is stable and won't invalidate any effects that use it. Passing it as an argument leaves it up to the caller how to handle changes to `interceptLocked`.

kswenson (changes requested):
👍 Heading in the right direction, but still a couple of remaining suggestions.

kswenson (approved):
👍 Changes look good

PR #1014, opened 2023-11-20
fix: User can add Least Squares Line to scatterplot (PT-181939979)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/181939979

This fixes a bug reported by QA that caused crashes when the LSRL adornment was activated in graphs with multiple subplots. The check for valid values in the model's `getLines` view wasn't precise enough to prevent invalid values from being used when adding new lines, which would lead to many unnecessary lines being added to the model's `lines` map. This change fixes that.


kswenson (approved):
👍 Looks good -- good use of `Number.isFinite`.

PR #1010, opened 2023-11-17
Refactor graph categories options and subplot helpers 

pjanik:
The goal of this PR is to clean up some elements of the `GraphDataConfigurationModel` so that it's possible to precalculate and cache subplot cases. The changes are as follows:

1. Cleaned up methods related to subplots and cells in `GraphDataConfigurationModel`. This model previously had multiple copies of methods with the same name (e.g., `isCaseInSubplot`), and two overlapping (possibly identical) concepts - graph cells and graph subplots. Following @bfinzer's suggestion, I removed redundant methods and left subplot naming. (@bfinzer, if you happen to review this PR, I'd mostly pay attention to the changes in `graph-data-configuration-model.ts`).
   
2. Moved `getAllCellKeys` to `GraphDataConfigurationModel`. This move appears to be necessary to enable the precalculation and caching of subplot cases.

3. Attempted to clean up and simplify the `IUpdateCategoriesOptions` passed around to Adornments models. This structure had significant overlap with the data available in `GraphDataConfigurationModel`. Consequently, I streamlined these options, making them slimmer and allowing adornments to rely more on data configuration (e.g., the newly added `categoriesOptions` view). This area is probably relevant to @emcelroy's work.

4. Moved some tests from Adornments to `GraphDataConfigurationModel` mainly because of item 3. Although mocking has become somewhat more complex, I hope it is acceptable.

I did some tests locally, and calculations seemed to be correct in various subplots. However, I can never be sure if I've triggered all possible divisions / subplots. :) If you have some robust smoke tests, I would appreciate checking them in this branch once it is deployed.

PR #1015, opened 2023-11-20
Fix graph data configuration model - rm risky and unnecessary caching

pjanik:
This PR reverses a change that was introduced as part of https://github.com/concord-consortium/codap/pull/1010.

Initially, it didn't seem harmful, but I encountered issues when working on another feature. `.categoriesOptions` gets cached because of a reaction that uses it, and I could no longer update it in the tests where I applied some mocking techniques (not pretty, but it's not straightforward with MST).

Moreover, I believe it's worth reverting, as these category options might indeed depend on case values / properties that are not observable. Therefore, this getter could possibly require some manual invalidation. I don't think it's necessary at this point, and if one of the underlying methods needs to be cached because of the performance issues, it should probably be done lower down.


kswenson (approved):
👍 LGTM

PR #1016, opened 2023-11-20
Cache subPlotCases, columnCases and rowCases in graph-data-configuration-model

pjanik:
https://www.pivotaltracker.com/story/show/186496477

This PR introduces caching for `subPlotCases`, `columnCases`, and `rowCases` in the `GraphDataConfigurationModel`. To keep things straightforward, these properties are evaluated lazily, and cache clearing is done sparingly.

@kswenson, I believe this pattern could be quite common in models with manual caching:

```javascript
addDisposer(self, reaction(
  () => self.getDataBasedOnObservables(),
  () => self.clearCache(),
  {
    name: "...",
    equals: comparer.structural
  }
))
// ...
handleDataSetAction(actionCall: ISerializedActionCall) {
  const cacheClearingActions = ["setCaseData", "addCases", "removeCases"]
  if (cacheClearingActions.includes(actionCall.name)) {
    self.clearCache()
  }
}
```

It's a simple concept, but I have minimal confidence in whether `cacheClearingActions` are sufficient or not. Thorough testing would entail going through all the UX scenarios. Perhaps confidence in this will come with time and more experience with CODAP. :wink: However, I believe it could be useful to offer some helpers in the long term - docs, conventions, helper methods, events/flags. I'm not entirely sure, but that's generally what I was trying to mention during the today's standup.

### DIFF
--- a/v3/src/components/data-display/models/data-configuration-model.ts
+++ b/v3/src/components/data-display/models/data-configuration-model.ts
@@ -407,20 +407,6 @@ export const DataConfigurationModel = types
       },
     }))
   .actions(self => ({
-    handleDataSetChange(data?: IDataSet) {
-      self.actionHandlerDisposer?.()
-      self.actionHandlerDisposer = undefined
-      self.clearFilteredCases()
-      if (data) {
-        self.actionHandlerDisposer = onAnyAction(data, this.handleAction)
-        self.filteredCases[0] = new FilteredCases({
-          source: data,
-          filter: self.filterCase,
-          collectionID: idOfChildmostCollectionForAttributes(self.attributes, data),
-          onSetCaseValues: this.handleSetCaseValues
-        })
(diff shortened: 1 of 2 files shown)


kswenson (approved):
👍 Looks good -- just one string change. While this is a clear improvement as written, it's still going to loop over all of the cases once for each cell, row, and column. Another approach would be to loop through the cases once and build up the row, column, and cell caches in one pass which seems like it would perform even better. Then the `subPlotCases`, `rowCases`, and `columnCases` view methods would just be cache lookups (and possibly rebuilding the cache if necessary).

pjanik:
I think the current solution essentially fixes O(n2) issue, but leaves us with bigger `k` in `k * O(n)` indeed. :) I'll try to implement your suggestion, thank you.

PR #1017, opened 2023-11-20
Drop attribute on map to color points or polygons

bfinzer:
This feature got broken during a recent refactor. The fix is to allow CodapMap to rerender more frequently so that the refs get updated properly.

[#186510502] Feature: User can drop an attribute on a map to color points or boundaries

* CodapMap no longer an observer so that it renders sufficiently at startup to create the legend drop zone
* Fix a bug whereby we were attempting to fit the bounds of the leaflet map before it had been assigned to the model


kswenson (approved):
👍 LGTM

PR #1019, opened 2023-11-21
Add `cachedFnFactory` MST util and use it to cache Attribute props 

pjanik:
https://www.pivotaltracker.com/story/show/186496489

This PR introduces manual caching for some of the Attribute props. I've implemented it as a small helper so that the pattern is easier to replicate in other models, making it clear that the function result is cached.

I opted not to use getters as I believe it may not be possible with this type of tool. Additionally, even if it were, there might be a conflict with Mobx caching. Developers might assume manual invalidation is sufficient, but Mobx could still cache the data.

Also, I left `self.changeCount // eslint-disable-line no-unused-expressions` as some tests were depending on this reactivity.  So, I assumed that some features might also rely on it.

As part of this task, I tried to understand the process and explore alternative options. My initial impression is that we have limited alternatives. Another potential approach could involve triggering Mobx caching using targeted observable blocks, such as by placing graph rendering in `autorun`. I experimented with this, but it seems to be a challenging task and could likely introduce other issues. You can find most of my conclusions in [this lengthy Slack thread](https://concord-consortium.slack.com/archives/C0303QUNU/p1700580463868519).

### DIFF
--- a/v3/src/models/data/attribute.test.ts
+++ b/v3/src/models/data/attribute.test.ts
@@ -81,7 +81,7 @@ describe("Attribute", () => {
 
     attribute.setUnits("m")
     expect(attribute.units).toBe("m")
-    expect(attribute.numericCount).toBe(0)
+    expect(attribute.getNumericCount()).toBe(0)
     expect(attribute.type).toBeUndefined()
 
     attribute.addValue("1")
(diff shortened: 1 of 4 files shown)


kswenson (approved):
👍 Looks great -- very elegant!

PR #1020, opened 2023-11-24
(Micro?) optimize various models, especially accessors used to retrieve case values

pjanik:
This PR includes several updates:

1. Various methods that I was investigating now have a more classic syntax, with each variable declared in its own statement (multiple `const` instead of just one). This change makes performance investigation more efficient (dev tools) and often leads to a more efficient implementation of the given method, as variables can be declared and calculated only when necessary. More details can be found [in this Slack thread](https://concord-consortium.slack.com/archives/C035J6RDAK0/p1700674913157909). While the old syntax might be optimized by the browser, there's never a guarantee it's going to happen, and we make it more challenging. 

2. This was especially important for critical methods like `Dataset#getCaseValue`, `#getValueAtIndex`, `#getStrValue`, `#getStrValueAtIndex`, etc. These methods are called so often that every inefficiency adds up quickly for a large number of cases.

3. I've also updated the volatile `caseIDMap` and `pseudoCaseMap` to be `Map` instances instead of an object literal. It was an experiment. This particular change didn't affect performance much, so it would be fine to revert it if you don't like it. Although it's generally recommended to use Maps for large dictionaries where keys are longer/complex. Also, this ensures better typing: `map = Record<string, number> = {}; map["nonExistentKey"]` will be considered a number, even though it's undefined. Proper `Map` doesn't have this problem, as the `.get()` method returns the map type or undefined (hence my changes in random files with fallback values).

Even though these changes seem small, they reduced the execution time of `setPointCoordinates` from ~420ms to ~270ms. It's a significant change, as my subsequent PR will make `setPointCoordinates` the main contributor to the poor performance of graph updates.

General benchmark progress:
```
Resizing the graph component: 2.1s -> 1.7s, 0.4s speed up
Rescaling either axis: ~0.1s speed up
Marquee select: ~0.1s speed up
Select by clicking on legend key: ~0.1s speed up
```

### DIFF
--- a/v3/src/models/data/data-set.ts
+++ b/v3/src/models/data/data-set.ts
@@ -152,11 +152,11 @@ export const DataSet = types.model("DataSet", {
 })
 .volatile(self => ({
   // map from case IDs to indices
-  caseIDMap: {} as Record<string, number>,
+  caseIDMap: new Map<string, number>(),
   // MobX-observable set of selected case IDs
   selection: observable.set<string>(),
   // map from pseudo-case ID to the CaseGroup it represents
-  pseudoCaseMap: {} as Record<string, CaseGroup>,
+  pseudoCaseMap: new Map<string, CaseGroup>(),
   transactionCount: 0
 }))
 .views(self => {
@@ -183,9 +183,9 @@ export const DataSet = types.model("DataSet", {
 .views(self => ({
   // map from attribute id to attribute
   get attrIDMap() {
-    const idMap: Record<string, IAttribute> = {}
+    const idMap = new Map<string, IAttribute>()
     self.attributes.forEach(attr => {
-      idMap[attr.id] = attr
+      idMap.set(attr.id, attr)
     })
     return idMap
   },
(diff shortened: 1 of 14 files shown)


bfinzer (approved):
👍🏻LGTM

v3/src/models/data/data-set.ts
```diff
@@ -666,61 +670,61 @@ export const DataSet = types.model("DataSet", {
...
-                : attr && (index != null) ? attr.value(index) : ""
+        if (self.isCaching) {
+          const caseID = self.cases[index]?.__id__
+          const cachedCase = self.caseCache.get(caseID)
+          if (cachedCase && Object.prototype.hasOwnProperty.call(cachedCase, attributeID)) {
+            return cachedCase[attributeID]?.toString()
+          }
+        }
+        const attr = self.attrIDMap.get(attributeID)
+        return attr?.value(index) || ""
```
kswenson, line 707:
```suggestion
        return attr?.value(index) ?? ""
```

pjanik:
Thanks @kswenson, I've merged or implemented all your suggestions.

PR #1021, opened 2023-11-24
Make caching of multi-args functions easier, cache expensive methods in DataConfigurationModel and GraphDataConfigurationModel

pjanik:
This pull request introduces `cachedFnWithArgsFactory`, which should simplify the process of adding caching for various methods. TypeScript types may seem intimidating, but I have made an effort to ensure that the same arguments are required in the `key`, `calculate`, and `invalidate` functions, and it appears to function as expected.

I used this helper in the `DataConfigurationModel` and `GraphDataConfigurationModel` to cache methods that I identified as causing the most significant slowdown in graph redraws.

While caching seems straightforward, the challenge lies in invalidation. I have minimal confidence that I am invalidating this cache every time it is necessary. In the long term, this could become a substantial issue. In my specific case, I use three distinct approaches to invalidation:

- `onAnyAction`
- MobX reaction
- The `_setAttribute` method, which I believe was designed for similar purposes

If we consider the current architecture plus caching as a solution to performance problems, I think we should streamline invalidations (via helpers, architecture, convention (?)).

Additionally, @kswenson, in another pull request, you mentioned:

> While this is a clear improvement as written, it's still going to loop over all of the cases once for each cell, row, and column. Another approach would be to loop through the cases once and build up the row, column, and cell caches in one pass, which seems like it would perform even better. Then the `subPlotCases`, `rowCases`, and `columnCases` view methods would just be cache lookups (and possibly rebuilding the cache if necessary).

My mini-refactoring of `DataConfigurationModel` still does not achieve this. `cachedFnWithArgsFactory` doesn't help in that regard, but I could easily merge all these methods into a single one and add another argument ("all" | "row" | "column") if you think it's worth it. However, my inclination is to prioritize the simplicity of caching and invalidation (usually through sparse invalidations) over targeted, precise, and more error-prone cache building and invalidating. This is only partially related to your comment, as merging these methods into a single one will likely address your feedback without much complication.

Re performance improvements:
```
Resizing the graph component: 1.7s -> 0.8s
Rescaling either axis: similar
Marquee select: similar
Select by clicking on legend key: similar
```


bfinzer (approved):
👍🏻LGTM

v3/src/components/graph/models/graph-data-configuration-model.ts
```diff
@@ -10,6 +10,7 @@ import {AttributeDescription, DataConfigurationModel, IAttributeDescriptionSnaps
   from "../../data-display/models/data-configuration-model"
 import {GraphAttrRole, graphPlaceToAttrRole, PrimaryAttrRoles} from "../../data-display/data-display-types"
 import {updateCellKey} from "../adornments/adornment-utils"
+import { cachedFnWithArgsFactory } from "../../../utilities/mst-utils"
```
kswenson, line 13: I would move this ⬆️ next to `js-utils` so the `../../../utilities` imports are together.

kswenson (approved):
👍 Looks good!

v3/src/components/graph/models/graph-data-configuration-model.ts
pjanik: Done, thanks.

PR #1022, opened 2023-11-27
WIP: avoiding reading of MST observables

pjanik:
@kswenson, @bfinzer, you might find this interesting. I don't intend to merge this PR in its current form, as it can be done in a cleaner way. However, my goal was to make absolutely minimal changes and show how it affects performance.

So, essentially, extracting `legendType` and `categorySet` from the case-number-dependent loop reduced the time of `setPointSelection` from 240ms to 150ms. It's a solid difference. Note that these methods don't really perform any significant calculations. They mostly involve accessing some properties and performing simple operations.

I think it's largely due to slow MobX model access. Also, this illustrates how these little things build up. In the perfect scenario, graph rendering code (e.g., D3) could use a temporary model built using plain JS structures and classes. However, maintaining this model would be a challenge.

### DIFF
--- a/v3/src/components/data-display/data-display-utils.ts
+++ b/v3/src/components/data-display/data-display-utils.ts
@@ -124,6 +124,8 @@ export function setPointSelection(props: ISetPointSelection) {
 
   if (!(dotsRef.current && dots)) return
 
+  const legendType = dataConfiguration.attributeType('legend') || ""
+  const categorySet = dataConfiguration.categorySetForAttrRole('legend')
   // First set the class based on selection
   dots
     .classed('graph-dot-highlighted', (aCaseData: CaseData) => !!dataset?.isCaseSelected(aCaseData.caseID))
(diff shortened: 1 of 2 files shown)


bfinzer (approved):
👍🏻LGTM

kswenson (commented):
> I don't intend to merge this PR in its current form, as it can be done in a cleaner way.

I look forward to that cleaner way.

pjanik:
I found it interesting to observe that such a small change can have a significant impact. However, my suggestion is not to address that issue at the moment.

I could extract various elements out of the loop and then pass them to methods that accept many parameters. But, I believe this is not the most elegant approach, and I don't have a clear idea of how to make it worthwhile. Perhaps we could come up with a more generic approach, such as using snapshots, plain data, bypassing MST proxies, etc. But it's a separate, way bigger task.

I can keep this PR open as a reminder for myself or others, as maybe we'll have time to dive into this area while we're focused on performance. But if it's still open in 2024, feel free to close it. :)

kswenson:
Closing -- it was always intended just to be a proof-of-concept.

PR #1026, opened 2023-11-28
fix: case table connecting lines [#186208228]

kswenson:
PT: [[#186208228]](https://www.pivotaltracker.com/story/show/186208228)


bfinzer (approved):
👍🏻LGTM
I don't pretend to understand why this fixes the problem, but perhaps I don't need to.

kswenson:
> 👍🏻LGTM I don't pretend to understand why this fixes the problem, but perhaps I don't need to.

The short version is that the `useEffect` within the hook function was updating the ref after the hook function ran and returned the content of the ref, so the client was receiving a `one-render-behind` value. The solution was to realize that we didn't need the ref at all.

PR #1027, opened 2023-11-29
fix: simplify/fix label menu rendering

kswenson:
Fixes several issues that led to confusing rendering issues with axis/legend attribute menus

- `Graph` wraps its callbacks in `useCallback`
- `AttributeLabel` is no longer an `observer` -- it should render when its parent re-renders
- `AxisOrLegendAttributeMenu` is no longer `memo`ized -- it should render when its parent re-renders
- `useOverlayBounds` computes the overlay bounds in a `useEffect` and calls `setState` when there is a change, thus triggering a re-render


kswenson:
@bfinzer incorporated the changes in this PR into #1028 manually, so this PR is no longer necessary.

PR #1028, opened 2023-11-30
Preparation for legends in maps

bfinzer:
So maps can have a legend for each layer, which means multiple legends in a single data display component. We want legends to work for both graphs and maps, so this PR implements a MultiLegend component that, in the case of graphs, only contains a single Legend.

[#186521999] Feature: Maps show a legend for each layer with a legend attribute

This PR does not yet bring about legends in maps, but it is preparing for that.

* Create MultiLegend component to contain multiple legends, one for each layer in a graph or map
* Legend no longer uses display layout, but gets its layout width as property and a function to set desired extent
* MultiLegend uses display layout to get tileWidth and to set desired extent
* Each legend lives in its own div contained by a Flex box
* Defined placeCanAcceptAttributeIDDrop at DataDisplayContentModel because both map and graph need it
* Renamed AttributeLabel as GraphAttributeLabel and defined similar LegendAttributeLabel


kswenson (approved):
👍 Looks good -- I added a few commits (reviewed synchronously with @bfinzer) and rebased to `main` and force-pushed (after consultation with @bfinzer). Remaining review issues are just relatively minor questions which may be easier dealt with in the next PR.

PR #1030, opened 2023-12-01
fix: intercept detached from origin on top split removal (PT-181939990)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/181939990

These changes fix a bug where using the Intercept Locked option with a Movable Line in a top split scatter plot, and then removing the top split attribute, the line no longer passes through the origin. I believe this also fixes [#186580777](https://www.pivotaltracker.com/story/show/186580777) and some other bugginess with the rendering of adornments and labels when top and right splits are removed.

### DIFF
--- a/v3/src/components/graph/adornments/adornment.tsx
+++ b/v3/src/components/graph/adornments/adornment.tsx
@@ -3,8 +3,7 @@ import { clsx } from "clsx"
 import { observer } from "mobx-react-lite"
 import { IAdornmentModel } from "./adornment-models"
 import { useGraphContentModelContext } from "../hooks/use-graph-content-model-context"
-import { useGraphLayoutContext } from "../hooks/use-graph-layout-context"
-import { useGraphDataConfigurationContext } from "../hooks/use-graph-data-configuration-context"
+import { useSubplotExtent } from "../hooks/use-subplot-extent"
 import { INumericAxisModel } from "../../axis/models/axis-model"
 import { getAdornmentComponentInfo } from "./adornment-component-info"
 import {transitionDuration} from "../../data-display/data-display-types"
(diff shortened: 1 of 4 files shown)


v3/src/components/graph/adornments/adornment.tsx
```diff
@@ -13,22 +14,16 @@ import "./adornment.scss"
...
-) {
-  const graphModel = useGraphContentModelContext(),
-    layout = useGraphLayoutContext(),
-    subPlotWidth = topCats.length > 0
-                     ? layout.plotWidth / topCats.length
-                     : layout.plotWidth,
-    subPlotHeight = rightCats.length > 0
-                      ? layout.plotHeight / rightCats.length
-                      : layout.plotHeight
-
```
emcelroy, line 31: I didn't get to the bottom of why, but when removing a top split, `topCats` was not changing. For example, if you were using Mammals and added Diet to the top split, `topCats` became `['meat', 'plants', 'both']`, but when you removed Diet from the top split, it would still be `['meat', 'plants', 'both']`. The same issue was true for `rightCats`, too. 

`topCats` and `rightCats` were being passed as props to `Adornment` by `Adornments` where they were set by calling `layout.getAxisMultiScale("top").categorySet` and `layout.getAxisMultiScale("rightSplit").categorySet`. The values weren't being properly updated there either.

It seems like the new method for setting `subPlotWidth` and `subPlotHeight` here is more concise anyway and follows a pattern already used elsewhere in CODAP v3. So I only mention the above in case it seems like there may be a larger problem with `useGraphLayoutContext`, although there may not be.

kswenson: I suspect it may be a caching/invalidation issue in `CategorySet` rather than a problem with `useGraphLayoutContext`. I'll add a PT story about it.

v3/src/components/graph/adornments/adornment.tsx
```diff
@@ -13,22 +14,16 @@ import "./adornment.scss"
...
+export const Adornment = observer(function Adornment({adornment, cellKey}: IProps) {
+  const graphModel = useGraphContentModelContext()
+  const layout = useGraphLayoutContext()
+  const dataConfig = useGraphDataConfigurationContext()
+  const numExtraPrimaryBands = dataConfig?.numRepetitionsForPlace("bottom") ?? 1
+  const numExtraSecondaryBands = dataConfig?.numRepetitionsForPlace("left") ?? 1
+  const subPlotWidth = layout.plotWidth / numExtraPrimaryBands
+  const subPlotHeight = layout.plotHeight / numExtraSecondaryBands
```
kswenson, line 26: This is fine as is, but it's also a good candidate for a targeted custom hook:
```typescript
export function useSubPlotExtent() {
  const layout = useGraphLayoutContext()
  const dataConfig = useGraphDataConfigurationContext()
  const numExtraPrimaryBands = dataConfig?.numRepetitionsForPlace("bottom") ?? 1
  const numExtraSecondaryBands = dataConfig?.numRepetitionsForPlace("left") ?? 1
  const subPlotWidth = layout.plotWidth / numExtraPrimaryBands
  const subPlotHeight = layout.plotHeight / numExtraSecondaryBands
  return { subPlotWidth, subPlotHeight }
}

kswenson (approved):
👍 LGTM

PR #1032, opened 2023-12-04
Version 3.0.0-pre.1455 - December 4, 2023

tejal-shah:
### Features/Improvements
- Table should scroll to show collapsed row when it is scrolled out of view [#185270348](https://www.pivotaltracker.com/story/show/185270348)
- Improved graph performance

### Bug Fixes
- Hierarchical case tables more than one level deep do not show grouping lines and expand/collapse buttons [#186208228](https://www.pivotaltracker.com/story/show/186208228)
- Case table contents disappears [#186510549](https://www.pivotaltracker.com/story/show/186510549)

### DIFF
--- a/v3/CHANGELOG.md
+++ b/v3/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Features/Improvements
 - Table should scroll to show collapsed row when it is scrolled out of view [#185270348](https://www.pivotaltracker.com/story/show/185270348)
+- Improved graph performance
 
 ### Bug Fixes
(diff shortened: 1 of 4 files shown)


kswenson (changes requested):
👍 Looks good -- In light of these PRs from @pjanik:

>(Micro?) optimize various models, especially accessors used to retrieve case values #1020
>Make caching of multi-args functions easier, cache expensive methods in DataConfigurationModel and GraphDataConfigurationModel #1021

I think you should add an item under features/improvements like:
```
- Improved graph performance
```
And in general I think it's a good idea to scan the PRs as well as PT because not every improvement is in response to a PT story.

PR #1035, opened 2023-12-06
Reimplement basic graph features using PixiJS as the points renderer instead of D3 SVG

pjanik:
https://www.pivotaltracker.com/story/show/186631176

This is just the first step. I've enabled rendering for basic graphs, allowing you to resize the graph tile, and the dots will re-render correctly. Additionally, I've implemented marquee selection.

To fix the Jest tests, I've included jest-webgl-canvas-mock (we already use jest-canvas-mock). This appears to be the easiest way to avoid mocking Pixi completely in the Jest environment.

There are a few `// TODO PIXI:` comments, and some may be worth reviewing. I've also temporarily kept the old SVG code for comparison purposes and for future use in implementing missing features such as animations or interactions.

In the next step, I'll be focusing on point-ID matching and animations/transitions.

### DIFF
--- a/v3/src/components/graph/components/graph-component.tsx
+++ b/v3/src/components/graph/components/graph-component.tsx
@@ -17,9 +17,10 @@ import {isGraphContentModel} from "../models/graph-content-model"
 import {Graph} from "./graph"
 import {DotsElt} from '../../data-display/d3-types'
 import {AttributeDragOverlay} from "../../drag-drop/attribute-drag-overlay"
+import {PixiPoints} from "../utilities/pixi-points"
 import "../register-adornment-types"
 
-export const GraphComponent = observer(function GraphComponent({tile}: ITileBaseProps) {
+export const GraphComponent = observer(function GraphComponent({ tile }: ITileBaseProps) {
   const graphModel = isGraphContentModel(tile?.content) ? tile?.content : undefined
 
   const instanceId = useNextInstanceId("graph")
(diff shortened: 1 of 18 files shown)


v3/src/components/graph/components/graph-component.tsx
```diff
@@ -17,24 +17,33 @@ import {isGraphContentModel} from "../models/graph-content-model"
...
   const { data } = useDataSet(graphModel?.dataset)
   const layout = useInitGraphLayout(graphModel)
   // Removed debouncing, but we can bring it back if we find we need it
   const graphRef = useRef<HTMLDivElement | null>(null)
-  const {width, height} = useResizeDetector<HTMLDivElement>({ targetRef: graphRef })
+  const { width, height } = useResizeDetector<HTMLDivElement>({ targetRef: graphRef })
   const dotsRef = useRef<DotsElt>(null)
+  const pixiPointsRef = useRef<PixiPoints>()
+  // TODO PIXI: PJ: probably should be fixed and become a ref, as memoization is meant for performance optimization
+  // and it's not guaranteed to be maintained across renders.
```
kswenson, line 35: Can also use something like https://www.npmjs.com/package/use-memo-one.

kswenson (approved):
👍 Looks like a good start!

v3/src/components/graph/components/graph-component.tsx
pjanik: It looks nice. I'm wondering if there's any advantage/difference to using refs?

PR #1036, opened 2023-12-07
fix: Squares of Residuals option (PT-181940029)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/181940029

The Movable Line's equation box content (including the "Sum of squares" part) wasn't being updated properly when you dragged the line from the middle portion. We were previously sending the initial `intercept` value to `refreshEquation` in the Movable Line component's `handleTranslate` function. Sending the `newIntercept` value instead appears to fix the problem.


bfinzer (approved):
👍🏻LGTM
Yup, that fixes it!

PR #1037, opened 2023-12-08
Map rescale

bfinzer:
[#186641108] Feature: Map inspector rescale button brings all map data into view

* Call fitMapBoundsToData from MapContentModel:resize

### DIFF
--- a/v3/src/components/data-display/models/data-configuration-model.ts
+++ b/v3/src/components/data-display/models/data-configuration-model.ts
@@ -544,6 +544,7 @@ export const DataConfigurationModel = types
     handleDataSetChange(data?: IDataSet) {
       self.actionHandlerDisposer?.()
       self.actionHandlerDisposer = undefined
+      self.clearCasesCache()
       self.clearFilteredCases()
       if (data) {
         self.actionHandlerDisposer = onAnyAction(data, self.handleDataSetAction)
--- a/v3/src/components/map/components/codap-map.tsx
+++ b/v3/src/components/map/components/codap-map.tsx
@@ -14,7 +14,6 @@ import {MapInterior} from "./map-interior"
 import {MultiLegend} from "../../data-display/components/legend/multi-legend"
 import {DroppableMapArea} from "./droppable-map-area"
 import {IDataSet} from "../../../models/data/data-set"
-import {mstAutorun} from "../../../utilities/mst-autorun"
 
 import 'leaflet/dist/leaflet.css'
 import "./map.scss"
@@ -29,6 +28,7 @@ export const CodapMap = observer(function CodapMap({mapRef}: IProps) {
     layout = useDataDisplayLayout(),
     mapHeight = layout.contentHeight,
     interiorSvgRef = useRef<SVGSVGElement>(null),
+    prevMapSize = useRef<{ width: number, height: number, legend: number }>({ width: 0, height: 0, legend: 0 }),
     forceUpdate = useForceUpdate()
 
   // trigger an additional render once references have been fulfilled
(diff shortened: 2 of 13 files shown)


kswenson:
@bfinzer's initial work on this PR constituted a one-line change to call Leaflet's `fitBounds` function when the user clicked the rescale button on the map's inspector panel. He then asked me to look into how to make that action undoable as part of my review. Thus began a saga that took over a week to resolve....

Leaflet's internal map state includes a center (latitude, longitude) and a zoom level (number). For serialization purposes, these values are stored in the `MapContentModel`. Therefore, we must synchronize the map state with the map model state for save/restore of documents and for undo/redo to work as expected. There are several things that play into and in some cases complicate this process, however:

1. The user can interact directly with the Leaflet map by clicking the zoom buttons or dragging/panning the map.
2. The map state can be changed indirectly via API, e.g. in response to clicking the rescale button.
3. Given the previous two points, there must be a mechanism to distinguish between the different sources of change so that changes triggered by the user interacting directly with the map can be captured in an undoable action, while changes that result from API calls can be made undoable (or not) as circumstances require.
4. Leaflet dispatches events to indicate changes in its internal state.
5. Leaflet's state changes are asynchronous.
6. When animating, these state changes can occur over an extended period of time (e.g. 300+ ms).
7. Leaflet state changes can result in move events being generated (center changes) and/or zoom changes at the same time.
8. Requests for map state changes may result in no events being generated, e.g. if the same center/zoom is requested in subsequent calls.

To accomplish all of this, this PR introduces the `LeafletMapState` class, which is a MobX class which listens for Leaflet state change events and keeps its own internal center/zoom state in sync with Leaflet. It also tracks any asynchronous actions that are in progress, e.g. whether the user appears to be interacting directly with the Leaflet map or whether the code is currently responding to a `fitBounds` request which is likely to result in a change in state. When tracking in-progress actions, debounced completion functions are used which both act as a flag to indicate that the process is ongoing as well as provide a mechanism for clearing the in-progress state once the action is completed. The completion functions are debounced so that any combination of move/zoom end events will result in the last event being the one that ends the process.

The `MapContentModel` then uses MobX `reaction`s to:
1. listen for `LeafletMapState` changes and synchronize those state changes to the `MapContentModel` and to capture them in undoable actions when appropriate.
2. listen for `MapContentModel` changes and synchronize those state changes to the Leaflet map when it is determined that those model changes were triggered externally (e.g. by undo/redo) rather than as a response to previously triggered Leaflet map changes.

Along the way, I also encountered several other tangentially related issues that had to be addressed before I could bring this PR to completion:

**Fix import of .csv files exported from CODAP v2** -- v2 .csv exports include comment lines starting with `#` which were not being ignored properly. This has been fixed.

**Fix `DataConfigurationModel.handleDataSetChange()` calls `clearCasesCache()`** -- Changing the `DataSet` clearly involves changing the cases.

**Fix map performance issue introduced in #1039** -- #1039 added `.on('end', refreshPointSelection)` to the implementation of the `refreshPoints()` function in the `MapPointLayer` component. Unfortunately, this call `refreshPointSelection()` for _each point_, which is an O(n^2) operation which absolutely kills performance. In this PR we just comment it out in the interest of finishing this PR. If we need to add back a call to `refreshPointSelection()` we will have to find another way to do so.

**Fix MST warning on access to defunct model after closing map tile** -- Added a call to `isAlive` in the map's `HideShowMenuList` component to avoid accessing a defunct model.

The result of all of this is that user changes initiated via the Leaflet map directly (e.g. panning/zooming) and user changes initiated via CODAP UI (e.g. rescale button, resizing tile) are properly undoable, while non-user actions (e.g. positioning, rescaling, etc. triggered internally) do not add undoable actions to the undo history.

Due to its complexity, this PR was reviewed synchronously with @emcelroy.

kswenson (approved):
👍 After @bfinzer's initial one-line commit, I added a bunch of additional code to handle undo/redo, etc. See comment explaining all the gory details elsewhere in this PR.

emcelroy (approved):
Looks good -- an impressive amount of work 👍 And thanks for the detailed overview/explanation.

PR #1038, opened 2023-12-08
PixiJS: add animations and improve point-caseID matching

pjanik:
This PR reimplements `PixiPoints.matchPointsToData` to be more efficient. There's a risk of O(n^2), but I'm assuming that we usually remove just a few points/cases (user interaction) or all of them (new dataset used in a graph). So, the current implementation will save us from destroying and re-creating everything (what could always be O(n), but possibly more expensive operations and chances for memory leaks) when a few points are removed or added. When we remove all the points, we should be also fine, as I start removal from the back. I think the most pessimistic scenario is when we remove the first half of the array, but probably it's not likely to happen.

The second part of this PR is about animations/transition support. I came up with my own pattern (enclosing regular updates in `.transition()` call), and  I didn't use any external libraries for transition in the end. I started with [pixi-animate](https://github.com/pixijs/animate), but I ran into issues. We'd need to define transition for each point separately what might have some overhead (possibly small, but still), and when I was clearing animations, it was taking forever as Animations.clear() is implemented in a O(n) way. Given that it's not rocket science, I believe it's safer to stick with a custom solution to avoid similar issues. Also, I suppose I'll need to rework this approach for the hover effect, as I can see that we can have multiple animations at the same time, which is not supported at the moment.

### DIFF
--- a/v3/src/components/graph/utilities/pixi-points.ts
+++ b/v3/src/components/graph/utilities/pixi-points.ts
@@ -1,10 +1,9 @@
 import * as PIXI from "pixi.js"
-import { Actions, Interpolations } from "pixi-actions"
 import { CaseData } from "../../data-display/d3-types"
+import { PixiTranstion } from "./pixi-transition"
 
 const DEFAULT_Z_INDEX = 0
 const RAISED_Z_INDEX = 1
-const TARGET_FPS = 60
 export interface IPixiPointMetadata {
   caseID: string
   plotNum: number
@@ -19,8 +18,6 @@ export interface IPixiPointStyle {
   strokeOpacity?: number
 }
 
-export type Interpolation = (input: number) => number
-
 export class PixiPoints {
   app: PIXI.Application = new PIXI.Application({
     resolution: window.devicePixelRatio,
@@ -31,14 +28,8 @@ export class PixiPoints {
   pointMetadata: IPixiPointMetadata[] = []
   pointIdToIndex = new Map<string, number>()
   textures = new Map<string, PIXI.Texture>()
-  transitionDuration = 0.5 // seconds
-  transitionInterpolation = Interpolations.smooth
 
-  constructor() {
-    this.app.ticker.add((delta) => {
-      Actions.tick(delta / TARGET_FPS)
-    })
-  }
+  currentTransition?: PixiTranstion
 
   get canvas() {
     return this.app.view as HTMLCanvasElement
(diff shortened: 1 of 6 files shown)


v3/src/components/graph/utilities/pixi-points.ts
```diff
@@ -26,8 +26,11 @@ export class PixiPoints {
     antialias: true
   })
   pointMetadata: IPixiPointMetadata[] = []
+  pointIdToIndex = new Map<string, number>()
   textures = new Map<string, PIXI.Texture>()
 
+  currentTransition?: PixiTranstion
```
kswenson, line 32:
```suggestion
  currentTransition?: PixiTransition
```

v3/src/components/graph/utilities/pixi-points.ts
```diff
@@ -40,6 +43,21 @@ export class PixiPoints {
...
+  transition(duration: number, callback: () => void) {
+    if (this.currentTransition) {
+      this.currentTransition.destroy()
+      this.currentTransition = undefined
+    }
+    if (duration === 0) {
+      callback()
+      return
+    }
+    this.currentTransition = new PixiTranstion(duration, this.points)
```
kswenson, line 55:
```suggestion
    this.currentTransition = new PixiTransition(duration, this.points)
```

v3/src/components/graph/utilities/pixi-points.ts
```diff
@@ -70,37 +88,96 @@ export class PixiPoints {
...
-    // TODO PIXI: optimize this method. We don't have to remove all children and re-add them.
-    this.app.stage.removeChildren()
-    this.pointMetadata = []
+    // This map will get invalid because of point removal.
+    this.pointIdToIndex.clear()
+    const texture = this.getPointTexture(style)
+
+    // First, remove all the old sprites. Go backwards, so it's less likely we end up with O(n^2) behavior (although
+    // still possible). If we expect to have a lot of points removed, we should just destroy and recreate everything.
+    // However, I believe than in most practical cases, we will only have a few points removed, so this is approach is
```
kswenson, line 97:
```suggestion
    // However, I believe that in most practical cases, we will only have a few points removed, so this is approach is
```

v3/src/components/graph/utilities/pixi-transition.ts
```diff
@@ -0,0 +1,112 @@
...
+const defaultInterpolation = smoother
+// const smooth: Interpolation = (x: number) => x * x * (3 - 2 * x)
+// const linear: Interpolation = (x: number) => x
+// const smooth2: Interpolation = (x: number) => smooth(smooth(x))
+// const pow2out: Interpolation = (x: number) => Math.pow(x - 1, 2) * (-1) + 1
+
+export type SupportedPropKey = "position" | "scale"
+export type SupportedPropValue = { x: number, y: number }
+
+export class PixiTranstion {
```
kswenson, line 16:
```suggestion
export class PixiTransition {
```

kswenson (approved):
👍 Looks good! Nice to see how straightforward the animation was.

bfinzer (approved):
👍🏻Looks good to me, too.

v3/src/components/graph/utilities/pixi-points.ts
pjanik: Thanks, fixed in the last commit.

PR #1039, opened 2023-12-12
Map inspector hiding and showing

bfinzer:
[#186647000] Feature: Eyeball icon for map component allows hiding and showing of cases

* Implemented view and action methods in MapContentModel
* Improve MapPointLayer response to change in number of cases
* Install use-debounce
* Debounce MapPolyognLayer:refreshPolygons, MapPointLayer:refreshPoints and usePlotResponders:callRefreshPointPositions
* In MapPointLayer:refreshPoints call refreshPointSelection after the transition so that selected points end up on top

### DIFF
--- a/v3/src/components/graph/hooks/use-plot.ts
+++ b/v3/src/components/graph/hooks/use-plot.ts
@@ -4,7 +4,6 @@ import {isAlive} from "mobx-state-tree"
 import {onAnyAction} from "../../../utilities/mst-utils"
 import {mstAutorun} from "../../../utilities/mst-autorun"
 import {mstReaction} from "../../../utilities/mst-reaction"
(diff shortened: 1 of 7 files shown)


v3/src/components/graph/hooks/use-plot.ts
```diff
@@ -55,37 +56,9 @@ export const usePlotResponders = (props: IPlotResponderProps) => {
     instanceId = useInstanceIdContext(),
     refreshPointPositionsRef = useCurrent(refreshPointPositions)
```
kswenson, line 57: You probably don't need the ref any more, i.e. you should be able to just pass `refreshPointPositions` directly to `useDebouncedCallback()`.

kswenson (approved):
👍 Looks good! One minor suggestion

PR #1040, opened 2023-12-13
feat: Connecting Lines b/w points (PT-181922446)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/181922446

### DIFF
--- a/v3/src/components/graph/components/scatter-plot-utils.ts
+++ b/v3/src/components/graph/components/scatter-plot-utils.ts
@@ -56,14 +56,10 @@ export function scatterPlotFuncs(layout: GraphLayout, dataConfiguration?: IGraph
     const xValue = getXCoord(caseID)
     const yValue = getYCoord(caseID, plotNum)
     if (isFinite(xValue) && isFinite(yValue)) {
-      const caseData = dataset?.getCase(caseID)
+      const caseData = dataset?.getCase(caseID, { numeric: false })
       if (caseData) {
         const lineCoords: [number, number] = [xValue, yValue]
-        return {
-          caseData,
-          lineCoords,
-          plotNum
-        }
+        return { caseData, lineCoords, plotNum }
       }
     }
   }
@@ -72,12 +68,10 @@ export function scatterPlotFuncs(layout: GraphLayout, dataConfiguration?: IGraph
     const lineDescriptions: IConnectingLineDescription[] = []
     const dataset = dataConfiguration?.dataset
     dataset?.cases.forEach(c => {
-      const line = connectingLine(c.__id__, 0)
-      line && lineDescriptions.push(line)
-      const y2line = yAttrIDs[1] && connectingLine(c.__id__, 1)
-      y2line && lineDescriptions.push(y2line)
-      const yRightLine = yAttrIDs[2] && connectingLine(c.__id__, 2)
-      yRightLine && lineDescriptions.push(yRightLine)
+      yAttrIDs.forEach((_yAttrID, plotNum) => {
+        const line = connectingLine(c.__id__, plotNum)
+        line && lineDescriptions.push(line)
+      })
     })
     return lineDescriptions
   }
--- a/v3/src/components/graph/components/scatter-plot-utils.ts
+++ b/v3/src/components/graph/components/scatter-plot-utils.ts
@@ -2,6 +2,13 @@ import { ScaleBand, ScaleLinear } from "d3"
 import { IGraphDataConfigurationModel } from "../models/graph-data-configuration-model"
 import { GraphLayout } from "../models/graph-layout"
 import { ILineDescription, ISquareOfResidual } from "../adornments/shared-adornment-types"
+import { ICase } from "../../../models/data/data-set-types"
+
+export interface IConnectingLineDescription {
+  caseData: ICase
+  lineCoords: [number, number]
+  plotNum: number
+}
 
 export function scatterPlotFuncs(layout: GraphLayout, dataConfiguration?: IGraphDataConfigurationModel) {
   const { dataset: data, yAttributeIDs: yAttrIDs = [], hasY2Attribute, numberOfPlots = 1 } = dataConfiguration || {}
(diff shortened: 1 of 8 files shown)


v3/src/components/graph/components/scatterdots.tsx
```diff
@@ -164,6 +175,125 @@ export const ScatterDots = observer(function ScatterDots(props: PlotProps) {
...
+      // attributes or the presence of a parent attribute and the number of unique values for that attribute. If there
+      // are multiple Y attributes, the number of groups matches the number of Y attributes. Otherwise, if there's a
+      // parent attribute, the number of groups matches the number of unique values for that attribute. If there's only
+      // one Y attribute and no parent attribute, then there's a only single group. The code below builds lists of
+      // connecting lines and case IDs for each group.
+      const lineGroups: Record<string, IConnectingLineDescription[]> = {}
+      const allLineCaseIds: Record<string, string[]> = {}
+      const yAttrCount = dataConfiguration?.yAttributeIDs?.length ?? 0
+      connectingLines.forEach((lineDescription: IConnectingLineDescription) => {
+        const parentAttrValue = parentAttrID ? importValueToString(lineDescription.caseData[parentAttrID]) : undefined
```
kswenson, line 249: Do we always want string values in `caseData`? If so, you can pass `{ numeric: false }` to the call to `getCase()` in the `connectingLine()` function in `scatter-plot-utils.ts`.

v3/src/components/graph/components/scatter-plot-utils.ts
```diff
@@ -44,6 +51,37 @@ export function scatterPlotFuncs(layout: GraphLayout, dataConfiguration?: IGraph
...
+
+  function connectingLinesForCases() {
+    const lineDescriptions: IConnectingLineDescription[] = []
+    const dataset = dataConfiguration?.dataset
+    dataset?.cases.forEach(c => {
+      const line = connectingLine(c.__id__, 0)
+      line && lineDescriptions.push(line)
+      const y2line = yAttrIDs[1] && connectingLine(c.__id__, 1)
+      y2line && lineDescriptions.push(y2line)
+      const yRightLine = yAttrIDs[2] && connectingLine(c.__id__, 2)
```
kswenson, line 79: I'm confused by these hard-coded indices given that there can be an arbitrary number of y attributes.

v3/src/components/graph/components/scatterdots.tsx
```diff
@@ -164,6 +175,125 @@ export const ScatterDots = observer(function ScatterDots(props: PlotProps) {
...
+          .call(dataTip)
+          .attr("fill", "none")
+          .attr("stroke", color)
+          .attr("stroke-width", allCasesSelected ? 4 : 2)
+          .style("cursor", "pointer")
+          .style("opacity", connectingLinesActivatedRef.current ? 1 : 0)
+          .transition()
+          .duration(1000)
+          .style("opacity", showConnectingLines ? 1 : 0)
+          .on("end", connectingLinesCleanUp)
```
kswenson, line 291: As I just found out recently, the `end` function is called once for each element being iterated which was a problem in a map with a bunch of points, but may not be as big a deal with a presumably small number of lines.

v3/src/components/graph/components/scatterdots.tsx
```diff
@@ -258,10 +389,19 @@ export const ScatterDots = observer(function ScatterDots(props: PlotProps) {
     }, { name: "ScatterDots.renderSquares" })
   }, [refreshSquares, showSquares])
 
+  // Call refreshConnectingLines when Connecting Lines option is switched on and when all
+  // points are selected.
+  useEffect(function renderConnectingLines() {
+    return autorun(() => {
+      refreshConnectingLines()
+    }, { name: "ScatterDots.renderSquares" })
```
kswenson, line 397:
```suggestion
    }, { name: "ScatterDots.refreshConnectingLines" })
```

kswenson (changes requested):
👍 Looks pretty good -- just a handful of suggestions to consider.

v3/src/components/graph/components/scatterdots.tsx
```diff
@@ -164,6 +174,125 @@ export const ScatterDots = observer(function ScatterDots(props: PlotProps) {
...
+      return `<p>${d}</p>`
+    })
+
+  const handleConnectingLinesHover = useCallback((
+    event: MouseEvent, caseIDs: string[], parentAttrName?: string, parentAttrValue?: string
+    ) => {
+    const caseIdCount = caseIDs?.length ?? 0
+    const datasetName = dataset?.name ?? ""
+    const parentAttrPart = `${parentAttrName}: ${parentAttrValue}<br />with `
+    const countPart = `${caseIdCount} points (${datasetName}) on this line`
```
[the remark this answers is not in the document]
emcelroy: As discussed elsewhere, I'm going to leave some work related to this for later. I changed it to use the existing  localizable string carried over from V2, but removed support for the new form of the tool tip we want in V3 when there isn't a parent attribute present. I added a todo comment about that, and I'll make a new PT story about it, too.

v3/src/components/graph/components/scatterdots.tsx
emcelroy: Thanks. I believe we do want string values in this case because if we use the value, it will be as a key on an object. I found that even with passing `{ numeric: false }` to `getCase()`, however, there'd still be a type error here unless I used `importValueToString` or `lineDescription.caseData[parentAttrID] as string`. Without doing that, the type of `parentAttrValue` will be `IValueType` which can be a string, number, boolean, or undefined.

Am I missing something obvious that should make that second step unnecessary? Or do you think that's potentially a problem with `getCase()`? I'm not sure. It returns a value with of type `ICase`, [the definition of which](https://github.com/concord-consortium/codap/blob/3cbd8539f1029ea79dcd075e1e681eca751a08fb/v3/src/models/data/data-set-types.ts#L12-L16) seems very reasonable, although I would expect items returned by `getCase()` to have properties whose values are either numbers or strings or maybe undefined, but not booleans.

v3/src/components/graph/components/scatter-plot-utils.ts
emcelroy: Thanks for pointing that out. I think I've improved/fixed this with the latest commit.

v3/src/components/graph/components/scatterdots.tsx
kswenson: With the `{ numeric: false }`, _you_ know that the values are always strings, the issue is just that the compiler doesn't know it. I think I would suggest just using `String(lineDescription.caseData[parentAttrID])` as a simpler conversion than `importValueToString()`.

kswenson (approved):
👍 Changes look good

PR #1042, opened 2023-12-14
fix: Treating numeric as categorical should update count adornment correctly (PT-186016145)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/186016145

`isCaseInSubPlot` was not always identifying matches between values because the value for `cellKey[key]` is always a string whereas `caseData[key]` values can be a string or numeric. A simple fix is to compare `cellKey[key]` with `caseData[key].toString()`. I'm not sure if that's entirely safe though. I can't think of a case where it would be problematic, but I could be missing something. If I am missing something, perhaps I should look into adjusting the type for `cellKey` so its property values can be numeric as well as strings?

### DIFF
--- a/v3/src/components/graph/components/scatter-plot-utils.ts
+++ b/v3/src/components/graph/components/scatter-plot-utils.ts
@@ -68,7 +68,7 @@ export function scatterPlotFuncs(layout: GraphLayout, dataConfiguration?: IGraph
         // If the line has a category and it does not match the categorical legend value,
         // do not render squares.
         if (category && legendValue !== category && legendType === "categorical") return
-        const fullCaseData = dataset?.getCase(caseData.__id__)
+        const fullCaseData = dataset?.getCase(caseData.__id__, { numeric: false })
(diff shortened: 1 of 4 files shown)


kswenson (changes requested):
It seems that before every call to `isCaseInSubPlot()`, the `caseData` is retrieved with a call to `DataSet.getCase()`, which takes an optional `options` argument. I think it would be preferable to use `DataSet.getCase(caseId, { numeric: false })` because then we'd be using the original string value rather than using a conversion that might or might not convert to the original string value.

emcelroy:
> It seems that before every call to `isCaseInSubPlot()`, the `caseData` is retrieved with a call to `DataSet.getCase()`, which takes an optional `options` argument. I think it would be preferable to use `DataSet.getCase(caseId, { numeric: false })` because then we'd be using the original string value rather than using a conversion that might or might not convert to the original string value.

I see. Thanks 👍

kswenson (approved):
👍 LGTM

PR #1043, opened 2023-12-14
fix: Graphs with subplots not rendered correctly on import (PT-185577446)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/185577446

The effect function in the `useInitGraphLayout repetitions` reaction wasn't being triggered on import, and so the number of repetitions on the axes were not being correctly set. Adding `fireImmediately: true` to the reaction's options ensures the effect function is triggered after the data function's first run.


kswenson (approved):
👍 LGTM

PR #1044, opened 2023-12-14
fix: Percent values in subplots not correct (PT-186682759)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/186682759

Previously, the `percentValue` view in the count adornment model was dividing `casesInPlot` by `dataConfig?.subPlotCases(cellKey)` to determine the percent. In graphs with multiple subplots, that caused the percent value to be 100% for each subplot.

This change ensures the view divides `casesInPlot` by _all_ the plotted cases so there's an accurate percent value shown for each subplot.


kswenson (approved):
👍 LGTM

PR #1045, opened 2023-12-15
Avoid excessive CPU workload caused by the Pixi render loop

pjanik:
https://www.pivotaltracker.com/story/show/186691239

This PR replaces the automatic render loop handling done by PIXI.Application with a custom one. Instead of running the loop all the time, the rendering loop is now started only when necessary (resize or transitions). This change should help with the CPU load observed by @bfinzer. While PIXI rendering is generally inexpensive, it is not necessary in our case. This might become noticeable when dealing with thousands of points or multiple graphs.

### DIFF
--- a/v3/src/components/graph/utilities/pixi-points.ts
+++ b/v3/src/components/graph/utilities/pixi-points.ts
@@ -19,42 +19,86 @@ export interface IPixiPointStyle {
 }
 
 export class PixiPoints {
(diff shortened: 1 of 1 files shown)


kswenson (approved):
👍 Looks good -- I'm not familiar with the pixi details, but it all seems reasonable.

PR #1046, opened 2023-12-18
Added CODAP app id (project id) to Google provider settings [PT-186571976]

dougmartin:
The upcoming new CFM Google provider requires the app id to be set.


kswenson (approved):
👍 LGTM

PR #1047, opened 2023-12-18
feat: Import V2 adornments (PT-186683021)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/186683021

### DIFF
--- a/v3/src/v2/codap-v2-types.ts
+++ b/v3/src/v2/codap-v2-types.ts
@@ -135,6 +135,12 @@ interface ICodapV2ProportionCoordinates {
   proportionY: number
 }
 
+interface ICodapV2ValueModel {
+  isVisible: boolean
+  enableMeasuresForSelection: boolean
+  values: Record<string, number>
+}
+
 interface ICodapV2CountAdornment {
   isVisible: boolean
   enableMeasuresForSelection: boolean
@@ -153,8 +159,9 @@ interface ICodapV2MovableValueAdornment {
   enableMeasuresForSelection: boolean
   isShowingCount: boolean
   isShowingPercent: boolean
-  valueModels: Record<string, number>[] // ??
+  valueModels: ICodapV2ValueModel[]
 }
+
 interface ICodapV2MeanAdornment {
   isVisible: boolean
   enableMeasuresForSelection: boolean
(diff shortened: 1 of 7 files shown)


v3/src/v2/codap-v2-types.ts
```diff
@@ -128,6 +128,13 @@ export interface ICodapV2TableStorage extends ICodapV2BaseComponentStorage {
 export interface ICodapV2PlotStorage {
   verticalAxisIsY2: boolean
   adornments: Record<string, any>
+  areSquaresVisible: boolean
+  isLSRLVisible: boolean
+  movableLineStorage: Record<string, any>
+  movablePointStorage: Record<string, any>
+  multipleLSRLsStorage: Record<string, any>
+  multipleMovableValues: Record<string, any>
```
kswenson, line 136: Could try to type these better at some point, but I don't know how important it is, so it doesn't need to hold up the PR.

kswenson (approved):
👍 Looks good!

v3/src/v2/codap-v2-types.ts
emcelroy: @kswenson I'm not sure it's important either, but it seemed easy enough to do. So I went ahead and did that. I also added support for Connecting Lines. That also seemed pretty straightforward, but I'm going to request a follow-up review just in case.

v3/src/v2/codap-v2-types.ts
```diff
@@ -125,9 +125,145 @@ export interface ICodapV2TableStorage extends ICodapV2BaseComponentStorage {
...
+  lsrls: ICodapV2LSRL[]
+}
+
+interface ICodapV2ValueModel {
+  isVisible: boolean
+  enableMeasuresForSelection: boolean
+  values: Record<string, number>
+}
+
+interface ICodapV2MultipleMovableValuesStorage {
```
kswenson, line 249: Do we need this and `ICodapV2MovableValueAdornment` above? They appear to be the same.

kswenson (approved):
👍 Looks good

v3/src/v2/codap-v2-types.ts
emcelroy: Thanks for catching that. I've double checked and we do _not_ need both. I'm going to remove `ICodapV2MultipleMovableValuesStorage`, and do a little clean up of `ICodapV2MovableValueAdornment`.

PR #1048, opened 2023-12-19
Refactor metadata storage in PixiPoints class

pjanik:
https://www.pivotaltracker.com/story/show/186645440

This PR refactors how point metadata is stored and accessed in PixiPoints. Previously, I was relying on point indices, but that approach is a bit more fragile, requires recalculations, and is not so easy to use in some scenarios. Here, it's replaced with a Map that uses a Sprite reference as the key. I could also extend PIXI.Sprite and store metadata directly there, but that would most likely not work that well with TypeScript and would require quite a bit of type casting. The Map approach might be a bit more verbose, but it's also more type-friendly.


kswenson (approved):
👍 LGTM

PR #1049, opened 2023-12-19
Implement hover effect in PixiJS graph

pjanik:
https://www.pivotaltracker.com/story/show/186645440

This PR implements a basic hover effect for points. Transitions can run in parallel now. Note that tooltips will be covered later. Additionally, it breaks the marquee selection. I spent far too much time attempting to pass events through the canvas to the background SVG rectangle and am still unsure about how to approach it. However, it seems to me that at this point, I should just reimplement the dragging handler using PixiJS.

### DIFF
--- a/v3/src/components/graph/utilities/pixi-points.ts
+++ b/v3/src/components/graph/utilities/pixi-points.ts
@@ -144,7 +144,7 @@ export class PixiPoints {
     graphics.drawCircle(0, 0, radius)
     graphics.endFill()
     const texture = this.renderer.generateTexture(graphics, {
-      // A trick to make sprites/textures look still sharp when they"re scaled up (e.g. during hover effect).
+      // A trick to make sprites/textures look still sharp when they're scaled up (e.g. during hover effect).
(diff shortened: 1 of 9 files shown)


v3/src/components/graph/utilities/pixi-points.ts
```diff
@@ -120,7 +143,12 @@ export class PixiPoints {
     graphics.lineStyle(strokeWidth, stroke, strokeOpacity ?? 0.4)
     graphics.drawCircle(0, 0, radius)
     graphics.endFill()
-    const texture = this.renderer.generateTexture(graphics)
+    const texture = this.renderer.generateTexture(graphics, {
+      // A trick to make sprites/textures look still sharp when they"re scaled up (e.g. during hover effect).
```
kswenson, line 147:
```suggestion
      // A trick to make sprites/textures look still sharp when they're scaled up (e.g. during hover effect).
```

kswenson (approved):
👍 LGTM

PR #1050, opened 2023-12-21
Implement graph point dragging and click selection using PixiJS, fix various bugs of the SVG behavior

pjanik:
This PR primarily focuses on implementing point dragging behavior using PixiJS. Compared to the previous SVG-based version, the current implementation addresses some issues and introduces/restores V2 features. To make QA and verification easier, I've added new PT stories:

- Basic dragging behavior [PT-186717142](https://www.pivotaltracker.com/story/show/186717142).
- Point selection via mouse click or shift + click [PT-186725017](https://www.pivotaltracker.com/story/show/186725017).
- Touch support for dragging [PT-186724975](https://www.pivotaltracker.com/story/show/186724975).
- Enabling dragging of graph points outside graph boundaries, similar to V2 [PT-186724979](https://www.pivotaltracker.com/story/show/186724979).
- Enhancing the reliability of point dragging, preventing 'lost' dragged points, unlike in the SVG version [PT-186724949](https://www.pivotaltracker.com/story/show/186724949).
- Fixing the animation of points returning to their initial position after dragging is completed [PT-186724971](https://www.pivotaltracker.com/story/show/186724971).

Additionally, dragging should now be much smoother, and the performance mode for rendering/dragging should also work better with large datasets.

Regarding the implementation details, I've tried to utilize as much of the existing drag handlers as possible to ensure no disruption of their functionality. I've also removed some radius logic, as it seems that the hover effect was already handling that (and that's now managed directly by the PixiPoints class).

### DIFF
--- a/v3/src/components/graph/components/casedots.tsx
+++ b/v3/src/components/graph/components/casedots.tsx
@@ -1,17 +1,18 @@
-import {randomUniform, select} from "d3"
+import {randomUniform} from "d3"
 import {mstReaction} from "../../../utilities/mst-reaction"
 import React, {useCallback, useEffect, useRef, useState} from "react"
+import * as PIXI from "pixi.js"
 import {CaseData} from "../../data-display/d3-types"
 import {IDotsRef} from "../../data-display/data-display-types"
 import {handleClickOnCase, setPointSelection} from "../../data-display/data-display-utils"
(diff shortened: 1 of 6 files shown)


kswenson (approved):
👍 Nice!

PR #1051, opened 2023-12-22
chore: cleanup references to IDotsRef and replace it with PixiPointsRef

pjanik:
[PT-186730431]

This PR removes references to IDotsRef and the commented-out SVG code, as I believe it is no longer necessary. This is also the first step towards reimplementing map dots that are currently not functioning.

### DIFF
--- a/v3/src/components/data-display/hooks/use-data-tips.ts
+++ b/v3/src/components/data-display/hooks/use-data-tips.ts
@@ -2,7 +2,7 @@ import {select} from "d3"
 import {useEffect} from "react"
 import {tip as d3tip} from "d3-v6-tip"
 import {IDataSet} from "../../../models/data/data-set"
-import {IDotsRef, transitionDuration} from "../data-display-types"
+import {transitionDuration} from "../data-display-types"
 import {CaseData} from "../d3-types"
 import {getCaseTipText} from "../data-display-utils"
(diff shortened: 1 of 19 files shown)


v3/src/components/data-display/hooks/use-data-tips.ts
```diff
@@ -73,11 +74,12 @@ export const useDataTips = ({dotsRef, dataset, displayModel}: IUseDataTips) => {
...
     if (urlParams.noDataTips === undefined) {
-      dotsRef.current && select(dotsRef.current)
-        .on('mouseover', showDataTip)
-        .on('mouseout', hideDataTip)
-        .call(dataTip)
+      // TODO PIXI: reimplement data tips
+      // dotsRef.current && select(dotsRef.current)
+      //   .on('mouseover', showDataTip)
+      //   .on('mouseout', hideDataTip)
+      //   .call(dataTip)
```
kswenson, line 81: Is there a PT story for this so we don't lose track of it?

kswenson (approved):
👍 LGTM

v3/src/components/data-display/hooks/use-data-tips.ts
pjanik: Yes: https://www.pivotaltracker.com/story/show/186725308

PR #1052, opened 2023-12-22
v2: fix: export csv boundaries

kswenson:
While working on the v3 map component I figured an easy way to generate test data with boundaries would be to export a document with boundaries from v2 as a .csv file. In doing so, I discovered that boundary data appears as `[Object object]` in .csv files exported from CODAP v2.  The fix here is to check for boundary data and convert it appropriately for export. If, for some reason, we don't want to export boundary data to .csv (e.g. it can result in very large files), then we should remove it from the export entirely.


bfinzer (approved):
👍🏻LGTM

PR #1053, opened 2023-12-25
Plotted function should also support use of the x-axis attribute name as a dummy variable 

pjanik:
This PR implements an overlooked feature of the plotted function formula and addresses @bfinzer's comment:
https://www.pivotaltracker.com/story/show/186437311/comments/239220433
Now, users can use the x-axis attribute name as a dummy variable.


kswenson (approved):
👍 LGTM

PR #1054, opened 2023-12-26
Formulas: fix aggregate context support

pjanik:
https://www.pivotaltracker.com/story/show/186485885

This PR addresses the support of various functions within the aggregate context. Previously, the following formulas would fail:
- `mean(lookupByIndex("Mammals", "LifeSpan", caseIndex))`
- `mean(if(LifeSpan < 30, 1, 2))`
- `mean(1, true)`, `mean(1, false)`

Now, all should execute correctly. I believe I've streamlined the workflow for adding standard functions by introducing `evaluateWithAggregateContextSupport`, which is automatically applied. Developers no longer need to handle arrays explicitly when using the `evaluate:` variant of function implementation. However, when using `evaluateRaw:`, handling arrays remains necessary, as seen in the `prev` and `next` functions where I included it.

### DIFF
--- a/v3/src/models/formula/formula-types.ts
+++ b/v3/src/models/formula/formula-types.ts
@@ -35,11 +35,13 @@ export type IFormulaDependency = ILocalAttributeDependency | IGlobalValueDepende
 
 export type FValue = string | number | boolean
 
+export type FValueOrArray = FValue | FValue[]
+
 export type EvaluateFunc = (...args: FValue[]) => FValue
 
-export type EvaluateFuncWithAggregateContextSupport = (...args: (FValue | FValue[])[]) => FValue | FValue[]
+export type EvaluateFuncWithAggregateContextSupport = (...args: (FValueOrArray)[]) => FValueOrArray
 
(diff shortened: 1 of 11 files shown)


v3/src/models/formula/formula-types.ts
```diff
@@ -35,7 +35,9 @@ export type IFormulaDependency = ILocalAttributeDependency | IGlobalValueDepende
 
 export type FValue = string | number | boolean
 
-export type EvaluateFunc = (...args: FValue[]) => FValue | FValue[]
+export type EvaluateFunc = (...args: FValue[]) => FValue
+
+export type EvaluateFuncWithAggregateContextSupport = (...args: (FValue | FValue[])[]) => FValue | FValue[]
```
kswenson, line 40: Defining a type like:
```typescript
export type FValueOrArray = FValue | FValue[]
```
would eliminate some redundancy here.

kswenson (approved):
👍 Looks good!

v3/src/models/formula/formula-types.ts
pjanik: Thanks, added in https://github.com/concord-consortium/codap/pull/1054/commits/afed1d74c66cb3fa37e4e7aa624eff5cc3194055

PR #1056, opened 2023-12-28
Reimplement dots rendering in maps using PixiJS + multiple changes to PixiPoints to make it possible

pjanik:
[[PT-186730431]](https://www.pivotaltracker.com/story/show/186730431)

This PR reimplements map points/dots rendering using PixiJS. This was necessary to avoid maintaining all the SVG dots rendering code and having duplicates of multiple rendering methods.

Again, I encountered problems with the fact that the Canvas element covers everything underneath it and intercepts all user events. This was problematic for interacting with map polygons. However, in this instance, I managed to pass events from PixiPoints to the underlying polygons. It's all implemented in the `PixiPoints.setupBackgroundEventDistribution()` method.

Also, I've added a resize observer, as there was no explicit code for resizing in the map code where I could simply invoke `PixiPoints.resize` (as with graphs).

There's one issue I didn't address. It doesn't seem possible to stop the propagation of events from the PixiJS canvas to the parent elements. Consequently, the click events are always captured by `use-map-model`. This was causing selected points to be immediately deselected. That's why I commented out the deselection line there. I've also added a new PT story to track this issue: https://www.pivotaltracker.com/story/show/186747322. I hope this compromise is minor enough to proceed.

Also, Cypress tests are still failing in this PR, but PR https://github.com/concord-consortium/codap/pull/1058 fixes them. So these changes can be tested at:
https://codap3.concord.org/branch/fix-transition-issues/index.html


kswenson (approved):
👍 Looks good!

PR #1057, opened 2023-12-28
Fix connecting lines interaction

pjanik:
https://www.pivotaltracker.com/story/show/186747801

This PR fixes interaction with connecting lines. That's another example of where PixiJS creates challenges. Since the canvas is covering the connecting lines, I had to use the background event redistribution mechanism added recently. However, if we have more layers that require interaction, this method might become difficult to continue and maintain.


kswenson (approved):
👍 Nice!

PR #1058, opened 2023-12-28
Handle scenario when PixiTransition tries to update a point that was deleted

pjanik:
https://www.pivotaltracker.com/story/show/186747821


kswenson (approved):
👍 LGTM

PR #1059, opened 2023-12-29
Fix dataset caching

pjanik:
[[PT-186751983]](https://www.pivotaltracker.com/story/show/186751983)

I've found this bug while working on PixiJS and transitions, but I think we should merge it into the `main` branch. Note that `get isCaching` was actually cached in some situations, resulting in incorrect value being returned. Also, I changed `views` to `volatile`, although it doesn't really affect the functionality (caching still works within the volatile block). However, considering this set of functionalities extends beyond just views, I hope 'volatile' will be more descriptive.


kswenson (approved):
👍 LGTM

PR #1060, opened 2023-12-29
Better support of concurrent transitions in PixiJS graphs

pjanik:
https://www.pivotaltracker.com/story/show/186751807

This PR significantly changes the transition implementation in the PixiJS graph. The previous version didn't work too well with multiple concurrent transitions or user interactions. This new approach should resolve those issues. It's a bit difficult to demonstrate with the default 1000ms transition time, but when I set it to 6s, the purpose becomes easier to understand:

https://github.com/concord-consortium/codap/assets/767857/74ed361b-6477-4a00-85cb-e439fb0a4750

While it might not seem that useful, this is just an example. It's relatively easy to trigger some artifacts in the previous version, and I thought I should work on it so that someone else doesn't spend a lot of time trying to track down this issue later.

Also, it includes commit from PR https://github.com/concord-consortium/codap/pull/1059, as otherwise things wouldn't work correctly. I'll deal with it while merging `main` into `pixi-main` later.

And the last thing to double-check is the update of `onDragEnd` in v3/src/components/graph/components/scatterdots.tsx. This was also necessary to make all that work, but I'm not 100% sure if I'm not breaking something else.

### DIFF
--- a/v3/src/components/graph/utilities/pixi-points.ts
+++ b/v3/src/components/graph/utilities/pixi-points.ts
@@ -123,7 +123,7 @@ export class PixiPoints {
     if (this.anyTransitionActive) {
       PixiTransition.transitionStep(this.targetProp, this.startProp)
     } else {
-      // The only reason for ticket to run is to handle ongoing transitions. If there are no transitions, we can stop.
+      // The only reason for ticker to run is to handle ongoing transitions. If there are no transitions, we can stop.
       this.ticker.stop()
     }
     this.renderer.render(this.stage)
--- a/v3/src/components/graph/utilities/pixi-points.ts
+++ b/v3/src/components/graph/utilities/pixi-points.ts
@@ -1,6 +1,6 @@
 import * as PIXI from "pixi.js"
 import { CaseData } from "../../data-display/d3-types"
-import { PixiTransition } from "./pixi-transition"
+import { PixiTransition, TransitionPropMap, TransitionProp } from "./pixi-transition"
 import { hoverRadiusFactor, transitionDuration } from "../../data-display/data-display-types"
 
 const DEFAULT_Z_INDEX = 0
@@ -72,8 +72,9 @@ export class PixiPoints {
 
   resizeObserver?: ResizeObserver
 
-  activeTransitions = 0
-  currentTransition?: PixiTransition
+  currentlySetupTransition?: PixiTransition
+  targetProp: TransitionPropMap = {}
+  startProp: TransitionPropMap = {}
 
   onPointClick?: PixiPointEventHandler
   onPointDragStart?: PixiPointEventHandler
@@ -81,7 +82,7 @@ export class PixiPoints {
   onPointDragEnd?: PixiPointEventHandler
 
   constructor(options?: IPixiPointsOptions) {
-    this.ticker.add(() => this.renderer.render(this.stage))
+    this.ticker.add(this.tick.bind(this))
     this.stage.addChild(this.background)
     this.stage.addChild(this.pointsContainer)
     // Enable zIndex support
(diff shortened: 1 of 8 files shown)


v3/src/components/graph/utilities/pixi-points.ts
```diff
@@ -114,46 +115,112 @@ export class PixiPoints {
...
-    return this.caseIDToPoint.get(caseId) as PIXI.Sprite
+  get anyTransitionActive() {
+    return PixiTransition.anyTransitionActive(this.targetProp)
+  }
+
+  tick() {
+    if (this.anyTransitionActive) {
+      PixiTransition.transitionStep(this.targetProp, this.startProp)
+    } else {
+      // The only reason for ticket to run is to handle ongoing transitions. If there are no transitions, we can stop.
```
kswenson, line 126:
```suggestion
      // The only reason for ticker to run is to handle ongoing transitions. If there are no transitions, we can stop.
```

v3/src/components/graph/utilities/pixi-transition.ts
```diff
@@ -10,109 +10,73 @@ const defaultInterpolation = smoother
...
-  setTargetXyProp(propKey: SupportedPropKey, point: PIXI.Sprite, x: number, y: number) {
-    let targetProp = this.targetProp[propKey]
-    let startProp = this.startProp[propKey]
-    if (!targetProp || !startProp) {
-      targetProp = this.targetProp[propKey] = new Map()
-      startProp = this.startProp[propKey] = new Map()
-    }
-    targetProp.set(point, { x, y })
-    startProp.set(point, { x: point[propKey].x, y: point[propKey].y })
+  getFactor(now: number) {
```
kswenson, line 73: Perhaps `getProgress`? `getFactor` seems sufficiently generic as to be unhelpful.

v3/src/components/graph/utilities/pixi-transition.ts
```diff
@@ -10,109 +10,73 @@ const defaultInterpolation = smoother
...
-    transitionFrame()
-  }
-
-  handleOnEnd() {
-    this.onEndCallback?.()
-    this.isFinished = true
-  }
+      for (const [point, target] of targetProp.entries()) {
+        const transition = target.transition
+        const factor = transition.getFactor(now)
```
kswenson, line 48: As suggested elsewhere, perhaps `progress` would be more intuitive than `factor`.

v3/src/components/graph/utilities/pixi-points.ts
```diff
@@ -72,16 +72,17 @@ export class PixiPoints {
 
   resizeObserver?: ResizeObserver
 
-  activeTransitions = 0
-  currentTransition?: PixiTransition
+  currentlySetupTransition?: PixiTransition
```
kswenson, line 75: Simplify name to `currentTransition`?

kswenson (approved):
👍 Looks good! Just a couple naming suggestions.

v3/src/components/graph/utilities/pixi-transition.ts
pjanik: Thanks, I changed it in https://github.com/concord-consortium/codap/pull/1060/commits/225f14e744826088a1a4009ea5f6ddc9d134a54c.

v3/src/components/graph/utilities/pixi-transition.ts
pjanik: Thanks, I changed it to progress in https://github.com/concord-consortium/codap/pull/1060/commits/225f14e744826088a1a4009ea5f6ddc9d134a54c.

v3/src/components/graph/utilities/pixi-points.ts
pjanik: That's what I used before. However, I think `currentTransition` implies it's a transition currently playing, which isn't the case here. This refers to the transition we're setting up via a provided callback. Once it starts, we actually set this property to undefined. Hence, I added `setup` to the name to clarify. 

I'm happy to improve this name, but given this context, do you have any suggestions? Or do you still think that `currentTransition` is better regardless?

kswenson: Good point, but after considering alternatives I'm inclined to think that `currentTransition` still seems best along with a comment that explains what you explained above, i.e. that it's the currently installed/configured transition, not necessarily that it's actively transitioning.

pjanik: @kswenson, I've pushed this change directly to pixi-main branch: https://github.com/concord-consortium/codap/commit/e43bbb299761587ae1efde176e2286bec68135f0

PR #1064, opened 2024-01-02
feat: Plotted Value should show units (PT-186681717)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/186681717

These changes will add units to labels and tool tips for the Plotted Value _and_ all other Univariate Measure adornments except the Box Plot. I believe that's fine even though the story didn't call for it since V2 shows units for all those adornments, too.


kswenson (approved):
👍 LGTM

PR #1065, opened 2024-01-03
fix: Movable Line, LSRL equation value format should match V2 (PT-186611743)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/186611743

These changes add two new functions to v3/src/components/graph/utilities/graph-utils.ts (`findNeededFractionDigits` and `roundToSignificantDigits`) that are lightly modified TypeScript versions of the V2 functions `findNeededFractionDigits` ([in plot_utilities.js](https://github.com/concord-consortium/codap/blob/b8df5a48e552c8d6d19ff8a3d90d8a553e57ac9d/apps/dg/components/graph/utilities/plot_utilities.js#L274-L339)) and `roundToSignificantDigits` ([in math_utilities.js](https://github.com/concord-consortium/codap/blob/b8df5a48e552c8d6d19ff8a3d90d8a553e57ac9d/apps/dg/utilities/math_utilities.js#L43-L95)). They're used to format the values in the Movable Line and LSRL adornments' equation strings.

There is also a restructuring of the existing `equationString` and `lsrlEquationString` functions, some minor changes to how some of the values were computed, and updates to related tests.

### DIFF
--- a/v3/src/components/graph/utilities/graph-utils.ts
+++ b/v3/src/components/graph/utilities/graph-utils.ts
@@ -10,6 +10,8 @@ import {defaultSelectedColor, defaultSelectedStroke, defaultSelectedStrokeWidth,
 import {IDataConfigurationModel} from "../../data-display/models/data-configuration-model"
 import { isFiniteNumber } from "../../../utilities/math-utils"
 import { IGraphDataConfigurationModel } from "../models/graph-data-configuration-model"
+import { GraphLayout } from "../models/graph-layout"
+import t from "../../../utilities/translation/translate"
 
 /**
  * Utility routines having to do with graph entities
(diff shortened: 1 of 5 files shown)


v3/src/components/graph/utilities/graph-utils.ts
```diff
@@ -198,49 +174,191 @@ export function percentString(value: number) {
...
+  function greatestDiff() {
+    return Math.max(
+      Math.abs(trialCoords.pt1.x - currCoords.pt1.x),
+      Math.abs(trialCoords.pt1.y - currCoords.pt1.y),
+      Math.abs(trialCoords.pt2.x - currCoords.pt2.x),
+      Math.abs(trialCoords.pt2.y - currCoords.pt2.y)
+    )
+  }
+
+  if (isNaN(slope) || isNaN(intercept) || !slope || !intercept) {
```
kswenson, line 252: Slope and intercept can validly be 0. Is treating them the same as NaNs correct here? Could use `isFiniteNumber()` instead.

kswenson (approved):
👍 LGTM -- given that you said the code is largely ported from v2, I didn't interrogate the implementation too closely.

v3/src/components/graph/utilities/graph-utils.ts
emcelroy: @kswenson I believe you're right that using `isFiniteNumber()` would be better. This part is _almost_ verbatim from [the V2 version](https://github.com/concord-consortium/codap/blob/b8df5a48e552c8d6d19ff8a3d90d8a553e57ac9d/apps/dg/components/graph/utilities/plot_utilities.js#L301-L303). I think I was misinterpreting how `SC.none()` works -- `SC.none(0)` returns `false` instead of `true` like I wrongly assumed.

`if (isNaN(iSlope) || isNaN(iIntercept) || SC.none(iSlope) || SC.none(iIntercept)) {`

I've changed this in the latest commit. Does that look all right to you?

kswenson: `SC.none()` is essentially like `== null`, i.e. it captures `null` and `undefined`, so `isFiniteNumber()` should be an appropriate replacement.

kswenson (approved):
👍 LGTM

PR #1066, opened 2024-01-04
fix: Percent adornment category should be remembered (PT-186016085)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/186016085

This removes the part of the autorun in `CountAdornment` that would change the model's `percentType` property to `cell` when the graph changed to a configuration that doesn't support the `row` or `column` options. Instead, the `percentValue` model view now checks the configuration and defaults to using `cell` in all instances where that's the only supported option. That means the `percentType` property value is only changed when the user changes it via the UI, and so the previously selected option will be remembered when the graph changes from a configuration that supports the percent type options, to one that doesn't, and then back to one that does.

### DIFF
--- a/v3/src/components/graph/adornments/count/count-adornment-component.tsx
+++ b/v3/src/components/graph/adornments/count/count-adornment-component.tsx
@@ -25,15 +25,8 @@ export const CountAdornment = observer(function CountAdornment({model, cellKey}:
   useEffect(() => {
     return autorun(() => {
       prf.begin("CountAdornment.autorun")
-      const shouldShowPercentOption = dataConfig ? dataConfig.categoricalAttrCount > 0 : false
-      const shouldShowPercentTypeOptions = dataConfig?.hasExactlyTwoPerpendicularCategoricalAttrs
-
-      // set percentType to 'cell' if attributes change to a configuration that doesn't support 'row' or 'column'
-      if (!shouldShowPercentTypeOptions && model?.percentType !== "cell") {
-        model.setPercentType("cell")
-      }
-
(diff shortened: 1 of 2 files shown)


v3/src/components/graph/adornments/count/count-adornment-component.tsx
```diff
@@ -25,15 +25,8 @@ export const CountAdornment = observer(function CountAdornment({model, cellKey}:
...
-      const shouldShowPercentOption = dataConfig ? dataConfig.categoricalAttrCount > 0 : false
-      const shouldShowPercentTypeOptions = dataConfig?.hasExactlyTwoPerpendicularCategoricalAttrs
-
-      // set percentType to 'cell' if attributes change to a configuration that doesn't support 'row' or 'column'
-      if (!shouldShowPercentTypeOptions && model?.percentType !== "cell") {
-        model.setPercentType("cell")
-      }
-
       // set showPercent to false if attributes change to a configuration that doesn't support percent
+      const shouldShowPercentOption = dataConfig ? dataConfig.categoricalAttrCount > 0 : false
```
kswenson, line 29: Could also be
```suggestion
      const shouldShowPercentOption = !!dataConfig?.categoricalAttrCount
```

kswenson (approved):
👍 LGTM

v3/src/components/graph/adornments/count/count-adornment-component.tsx
emcelroy: Yeah, that's better. Thanks 👍

PR #1067, opened 2024-01-04
fix: Import V2 adornments (PT-186683021)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/186683021

This fixes a problem QA found where Movable Points and Movable Lines weren't being correctly rendered when there were categorical attributes on both the top and right splits.


kswenson (approved):
👍 LGTM

PR #1068, opened 2024-01-05
Fix graph rendering in Safari (incorrect offset) 

pjanik:
https://www.pivotaltracker.com/story/show/186784214

This PR fixes incorrect PixiJS layer rendering observed by @bfinzer. I hope the comments explain why it had to be fixed that way.

Generally, if any parent of the foreignObject, such as an SVG element or a G element, has any kind of translation (either via transform or x/y attributes), this breaks rendering in Safari. The only workaround I found was to apply the X/Y attributes directly to the foreignObject element. Therefore, I had to split the previous SVG parent into two G groups around the foreign object.

I've also replaced the SVG tag with a G tag and dropped setting width and height, as both adjustments seemed cleaner to me. Initially, I was unsure whether embedding SVG within SVG was allowed, but after checking, I found that it is. However, please let me know if I'm overlooking any reasons why SVG was used previously.

### DIFF
--- a/v3/src/components/graph/components/graph.tsx
+++ b/v3/src/components/graph/components/graph.tsx
@@ -58,7 +58,8 @@ export const Graph = observer(function Graph({graphController, graphRef, pixiPoi
     layout = useGraphLayoutContext(),
     xScale = layout.getAxisScale("bottom"),
     svgRef = useRef<SVGSVGElement>(null),
-    plotAreaSVGRef = useRef<SVGSVGElement>(null),
+    plotArea1Ref = useRef<SVGGElement>(null),
+    plotArea2Ref = useRef<SVGGElement>(null),
     backgroundSvgRef = useRef<SVGGElement>(null),
     pixiContainerRef = useRef<SVGForeignObjectElement>(null),
     xAttrID = graphModel.getAttributeID('x'),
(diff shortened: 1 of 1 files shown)


v3/src/components/graph/components/graph.tsx
```diff
@@ -228,12 +235,22 @@ export const Graph = observer(function Graph({graphController, graphRef, pixiPoi
...
+            element content incorrectly if it or its parent is translated. The only workaround, as of January 2024, is
+            to use the X and Y attributes of the foreignElement tag itself. See:
+            https://www.pivotaltracker.com/story/show/186784214
+          */}
+          <g ref={plotArea1Ref}>
+            {/* Components rendered below the dots/points should be added to this group. */}
             {renderPlotComponent()}
-            <foreignObject ref={pixiContainerRef} x={0} y={0} width="100%" height="100%"/>
+          </g>
+          <foreignObject ref={pixiContainerRef} width="100%" height="100%"/>
```
kswenson, line 249: Not new to this PR, but I think the width and height are incorrect here -- this makes the pixi container the size of the entire component when it need only be the size of the plot area.

kswenson (approved):
👍 Looks good -- I noted one issue about the size of the pixi container that should be addressed at some point, but need not hold up this PR given that it wasn't introduced here, but warrants at least a TODO comment or PT story.

v3/src/components/graph/components/graph.tsx
pjanik: I've just fixed it in the last commit.

pjanik:
> 👍 Looks good -- I noted one issue about the size of the pixi container that should be addressed at some point, but need not hold up this PR given that it wasn't introduced here, but warrants at least a TODO comment or PT story.

I've addressed that and added links to bug reports that you mentioned in PT story. So, I'm merging this PR.

PR #1069, opened 2024-01-05
fix: count adornment responds to data changes [#186513124]

kswenson:
[[#186513124]](https://www.pivotaltracker.com/story/show/186513124) Count adornment does not immediately update when case data modified

Alternative to the fix in #1063.
>I'm not sure calling the data configuration model's _invalidateCases action from the count adornment component is the best thing to do

@emcelroy Your misgivings were warranted. Ultimately, the problem was with the caching in the `GraphDataConfigurationModel`. `GraphDataConfigurationModel.allPlottedCases` wasn't being invalidated/updated on value changes. In addition to caching the result from `allPlottedCases` with `cachedFnFactory`, this PR also changes the caching of `allPlottedCases`, `subPlotCases`, `rowCases`, and `columnCases` so that they cache and return arrays of case ids rather than entire case objects. Many clients don't need the case data and those that do should request it themselves from the DataSet rather than caching what potentially amounts to an entire copy of the DataSet locally. Finally, we add an observable volatile `casesChangeCount` property to the `DataConfigurationModel` which is updated when `clearCasesCache` is called and reference it in the `CountAdornmentComponent` to trigger a re-render. Other adornments may need to do the same.

### DIFF
--- a/v3/src/components/graph/adornments/count/count-adornment-component.tsx
+++ b/v3/src/components/graph/adornments/count/count-adornment-component.tsx
@@ -21,6 +21,8 @@ export const CountAdornment = observer(function CountAdornment({model, cellKey}:
   const casesInPlot = dataConfig?.subPlotCases(cellKey)?.length ?? 0
   const percent = model.percentValue(casesInPlot, cellKey, dataConfig)
   const displayPercent = model.showCount ? ` (${percentString(percent)})` : percentString(percent)
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const rerenderOnCasesChange = dataConfig?.casesChangeCount
 
   useEffect(() => {
     return autorun(() => {
--- a/v3/src/components/data-display/models/data-configuration-model.ts
+++ b/v3/src/components/data-display/models/data-configuration-model.ts
@@ -50,7 +50,8 @@ export const DataConfigurationModel = types
     actionHandlerDisposer: undefined as (() => void) | undefined,
     filteredCases: [] as FilteredCases[],
     handlers: new Map<string, (actionCall: ISerializedActionCall) => void>(),
-    pointsNeedUpdating: false
+    pointsNeedUpdating: false,
+    casesChangeCount: 0
   }))
   .views(self => ({
     get isEmpty() {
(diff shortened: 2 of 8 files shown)


v3/src/components/graph/adornments/count/count-adornment-component.tsx
```diff
@@ -21,6 +21,8 @@ export const CountAdornment = observer(function CountAdornment({model, cellKey}:
   const casesInPlot = dataConfig?.subPlotCases(cellKey)?.length ?? 0
   const percent = model.percentValue(casesInPlot, cellKey, dataConfig)
   const displayPercent = model.showCount ? ` (${percentString(percent)})` : percentString(percent)
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const rerenderOnCasesChange = dataConfig?.casesChangeCount
```
emcelroy, line 25: Instead of having this unused variable, would it be at all better to do something like add a ref...

`const displayCount = useRef(casesInPlot)`

...and then have a `useEffect` like the below?

```
  useEffect(function handleCaseChanges() {
    return mstReaction(
      () => dataConfig?.casesChangeCount,
      () => displayCount.current = dataConfig?.subPlotCases(cellKey)?.length ?? 0,
      {name: "CountAdornment.handleCaseChanges"}, model)
  }, [cellKey, dataConfig, dataConfig?.casesChangeCount, model])
```
I tried it out and that seems to work fine. I'm not sure it's much better than having the unused variable, though. It's certainly a lot more code.

emcelroy (approved):
Looks good and makes sense -- thank you! I left one comment on the change to `CountAdornment` but I'm not sure what I suggest would be an improvement.

v3/src/components/graph/adornments/count/count-adornment-component.tsx
kswenson: I thought about that, but in addition to being more code it's also a lot more overhead of spinning up reactions and effects. The fact that the component is declared as an `observer` is automatically setting up the reaction we need. All that's necessary is that we reference an observable property within the component, hence the unused variable.

PR #1070, opened 2024-01-05
chore: refactor AxisModel to eliminate need for dispatcher

kswenson:
Doing some beginning of year house-cleaning of some branches that have been sitting around on my machine...

Using `types.literal` allows MST to infer the appropriate type from the snapshot and obviates the need for a dispatcher.


bfinzer (approved):
👍🏻LGTM

PR #1072, opened 2024-01-05
feat: Count labels change their font size to fit (PT-181914458)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/181914458

All instances of the count adornment need to have the same font size at all times, but it may be the case that only one instance's text is too wide for its subplot while all others fit fine. So in order to dynamically change the font size of _all_ instances at the same time, these changes add a `defaultFontSize` property to `AdornmentsStore` and have the `CountAdornment` component modify that as needed when the graph is resized.

(BTW, I was hoping there'd be a pure CSS solution for this, but I couldn't find anything that would work.)


kswenson (commented):
>This is still a work in progress but I'm hoping for some thoughts on the general approach.

Adding `defaultFontSize` to the `AdornmentsStore` seems like a reasonable approach.

kswenson (approved):
👍 LGTM

PR #1073, opened 2024-01-08
chore: improve eslint configuration

kswenson:
Added
```
    "componentWrapperFunctions": [
      // The name of any function used to wrap components, e.g. Mobx `observer` function.
      // If this isn't set, components wrapped by these functions will be skipped.
      "observer"
    ]
```
to our ESLint configuration so that components wrapped in `observer` are linted correctly.


kswenson:
Merging without review as it's a simple configuration change.

PR #1074, opened 2024-01-09
Adding map and updating legend cypress tests

tejal-shah:


v3/cypress/e2e/map.spec.ts
```diff
@@ -0,0 +1,198 @@
...
+    mlh.selectCategoryColorForCategoricalLegend(arrayOfValues[0].values[1])
+    mlh.verifyCategoricalLegendKeySelected(arrayOfValues[0].values[1], 67)
+
+    // TODO: unselecting by clicking away in map doesn't work
+    // Uncomment the below two lines once it's fixed
+    // mlh.unselectLegendCategory()
+    // mlh.verifyNoLegendCategorySelectedForCategoricalLegend()
+
+    // TODO: This drag legend doesn't work for maps
+    // Uncomment line 95 and remove lines 96, 97 once it starts working
```
emcelroy, line 94: This is picky, sorry, but would it be better to word these instructions more like you do elsewhere, something like "Uncomment the next line and remove the two after it" just in case the line numbers change?

emcelroy (approved):
Looks good 👍 I made one small suggestion for a comment, but I don't think implementing that is critical.

tejal-shah:
@emcelroy , Thanks! I've updated my comments now and have also included PT bug ids int hose comments now.

PR #1075, opened 2024-01-10
fix: Style CFM dialog provider message

dougmartin:
Without this fix the provider message, which contains the username, is invisible as it is white on a white background.  This also sets the font-size to be the same as the provider message in SageModeler.


eireland (approved):
👍 LGTM

PR #1077, opened 2024-01-10
fix: Percent values not always correct (PT-186822118)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/186822118

The `percentValue` view in the count adornment model was using `dataConfig?.allPlottedCases.length` to get the number of plotted cases, but that was always equal to `0`. Changing it to `dataConfig?.allPlottedCases().length` results in getting the actual number of plotted cases, and so the correct percent values.


kswenson (approved):
👍 LGTM -- I'm surprised this didn't generate a type error when `allPlottedCases()` was changed to no longer be an accessor. 🤷

PR #1078, opened 2024-01-11
feat: Show both Mean & Median tips when overlapped (PT-186150345)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/186150345

Introduces a minor feature new in V3. If the Mean and Median adornments are both active and have the same value or values that result in their lines being rendered within two pixels of each other, hovering over one adornment's line will result in the other adornment's text tip appearing (in addition to its own).

### DIFF
--- a/v3/src/components/graph/adornments/univariate-measures/univariate-measure-adornment-simple-component.tsx
+++ b/v3/src/components/graph/adornments/univariate-measures/univariate-measure-adornment-simple-component.tsx
@@ -44,6 +44,8 @@ export const UnivariateMeasureAdornmentSimpleComponent = observer(
     const valueRef = useRef<SVGGElement>(null)
     const valueObjRef = useRef<IValue>({})
     const labelRef = useRef<HTMLDivElement>(null)
+    const isBlockingOtherMeasure = dataConfig &&
+      helper.blocksOtherMeasure({adornmentsStore, attrId, dataConfig, isVertical: isVertical.current})
 
     const highlightCovers = useCallback((highlight: boolean) => {
       const covers = selectAll(`#${helper.measureSlug}-${containerId} .${helper.measureSlug}-cover`)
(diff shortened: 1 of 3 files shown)


v3/src/components/graph/adornments/univariate-measures/univariate-measure-adornment-model.ts
```diff
@@ -27,6 +27,7 @@ export const UnivariateMeasureAdornmentModel = AdornmentModel
   .props({
     measures: types.map(MeasureInstance),
     showMeasureLabels: false,
+    showTipOnOverlapHover: false,
```
emcelroy, line 30: I'm now wondering if adding this property to the models is at least flirting with overkill? The component could check whether the adornment is of type `Mean` or `Median` instead of using this property.

kswenson: The primary reason for storing something in an MST model is if it's user-modifiable state that should be serialized with the document. If it's not user-modifiable state, then storing it with the document just bloats the document unnecessarily. In this case it seems like a static property of the adornment determinable from the type, in which case it could be handled with a utility function that checked the type of the adornment or by adding an optional property to the adornment registration.

v3/src/components/graph/adornments/univariate-measures/univariate-measure-adornment-simple-component.tsx
```diff
@@ -258,7 +269,7 @@ export const UnivariateMeasureAdornmentSimpleComponent = observer(
       if (measure) {
         addAdornmentElements(measure, valueObjRef.current, newLabelObj)
       }
-    }, [model, helper.instanceKey, labelRef, addAdornmentElements])
+    }, [addAdornmentElements, helper.instanceKey, model.isVisible, model?.measures])
```
kswenson, line 272: I think it would be sufficient to use:
```suggestion
    }, [addAdornmentElements, helper.instanceKey, model])
```
The value of `model` is captured by the function such that the function would not work if the `model` changed and the callback were not invalidated. Changes to the properties of the model don't invalidate the function in the same way, i.e. the callback behaves as intended after such changes.

kswenson (changes requested):
It seems that `showTipOnOverlapHover` need not be a property on the model.

v3/src/components/graph/adornments/univariate-measures/univariate-measure-adornment-model.ts
emcelroy: @kswenson That makes a lot of sense, thanks 👍

I'm not sure how much this changes anything, but I only just realized that in my zeal to simplify things, I accidentally lost an important part of the feature -- the part that checks if the other adornment's line is within 2px of the current adornment's line. Sorry for the premature review request. I've switched this to a draft PR and temporarily removed @bfinzer from the Reviewers list. I'll work on getting that missing piece back.

emcelroy: @kswenson @bfinzer Sorry again for the mix up. I've updated the PR with a new commit that drops the model changes entirely and adds a function to `UnivariateMeasureAdornmentHelper` for determining if an adornment is blocking another from receiving mouseover events.

kswenson (approved):
👍 LGTM

PR #1081, opened 2024-01-15
Fix PixiJS point style update

pjanik:
This PR fixes two bugs found by Tejal:

https://www.pivotaltracker.com/story/show/186828410
https://www.pivotaltracker.com/story/show/186828354

The problem is that `style: Partial<IPixiPointStyle>` sometimes has the `fill` property equal to an empty string, and PixiJS doesn't know how to render an empty fill color. I bet the client code meant not to update the fill. The fix might do more than necessary, but I had very limited time, and at least it should fix all the similar, possibly hidden issues.


kswenson (approved):
👍 LGTM

PR #1082, opened 2024-01-16
fix: Adornment elements outside plot should not cover other elements (PT-186035970)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/186035970

Ensures the graph layout is updated to accommodate for adornment banners when they're present, and that the adornment banners don't hide the highlighted top split drop zone when an attribute is being dragged.


v3/src/components/graph/adornments/adornments.tsx
```diff
@@ -12,16 +13,37 @@ import { useGraphDataConfigurationContext } from "../hooks/use-graph-data-config
...
+  
+    updateLayout()
+  
+    return () => {
+      mstAutorun(
+        updateLayout,
+        { name: "Graph.handleAdornmentBannerCountChange" }, graphModel
+      )
+    }
+  }, [layout, graphModel, bannerCount])
```
kswenson, line 46: I think what you want here is this simpler version:
```suggestion
  const { left, top, width, height } = layout.computedBounds.plot

  useEffect(function handleAdornmentBannerCountChange() {
    return mstAutorun(() => {
      const bannerCount = graphModel.adornmentsStore.activeBannerCount
      const bannersHeight = bannerCount * kGraphAdornmentsBannerHeight
      layout.setDesiredExtent("top", bannersHeight)
      }, { name: "Graph.handleAdornmentBannerCountChange" }, graphModel
    )
  }, [graphModel, layout])
```
There are several issues with your version:
1. the `defaultTop` ref will be computed on every render, but only the the first value will be retained by the ref.
2. the purpose of the `mstAutorun` is apparently to recompute the banner height when the number of banners changes, but the `bannerCount` is retrieved outside the `mstAutorun`, so changes won't trigger it.
3. the return value of the effect is a function which calls `mstAutorun`, which means that when the effect's cleanup function is run the autorun will be installed but will never be cleaned up.
4. when the `bannerCount` changes, the effect will run, which will a) execute the cleanup function, installing an autorun that will execute `updateLayout`, b) call `updateLayout` again as part of the effect, and c) return a new cleanup function.

With my version, the effect is only re-run when the `layout` or the `graphModel` change, which is effectively never. The `mstAutorun` accesses all of the observables it cares about, primarily the `bannerCount`, so it will re-run on change. And the return value of the effect is a cleanup function which disposes of the autorun, because that's the return value of `mstAutorun`. Finally, I don't think it's necessary to subtract the banner height from the top and then add it back later -- I'm pretty sure the top value we want is just the total banner height. Check me to make sure that this all still works.

v3/src/components/graph/components/graph.scss
```diff
@@ -111,6 +111,7 @@
 
   &.active {
     border: 5px solid rgb(255, 216, 102);
+    z-index: 10;
```
kswenson, line 114: At some point we need to stop playing these ad hoc z-index games, but I guess now is not the time.

kswenson (changes requested):
Made some suggestions to simplify the code and make it easier to reason about.

v3/src/components/graph/adornments/adornments.tsx
emcelroy: @kswenson Thank 👍 As discussed elsewhere, I made the change you suggested and created [a new story about addressing the top split issue](https://www.pivotaltracker.com/story/show/186864009).

v3/src/components/graph/adornments/adornments.tsx
```diff
@@ -1,6 +1,7 @@
-import React from "react"
+import React, { useEffect, useRef } from "react"
```
kswenson, line 1:
```suggestion
import React, { useEffect } from "react"
```

kswenson (approved):
👍 LGTM -- just one lint warning to address.

PR #1083, opened 2024-01-17
fix: LSRL does not retain equation location (PT-186834980)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/186834980

Previously, `updateLines` wasn't setting a value for `equationCoords` when creating/updating lines. That resulted in moved equation boxes reverting to the default position when their associated lines were updated.

Model actions that can modify lines now check for a valid `equationCoord` value and pass it to `updateLines` so the information isn't lost.

### DIFF
--- a/v3/src/components/graph/adornments/lsrl/lsrl-adornment-model.ts
+++ b/v3/src/components/graph/adornments/lsrl/lsrl-adornment-model.ts
@@ -8,7 +8,7 @@ import { ScaleNumericBaseType } from "../../../axis/axis-types"
 import { ILineDescription } from "../shared-adornment-types"
 import { isFiniteNumber } from "../../../../utilities/math-utils"
 
-interface ILSRLLine {
+interface ILSRLine {
   category?: string
   equationCoords?: Point
   intercept?: number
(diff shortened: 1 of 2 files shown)


v3/src/components/graph/adornments/lsrl/lsrl-adornment-model.ts
```diff
@@ -131,20 +140,19 @@ export const LSRLAdornmentModel = AdornmentModel
...
-    line: {category?: string, intercept?: number, rSquared?: number, slope?: number, sdResiduals?: number},
-    key="", index?: number
-  ) {
+  updateLines(line: ILSRLLine, key="", index?: number) {
+    const { category, equationCoords, intercept, rSquared, sdResiduals, slope } = line
     const existingLines = self.lines.get(key)
     const newLines = existingLines ? [...existingLines] : []
     // Remove any pre-existing line in newLines at specified index, otherwise we can end up with duplicates.
-    index && newLines.splice(index, 1)
+    isFiniteNumber(index) && newLines.splice(index, 1)
```
emcelroy, line 148: This line wasn't working when `index` was `0`. Fixing that meant the former line 168 `self.lines.clear()` was no longer needed. I believe I must've originally added that because this line (148) wasn't doing what I thought it was but was "failing" silently.

v3/src/components/graph/adornments/lsrl/lsrl-adornment-model.ts
```diff
@@ -8,6 +8,15 @@ import { ScaleNumericBaseType } from "../../../axis/axis-types"
 import { ILineDescription } from "../shared-adornment-types"
 import { isFiniteNumber } from "../../../../utilities/math-utils"
 
+interface ILSRLLine {
```
kswenson, line 11: A nit, but seems like this should be `ILSRLine`.

v3/src/components/graph/adornments/lsrl/lsrl-adornment-model.ts
kswenson: Since it's not really an issue of floating-point finiteness, `;(index != null) && newLines.splice(index, 1)` might be clearer.

kswenson (approved):
👍 LGTM -- just a couple of small suggestions.

PR #1085, opened 2024-01-18
feat: Added dst sample csv [PT-186630112]

dougmartin:
This is the NOAA Weather sample data exported from this shared CODAP V2 file:

https://codap.concord.org/app/static/dg/en/cert/index.html#shared=https%3A%2F%2Fcfm-shared.concord.org%2FRnkig4wxGKOn6z48oZNN%2Ffile.json

This can be loaded in V3 using the `?sample=dst` query parameter.


dougmartin:
Closing without merging as this adds too much data to the bundle as it is a large csv

PR #1086, opened 2024-01-18
chore: Memoize cellKey prop for all adornments (PT-186866222)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/186866222


kswenson (approved):
👍 LGTM

PR #1087, opened 2024-01-18
fix: Separate counts for each region defined by movable values (PT-185940746)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/185940746


kswenson (approved):
👍 LGTM

PR #1088, opened 2024-01-19
chore: improve adornment types

kswenson:
In reviewing a bunch of adornment-related PRs from @emcelroy recently, I noticed a couple of different areas where the types could be improved:
1. Each adornment component was declaring its own props even though the same props are passed to all adornment components.
2. On a related note, in `IAdornmentComponentInfo`, the `Component`, `Controls`, and `BannerComponent` properties were all declared as `React.ComponentType<any>` with a `TODO` comment indicating that it should be fixed at some point.
3. This PR defines `IAdornmentComponentProps`, `IAdornmentControlsProps`, and `IAdornmentBannerComponentProps` interfaces which are then used by any implementing components that need them.
4. The `Adornment` component was casting its axes to be `NumericAxisModel`s even though not all axes are numeric. With this PR, non-numeric axes are passed as `undefined` which eliminates the potential of inadvertently accessing a categorical axis as though it were numeric.
5.  A number of different adornments were using the same inelegant code for accessing axis domains within MobX autoruns/reactions so that they would trigger when the axis domains changed. This PR adds a `getAxisDomains()` utility function which eliminates the redundant/inelegant code.
6. The `GraphMeasurePanel` contained an unnecessary cast which resulted in a loss of type information.
7. The process of standardizing the use of `IAdornmentComponentProps` (#s 1-3 above) led to the realization that the `MovableValueAdornmentComponent` claimed to be receiving (and making use of) a `transform` property that is never passed to it. This phantom property has been eliminated without any apparent ill effect.

### DIFF
--- a/v3/src/components/graph/adornments/adornment-component-info.ts
+++ b/v3/src/components/graph/adornments/adornment-component-info.ts
@@ -1,9 +1,30 @@
 import React from "react"
+import { AdornmentModel, IAdornmentModel } from "./adornment-models"
+import { INumericAxisModel } from "../../axis/models/axis-model"
+
+export interface IAdornmentComponentProps {
+  cellKey: Record<string, string>
+  containerId: string
+  model: IAdornmentModel
+  plotHeight: number
+  plotWidth: number
+  xAxis?: INumericAxisModel
+  yAxis?: INumericAxisModel
+}
+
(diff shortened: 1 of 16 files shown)


emcelroy (approved):
Looks great -- thank you!

PR #1093, opened 2024-01-22
feat: Adornments should tell GraphLayout how much space is needed for adornment banners (PT-186864009)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/186864009

### DIFF
--- a/v3/src/components/axis-graph-shared.ts
+++ b/v3/src/components/axis-graph-shared.ts
@@ -3,6 +3,8 @@ import {AxisPlaces} from "./axis/axis-types"
 export const GraphPlaces = [...AxisPlaces, "yPlus", "plot", "legend"] as const
 export type GraphPlace = typeof GraphPlaces[number]
 
+export type GraphExtentsPlace = GraphPlace | "banners"
+
 export function isVertical(place: GraphPlace) {
   return ["left", "rightCat", "rightNumeric", "yPlus"].includes(place)
 }
--- a/v3/src/components/axis-graph-shared.ts
+++ b/v3/src/components/axis-graph-shared.ts
@@ -3,8 +3,6 @@ import {AxisPlaces} from "./axis/axis-types"
 export const GraphPlaces = [...AxisPlaces, "yPlus", "plot", "legend"] as const
 export type GraphPlace = typeof GraphPlaces[number]
 
-export type GraphExtentsPlace = GraphPlace | "banners"
-
 export function isVertical(place: GraphPlace) {
   return ["left", "rightCat", "rightNumeric", "yPlus"].includes(place)
 }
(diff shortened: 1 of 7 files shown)


v3/src/components/axis-graph-shared.ts
```diff
@@ -3,6 +3,8 @@ import {AxisPlaces} from "./axis/axis-types"
 export const GraphPlaces = [...AxisPlaces, "yPlus", "plot", "legend"] as const
 export type GraphPlace = typeof GraphPlaces[number]
 
+export type GraphExtentsPlace = GraphPlace | "banners"
```
emcelroy, line 6: Not sure this is the perfect spot for this, but I couldn't think of a more appropriate location.

v3/src/components/graph/models/graph-layout.ts
```diff
@@ -124,27 +124,74 @@ export class GraphLayout extends DataDisplayLayout implements IAxisLayout {
...
-        left: {left: 0, top: topAxisHeight, width: leftAxisWidth, height: plotHeight},
-        top: {left: leftAxisWidth, top: 0, width: tileWidth - leftAxisWidth - rightAxisWidth, height: topAxisHeight},
-        plot: {left: leftAxisWidth, top: topAxisHeight, width: plotWidth, height: plotHeight},
-        bottom: {left: leftAxisWidth, top: topAxisHeight + plotHeight, width: plotWidth, height: bottomAxisHeight},
-        legend: {left: 6, top: tileHeight - legendHeight, width: tileWidth - 6, height: legendHeight},
-        rightNumeric: {left: leftAxisWidth + plotWidth, top: topAxisHeight, width: v2AxisWidth, height: plotHeight},
-        rightCat: {left: leftAxisWidth + plotWidth, top: topAxisHeight, width: rightAxisWidth, height: plotHeight},
-        yPlus: {left: 0, top: topAxisHeight, width: leftAxisWidth, height: plotHeight} // This value is not used
+      plotHeight = tileHeight - bannersHeight - topAxisHeight - bottomAxisHeight - legendHeight,
+      newBounds: Record<GraphExtentsPlace, Bounds> = {
```
emcelroy, line 135: With the introduction of `bannersHeight` some lines in the definition of `newBounds` became too long according to the linting rules. Breaking up just those particular lines into two lines didn't help with legibility, in my opinion. Ultimately, it seemed to me that the below made the various calculations easier to grok. If there's a good reason we should try to minimize the line breaks, though, I can bring it back more inline with the previous state.

v3/src/components/axis-graph-shared.ts
kswenson: My inclination would be to define it in `data-display-layout.ts`. It has no meaning outside of the layout context. As a separate matter, I noticed in the code below that we return a value for `yPlus` along with a comment indicating that it is unused. It should be possible to do this instead:
```typescript
export type GraphExtentsPlace = Exclude<GraphPlace | "banners", "yPlus">
```
where `Exclude` comes from the `type-fest` library, and then the extraneous property wouldn't be necessary.

kswenson (approved):
👍 Looks good -- made a couple of suggestions

PR #1094, opened 2024-01-22
fix: axis repetitions should update when data values change

kswenson:
PT: [[#186781301]](https://www.pivotaltracker.com/story/show/186781301)

The main change that fixes the bug here is referencing the `casesChangeCount` inside the MobX `reaction` installed by `useInitGraphLayout()`. Along the way, I also turned `DataConfigurationModel.categoryArrayForAttrRole()` into a self-caching function, which may not have been necessary to fix the original bug but seemed like a reasonable idea nonetheless.


emcelroy (approved):
Looks good and works as expected in the browser -- thanks!

PR #1096, opened 2024-01-23
fix: show percent options with split plots (PT-186789144)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/186789144

These changes ensure that the Percent adornment option in the Graph's Measures menu displays the Row, Column, and Cell options when the plot is split. They also update the associated `GraphDataConfigurationModel` views accordingly to determine the correct values.


kswenson (approved):
👍 LGTM

PR #1098, opened 2024-01-24
fix: Counts wrong when movable values and top split present (PT-186874564)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/186874564

Previously, the `regionCounts` view on `CountAdornmentModel` was using the full width and height of the plot when calculating the coordinates of the boundaries defining each region created by a Movable Value. When the plot was divided into subplots by a categorical split, that would mean those calculations were wrong, resulting in incorrect count values and placement.

These changes update `regionCounts` to use a copy of the `scale` which we modify to use a range defined by the subplots' width and height. This would be the second time we've done this in the adornment code. The first is [here in the Movable Line component](https://github.com/concord-consortium/codap/blob/230a3f958c764a09a98ee329657a1315eefefac7/v3/src/components/graph/adornments/movable-line/movable-line-adornment-component.tsx#L67-L68). It makes me wonder if there should be a version of `scale` that takes into account the smaller dimensions of subplots? Or maybe there already is one and I'm not seeing it?

### DIFF
--- a/v3/src/components/graph/adornments/count/count-adornment-component.tsx
+++ b/v3/src/components/graph/adornments/count/count-adornment-component.tsx
@@ -74,6 +74,8 @@ export const CountAdornment = observer(function CountAdornment(props: IAdornment
     const regionCountParams: IRegionCountParams = {
       cellKey,
       dataConfig,
+      plotHeight,
+      plotWidth,
       scale,
       subPlotRegionBoundaries: subPlotRegionBoundariesRef.current,
     }
@@ -93,7 +95,8 @@ export const CountAdornment = observer(function CountAdornment(props: IAdornment
         })}
       </>
     )
-  }, [casesInPlot, cellKey, dataConfig, graphModel.plotType, model, primaryAttrRole, scale, textContent])
+  }, [casesInPlot, cellKey, dataConfig, graphModel.plotType, model, plotHeight, plotWidth, primaryAttrRole,
+      scale, textContent])
 
   useEffect(function resizeTextOnCellWidthChange() {
     return mstAutorun(() => {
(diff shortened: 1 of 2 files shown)


kswenson:
> These changes update `regionCounts` to use a copy of the `scale` which we modify to use a range defined by the subplots' width and height. This would be the second time we've done this in the adornment code. The first is [here in the Movable Line component](https://github.com/concord-consortium/codap/blob/230a3f958c764a09a98ee329657a1315eefefac7/v3/src/components/graph/adornments/movable-line/movable-line-adornment-component.tsx#L67-L68). It makes me wonder if there should be a version of `scale` that takes into account the smaller dimensions of subplots? Or maybe there already is one and I'm not seeing it?

I agree that there's a problem to be addressed here. Another symptom is that many of the adornments have `MobX` `autorun`s or `reaction`s for responding to changes, but because subplot width and height are not observable, they end up as `useEffect` dependencies which require destroying and reinstalling the reactions. I suspect the solution involves making the `GraphLayout` responsible for subplot extents and scales the same way it's currently responsible for overall plot extents and scales. Then the subplot extents would be observable and clients could access the subplot scales through the layout as well. That should probably be a separate story rather than holding up this one, however.

kswenson (approved):
👍 Looks good -- with the proviso that there's a need for a larger refactor here that is described in a separate comment.

emcelroy:
> adornments have `MobX` `autorun`s or `reaction`s for responding to changes, but because subplot width and height are not observable, they end up as `useEffect` dependencies which require destroying and reinstalling the reactions.



> 👍 Looks good -- with the proviso that there's a need for a larger refactor here that is described in a separate comment.

I started a [new story for that issue](https://www.pivotaltracker.com/story/show/186908446).

PR #1100, opened 2024-01-25
Add toolbar cypress tests

tejal-shah:


v3/cypress/support/elements/toolbar-elements.ts
```diff
@@ -0,0 +1,41 @@
...
+        el = "[data-testid=tool-shelf-button-Slider]"
+        break
+      case "calculator":
+        el = "[data-testid=tool-shelf-button-Calc]"
+        break
+      case "table":
+        el = "[data-testid=tool-shelf-button-table]"
+        break
+      case "map":
+        el = "[data-testid=tool-shelf-button-Map]"
```
emcelroy, line 18: Not super important, but should we make these IDs consistent? All of them capitalize the component name part except`tool-shelf-button-table`. If it were up to me, I'd make all of them all lowercase, but maybe there's a good reason some use uppercase for the component name?

emcelroy (changes requested):
This is looking good 👍 There are what I'm pretty sure are unwanted semicolons in v3/cypress/e2e/toolbar.spec.ts, though. I also left a comment about possibly making some data-testid values more consistent.

tejal-shah:
@emcelroy , I've removed the unwanted semicolons and have added the data-testids in the code which makes selecting those elements much simpler now. Thank you for suggesting that. Additionally, those data-testids were also being used in the jest tests which I fixed as well.

emcelroy (approved):
Looks good 👍 Sorry I didn't notice this before, but the `data-testid` value referenced on line 21 of v3/cypress/support/elements/toolbar-elements.ts is`Delete Data Set-button`. If we want to be super consistent, that would be better as `delete-data-set-button` or at least `Delete-Data-Set-button`. 

I don't think it's absolutely necessary to change that if it'd mean having to update the code in an awkward way just for the `data-testid` value. I believe technically spaces are allowed in `data-testid` values, right?

tejal-shah:
@emcelroy , I've now modified this data-testid to `delete-data-set-button-delete`. I had to add the extra `delete` at the end because there are two of these buttons - one to confirm delete and the other one to cancel and the trailing `delete` defines the classname that distinguishes them.

emcelroy:
> @emcelroy , I've now modified this data-testid to `delete-data-set-button-delete`. I had to add the extra `delete` at the end because there are two of these buttons - one to confirm delete and the other one to cancel and the trailing `delete` defines the classname that distinguishes them.

Thank you 👍

PR #1108, opened 2024-02-01
feat: load configuration from environment

kswenson:
- changes how google drive credentials are handled
- configures GitHub actions to use production google drive credentials
- updates many GitHub actions to their latest version
- tweaks the webpack configuration so that more asset files get appropriate names and end up in the right folders


emcelroy (approved):
Looks good 👍

PR #1109, opened 2024-02-02
fix: Case dot size should animate smoothly when Connecting Lines are activated/deactivated (PT-186762653)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/186762653

The first fix for this issue was made obsolete by the switch to PixiJS.

In addition to enabling the smooth animation of case dots, this PR includes changes to the `Scatterdots` component that fix a bug preventing the Point Size control from working correctly when Connecting Lines were present. The fix is to ensure the case dot size change occurs only once when the Connecting Lines are first activated.


v3/src/components/graph/utilities/pixi-points.ts
```diff
@@ -454,10 +457,17 @@ export class PixiPoints {
...
+        metadata.style = newStyle
       }
     }
 
+    // If we're animating the change, we need to transition the point scale.
+    animateChange && this.points.forEach((point) => {
+      this.transition(() => {
+        this.setPointScale(point, style.radius / currentRadius)
+      }, { duration: transitionDuration })
+    })
```
pjanik, line 469: Would it be possible to call these lines outside of this function? It seems strange to me that `animateChange` triggers the animation of only one specific style property (point radius). Seeing a call to `matchPointsToData` with the `animateChange` parameter set to true, I would expect that all updated style properties would be animated.

v3/src/components/graph/utilities/pixi-points.ts
```diff
@@ -411,8 +411,11 @@ export class PixiPoints {
     })
   }
 
-  matchPointsToData(caseData: CaseData[], style: IPixiPointStyle) {
-    const texture = this.getPointTexture(style)
+  matchPointsToData(caseData: CaseData[], style: IPixiPointStyle, animateChange = false) {
+    // If change should be animated, we will modify the point radius with a transition after modifying everything else.
+    const currentRadius = this.points[0] ? this.getMetadata(this.points[0]).style.radius : 6
```
pjanik, line 416: This is probably a theoretical issue, but is the first point always the source of truth for the point radius? In theory, a user can hover over it, which affects the radius (but I'm not sure if it affects this property or the scale?).

Maybe extracting that to a separate method like `getCurrentPointRadius` together with a short comment could be a bit safer long-term.

v3/src/components/graph/components/scatterdots.tsx
```diff
@@ -292,9 +285,20 @@ export const ScatterDots = observer(function ScatterDots(props: PlotProps) {
...
+    } else if (!showConnectingLines) {
+      const scaleFactor = origPointSizeMultiplier.current / pointSizeMultiplier
+      await pixiPointsRef.current?.setAllPointsScale(scaleFactor, transitionDuration)
+      graphModel.pointDescription.setPointSizeMultiplier(origPointSizeMultiplier.current)
+    }
   }, [dataConfiguration, dataTip, dataset?.collections, graphModel.pointDescription,
-    handleConnectingLinesClick, handleConnectingLinesMouseOut, handleConnectingLinesMouseOver, layout,
-    showConnectingLines])
+      handleConnectingLinesClick, handleConnectingLinesMouseOut, handleConnectingLinesMouseOver, layout,
+      showConnectingLines, pointSizeMultiplier, connectingLinesActivatedRef, pixiPointsRef])
```
kswenson, line 301:
```suggestion
      handleConnectingLinesClick, handleConnectingLinesMouseOut, handleConnectingLinesMouseOver,
      layout, pixiPointsRef, pointSizeMultiplier, showConnectingLines])
```
Don't need `connectingLinesActivatedRef`; others may as well be alphabetized.

kswenson (approved):
👍 Looks good!

PR #1114, opened 2024-02-07
fix: CFM style tweaks; fix font configuration

kswenson:
- resolves some CFM styling issues [[PT-186987479]](https://www.pivotaltracker.com/story/show/186987479)
- fixes the font configuration in CODAP v3 globally

### DIFF
--- a/v3/src/styles/menu-style.ts
+++ b/v3/src/styles/menu-style.ts
@@ -6,7 +6,7 @@ const helpers = createMultiStyleConfigHelpers(["menu", "list", "item"])
 export const Menu = helpers.defineMultiStyleConfig({
   baseStyle: {
     menu: {
-      font: `"Montserrat-Regular", sans-serif`,
+      font: `"Montserrat", sans-serif`,
       boxShadow: "md",
       rounded: "md",
       flexDirection: "column",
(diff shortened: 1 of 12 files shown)


v3/src/styles/menu-style.ts
```diff
@@ -6,7 +6,7 @@ const helpers = createMultiStyleConfigHelpers(["menu", "list", "item"])
 export const Menu = helpers.defineMultiStyleConfig({
   baseStyle: {
     menu: {
-      font: `"Montserrat-Regular", sans-serif`,
+      font: `"Montserrat", sans-serif`,
```
emcelroy, line 9: Super nitpicky, sorry, but it'd be a little nicer if this was changed to...

`font: "Montserrat, sans-serif",`

Similarly, maybe for consistency remove all the `"` and `'` instances around `Montserrat` in the SCSS files, too, since they're no longer needed?

None of that's required, of course, if you don't think it's worth bothering over. The fonts are all being applied appropriately as is.

emcelroy (approved):
Looks great! I left one non-blocking suggestion.

v3/src/styles/menu-style.ts
kswenson: Good point -- done.

PR #1115, opened 2024-02-08
Hierarchical ruler menu for univariate plots

bfinzer:
[#186948611] Bug fix: The ruler menu for univariate plots is hierarchical
Note that there is a 'todo' in movable-value-adornment-registration.tsx and a PT story (#187005556) relating to needed improvements.
* In adornment-types.ts change the definition of the "dotPlot" element of the `measures` array to define the hierarchy of options.
* Break `AdornmentsStore` into `AdornmentsBaseStore` containing state information and `AdornmentsStore` containing UI information
  * Ruler state visibility is stored, for now, in the `rulerState` property of the `AdornmentStore`. This will probably move to document-wide storage soon.
* ~~Installed `react-icons` to get access to disclosure triangle icons to use with chakra button~~
* Defined `GraphMeasureGroup` component to display disclosure button and the items within it
* Changed movable-value-adornment-registration.tsx so that instead of a MenuButton, we have two buttons, one for adding and one for removing movable values


v3/src/components/graph/adornments/adornments-store-utils.ts
```diff
@@ -0,0 +1,154 @@
...
+        disabled,
+        title: "DG.Inspector.graphInterceptLocked",
+        type: "control",
+        clickHandler: theStore.toggleInterceptLocked
+      })
+    }
+    return true // The Connecting Lines checkbox was either already added or just added
+  }
+
+  // The sum of squares of residuals is only shown for scatter plots, and only when the LSRL or Movable Line is
```
emcelroy, line 113: Is this comment unfinished? Seems like it's maybe at least missing something like "present" at the end?

v3/src/components/graph/adornments/movable-value/movable-value-adornment-registration.tsx
```diff
@@ -35,25 +36,26 @@ const Controls = () => {
...
     )
   }
-
+  /**
+   * Todo: The two buttons are problematic wrt style.
+   * 1. The className is not being applied to the buttons.
+   * 2. The font-weight defined in buttonStyle is not being applied to the buttons.
+   * 3. There should be a space between the two buttons.
+   * Note that moving this code to the GraphMeasureGroup component doesn't solve the problem.
+   */
```
emcelroy, line 45: Does this comment need to be updated since you fixed that style selector issue? I think at least #1 is fixed, and that note at the end doesn't matter anymore?

v3/src/components/graph/components/inspector-panel/measure-panel.scss
```diff
@@ -0,0 +1,22 @@
+.measure-group-button {
+  justify-content: left!important;
```
emcelroy, line 2: Are all the instances of `!important` in this file necessary? Or are they just remnants of trying to troubleshoot the styling issue that has since been fixed?

emcelroy (changes requested):
This is looking good to me.  I left a few comments/questions about some small things to possibly clean up. There are also some failing Cypress tests in adornments.spec.ts. I did a quick check and it seems like the UI is working fine, it’s just that the tests need to be updated to account for the fact that the checkboxes aren’t always immediately accessible.

v3/src/components/graph/components/inspector-panel/graph-measure-group.scss
kswenson, line 1: I agree with @emcelroy's comment, which may have gotten lost in a rebase -- These `!important`s should not be necessary. Remove them if they can be removed. If not, we should figure out why and fix it properly.

kswenson (changes requested):
Looks pretty good at this point. I like the new look. I added a commit which fixed some issues:
- removed extraneous commits from the PR with a rebase
- removed import of `react-icons` in favor of downloading the two icon files we need directly
- rearranged some of the type definitions and eliminated some redundant type definitions
- removed some of the gratuitous white-space changes (we need to figure out how to turn those off in your IDE)
- rearranged some of the `.scss` files and their imports

There are still some remaining issues pointed out by @emcelroy and myself in separate comments remaining for you to address, including:
- use of `!important` where it shouldn't be necessary
- some TypeScript type improvements

v3/src/components/graph/adornments/adornments-store-utils.ts
```diff
@@ -0,0 +1,154 @@
...
+}
+
+export function isMeasureMenuItem(item: IMeasureMenuItem | IGroupItem): item is IMeasureMenuItem {
+  return item.type !== 'Group'
+}
+
+export interface IGroupItem {
+  title: string
+  type: string
+  rulerStateKey: RulerStateKey | undefined
```
kswenson, line 26:
```suggestion
  rulerStateKey?: RulerStateKey
```
would be more conventional.

v3/src/components/graph/components/inspector-panel/graph-measure-group.scss
```diff
@@ -0,0 +1,22 @@
+.measure-group-button {
+  justify-content: left!important;
+  font-size: 12px!important;
+  font-weight: 400!important;
+  background-color: #eeeeee!important;
+  height: 24px!important;
+  padding-inline-start: 0!important;
+  border: none!important;
```
kswenson, line 8: What about these `!important`s?

kswenson (approved):
👍 Looks good -- just a couple of remaining suggestions.

v3/src/components/graph/adornments/adornments-store-utils.ts
emcelroy: I hate to be so picky but I still find this comment confusing as written. Either there's a word or two missing at the end or it's assumed that it'll be understood that "shown" in the first clause applies to "is" in the second. The latter is not obvious in my opinion. If nobody else feels that way, though, this is the last time I'll mention it.

emcelroy (approved):
Looks good 👍 I made a non-blocking, nit-picky comment about a comment. There's also a lint warning in adornments.spec.ts about line 242 being too long that it'd be nice to fix, but also probably not super important.

PR #1117, opened 2024-02-09
feat: save/restore and undo/redo of column widths

kswenson:


v3/src/components/case-table/collection-table.tsx
```diff
@@ -73,6 +79,42 @@ export const CollectionTable = observer(function CollectionTable(props: IProps)
...
+        columnWidths.current = newColumnWidths
+        forceUpdate()
+      },
+      { name: "CollectionTable.updateColumnWidths", fireImmediately: true, equals: comparer.structural },
+      metadata)
+  }, [columns, forceUpdate, metadata])
+
+  // respond to column width changes from RDG
+  const handleColumnResize = useCallback(
+    function handleColumnResize(idx: number, width: number, isComplete?: boolean | undefined) {
```
tealefristoe, line 108: Do you need the `| undefined` on `isComplete` here?

tealefristoe (approved):
Looks good 👍

v3/src/components/case-table/collection-table.tsx
kswenson: Good catch -- not sure how that snuck in there.

PR #1118, opened 2024-02-10
Add Web View Tile to v3

tealefristoe:
This PR adds the web view tile to CODAP v3. Behavior is duplicated from v2.
- The web tile can be added using the options toolbar button or by dragging a url onto the CODAP workspace.
- A button on the tile's toolbar allows the user to change which url is being displayed in the tile. In v2 this action was not undoable, but it is in v3.
- A background message is displayed to warn the user that their website might not have loaded properly.

Important additional change:
- `useDocumentContext` now provides the document model, rather than the document content. The content can still easily be accessed via `useDocumentContext()?.content`. Additionally, I went through and cleaned up parts of the code where `useDocumentContext` could be used instead of prop drilling.


v3/src/components/plugin/plugin.scss
```diff
@@ -0,0 +1,9 @@
+.codap-plugin {
+  width: 100%;
+  height: calc(100% - 25px);
```
bfinzer, line 3: There is a variable for the 25px: $title-bar-height

v3/src/components/tool-shelf/options-button.tsx
```diff
@@ -0,0 +1,63 @@
...
+import t from "../../utilities/translation/translate"
+import { kWebViewTileType } from "../web-view/web-view-defs"
+import { isWebViewModel } from "../web-view/web-view-model"
+import { WebViewUrlModal } from "../web-view/web-view-url-modal"
+
+import "./tool-shelf.scss"
+
+export const OptionsShelfButton = () => {
+  const document = useDocumentContext()
+  const formulaModal = useDisclosure()
```
bfinzer, line 14: I don't understand "formula" in the name. "urlModal"?

v3/src/components/tool-shelf/tool-shelf.tsx
```diff
@@ -76,7 +76,8 @@ export const ToolShelf = observer(function ToolShelf({ document }: IProps) {
     {
       icon: <OptionsIcon className="icon-options"/>,
       label: t("DG.ToolButtonData.optionMenu.title"),
-      hint: t("DG.ToolButtonData.optionMenu.toolTip")
+      hint: t("DG.ToolButtonData.optionMenu.toolTip"),
```
bfinzer, line 79: In V2 the tooltip is "Display a website, configure guide…" But here I'm just getting "Options"

v3/src/components/web-view/web-view-url-modal.tsx
```diff
@@ -0,0 +1,71 @@
...
+import t from "../../utilities/translation/translate"
+
+interface IProps {
+  currentValue?: string
+  isOpen: boolean
+  onAccept: (value: string) => void
+  onClose: () => void
+}
+
+export const WebViewUrlModal = ({ currentValue="", isOpen, onAccept, onClose }: IProps) => {
```
bfinzer, line 15: I notice that pressing <return> doesn't accept value and close the modal. Similarly, <esc> should close without accepting any edits.

bfinzer (approved):
👍🏻LGTM
* Just some minor suggestions.

v3/src/components/tool-shelf/options-button.tsx
tealefristoe: Good catch, that's left over from copying and pasting.

v3/src/components/web-view/web-view-url-modal.tsx
tealefristoe: I was able to address your other three suggestions, but I'm not sure what you mean by this one. Are the buttons of the modal not working for you?

v3/src/components/tool-shelf/tool-shelf.test.tsx
kswenson, line 1: With suggestions elsewhere, I think this can go back to the way it was.

v3/src/components/web-view/web-view-inspector.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+  const handleSetWebViewUrlOpen = () => {
+    webViewModal.onOpen()
+    handleModalOpen(true)
+  }
+
+  const handleSetWebViewUrlClose = () => {
+    webViewModal.onClose()
+    handleModalOpen(false)
+  }
```
kswenson, line 33: I think you only need one of these mechanisms -- either `useDisclosure/onOpen/onClose` or the `isModalOpen/isModalClosed` state.

v3/src/components/web-view/web-view-registration.ts
```diff
@@ -0,0 +1,26 @@
+import { registerTileComponentInfo } from "../../models/tiles/tile-component-info"
+import { registerTileContentInfo } from "../../models/tiles/tile-content-info"
+import { kWebViewTileType } from "./web-view-defs"
+import { WebViewModel } from "./web-view-model"
+import { WebViewComponent } from "./web-view"
+import { WebViewInspector } from "./web-view-inspector"
+import { WebViewTitleBar } from "./web-view-title-bar"
+
+export const kWebViewIdPrefix = "WEBVIEW"
```
kswenson, line 9: By convention we generally keep these to four characters, so I'd probably go with:
```suggestion
export const kWebViewIdPrefix = "WEBV"
```

v3/src/components/web-view/web-view-registration.ts
```diff
@@ -0,0 +1,26 @@
...
+})
+
+registerTileComponentInfo({
+  type: "CodapWebView",
+  TitleBar: WebViewTitleBar,
+  Component: WebViewComponent,
+  InspectorPanel: WebViewInspector,
+  tileEltClass: "codap-web-view",
+  defaultWidth: 300,
+  defaultHeight: 300
```
kswenson, line 25: Should check to see what v2 uses as the default size for web views.

v3/src/components/web-view/web-view-url-modal.tsx
kswenson: Also there's some styling niceties like indicating the OK button is the default and making it disabled when nothing has been entered but enabled when something has been entered that would be nice to observe. That said, I'd also be okay with putting off styling tweaks for a subsequent PR if you'd rather get the basic functionality PR wrapped up first.

v3/src/components/web-view/web-view.scss
```diff
@@ -0,0 +1,44 @@
...
+  position: relative;
+  width: 100%;
+
+  .codap-web-view-backdrop {
+    align-items: center;
+    display: flex;
+    flex-direction: column;
+    height: 100%;
+    position: absolute;
+    transform: rotate(-30deg) translateY(35%);
```
kswenson, line 16: This confused me initially -- perhaps a comment about it being used for the loading error message.

kswenson (changes requested):
👍 Generally looks good -- a smattering of minor suggestions, but the larger issue that we should resolve before merging this is the question raised on slack of whether we need separate component types for plugins and web views or whether they can share the same underlying type. I also suggested reverting `useDocumentContext` in the interest of dependency management. We can discuss offline if that requires further explication.

v3/cypress/e2e/toolbar.spec.ts
```diff
@@ -39,4 +40,22 @@ context("codap toolbar", () => {
...
+      toolbar.getWebViewButton().click()
+      WebViewTileElements.getUrlModal().should("exist")
+      WebViewTileElements.enterUrl(url)
+      cy.wait(8000)
+      WebViewTileElements.getUrlModal().should("not.exist")
+      WebViewTileElements.getIFrame().find(`.central-textlogo`).should("be.visible")
+      WebViewTileElements.getEditUrlButton().click()
+      WebViewTileElements.getUrlModal().should("exist")
+      WebViewTileElements.enterUrl(url2)
+      cy.wait(8000)
```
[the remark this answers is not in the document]
tealefristoe: I guess I was being overly conservative there--just tested with 1 second wait times and it worked locally.

v3/src/hooks/use-document-context.ts
```diff
@@ -1,7 +1,7 @@
 import { createContext, useContext } from "react"
-import { IDocumentContentModel } from "../models/document/document-content"
+import { IDocumentModel } from "../models/document/document"
 
-export const DocumentContext = createContext<IDocumentContentModel | undefined>(undefined)
+export const DocumentContext = createContext<IDocumentModel | undefined>(undefined)
```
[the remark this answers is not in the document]
tealefristoe: In order to make updating the web view's url undoable, the `IDocumentModel` must be passed into `OptionsShelfButton`, so there are now two places where we need to prop drill it.

I think it's quite confusing that `useDocumentContext` provides the content rather than the document itself. This leads to situations where variables called `document` sometimes refer to `IDocumentModel` and sometimes to `IDocumentContentModel`. What do you think about changing `useDocumentContext` to `useDocumentContent` and naming variables that contain the content `documentContent`?

v3/src/utilities/translation/lang/en-US.json5
```diff
@@ -18,6 +18,12 @@
     "V3.Inspector.rescale.noRescale.toolTip": "All the data are already visible",
     "V3.Inspector.rescale.casePlot.toolTip": "Move the points to new random positions",
 
+    // V3 web view strings that are not present in V2 and will eventually require translation.
+    "V3.WebView.Modal.PlaceholderText": "URL",
+    "V3.WebView.Modal.okBtnTitle": "OK",
+    "V3.Undo.webView.changeUrl": "Undo changing webView's url",
+    "V3.Redo.webView.changeUrl": "Redo changing webView's url",
```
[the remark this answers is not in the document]
tealefristoe: I based these on the messages for adding the web view, which were already in the file. Should I update these as well?

```
"DG.Undo.webView.show": "Undo adding webView component",
"DG.Redo.webView.show": "Redo adding webView component",
```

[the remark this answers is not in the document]
kswenson: Sigh -- looking at the documentation (or even just the UI), the correct text would seem to be `Undo/redo changing web page url`. We haven't really settled on how to handle potential changes to strings from v2. How about adding new `V3` strings for `Undo/Redo adding web page component` and we'll deal with the question of when/how to sync the v2 strings another day. 🤷

v3/src/components/web-view/web-view-url-modal.tsx
```diff
@@ -0,0 +1,94 @@
...
+      </ModalHeader>
+      <ModalBody>
+        <FormControl display="flex" flexDirection="column" data-testid="web-view-url-form">
+          <Input
+            data-testid="web-view-url-input"
+            onChange={handleValueChange}
+            onKeyDown={e => e.stopPropagation()}
+            onKeyUp={handleKeyUp}
+            placeholder={t("V3.WebView.Modal.PlaceholderText")}
+            ref={r => inputRef.current = r}
```
kswenson, line 75: Do you need the functional form here? Seems like the simpler form should work (unless Chakra's forwarding of the ref doesn't work as expected):
```suggestion
            ref={inputRef}
```

kswenson (approved):
👍 Looks good -- just a couple of remaining suggested tweaks. **BUT**, before you merge, let's have a conversation about whether we need both web view and plugin components at today's planning meeting. I'll frame the discussion in a followup comment on this PR.

PR #1120, opened 2024-02-12
added lat/long data to roller-coasters.csv in ./sample-data

nstclair-cc:


kswenson (approved):
👍 Looks good, although I'm a little surprised we dropped the `state` attribute. 🤔

tejal-shah:
@nstclair-cc , when I try https://codap3.concord.org/branch/add-roller-coasters/?sample=roller-coasters, it doesn't load the roller coasters data, which it should have. There seems to be some issue with this csv file. I'm unsure what it is, but it doesn't load the sample data in it. I'm attaching another roller-coasters.csv file with this comment that seems to work if you want to include this one instead. (It has the `State` attribute but is missing is `latitude` and `longitude` attributes.)
[roller-coasters.csv](https://github.com/concord-consortium/codap/files/14302052/roller-coasters.csv)

tejal-shah (changes requested):
Left a comment.

kswenson:
> > @nstclair-cc , when I try https://codap3.concord.org/branch/add-roller-coasters/?sample=roller-coasters, it doesn't load the roller coasters data, which it should have. There seems to be some issue with this csv file. I'm unsure what it is, but it doesn't load the sample data in it. I'm attaching another roller-coasters.csv file with this comment that seems to work if you want to include this one instead. (It has the `State` attribute but is missing is `latitude` and `longitude` attributes.) [roller-coasters.csv](https://github.com/concord-consortium/codap/files/14302052/roller-coasters.csv)
> 
> Hi @tejal-shah - So I ran into this issue and had an exchange with Kirk, and I think the URL we landed on is https://codap3.concord.org/branch/add-roller-coasters/?sample=coasters. I thought that `?sample=coasters` was the link we wanted to use but can change it to `roller-coasters` if that's the preferred convention - just let me know!
> 
> I also added the attribute `states` in the last commit.

Yes, `sample=coasters` is the correct URL param.

tejal-shah (approved):
LGTM 👍

PR #1121, opened 2024-02-13
Drop attribute on map bug fix

bfinzer:
[#187025508] Bug fix: Dragging an attribute to a map doesn't work properly
* The `CodapMap` component was no longer passing a valid DOM ref to `DroppableMapArea`
* The drop area highlighting wasn't showing above the map, so I had to adjust the z-index (ugh!)


kswenson (approved):
👍 LGTM

PR #1122, opened 2024-02-14
fix: MST warnings/errors on delete DataSet [#187041114]

kswenson:
[[#187041114]](https://www.pivotaltracker.com/story/show/187041114)

The crash was from the history system failing to find a shared model that had been deleted. Beyond that there were still some warnings for accessing deleted models that were fixed by unregistering DataSet-dependent formulas immediately.

A couple of remaining issues are not addressed by this PR:
1. Undo of deleting the DataSet doesn't work correctly.
2. There are no formulas in the document in question, so all of the formulas being registered/unregistered are empty default formulas. Would probably be better to not create formulas for attributes without formulas.

### DIFF
--- a/v3/src/models/formula/formula-manager.ts
+++ b/v3/src/models/formula/formula-manager.ts
@@ -86,6 +86,15 @@ export class FormulaManager {
   }
 
   @action removeDataSet(dataSetId: string) {
+    // unregister all formulas associated with the removed DataSet
+    const formulasToUnregister: string[] = []
+    this.extraMetadata.forEach((extraMetadata, formulaId) => {
+      if (extraMetadata.dataSetId === dataSetId) {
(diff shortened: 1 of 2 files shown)


pjanik (approved):
Looks good to me 👍 

> There are no formulas in the document in question, so all of the formulas being registered/unregistered are empty default formulas. Would probably be better to not create formulas for attributes without formulas.

It makes sense. I think this stems from my very first days in CODAP, where formulas were created by default in a manner similar to the previous placeholder or some other attributes. Back then, I felt it was safer to follow whatever pattern we had in place. 

I feel I gave it a second thought later and still decided to leave it as it was, but I'm not sure why. Maybe it had something to do with observing changes/updates in the formula.

PR #1124, opened 2024-02-14
chore: don't create default empty formula for attributes

kswenson:


tealefristoe (approved):
Looks good 👍

pjanik:
Note that this PR likely caused these warnings:

Screenshot 2024-09-19 at 15 53 25

I believe the system was initially designed with the assumption that there would only be one scenario where a formula is absent - an empty formula. Now, we have two scenarios: an undefined formula or an empty formula (at least in theory, since I think the formula is deleted before it can become empty). This issue can likely be fixed, but given the similar issue with DataSet filter formulas, I wonder what the main reason was for avoiding these empty objects? To prevent them from being included in the serialized JSON?

PR #1125, opened 2024-02-14
feat: User can configure bar chart (PT-186948896)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/186948896


v3/src/components/graph/components/dotplotdots.tsx
```diff
@@ -211,41 +215,100 @@ export const DotPlotDots = observer(function DotPlotDots(props: PlotProps) {
...
+        secondaryAttrID && (key[secondaryAttrID] = category)
+        extraSecondaryAttrID && (key[extraSecondaryAttrID] = extraCategory)
+        extraPrimaryAttrID && (key[extraPrimaryAttrID] = extraPrimaryCategory)
+        return key
+      }
+
+      const computeBarDimension = (anID: string) => {
+        // The returned value is the width or height of the plot (depending the primary axis) divided by the
+        // number of cases in the same subplot as the case whose ID matches anID.
+        const subPlotCaseCount = dataConfig?.subPlotCases(subPlotKey(anID)).length ?? 0
```
emcelroy, line 239: I'm not sure this use of `subPlotCases` is the best thing to do, but if so, maybe the related `subPlotKey` function above should be made a utility function or a view that could be used elsewhere? It seemed like the obvious thing to do to me coming from the adornment work. I think it'd be possible to get the same information, though, without using `subPlotCases` with something like the below which would be more in keeping with the existing code in this component.

```
const subPlotCaseCount = dataConfig?.caseDataArray.filter((c: CaseData) => {
           const cCat = dataset?.getStrValue(c.caseID, secondaryAttrID) ?? '__main__'
           const cExtraCat = dataset?.getStrValue(c.caseID, extraSecondaryAttrID) ?? '__main__'
           const cExtraPrimaryCat = dataset?.getStrValue(c.caseID, extraPrimaryAttrID) ?? '__main__'
           return cCat === category && cExtraCat === extraCategory && cExtraPrimaryCat === extraPrimaryCategory
         }).length ?? 0
```

v3/src/components/graph/components/dotplotdots.tsx
```diff
@@ -211,41 +215,84 @@ export const DotPlotDots = observer(function DotPlotDots(props: PlotProps) {
...
+        const extraPrimaryCoord = extraPrimaryValue ? extraPrimaryAxisScale(extraPrimaryValue ?? '__main__') ?? 0 : 0
+        return { primaryCoord, extraPrimaryCoord }
+      }
+
+      const getSubPlotDetails = (anID: string) => {
+        const subPlotKey = dataset?.subPlotKey(anID, secondaryAttrID, extraSecondaryAttrID, extraPrimaryAttrID) ?? {}
+        const casesInCategory = dataConfig?.subPlotCases(subPlotKey) ?? []
+        const caseIndex = casesInCategory.findIndex((id) => id === anID)
+        return { subPlotKey, casesInCategory, caseIndex }
+      }
```
kswenson, line 230: The id-to-index lookup here is O(2n^2), since `getSubPlotDetails()` is called twice for each case, which could be a potential performance issue.  I played around a bit and the following caches the results on the first call which should improve performance, particularly on large data sets.
```suggestion
      interface ISubPlotDetails {
        cases: string[]
        indices: Record<string, number>
      }
      const subPlotDetails = new Map<string, ISubPlotDetails>()

      const getSubPlotDetails = (anID: string) => {
        const subPlotKey = dataset?.subPlotKey(anID, secondaryAttrID, extraSecondaryAttrID, extraPrimaryAttrID) ?? {}
        const subPlotMapKey = JSON.stringify(subPlotKey)
        let details: ISubPlotDetails | undefined = subPlotDetails.get(subPlotMapKey)
        if (!details) {
          const cases = dataConfig?.subPlotCases(subPlotKey) ?? []
          const indices: Record<string, number> = {}
          cases.forEach((caseId, index) => indices[caseId] = index)
          details = { cases, indices }
          subPlotDetails.set(subPlotMapKey, details)
        }
        return { subPlotKey, casesInCategory: details.cases, caseIndex: details.indices[anID] }
      }
```
This suggestion doesn't include a related point made elsewhere, that the `subPlotKey` method should be on the `GraphDataConfiguration` rather than on the `DataSet`.

v3/src/components/graph/components/inspector-panel/panel.scss
kswenson, line 1: `panel.scss` is too generic. Call it something like `inspector-panel.scss`.

v3/src/components/axis/hooks/use-sub-axis.ts
```diff
@@ -93,9 +95,13 @@ export const useSubAxis = ({
         },
         renderNumericAxis = () => {
           select(subAxisElt).selectAll('*').remove()
-          const numericScale = d3Scale as unknown as ScaleLinear<number, number>,
-            axisScale = axis(numericScale).tickSizeOuter(0).tickFormat(format('.9')),
-            duration = isAnimating() ? transitionDuration : 0
+          const numericScale = d3Scale as unknown as ScaleLinear<number, number>
+          // When displaying bars, set the domain to [0, 100], otherwise use the default.
+          if (graphModel.pointDisplayType === "bars") {
```
kswenson, line 100: It didn't show up in the CodeCov report for some reason, but eslint reports `graphModel.pointDisplayType` as a missing dependency in this callback.

v3/src/components/graph/utilities/pixi-points.ts
```diff
@@ -470,6 +501,14 @@ export class PixiPoints {
...
       const point = this.points[i]
       if (point.texture !== texture) {
+        // The anchor should be set according to the point's shape.
+        // Circle: center (0.5)
+        // Horizontal Bar: bottom left corner (1,0)
+        // Vertical Bar: top left corner (0,0)
+        point.anchor.set(
+          this.displayType === "bars" ? (this.barOrientation === "horizontal" ? 1 : 0) : 0.5,
+          this.displayType === "bars" ? 0 : 0.5
+        )
```
kswenson, line 511: On slack you asked whether there might be a way to avoid passing the `barOrientation` into this class. It's not a huge win, but you could pass the `anchor` in instead of the `barOrientation`. You could define `anchor` constants somewhere, e.g.
```typescript
// The anchor should be set according to the point's shape.
export const circleAnchor: PIXI.IPointData = { x: 0.5, y: 0.5 }
export const hBarAnchor: PIXI.IPointData = { x: 1, y: 0 }
export const vBarAnchor: PIXI.IPointData = { x: 0, y: 0 }
```
and then the client would pass the appropriate `anchor` value and this code would become:
```suggestion
        point.anchor.copyFrom(this.anchor)
```
This might make it easier to extend support to other shapes, for instance.

kswenson (changes requested):
👍 This is looking much better and is much easier to reason about. A bunch of suggestions, but none of them should require too much work.

kswenson:
As a separate matter, there are issues with rendering large numbers of bars (visible with the `abalone` sample data) that I think we can deal with in a subsequent PR.

v3/src/components/data-display/components/data-tip.tsx
```diff
@@ -20,20 +20,17 @@ interface IDataTipHelperProps {
...
   return {
     getBoundingClientRect: () => ({
       left: pointLeft,
       top: pointTop,
       right: pointLeft + sprite.width,
       bottom: pointTop + sprite.height,
-      width: sprite.width,
-      height: sprite.height,
+      width: 1,
+      height: 1,
```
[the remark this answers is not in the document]
emcelroy: Thanks for catching the issue with `right` and `bottom`. I've fixed that and added a comment. My reasoning for changing this was that the current code wasn't working correctly for bars. The tips would appear some distance away from longer bars. I could correct that by conditionally adjusting the positioning for the bar display, but it seemed simpler to just always position the tip in relation to the mouse/pointer. Does that make sense? Or am I overlooking a reason why that won't work well?

v3/src/components/graph/components/dotplotdots.tsx
emcelroy: This works well -- thanks!

v3/src/components/graph/utilities/pixi-points.ts
emcelroy: I like this idea as a small improvement. In practice though I couldn't get it to work. Even though I set and updated an `anchor` property in essentially all the same places I set and updated the `barOrientation` property, `PixiPoints` would not honor the updated anchor value when rendering bars. I'm not sure why. I do want to work that out, but if it's OK with you, I'd prefer to let it go for this PR and figure it out in the next PR for getting the bars to appear in a smooth transition.

emcelroy:
> As a separate matter, there are issues with rendering large numbers of bars (visible with the `abalone` sample data) that I think we can deal with in a subsequent PR.

Noted. I'm pretty sure this has to do with the dimensions of the bar in relation to the stroke size. The latter probably needs to be scaled somehow.

v3/src/components/graph/utilities/pixi-points.ts
```diff
@@ -283,15 +290,30 @@ export class PixiPoints {
...
     }
     const graphics = new PIXI.Graphics()
     graphics.beginFill(fill)
     graphics.lineStyle(strokeWidth, stroke, strokeOpacity ?? 0.4)
-    graphics.drawCircle(0, 0, radius)
+    if (this.displayType === "bars" && width && height) {
+      graphics.drawRect(0, 0, width, height)
+    } else {
+      graphics.drawCircle(0, 0, radius)
+    }
```
kswenson, line 316: Minor point because we presumably don't expect it to happen, but if there is no width or height it seems like it might be better to draw a rectangle by default, so perhaps this should be something like:
```suggestion
    if (this.displayType === "bars") {
      graphics.drawRect(0, 0, width ?? radius, height ?? radius)
    } else {
      graphics.drawCircle(0, 0, radius)
    }
```

v3/src/components/data-display/components/data-tip.tsx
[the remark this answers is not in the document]
kswenson: If the goal is simply to position the data tip relative to the pointer, then I would think it should be:
```suggestion
      right: pointLeft + 1,
      bottom: pointTop + 1,
      width: 1,
      height: 1,
```
If the goal is to make sure it doesn't overlap the sprite then we should incorporate the sprite dimensions into the computation. In any case, the return value should be a valid rectangle in which `right === left + width` and `bottom === top + height`.

kswenson (changes requested):
👍 Changes generally look good -- the virtual element computation still doesn't seem quite right.

v3/src/components/data-display/components/data-tip.tsx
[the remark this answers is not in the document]
emcelroy: Ah, you're right. I somehow got it into my head that the value for `right` should be the distance from the right side of the container to the pointer, and `bottom` should likewise be the distance from the bottom of the container. I double checked how `getBoundingClientRect` should work, and those values should be as you described. I've made that change.

kswenson (approved):
👍 Looks good

PR #1128, opened 2024-02-16
187057621 add sampler translation strings

eireland:
Adds back The Sampler Measures translation strings that got accidentally removed by one of the builds.


bfinzer:
When I go back in time in CODAP releases (e.g. build 0680) I find that the Sampler is translated. For example I switch language to Spanish and the Sampler shows up with translated strings. But I don't see those translations in this pull request; i.e. strings.js in es.lproj are all in English. I'm guessing this is not as it should be?

bfinzer (changes requested):
See comment about translation strings not being present.

eireland:
I don't think those keys have been translated which is why they come up as English. If you scroll up to lines 901-918, you will see the keys that have been translated into Spanish.

bfinzer (approved):
👍🏻LGTM

PR #1129, opened 2024-02-20
187057225 First Contact

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187057204

This PR establishes the first connection between web views in CODAP v3 and plugins. Currently, very little communication is handled, and that's hard coded in the `useDataInteractiveController` hook. In future PRs this will be greatly expanded and written in a more elegant way. In this PR, you should expect:
- CODAP to try to establish a connection with any website loaded in a web view using `iframe-phone`.
- Messages received from plugins to be logged in the console as well as displayed on a toast popup.
- CODAP to update the web view's title when a plugin requests it to do so.

The CODAP API Tester is a great plugin to test. Its url: https://concord-consortium.github.io/codap-data-interactives/DataInteractiveAPITester/index.html?lang=en

Additionally, a `key` was added to the options button in the `ToolShelf` to suppress an error that was showing up in the console from a previous PR.

### DIFF
--- a/v3/src/components/web-view/use-data-interactive-controller.tsx
+++ b/v3/src/components/web-view/use-data-interactive-controller.tsx
@@ -7,10 +7,15 @@ import { isWebViewModel } from "./web-view-model"
 
 function extractOrigin(url?: string) {
   if (!url) return
-  return new URL(url).origin
+  // TODO It would probably be better to confirm that the url is legal before trying to create a URL from it
+  try {
+    return new URL(url).origin
+  } catch (e) {
+    debugLog(DEBUG_PLUGINS, `Could not determine origin from illegal url:`, url)
+  }
 }
 
-export function useDataInteractiveController(iframeRef: React.MutableRefObject<HTMLIFrameElement|null>, tile?: ITileModel) {
+export function useDataInteractiveController(iframeRef: React.RefObject<HTMLIFrameElement>, tile?: ITileModel) {
   const toast = useToast()
   const webViewModel = tile?.content
   const url = isWebViewModel(webViewModel) ? webViewModel.url : undefined
--- a/v3/src/components/web-view/use-data-interactive-controller.tsx
+++ b/v3/src/components/web-view/use-data-interactive-controller.tsx
@@ -1,0 +1,76 @@
+import { useToast } from "@chakra-ui/react"
+import iframePhone from "iframe-phone"
+import React, { useEffect } from "react"
+import { ITileModel } from "../../models/tiles/tile-model"
+import { isWebViewModel } from "./web-view-model"
+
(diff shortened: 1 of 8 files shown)


kswenson (changes requested):
👍 Looks good! Just a handful of suggested tweaks.

v3/src/lib/debug.ts
```diff
@@ -16,6 +16,7 @@ export const DEBUG_MAP = debugContains("map")
 export const DEBUG_SAVE = debugContains("save")
 export const DEBUG_STORES = debugContains("stores")
 export const DEBUG_UNDO = debugContains("undo")
+export const DEBUG_PLUGINS = debugContains("plugins")
```
kswenson, line 19: This list is alphabetized, so move this ⬆️ . (Given your predilection for alphabetizing lists I figured you'd notice that. 😉)

v3/src/components/web-view/use-data-interactive-controller.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { DEBUG_PLUGINS, debugLog } from "../../lib/debug"
+import { ITileModel } from "../../models/tiles/tile-model"
+import { isWebViewModel } from "./web-view-model"
+
+function extractOrigin(url?: string) {
+  if (!url) return
+  return new URL(url).origin
+}
+
+export function useDataInteractiveController(iframeRef: React.MutableRefObject<HTMLIFrameElement|null>, tile?: ITileModel) {
```
kswenson, line 13: I think you can use:
```suggestion
export function useDataInteractiveController(iframeRef: React.RefObject<HTMLIFrameElement>, tile?: ITileModel) {
```
which has the advantage that it makes the line <120. 😉

v3/src/components/web-view/use-data-interactive-controller.tsx
```diff
@@ -0,0 +1,74 @@
+import { useToast } from "@chakra-ui/react"
+import iframePhone from "iframe-phone"
+import React, { useEffect } from "react"
+import { DEBUG_PLUGINS, debugLog } from "../../lib/debug"
+import { ITileModel } from "../../models/tiles/tile-model"
+import { isWebViewModel } from "./web-view-model"
+
+function extractOrigin(url?: string) {
+  if (!url) return
+  return new URL(url).origin
```
kswenson, line 10: I realize this was my suggestion (and I stick by it), but it occurred to me after I made it that it's worth checking what the behavior of `URL` is when given an invalid url string. May need to wrap this in `try`/`catch` or something like that. If the answer is non-obvious then perhaps a TODO would suffice for now.

kswenson (approved):
👍 Looks good -- just a few last suggestions.

PR #1130, opened 2024-02-21
fix: cfm-wrapping for create a copy (update CFM version)

kswenson:
[[#187083507]](https://www.pivotaltracker.com/story/show/187083507)


kswenson:
Merging without review as it's just a dependency version update.

PR #1132, opened 2024-02-21
feat: provide access to v2 `example-documents`

kswenson:
[[#186945881]](https://www.pivotaltracker.com/story/show/186945881)

Currently, `example-documents` are deployed to a folder alongside the v2 application as part of the CODAP v2 build process. Ultimately, there should be a GitHub Actions-based deployment process attached to the [codap-data](https://github.com/concord-consortium/codap-data) repository that is responsible for deploying the latest `example-documents` to s3. In the meantime, I have manually copied the contents of the `example-documents/src` folder in the [codap-data](https://github.com/concord-consortium/codap-data) repository to the `example-documents` folder of the `codap-resources` s3 bucket and updated the local CFM URL to point to the updated example documents.

Note that several of the documents don't load correctly in v3, but that's a topic for another PR.

### DIFF
--- a/v3/cypress/e2e/cfm.spec.ts
+++ b/v3/cypress/e2e/cfm.spec.ts
@@ -1,0 +1,33 @@
+import { CfmElements as cfm } from "../support/elements/cfm"
+
+context("CloudFileManager", () => {
+  beforeEach(function () {
+      const queryParams = "" // "?sample=mammals&dashboard&mouseSensor"
+      const url = `${Cypress.config("index")}${queryParams}`
+      cy.visit(url)
+      cy.wait(2500)
+  })
+  it("Opens Mammals example document via CFM Open dialog", () => {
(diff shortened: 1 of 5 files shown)


v3/cypress/e2e/cfm.spec.ts
```diff
@@ -0,0 +1,33 @@
+import { CfmElements as cfm } from "../support/elements/cfm"
+
+context("CloudFileManager", () => {
+  beforeEach(function () {
+      const queryParams = "" // "?sample=mammals&dashboard&mouseSensor"
```
emcelroy, line 5: You very probably left the commented out query params here for reference since they could potentially be useful to see as an example, but I'm noting them just in case you didn't mean to leave them here.

emcelroy (approved):
Looks good!

PR #1133, opened 2024-02-22
Map base layer control

bfinzer:
[#187098593] Feature: The map's **Layers** inspector menu gives users control over which base layer is showing
* Added MapLayersPalette that comes up when the user clicks the Layers icon on the map inspector panel
* Provide a checkbox for base layer visibility and a segmented control for choosing the desired base layer
* Added `baseMapLayerName` property to `MapContentModel`
* ~~When base layer name or visibility changes, as a kludge, also rescale the map to force the base layer to display in the right lat/long~~


kswenson (approved):
👍 Looks good -- I added a couple of commits which fixes the behavior when changing base layer and cleans up a few other things.

PR #1134, opened 2024-02-23
feat: When user configures numeric univariate plot to display a bar for each point, points should change dimension and shape in a smooth transition (PT-187032200)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/187032200

When the `Bar for Each Point` option is selected for univariate plot graphs, these changes make the switch from points to bars happen in a smooth transition instead of instantly. The transition is achieved in the `PixiPoints` class by first applying a common rectangle texture to all the point sprites that is square in shape but has the same dimensions as the default circular texture, then scaling each point sprite's square to what will be the width and height of its bar, and then creating and applying unique rectangle textures for each point sprite once the transition is complete. The process is reversed when the display type is changed from bars to points, minus the creation of new textures for each point.


kswenson (approved):
👍 Looks good -- comments are very helpful. I can't really evaluate things like whether this is the best way of handling PIXI textures in this case, but otherwise the code seems clear.

PR #1140, opened 2024-02-28
feat: point size control disabled for bars

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/187127970

When points are changed to bars, the "Point size" control is disabled so bars cannot be resized.


kswenson (approved):
👍 LGTM

PR #1145, opened 2024-03-03
chore: improve infrastructure for importing v2 tiles

kswenson:
- component importers build snapshots rather than tiles to better support MST references
- component content `defaultContent()` implementations return snapshots rather than tiles 
- graphs and maps synchronize from data configurations to the shared model manager, i.e. the data configurations are the source of truth (unlike tables for which the shared model manager is the source of truth because tables don't store redundant shared model references).
- flesh out graph component unit tests
- declare some v2 document properties optional, since they're not always present
- switch from using `Record<>` to `Map<>` for some internal v2 document properties for better type safety

@bfinzer Given the size of this PR, I suggest looking it over async first and then we can do a sync review to go over the main points and answer any questions.

### DIFF
--- a/v3/cypress/e2e/graph.spec.ts
+++ b/v3/cypress/e2e/graph.spec.ts
@@ -64,35 +64,35 @@ context("Graph UI", () => {
     // the Collection Name. Blocker: PT #187033159
 
   })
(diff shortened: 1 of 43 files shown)


kswenson:

PR #1147, opened 2024-03-05
feat: remove dev-only url params when opening/closing files

kswenson:
[[PT-187169360](https://www.pivotaltracker.com/story/show/187169360)


tealefristoe (approved):
Looks good 👍

PR #1148, opened 2024-03-05
feat: show color swatches in case table

kswenson:


v3/src/global.d.ts
```diff
@@ -4,6 +4,14 @@ declare module "*.png";
 declare module "*.svg";
 declare module "d3-v6-tip";
 
+declare module "color-parse" {
+  export default function parse(cstr: string): {
+    space: string | undefined
```
tealefristoe, line 9:
```suggestion
    space?: string
```

v3/src/global.d.ts
```diff
@@ -4,6 +4,14 @@ declare module "*.png";
 declare module "*.svg";
 declare module "d3-v6-tip";
 
+declare module "color-parse" {
+  export default function parse(cstr: string): {
```
tealefristoe, line 8: Any way to use the `ParseColorResult` interface defined in `color-parse-strict.ts` for the return type here?

tealefristoe (approved):
Looks good 👍 

I left a couple of minor comments. I also notice that there are some stylistic differences between `color-parse-strict.ts` and the rest of the code base. I'm not sure if these are worth addressing since it's borrowed code, but I'd consider at least replacing `var` with `let`.

tealefristoe:
I also just noticed that the `attribute-types.spec.ts` cypress test is failing.

bfinzer:
Bug: I made a new table, renamed the attribute to `color` and typed "red" into the first cell. The red color swatch appeared. It shouldn't. When I added more cases, that cell value reverted to string "red" as it should be.

kswenson:
>I made a new table, renamed the attribute to `color` and typed "red" into the first cell. The red color swatch appeared. It shouldn't.

That's bizarre -- that was one of my test cases and it was working for me. 🤔 I'll take a look.

v3/src/components/case-table/use-columns.tsx
```diff
@@ -25,6 +27,28 @@
   return formatter
 }
 
+export function renderValue(str = "", num = NaN, attr?: IAttribute) {
+  const type = attr?.type
+
+  // colors
+  if (type === "color" && parseColor(str).space) {
```
bfinzer, line 34: Looks like if the type is not "color" the swatch will never show. But we want "strict" colors to display as swatches always, right?

v3/src/utilities/color-parse-strict.ts
```diff
@@ -0,0 +1,169 @@
...
+/**
+ * Parse color from the string passed
+ *
+ * @return {Object} A space indicator `space`, an array `values` and `alpha`
+ */
+export function parseColorStrict(cstr: string): ParseColorResult {
+	let m: RegExpExecArray | null
+	var parts: Array<string | number> = [], alpha = 1
+	let space = ''
+
```
bfinzer, line 40: Get rid of commented out code below?

bfinzer (changes requested):
I noted a number of bugs. Some (or all?) are case table refresh bugs and could be set aside in PT as you see fit.

v3/src/components/case-table/use-columns.tsx
kswenson: That wasn't my understanding from our previous conversation, but we can discuss at the planning meeting today.

kswenson:
> I also just noticed that the `attribute-types.spec.ts` cypress test is failing.

I fixed the cypress error which was preventing deployment, which may also explain some of the bugs @bfinzer found.

>I made a new table, renamed the attribute to `color` and typed "red" into the first cell. The red color swatch appeared. It shouldn't.

@bfinzer I also fixed one internal invalidation bug that explained some of this, but I think some of the other bugs you found were the result of the previous build not deploying due to the failing cypress test, so you may want to test again.

bfinzer:
I checked out the branch directly. Wouldn’t that mean deployment is not an issue?

bfinzer:
I notice the branch name is missing the `v3` suffix.

bfinzer (approved):
👍🏻LGTM
The bugs I noted are no longer present. Nice!

PR #1150, opened 2024-03-06
fix: bar for each point breaks axis (PT-187142716)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/187142716

Removes the setting of domain to static values in `useSubAxis` when the point display type changes to bars, and changes `FreeDotPlotDots` so it updates the axis to use a domain appropriate to the dataset's value range and the point display type. Includes changes to the `setNiceDomain` function in graph-utils.ts so it computes correct values according to point display type, and makes some adjustments to the `FreeDotPlotDots` component's `computePrimaryCoord` and `getBarPositionInSubPlot` functions.

### DIFF
--- a/v3/cypress/e2e/axis.spec.ts
+++ b/v3/cypress/e2e/axis.spec.ts
@@ -153,23 +153,19 @@ context("Test graph axes with various attribute types", () => {
   })
   it("will adjust axis domain when points are changed to bars", () => {
     // When there are no negative numeric values, such as in the case of Height, the domain for the primary
-    // axis of a univariate plot showing bars should start at zero. So for this test we essentially subtract
-    // the first tick value from arrayOfValues[3].values (i.e. -0.5) to test that the domain starts at zero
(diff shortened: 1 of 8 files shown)


kswenson (approved):
👍 Generally looks good -- I added a commit which refactors things a bit and eliminates some redundancy. See what you think and make sure it still works as expected before merging. 😉

emcelroy:
> 👍 Generally looks good -- I added a commit which refactors things a bit and eliminates some redundancy. See what you think and make sure it still works as expected before merging. 😉

I see what you did there -- thanks! Looks good and everything still works. There was just a tiny bit of additional cleanup possible, so I made one more commit in which I removed `primaryAttribute` from the new `useEffect` because it was no longer needed, and then put `primaryAttrID` back where it had been inside `refreshPointPositions` because it wasn't needed to for helping define `primaryAttribute` in that `useEffect` anymore.

PR #1151, opened 2024-03-06
187176366 Handle Plugin Update Attribute Requests

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187176366

### DIFF
--- a/v3/src/data-interactive/data-interactive-types.ts
+++ b/v3/src/data-interactive/data-interactive-types.ts
@@ -1,9 +1,10 @@
 import { RequireAtLeastOne } from "type-fest"
 import { IAttribute } from "../models/data/attribute"
+import { IAttributeArchive } from "../models/data/attribute-types"
+import { ICollectionModel } from "../models/data/collection"
 import { IDataSet } from "../models/data/data-set"
 import { IGlobalValue } from "../models/global/global-value"
 import { ITileModel } from "../models/tiles/tile-model"
-import { ICollectionModel } from "../models/data/collection"
 
 export type DICase = unknown
 export type DIComponent  = unknown
@@ -31,6 +32,7 @@ export interface DIResources {
 }
 
 export interface DIValues {
+  attrs?: IAttributeArchive[]
   blockDisplayOfEmptyCategories?: boolean
   _categoryMap?: unknown
   cid?: string
--- a/v3/src/data-interactive/handlers/attribute-handler.ts
+++ b/v3/src/data-interactive/handlers/attribute-handler.ts
@@ -1,18 +1,31 @@
-import { DIHandler, DIResources, diNotImplementedYet } from "../data-interactive-types"
+import { DIHandler, DIResources, DIValues, diNotImplementedYet } from "../data-interactive-types"
 import { registerDIHandler } from "../data-interactive-handler"
 
+const attributeNotFoundResult = {success: false, values: {error: 'Attribute not found'}}
 export const diAttributeHandler: DIHandler = {
   get(resources: DIResources) {
     const { attribute } = resources
-    return attribute
-      ? {
-          success: true,
-          values: attribute.toArchive
-        }
-      : {success: false, values: {error: 'Attribute not found'}}
+    if (!attribute) return attributeNotFoundResult
+    return {
+      success: true,
+      values: attribute.toArchive
+    }
   },
   create: diNotImplementedYet,
-  update: diNotImplementedYet,
+  update(resources: DIResources, values?: DIValues) {
+    const { attribute } = resources
+    if (!attribute) return attributeNotFoundResult
+    if (values?.name !== undefined) attribute.setName(values?.name)
+    if (values?.title !== undefined) attribute.setTitle(values?.title)
+    return {
+      success: true,
+      values: {
+        attrs: [
+          attribute.toArchive
+        ]
+      }
+    }
+  },
   delete: diNotImplementedYet
 }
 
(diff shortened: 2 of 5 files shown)


kswenson (changes requested):
As mentioned in the previous PR I think the goal should be to minimize v2 impact on v3 code, so many of these changes should move elsewhere.

tealefristoe:
@kswenson Do you have any ideas on how to handle the actual update action outside of the attribute model? I figure this should happen in a single action and `withoutUndo`.

kswenson:
> @kswenson Do you have any ideas on how to handle the actual update action outside of the attribute model? I figure this should happen in a single action and `withoutUndo`.

It's somewhat counter-intuitive given the naming, but I would think that `applyUndoableAction()` would serve to wrap all of the component actions into a single action and then `withoutUndo()` could be used to disable the undo/redo if desired.

v3/src/data-interactive/handlers/attribute-handler.ts
```diff
@@ -1,42 +1,79 @@
...
+import { DIHandler, DIResources, DIValues, diNotImplementedYet } from "../data-interactive-types"
 
+function convertAttributeToV2(resources: DIResources) {
+  const { attribute, dataContext } = resources
+  const metadata = (dataContext && getSharedCaseMetadataFromDataset(dataContext))
+  if (attribute) {
+    const { name, type, title, description, editable, id, precision } = attribute
+    return {
+      name,
+      type, // TODO This won't return "none", which v2 sometimes does
```
kswenson, line 15: I think this is fine. I looked into this recently and it seems like v2 used `none`/`null` interchangeably, so I think it's safe to eliminate `none` from consideration.

v3/src/data-interactive/handlers/attribute-handler.ts
```diff
@@ -1,42 +1,79 @@
...
+      title,
+      // cid: self.cid, // TODO What should this be?
```
kswenson, line 17: I looked into v2's use of ids recently as well. I think this should be:
```suggestion
      cid: attribute.id,
```
and eventually `id` and `guid` below will be something different, but that doesn't need to hold up this PR.

v3/src/data-interactive/handlers/attribute-handler.ts
```diff
@@ -1,42 +1,79 @@
...
+    attribute.applyUndoableAction(() => {
+      withoutUndo()
+      if (values?.description !== undefined) attribute.setDescription(values.description)
+      if (values?.editable !== undefined) attribute.setEditable(values.editable)
+      if (values?.formula !== undefined) attribute.setDisplayExpression(values.formula)
+      if (values?.name !== undefined) attribute.setName(values.name)
+      if (values?.precision !== undefined) attribute.setPrecision(values.precision)
+      if (values?.title !== undefined) attribute.setTitle(values.title)
+      if (values?.type && isAttributeType(values.type)) attribute.setUserType(values.type)
+      if (values?.unit !== undefined) attribute.setUnits(values.unit)
```
kswenson, line 62: Since these values are coming from outside, I'd be inclined to use the broader `!= null` so that both `null` and `undefined` values are handled.

v3/src/models/data/attribute-types.ts
kswenson, line 1: We already have a very similar type `ICodapV2Attribute` in `codap-v2-types.ts`.

v3/src/data-interactive/data-interactive-types.ts
```diff
@@ -31,6 +32,7 @@ export interface DIResources {
 }
 
 export interface DIValues {
+  attrs?: IAttributeArchive[]
```
kswenson, line 35: As mentioned elsewhere, we have a similar type `ICodapV2Attribute` in `codap-v2-types.ts`. If the DI usage is somewhat different perhaps that type could be extended/modified locally rather then defining an entirely new type.

kswenson (changes requested):
👍 Generally looks good aside from a few selected tweaks. Note that I've been looking into our usage of some of the less clear properties and some of this will probably need to change at some point but I think those adjustments can wait until the corresponding PR lands rather than holding up this PR.

tealefristoe:
@kswenson Using `ICodapV2Attribute` worked well and allowed me to get rid of `IAttributeArchive`. However, it wasn't a perfect match and I wanted to highlight a couple of changes:
- A number of fields were made optional in `ICodapV2Attribute` and one was added: `hidden` (added), `description`, `formula`, `precision`.
- `id` and `guid` are `numbers` in `ICodapV2Attribute`, which makes using `attribute.id` not work for them. I'm still using `attribute.id` by casting it as `number`, but this basically always returns `null`. Since `id` and `guid` were already not correct, I figure this is ok for now, but it seems like something to address sooner rather than later.

kswenson (approved):
👍 LGTM

kswenson:
> @kswenson Using `ICodapV2Attribute` worked well and allowed me to get rid of `IAttributeArchive`. However, it wasn't a perfect match and I wanted to highlight a couple of changes:
> 
> * A number of fields were made optional in `ICodapV2Attribute` and one was added: `hidden` (added), `description`, `formula`, `precision`.

I noted the lack of `hidden` myself and have added it locally. I'm fine with making the others optional for now. If there's a difference between what's actually in documents vs. what's appropriate for the DI API that adjustment can be made separately.

> * `id` and `guid` are `numbers` in `ICodapV2Attribute`, which makes using `attribute.id` not work for them. I'm still using `attribute.id` by casting it as `number`, but this basically always returns `null`. Since `id` and `guid` were already not correct, I figure this is ok for now, but it seems like something to address sooner rather than later.

This will be part of an upcoming PR that tries to make sense of the multiple ids and name/title issues. The short version is that at least for imported v2 documents I think we need to preserve the v2 integer id, but this is fine for now.

PR #1152, opened 2024-03-08
stage build 3.0.0-pre.1517

nstclair-cc:
This staging build includes:

### Features/Improvements
- PT-181961973: The map **Layers** inspector menu gives users control over map layers
- PT-187005635: Users can specify that an attribute has a **color** data type
- PT-187014891: Color values are displayed in case table and case card as color swatches
- PT-187169360: Remove `sample`/`dashboard` url parameters on open/close file
- PT-187005724: A data value in various formats is automatically recognized as a color
- PT-187014833: When an attribute is user-assigned the color type, all 140 web color names are recognized

### DIFF
--- a/v3/CHANGELOG.md
+++ b/v3/CHANGELOG.md
@@ -1,5 +1,21 @@
 # Changelog
 
+## Version 3.0.0-pre.1517 - March 7, 2024
+
+### Features/Improvements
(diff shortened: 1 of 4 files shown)


kswenson (approved):
👍 LGTM

PR #1153, opened 2024-03-08
fix: graph handling of `color` attributes

kswenson:
[[PT-187203297]](https://www.pivotaltracker.com/story/show/187203297)

- Attributes with only empty values are incorrectly interpreted as `color` attributes.
- Attributes with color type are not handled in graph.


kswenson:

PR #1154, opened 2024-03-08
fix: clicking on points in dot plot fails to select them (PT-187153567)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/187153567

### DIFF
--- a/v3/src/components/graph/components/graph-component.tsx
+++ b/v3/src/components/graph/components/graph-component.tsx
@@ -65,24 +65,23 @@ export const GraphComponent = observer(function GraphComponent({ tile }: ITileBa
   if (!graphModel) return null
 
   return (
-    pixiPointsRef.current &&
-      <DataSetContext.Provider value={data}>
-        <InstanceIdContext.Provider value={instanceId}>
-          <GraphLayoutContext.Provider value={layout}>
-            <AxisLayoutContext.Provider value={layout}>
-              <GraphContentModelContext.Provider value={graphModel}>
-                <AxisProviderContext.Provider value={graphModel}>
-                  <Graph
-                    graphController={graphController}
-                    graphRef={graphRef}
-                    pixiPointsRef={pixiPointsRef}
-                  />
-                </AxisProviderContext.Provider>
-                <AttributeDragOverlay activeDragId={overlayDragId} />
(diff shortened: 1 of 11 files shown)


v3/src/components/graph/components/graph-component.tsx
```diff
@@ -65,23 +65,24 @@ export const GraphComponent = observer(function GraphComponent({ tile }: ITileBa
...
-                  pixiPointsRef={pixiPointsRef}
-                />
-              </AxisProviderContext.Provider>
-              <AttributeDragOverlay activeDragId={overlayDragId} />
-            </GraphContentModelContext.Provider>
-          </AxisLayoutContext.Provider>
-        </GraphLayoutContext.Provider>
-      </InstanceIdContext.Provider>
-    </DataSetContext.Provider>
+    pixiPointsRef.current &&
```
kswenson, line 68: Should be able to remove this now, right?

v3/src/components/graph/utilities/graph-utils.ts
```diff
@@ -382,7 +382,7 @@ export interface ISetPointSelection {
 export interface ISetPointCoordinates {
   anchor?: Point
   dataset?: IDataSet
-  pixiPointsRef: IPixiPointsRef
+  pixiPoints: PixiPoints
```
kswenson, line 385:
```suggestion
  pixiPoints?: PixiPoints
```
If you make this optional, then you shouldn't need to change as many callers. The code is already testing for existence in line 427 below.

kswenson (changes requested):
👍 Looks good -- just a couple of remaining things to clean up a bit.

v3/src/components/graph/components/graph-component.tsx
emcelroy: Weird. I swear I removed that previously, but I must not have committed the change somehow. 

I retesting things after removing it again and the bug is still fixed.

v3/src/components/graph/utilities/graph-utils.ts
emcelroy: Yes, thanks 👍

kswenson (approved):
👍 LGTM

PR #1155, opened 2024-03-11
fix: Edit Attribute Properties dialog [PT-187188045]

kswenson:
[[PT-187188045]](https://www.pivotaltracker.com/story/show/187188045)

The PT story is specifically about unselecting a user-selected attribute type, but it turned out there were several systemic problems in the dialog:
- clearing of property values generally didn't work (not just `color`)
- re-opening the dialog didn't refresh the dialog contents so that, for instance, if attribute properties were changed via undo/redo those changes weren't reflected in the dialog

In discussion with @bfinzer, some other changes were made to the v3 version of the dialog relative to v2:
- never store the `none` attribute type back to the model
- use "none" for the empty/unspecified attribute type label (rather than the empty string, i.e. nothing) 
- use "Yes"/"No" for radio button labels for editability rather than "True"/"False"
- add "Yes"/"No" to the list of translatable strings ("True"/"False" is hard-coded in v2 and neither that nor "Yes"/"No" were previously available as translatable strings).

### DIFF
--- a/v3/src/utilities/translation/lang/en-US.json5
+++ b/v3/src/utilities/translation/lang/en-US.json5
@@ -1,7 +1,5 @@
 {
     // CODAP V3 temporary strings (not for translation, at least not yet)
-    // code uses this dynamically, i.e. `DG.CaseTable.attribute.type.${type}`
-    "DG.CaseTable.attribute.type.color": "color",
     "V3.caseTable.noCases": "no cases",
     "V3.summary.parseResults": "Parsed \"%@1\" with %@2 %@3 (%@4 selected) and...",
     "V3.summary.noData": "No data",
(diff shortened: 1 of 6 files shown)


tealefristoe (approved):
Looks good to me 👍

bfinzer:
This could be a separate PT story: Editing the description changes the hover tip for the attribute in the case table. But *undoing* after that change the hover tip remains as it was.

v3/src/utilities/translation/lang/en-US.json5
```diff
@@ -20,6 +23,8 @@
     "V3.Undo.map.inspector.changeMapBaseLayer": "Undo changing base map",
     "V3.Redo.map.inspector.changeMapBaseLayer": "Redo changing base map",
 
+    // V3 case table strings that are not present in V2 and will eventually require translation.
+    "DG.CaseTable.attribute.type.color": "color", // used dynamically (`DG.CaseTable.attribute.type.${type}`)
```
bfinzer, line 27: Should be V3.CaseTable?

bfinzer (approved):
👍🏻LGTM

v3/src/utilities/translation/lang/en-US.json5
kswenson: That's the reason for the comment at the end of the line. The code builds the key dynamically with a template string so it wouldn't work with `V3`.

PR #1158, opened 2024-03-12
187194068 V3 Interactive State

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187194068

This PR makes CODAP V3 restore plugin state when documents are saved and loaded.

This work involved a lot of juggling asynchronous calls, which is something I have a tough time wrapping my head around. I wouldn't be surprised if there are unnecessary `async` and `awaits`, or if there were other aspects that are less than optimal.


kswenson (approved):
👍 Looks good! @tealefristoe and I worked on it together and added a few more commits on code review.

PR #1162, opened 2024-03-14
187230453 V3 Handle Plugin Attribute Create and Delete Requests

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187230453

This PR adds support for handling attribute create and delete requests from plugins.


v3/src/utilities/translation/lang/en-US.json5
```diff
@@ -59,6 +59,8 @@
     "V3.DI.Error.unsupportedAction": "Unsupported action: %@/%@",
     "V3.DI.Error.undefinedResponse": "Action handler returned undefined.",
     "V3.DI.Error.attributeNotFound": "Attribute not found",
+    "V3.DI.Error.fieldRequired": "%@ %@: %@ required",
```
kswenson, line 62: Order might be different in different languages -- this allows translators to rearrange  the strings, if necessary.
```suggestion
    "V3.DI.Error.fieldRequired": "%@1 %@2: %@3 required",
```

v3/src/data-interactive/handlers/interactive-frame-handler.ts
```diff
@@ -37,6 +37,8 @@ export const diInteractiveFrameHandler: DIHandler = {
     // TODO: Expand to handle additional values
     const { interactiveFrame } = resources
     if (!interactiveFrame) return noIFResult
+    if (Array.isArray(values)) return { success: true } // CODAP v2 seems to ignore interactiveFrame updates when an array is passed for values
```
kswenson, line 40:
```suggestion
    // CODAP v2 seems to ignore interactiveFrame updates when an array is passed for values
    if (Array.isArray(values)) return { success: true }
```
line length

v3/src/data-interactive/handlers/attribute-handler.ts
```diff
@@ -74,96 +109,17 @@ export const diAttributeHandler: DIHandler = {
...
     return attributeNotFoundResult
   },
-  delete: diNotImplementedYet
+  delete(resources: DIResources) {
+    const { attribute, dataContext } = resources
+    if (!attribute) return attributeNotFoundResult
+    if (!dataContext) return dataContextNotFoundResult
+
+    dataContext.applyUndoableAction(() => {
+      withoutUndo()
```
kswenson, line 118: Ok for now, but this is the kind of operation that should either be undoable or should clear the undo history entirely because attempting to undo past an unrecorded attribute deletion can lead to hard crashes.

v3/src/data-interactive/handlers/attribute-handler.ts
```diff
@@ -46,10 +52,39 @@ export const diAttributeHandler: DIHandler = {
...
   },
-  create: diNotImplementedYet,
+  create(resources: DIResources, values?: DIValues) {
+    const { dataContext } = resources
+    if (!dataContext) return dataContextNotFoundResult
+
+    // Wrap single attribute in array and bail if any new attributes are missing names
+    const attributeValues = Array.isArray(values) ? values : [values]
+    const attributeErrors = attributeValues.map(singleValue => {
+      if (!singleValue || singleValue.name == null) {
```
kswenson, line 62:
```suggestion
      if (!singleValue?.name) {
```
An empty string is not a valid name, either.

kswenson (changes requested):
👍 Looks good -- just a few suggestions and some lint warnings to fix.

tealefristoe:
@kswenson My plan to use `DIValues` as an `IAttributeSnapshot` won't work because there are a number of discrepancies between the two:
- Attributes have a `units` prop, but values have an equivalent `unit` field.
- Attributes have `userType`, but values have `type`. (Plus `type` is any string, but `userType` is one of a set number of particular strings.)
- Attribute's `id` is equivalent to values' `cid`, not its `id`. (And should we allow plugins to specify an attribute's `id` anyway?)
- The equivalent of values' `formula` is buried in `attribute.formula.display`.
- The equivalent of values' `hidden` isn't in the attribute at all, but in the `SharedCaseMetadata`. (Realizing this allowed me to improve the `attribute.update` handler in this PR.)

This is all explanation for why I changed so much since you last looked over this PR.

v3/src/data-interactive/handlers/attribute-handler.ts
```diff
@@ -1,43 +1,76 @@
...
-  const metadata = (dataContext && getSharedCaseMetadataFromDataset(dataContext))
   if (attribute) {
-    const { name, type, title, description, editable, id, precision } = attribute
+    return convertAttributeToV2(attribute, dataContext)
+  }
+}
+
+function convertValuesToAttributeSnapshot(values: DISingleValues): IAttributeSnapshot | undefined {
+  if (values.name) {
+    const userType = values.type && isAttributeType(values.type) ? values.type : undefined
```
kswenson, line 44: If you define `isAttributeType()` to return `false` for `undefined`, then this can be simplified to:
```suggestion
    const userType = isAttributeType(values.type) ? values.type : undefined
```

v3/src/data-interactive/handlers/attribute-handler.ts
```diff
@@ -1,43 +1,76 @@
...
-      precision,
-      unit: attribute.units
+      name: values.name,
+      userType,
+      title: values.title,
+      id: values.cid, // TODO Should we allow the values to specify an id at all?
+      // defaultMin: values.defaultMin, // TODO defaultMin not a part of IAttribute yet
+      // defaultMax: values.defaultMax, // TODO defaultMax not a part of IAttribute yet
+      description: values.description,
+      // categoryMap: values._categoryMap, // TODO categoryMap not part of IAttribute. Should it be?
```
kswenson, line 53: Likely to end up as part of metadata, but fine here for now.

v3/src/data-interactive/handlers/attribute-handler.ts
```diff
@@ -46,10 +79,49 @@ export const diAttributeHandler: DIHandler = {
...
+      }
+      return { success: true }
+    }).filter(error => !error.success)
+    if (attributeErrors.length > 0) return attributeErrors[0]
+
+    // Create the attributes
+    const attributes: IAttribute[] = []
+    const createAttribute = (value: DISingleValues) => {
+      dataContext.applyUndoableAction(() => {
+        withoutUndo()
```
kswenson, line 104: As mentioned elsewhere, we need to come up with a consistent strategy for dealing with undo/redo of plugin-driven actions. Since first reviewing this PR I had a meeting with @eireland at which we determined that in v2 some plugin-driven actions are undoable and some are not, depending entirely on the vagaries of the implementation. So there isn't a consistent v2 strategy to follow. Need not hold up this PR, however.

kswenson (approved):
👍 Looks good -- one suggested simplification and a couple of other comments.

PR #1164, opened 2024-03-15
feat: User can change bin width and alignment (PT-187125519)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/187125519


v3/src/components/graph/components/binneddotplotdots.tsx
```diff
@@ -153,6 +156,13 @@ export const BinnedDotPlotDots = observer(function BinnedDotPlotDots(props: Plot
 
   usePlotResponders({pixiPoints, refreshPointPositions, refreshPointSelection})
 
+  // respond to binAlignment and binWidth changes because the axis may need to be updated
+  useEffect(function respondToGraphBinSettings() {
+    return mstAutorun(() => {
+      refreshPointPositions(false)
+    }, {name: "respondToGraphBinSettings"}, graphModel)
+  }, [dataset, graphModel, refreshPointPositions])
```
kswenson, line 164: This is likely to result in redundant calls to `refreshPointPositions()`. As an autorun, this function will trigger when any observable value referenced in `refreshPointPositions()` is changed. The `usePlotResponders()` call in line 257 sets up a bunch of other observers that could trigger at the same time. The reactions set up in `usePlotResponders()` are debounced so that (in theory) multiple calls won't result in multiple runs, but this call won't be debounced with the others. If the issue you're trying to solve here is really about `binAlignment` and `binWidth` settings, then perhaps a `reaction` that limited the response to those values would serve the purpose. I suspect there's a larger refactor lurking here in which there could be a single `autorun` in `use-plot.ts` which re-runs any time any observable value used in the function changes and then additional `reaction`s which just trigger on changes to properties not directly referenced in the `autorun`, but that's presumably outside the scope of this PR.

kswenson (changes requested):
👍 Generally looks good -- a few suggestions to follow up on.

v3/src/components/graph/models/graph-data-configuration-model.ts
```diff
@@ -397,27 +397,30 @@ export const GraphDataConfigurationModel = DataConfigurationModel
...
       const kDefaultNumberOfBins = 4
       const minValue = self.caseDataArray.reduce((min, aCaseData) => {
         return Math.min(min, self.dataset?.getNumeric(aCaseData.caseID, self.primaryAttributeID) ?? min)
-      }, 0) ?? 0
+      }, Infinity)
       const maxValue = self.caseDataArray.reduce((max, aCaseData) => {
         return Math.max(max, self.dataset?.getNumeric(aCaseData.caseID, self.primaryAttributeID) ?? max)
-      }, 0) ?? 0
+      }, -Infinity)
       const binRange = (maxValue - minValue) / kDefaultNumberOfBins
```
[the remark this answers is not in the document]
emcelroy: I had some trouble with this part and very probably didn't nail it with my solution. I'm pretty sure though that the initial value of min can't be 0 or any other finite number since we don't know what the minimum value is until we iterate over the cases. If we set the initial value to 0 and the minimum value in `caseDataArray` is 3, for example,  we'd end up getting 0 as the minimum value instead of 3.

The latest changes are an improvement, I _think_. Instead of using `Infinity` and `-Infinity`, I initialize both min and max as `NaN` which is at least a little clearer? I believe with using `NaN` here you'd get the same results as with `Infinity` when there are no cases -- both `binWidth` and `binAlignment` would be set to `NaN`. I'm not 100% sure because I can't seem to test that. When there are no cases, you don't get the option to group points into bins. So maybe it's a moot point?

The latest version is a bit more verbose. Assuming it's a definite improvement, I'm wondering if it'd make sense to make one or more separate view methods for getting the min and max numeric values in `caseDataArray`? I'm not sure if that would be useful anywhere else, but it would help keep `binDetails` a little more concise. Do you think that'd be worth doing in this instance? Or is it OK as is?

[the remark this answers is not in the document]
kswenson: I did not mean to suggest that there's necessarily anything wrong with using `Infinity` in the way you were using it before. I merely meant that it would be bad if those min/max values propagated through and resulted in an attempt to render an axis with range `(Infinity, -Infinity)`, so there needs to be a check somewhere in the code and possibly a decision about what to render in that case. The fact that you said you can't get here when there's no cases goes a long way toward addressing the issue. Did you try grouping points into bins and then deleting the existing cases? In any case, I think the previous version with the infinities is probably clearer than the NaN version, as long as it doesn't propagate out to users.

v3/src/components/graph/components/inspector-panel/display-config-palette.tsx
```diff
@@ -27,6 +27,32 @@ export const DisplayConfigPalette = observer(function DisplayConfigPanel(props:
     }
   }
 
+  const setBinOption = (option: string, value: number) => {
```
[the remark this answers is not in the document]
kswenson: Since the constant is only used in this file, I'd just define it in this file rather than in `data-display-types.ts`.

v3/src/components/data-display/data-display-types.ts
```diff
@@ -6,6 +6,7 @@ export type Point = { x: number, y: number }
 export type CPLine = { slope: number, intercept: number, pivot1?: Point, pivot2?: Point }
 export const kNullPoint = {x: -999, y: -999}
 
+export type BinOption = "binWidth" | "binAlignment"
```
kswenson, line 9: I would leave this definition in `display-config-palette.tsx` since that's the only place it's used. It's not a constant meant to be shared across modules.

v3/src/components/graph/models/graph-data-configuration-model.ts
```diff
@@ -397,27 +397,40 @@ export const GraphDataConfigurationModel = DataConfigurationModel
...
+          minValue = isNaN(minValue) ? numericValue : Math.min(minValue, numericValue)
+        }
+      }
+      let maxValue = NaN
+      for (const aCaseData of self.caseDataArray) {
+        const numericValue = self.dataset?.getNumeric(aCaseData.caseID, self.primaryAttributeID)
+        if (isFiniteNumber(numericValue)) {
+          maxValue = isNaN(maxValue) ? numericValue : Math.max(maxValue, numericValue)
+        }
+      }
```
kswenson, line 416: As mentioned elsewhere, there was nothing wrong with the version that used `Infinity`. The only requirement is that none of these `Infinity` or `NaN` values gets propagated to the user.

kswenson (approved):
👍 Changes look good -- I'm approving with a few remaining minor suggestions:
1. Move a constant
2. Consider going back to the `Infinity` version of the bounds-checking code (not a requirement -- pick the one that makes the most sense, although the `Infinity` version seemed a bit simpler at first blush).
3. Test what happens if you delete the last case(s) after enabling binning. If it works and/or is easy to fix, feel free to address it. If it is going to be more work than that you can write a PT story.

v3/src/components/graph/models/graph-data-configuration-model.ts
```diff
@@ -397,27 +397,44 @@ export const GraphDataConfigurationModel = DataConfigurationModel
         }
       })
     },
-    binDetails() {
+    binDetails(_binAlignment?: number, _binWidth?: number, reset = false) {
```
kswenson, line 400: It doesn't appear that `reset` is ever used, in which case it can be removed.

emcelroy: Yes, I forgot to get rid of that. Thanks for pointing it out. I've removed it in the draggable bin boundaries branch I'm working on.

PR #1165, opened 2024-03-15
fix: case table display of color swatches with out-of-range values

kswenson:
[[PT-187213848]](https://www.pivotaltracker.com/story/show/187213848)

The essence of the bug was that after parsing a string as a valid color string we were then passing it directly to the browser. In strings with out-of-range values, however, the parser identified it as a valid string but the browser didn't display it properly. Meanwhile, in my investigation of color palettes, I came across another library ([colord](https://www.npmjs.com/package/colord)) which performs color conversions in addition to parsing and is more configurable than the [color-parse](https://www.npmjs.com/package/color-parse) library we were using before and even has built-in types.

Therefore, we take this opportunity to switch to the [colord](https://www.npmjs.com/package/colord) library internally and fix the bug by not simply parsing the user's string for validity but also canonicalizing it so that it works reliably in the browser.


v3/src/components/case-table/use-columns.tsx
```diff
@@ -32,10 +31,11 @@ export function renderValue(str = "", num = NaN, attr?: IAttribute) {
   const { type, userType } = attr || {}
 
   // colors
-  if ((type === "color" && parseColor(str).space) || (!userType && parseColorStrict(str).space)) {
+  const color = type === "color" || !userType ? parseColor(str, { colorNames: type === "color" }) : ""
+  if (color) {
     return (
       <div className="cell-color-swatch" >
-        <div className="cell-color-swatch-center" style={{ background: str }} />
+        <div className="cell-color-swatch-center" style={{ background: color }} />
```
tealefristoe, line 38: I think this will make `background: true` if `attr.type === "color"`, which seems wrong.

tealefristoe (changes requested):
Generally looks good, but I think there's a bug in `use-columns.tsx`.

tealefristoe (approved):
I see that my concern was based on a misunderstanding of the order of operations between `||` and `? :` and that your logic should work.

PR #1166, opened 2024-03-15
fix: graph import bug [PT-187250605]

kswenson:
[[PT-187250605]](https://www.pivotaltracker.com/story/show/187250605)

The bug occurred when the `useSubAxis` hook attempted to render an axis before it was fully initialized. In particular, the correct `NumericAxisModel` had been restored from the document but the `MultiScale`'s `scaleType` hadn't been set yet -- this occurs in `GraphController.initializeGraph`. The fix is simply to have `useSubAxis`'s `renderNumericAxis` function bail if the axis model and its scale aren't appropriate (yet).

I also fixed an unrelated MobX warning that appeared in loading the same document which simply required adding the `@action` decorator to a couple of MobX class methods that needed it.

### DIFF
--- a/v3/src/components/axis/hooks/use-sub-axis.ts
+++ b/v3/src/components/axis/hooks/use-sub-axis.ts
@@ -94,9 +94,11 @@ export const useSubAxis = ({
             .style("stroke-opacity", "0.7")
         },
         renderNumericAxis = () => {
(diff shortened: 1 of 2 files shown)


bfinzer (approved):
👍🏻LGTM

PR #1170, opened 2024-03-18
Removed 4 flaky tests to fix main. See PT #187262957.

nstclair-cc:
We've spotted four flaky tests in Cypress, and it looks like they might be tied to the redo button suddenly becoming disabled. If I were to venture a guess, it seems likely related to PT #187262957. In any case, I've created a pull request to address the main failing issue.


kswenson (approved):
👍 LGTM

PR #1171, opened 2024-03-19
187262806 v3 Import v2 Doc Plugins

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187262806

This PR allows web views and game views in v2 documents to be loaded as v3 web views.

For now, I'm not loading guide views, since we haven't yet decided how to handle guides in v3.

Additional fixes:
- Tiles imported from v2 documents that don't have `left` and `top` specified are getting defaults of 0. Leaving these undefined was causing a bug that ignored their `height` and `width`, resulting in tiles with no height or width that couldn't be resized.
- `iframe-phone` connections made in `webViews` are now properly disposed.
- `v2Attributes` can now have `editable` be a `boolean[]`. This is really weird, but was necessary to load the Parachute Model v2 doc, which has `[true]` for several `attribute.editable`.

This PR doesn't quite hit the test coverage target because I don't think any of the example documents has a web view to import, just plugins and guides.


v3/src/v2/codap-v2-types.ts
```diff
@@ -15,7 +15,7 @@ export interface ICodapV2Attribute {
   colormap?: any
   _colormap?: any
   blockDisplayOfEmptyCategories?: boolean
-  editable: boolean
+  editable: boolean | boolean[]
```
kswenson, line 18: Unless there's some evidence that this is also true for saved documents (which I don't think there is), it should be possible to handle this in the definition of `DIAttribute` without having to change `ICodapV2Attribute` or `CodapV2Document`. Something like:
```typescript
export interface DIAttribute extends Partial<Omit<ICodapV2Attribute, "editable">> {
  // some plugins, e.g. parachute, inexplicably send an array here ¯\_(ツ)_/¯
  editable?: boolean | boolean[]
}
```
Moving forward, the idea should be that the `codap-v2-types.ts` accurately reflect the types for v2 documents while the `data-interactive-types.ts` reflect the types for the plugin API, which can be derived from the document types but need not mirror them precisely.

v3/src/v2/import-v2-document.ts
```diff
@@ -32,7 +32,7 @@ export async function importV2Document(v2Document: CodapV2Document) {
       if (row && tile) {
         const info = getTileComponentInfo(tile.content.type)
         if (info) {
-          const { left, top, width, height } = v2Component.layout
+          const { left = 0, top = 0, width, height } = v2Component.layout
```
kswenson, line 35: Out of curiosity, did you happen to check how v2 handles this situation? It seems odd to me that these values can be undefined.

kswenson (changes requested):
👍 Generally looks good -- one suggestion to restrict plugin weirdness to the plugin code.

v3/src/v2/import-v2-document.ts
tealefristoe: I didn't dig into the code for v2 about this, but I did check that it makes the Markov plugin appear at about 0,0.

The main problem this is solving is that `isFreeTileInRowOptions` in `free-tile-row.ts` checks for `x` and `y` to determine if an object is an `IFreeTileInRowOptions`. `FreeTileRow.insertTile` uses this function to make sure the options passed to it are actually `IFreeTileInRowOptions`, and if they aren't it uses an empty object for the options, which results in `height` and `width` of `undefined` (even if these values are included in the options that were deemed the wrong type). Setting `x` and `y` to 0 as the default here ensures that the options passed to `content?.insertTileSnapshotInRow` on line 39 of `import-v2-document.ts` will pass the `isFreeTileInRowOptions` check, so their `height` and `width` won't be ignored. I figured this was the safest, simplest way to avoid this problem, but maybe there's a better solution?

v3/src/v2/codap-v2-types.ts
tealefristoe: This is an issue with the saved document, not just the plugin. Here's the document:
```{"name":null,"guid":1,"components":[{"type":"DG.GameView","guid":2,"componentStorage":{"currentGameName":"Next Gen MW","currentGameUrl":"https://lab.concord.org/embeddable.html#interactives/inquiry-space/parachute.json","allowInitGameOverride":true,"_links_":{"context":{"type":"DG.DataContextRecord","id":3}},"title":"parachute","name":"parachute"},"layout":{"width":640,"height":480,"x":5,"y":5,"isVisible":true,"left":5,"top":5,"zIndex":2},"savedHeight":null},{"type":"DG.TableView","guid":13,"componentStorage":{"_links_":{"context":{"type":"DG.DataContextRecord","id":3}},"attributeWidths":[],"title":"runs/measurements","name":"runs/measurements"},"layout":{"width":471,"height":200,"left":640,"top":0,"zIndex":10,"isVisible":true},"savedHeight":null}],"contexts":[{"type":"DG.GameContext","document":1,"guid":3,"flexibleGroupingChangeFlag":false,"name":"Data_Set_1","title":"runs/measurements","collections":[{"areParentChildLinksConfigured":true,"attrs":[{"name":"Run","type":"nominal","title":"Run","description":"","blockDisplayOfEmptyCategories":true,"editable":false,"hidden":false,"guid":5,"precision":2,"unit":null},{"name":"mass-of-jumper","type":"none","title":"mass-of-jumper","description":null,"blockDisplayOfEmptyCategories":true,"editable":[true],"hidden":false,"guid":6,"precision":null,"unit":"g"},{"name":"parachute-size","type":"none","title":"parachute-size","description":null,"blockDisplayOfEmptyCategories":true,"editable":[true],"hidden":false,"guid":7,"precision":null,"unit":"cm^2"}],"cases":[],"caseName":null,"childAttrName":null,"collapseChildren":true,"guid":4,"labels":{"singleCase":"run","pluralCase":"runs","singleCaseWithArticle":"a run","setOfCases":"set","setOfCasesWithArticle":"a set"},"name":"runs","title":"runs","type":"DG.Collection"},{"areParentChildLinksConfigured":true,"attrs":[{"name":"Time","type":"none","title":"Time","description":null,"blockDisplayOfEmptyCategories":true,"editable":[true],"hidden":false,"guid":9,"precision":null,"unit":"s"},{"name":"Distance to surface","type":"none","title":"Distance to surface","description":null,"blockDisplayOfEmptyCategories":true,"editable":[true],"hidden":false,"guid":10,"precision":null,"unit":"m"},{"name":"Velocity","type":"none","title":"Velocity","description":null,"blockDisplayOfEmptyCategories":true,"editable":[true],"hidden":false,"guid":11,"precision":null,"unit":"m/s"}],"cases":[],"caseName":null,"childAttrName":null,"collapseChildren":null,"guid":8,"labels":{"singleCase":"measurement","pluralCase":"measurements","singleCaseWithArticle":"a measurement","setOfCases":"time series","setOfCasesWithArticle":"a time series"},"name":"measurements","parent":4,"title":"measurements","type":"DG.Collection"}],"contextStorage":{"gameName":"Next Gen MW","gameUrl":"https://lab.concord.org/embeddable.html#interactives/inquiry-space/parachute.json"}}],"globalValues":[],"appName":"DG","appVersion":"2.0","appBuildNum":"0434","metadata":{"changeCount":0}}
```

kswenson: 🤦 Ok, so you're right to deal with it in `codap-v2-types.ts`. I've spent more time thinking about this than it probable deserves, but this is where I come down:
1. If it's in the document, we need to deal with it and the types should reflect it.
2. I checked and there's no evidence that this is dealt with granularly in the v2 code, i.e. I suspect `[true]` would behave the same as `[false]` in v2 because they would both just be treated as truthy. (This hypothesis could be tested.)
3. The issue only arises for v3 because MST's type requirements are more rigorous than SproutCore's (non-existent) type requirements.
4. While it is necessary for v3 to accommodate v2 documents as they exist in the field, it is not necessary that v3 perform better than v2 with such documents nor that it accommodate the original plugin bug (I assume that's the origin of the problem) that led to the invalid values.

Therefore, my proposal is that this be:
```suggestion
  // plugin bugs have led to documents in the field with values like `[true]`
  editable: boolean | unknown
```
and then use `editable: !!value` in the code. In other words, we'll accommodate the fact that there may be invalid values there, but we're not going to add code to handle specific invalid values. My reading of the code is that that's effectively how v2 is handling it.

v3/src/v2/import-v2-document.ts
kswenson: I think you're right that this is probably the best way to handle it. I'm just surprised that those properties are written out as `undefined`. The reason I wrote the free tile check that way is because I assumed that surely every CODAP component must save its position. If v2 handles such components by placing them in the top left, then defaulting to zero seems like the right approach.

v3/src/v2/codap-v2-types.ts
kswenson: >but we're not going to add code to handle specific invalid values

or indicate that any other values should be considered valid.

tealefristoe: I suspect that you're right about v2 treating any editable value as truthy, rather than expecting any specific value aside from `true` or `false`. Given that, I think your approach makes the most sense 👍

tealefristoe: One potential downside to this approach is that `editable` defaults to `false`, since `!!undefined` is `false`. Probably not a huge deal, but I thought it was worth pointing out.

kswenson: That is a good point. So it should probably be `value == null || !!value`.

kswenson (approved):
👍 Changes look good

PR #1172, opened 2024-03-19
187230453 v3 Fix Create Attribute Bug

tealefristoe:
PT story: https://www.pivotaltracker.com/story/show/187230453

This PR fixes a bug where creating an attribute would try to directly modify an MST model, which was throwing an error. This bug was also present in the update attribute call which I caught earlier, but I somehow missed it here.


kswenson (approved):
👍 LGTM

PR #1173, opened 2024-03-20
chore: improve v2 attribute import and plugin typing

kswenson:
[[PT-187277525]](https://www.pivotaltracker.com/story/show/187277525)

- Improves import of v2 attributes
- Improves the TypeScript types for: `ICodapV2Attribute`, `DISingleValues`
- Adds new TypeScript types: `DIAttribute`, `DIAttributes`, `DIInteractiveFrame`
- Adds a document that starts the documentation for the v2 document format

### DIFF
--- a/v3/doc/v2-document-format.md
+++ b/v3/doc/v2-document-format.md
@@ -1,0 +1,35 @@
+# Notes on the CODAP v2 Document Format
+
+## Introduction and Notes
+
+### Identifiers
+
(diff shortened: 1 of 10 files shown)


tealefristoe (approved):
This all looks good to me 👍 

I was wondering if we might want to add type checking before casting `DIValues` as various more specific types, but I now see that all properties of all types seem to be optional, so I don't think this would actually make any difference.

PR #1175, opened 2024-03-20
Add sort plugin api

eireland:
Adds a plugin API for sorting an attribute.

### DIFF
--- a/apps/dg/controllers/document_helper.js
+++ b/apps/dg/controllers/document_helper.js
@@ -207,11 +207,8 @@ DG.DocumentHelper = SC.Object.extend(
                   } else if (prop === 'rerandomize' && !!iValues[prop]) {
                     context.doRerandomizeAll();
                   } else if (prop === 'sort') {
-                    // example message: iValues['sort] = {attrID: 1, isDescending: true}
-                    // last resort: Or in the plugin - get the table for that context, and pass it in iValues with AttrId and asc/desc, so that you can call sort on it.
                     var controllersMap = DG.currDocumentController().get('componentControllersMap');
                     var tableAdapters = [];
-
                     for (var key in controllersMap) {
                       if (controllersMap[key].hasOwnProperty('caseTableAdapters')) {
                         tableAdapters.push(controllersMap[key]);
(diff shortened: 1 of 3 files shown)


bfinzer:
Two problems related to finding the relevant case table adapter and delegating the sort to it:
1. This adds an undoable action to CODAP's undo history.
2. If a case table has never been created; i.e. only a case card exists (which I'm not sure is possible) there may be no case table adapter to carry out the sort.

Given that we're nearing the end of V2's life and that this API call will probably only be used by one plugin, I think we can live with these two problems.

bfinzer (approved):
👍🏻Looks Good Enough To Me
See comment

apps/dg/controllers/document_helper.js
```diff
@@ -206,6 +206,25 @@ DG.DocumentHelper = SC.Object.extend(
...
+                      if (controllersMap[key].hasOwnProperty('caseTableAdapters')) {
+                        tableAdapters.push(controllersMap[key]);
+                        break;
+                      }
+                    }
+                    tableAdapters.forEach(function (tableController) {
+                      var tableContext = tableController.get('dataContext');
+                      var attribute = typeof(iValues[prop].attrId) === "string"
+                                        ? tableContext.getAttribute(iValues[prop].attrId).id
+                                        : iValues[prop].attrId;
```
kswenson, line 222: Do we need to `id => attribute => id` rather than just passing `iValues[prop].attrId` to `sortAttribute` below?

kswenson (approved):
👍 LGTM -- just one question about a possible code simplification

kswenson:
> * This adds an undoable action to CODAP's undo history.

In talking about this earlier with @eireland, there are other examples of plugin actions which result in undoable actions for CODAP. V2 is inconsistent in this respect.

apps/dg/controllers/document_helper.js
eireland: This was to take into account that users can pass either the attribute name or the id.

kswenson: I don't see a `getAttribute` method in the code, but there is a `getAttributebyName()` method -- and then I'd suggest adding a comment about the fact that `attrId` can be a name or an id because otherwise one is tempted to assume that it's an id for some strange reason. 😉

PR #1176, opened 2024-03-21
Restore graph ability to have multiple y attributes

bfinzer:
[#187286536] Bug fix: Failure to plot multiple y-attributes

* `PixiPoints` was ignoring the `plotNum` property of `CaseData` and thereby regarding cases plotted in multiple plots as duplicates. We fix this by regarding `CaseData` rather than just `caseID` as the key to points in `PixiPoints`

**Note**: In `DisplayItemDescriptionModel` I encountered the need to move the convenience getters that refer to points rather than items into a new view section. Not sure why that need occurred now.

### DIFF
--- a/v3/src/components/data-display/models/display-item-description-model.ts
+++ b/v3/src/components/data-display/models/display-item-description-model.ts
@@ -36,19 +36,22 @@ export const DisplayItemDescriptionModel = types
     get itemStrokeSameAsFill() {
       return self._itemStrokeSameAsFill
     },
+  }))
+  .views(self => ({
     // Convenience methods referring to points, especially for use by graphs
     pointColorAtIndex(plotIndex = 0) {
-      return this.itemColorAtIndex(plotIndex)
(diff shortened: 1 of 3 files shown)


kswenson (approved):
👍 Looks good after I added a couple of commits. `Map`s/`Set`s with object keys use shallow/reference equality rather than deep equality of keys, so the original changes wouldn't have worked as expected. I added a couple of commits to use string keys derived from `CaseData` objects, which should work as expected.

emcelroy (commented):
This generally looks good to me. I see a bug when checking it in the browser, though.

If you put multiple attributes on the Y, then try dragging various points, sometimes the wrong point moves. It's not immediately obvious to me why that is when looking at the code, but it seems like something isn't being mapped quite right. But maybe that should be considered a separate bug if it's not easy to fix?

bfinzer:
@emcelroy@concord.org Yes, I'm going to create a PT story for this bug.

PR #1177, opened 2024-03-21
feat: graph bin boundaries are draggable (PT-187125534)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/187125534

Makes the bin boundaries for binned points draggable in order to modify the bin width and alignment values. One difference from V2 introduced in these changes, after discussion with @bfinzer, is that the bin width is limited to a default minimum when it's set by dragging _and_ when it's set by entering a value into the "Bin width" text field. In V2, the default minimum limitation is only applied when the width is set by dragging.


v3/src/components/graph/components/binneddotplotdots.tsx
```diff
@@ -15,27 +17,120 @@ import { useInstanceIdContext } from "../../../hooks/use-instance-id-context"
...
+    lowerBoundaryRef.current = primaryAxisScale(newBinAlignment)
+    graphModel.setDynamicBinAlignment(newBinAlignment)
+  }, [dataConfig, graphModel, primaryAxisScale])
+
+  const handleDragBinBoundary = useCallback((event: MouseEvent) => {
+    if (!dataConfig) return
+    const dragValue = primaryIsBottom ? event.x : event.y
+    const screenBinWidth = Math.max(kMinBinPixelWidth, dragValue - lowerBoundaryRef.current)
+    const worldBinWidth =
+      Math.abs(primaryAxisScaleCopy.current.invert(screenBinWidth) - primaryAxisScaleCopy.current.invert(0))
```
kswenson, line 103: Use `screenWidthToWorldWidth()` utility function as described elsewhere.

v3/src/components/graph/components/binneddotplotdots.tsx
```diff
@@ -15,27 +17,120 @@ import { useInstanceIdContext } from "../../../hooks/use-instance-id-context"
...
+    graphModel.setDynamicBinWidth(worldBinWidth)
+  }, [dataConfig, graphModel, primaryIsBottom])
+
+  const handleDragBinBoundaryEnd = useCallback(() => {
+    graphModel.applyUndoableAction(
+      () => {
+        if (graphModel.binAlignment && graphModel.binWidth) {
+          // Round binWidth to the nearest multiple of .25
+          const newBinWidth = Math.round(graphModel.binWidth * 4) / 4
```
kswenson, line 112: Seems like .25 assumption would depend on axis domain.

v3/src/components/graph/components/binneddotplotdots.tsx
```diff
@@ -151,23 +236,52 @@ export const BinnedDotPlotDots = observer(function BinnedDotPlotDots(props: Plot
...
+  }, [graphModel, primaryAxisScale])
 
   return (
-    <>
-      <g data-testid={`bin-ticks-${instanceId}`} className="bin-ticks" ref={binTicksRef}/>
-    </>
+    plotArea2Ref?.current && createPortal(
+      <><g data-testid={`bin-ticks-${instanceId}`} className="bin-ticks" ref={binBoundariesRef}/></>,
+      plotArea2Ref.current
+    )
```
kswenson, line 285: I would suggest using Chakra's `<Portal>` here, which should behave the same but make the code a bit tidier. Also probably worth a comment about the fact that we're portaling to put the user-manipulable bin boundaries above the points.

kswenson (changes requested):
👍 Generally looks good -- still some remaining issues with coordinate systems and transforming between them. In another comment I suggest some utility functions that should help and also might want to improve some variable names so it's clear when we're dealing with world coordinates and when we're dealing with screen coordinates.

kswenson:
>One thing to especially consider in review is my use of `createPortal` to place the bin boundaries in the `plotArea2` SVG group. If I understand correctly, the bin boundary elements need to be in that group to be manipulatable by the user. But since plot components are rendered in `plotArea1`, `BinnedDotPlotDots` has to get them into `plotArea2` somehow. I'm not sure `createPortal` is the optimal method for doing that, but I couldn't think of anything else more straightforward.

I think portaling is the right approach. I suggested using Chakra's `<Portal>` component to simplify the code a bit and some renaming to make the code a bit clearer.

v3/src/components/graph/components/binneddotplotdots.tsx
```diff
@@ -151,23 +236,52 @@ export const BinnedDotPlotDots = observer(function BinnedDotPlotDots(props: Plot
...
+   pointDisplayType, addBinBoundaryDragHandlers])
 
   usePlotResponders({pixiPoints, refreshPointPositions, refreshPointSelection})
 
-  // respond to binAlignment and binWidth changes because the axis may need to be updated
+  // Respond to binAlignment and binWidth changes because the axis may need to be updated.
   useEffect(function respondToGraphBinSettings() {
     return mstReaction(
-      () => [graphModel.binAlignment, graphModel.binWidth],
+      () => [graphModel._binAlignment, graphModel._binWidth, graphModel.binAlignment, graphModel.binWidth],
```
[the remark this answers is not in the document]
emcelroy: I found the underscore properties were necessary. When I remove them, calling the graph model's `endBinBoundaryDrag` method doesn't have the desired effect of removing the drag classes from the boundaries and reapplying the drag-handling methods to them. I think that's because the non-underscore versions' values don't appear to change? In other words, when you've dragged a boundary and the `binWidth` is at say `5` and you stop dragging, `binWidth` will still be `5` and therefore won't trigger the reaction. We have to watch `_binWidth` for the actual change we're concerned with.  I may be wrong about that and/or there may be a more elegant way of handling that issue, though.

v3/src/components/graph/components/binneddotplotdots.tsx
```diff
@@ -15,27 +17,120 @@ import { useInstanceIdContext } from "../../../hooks/use-instance-id-context"
...
   const { onDrag, onDragEnd, onDragStart } = useDotPlotDragDrop()
   usePixiDragHandlers(pixiPoints, {start: onDragStart, drag: onDrag, end: onDragEnd})
 
+  const drawBinBoundaries = useCallback(() => {
+    if (!dataConfig || !isFiniteNumber(graphModel.binAlignment) || !isFiniteNumber(graphModel.binWidth)) return
+    // Round binWidth to one decimal place.
+    const binWidth = Math.round(graphModel.binWidth * 10) / 10
+    const secondaryPlace = primaryIsBottom ? "left" : "bottom"
+    const secondaryAxisScale = layout.getAxisScale(secondaryPlace) as ScaleBand<string>
+    const secondaryAxisExtent = Math.abs(Number(secondaryAxisScale.range()[0] - secondaryAxisScale.range()[1]))
```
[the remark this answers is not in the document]
emcelroy: No, `Number` wasn't necessary and I've removed that. I must've added it while trying to work out a problem and forgot to check if it was actually needed.

v3/src/components/graph/components/binneddotplotdots.tsx
```diff
@@ -15,27 +17,120 @@ import { useInstanceIdContext } from "../../../hooks/use-instance-id-context"
...
+    primaryAxisScaleCopy = useRef<ScaleNumericBaseType>(primaryAxisScale.copy()),
+    lowerBoundaryRef = useRef<number>(0)
 
   const { onDrag, onDragEnd, onDragStart } = useDotPlotDragDrop()
   usePixiDragHandlers(pixiPoints, {start: onDragStart, drag: onDrag, end: onDragEnd})
 
+  const drawBinBoundaries = useCallback(() => {
+    if (!dataConfig || !isFiniteNumber(graphModel.binAlignment) || !isFiniteNumber(graphModel.binWidth)) return
+    // Round binWidth to one decimal place.
+    const binWidth = Math.round(graphModel.binWidth * 10) / 10
```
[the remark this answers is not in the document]
emcelroy: Thanks for checking that. I was incorrect in my assessment that the bin width values were being rounded in V2. After re-checking things and discussing with Bill a little, I determined these instances of rounding are not needed. So I've removed all of them.

v3/src/components/graph/components/binneddotplotdots.tsx
emcelroy: [This was resolved in an off-Github conversation where we determined we would stick with using `createPortal` for now since the target is an SVG `g` element, and Chakra's `Portal` requires an HTML element.]

emcelroy:
Note: because it was so simple and related to the binned points work, I included a fix for [CODAP crashes upon graph component resize when "Group into Bins" is applied to Lifespan Data](https://www.pivotaltracker.com/story/show/187296306) here.

v3/src/components/graph/components/binneddotplotdots.tsx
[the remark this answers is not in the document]
emcelroy: Thinking about this a bit more, I should at least add something about the above to the comment for this `useEffect`, or maybe make a separate `useEffect` specific to the underscore properties? 

I think ideally we wouldn't need to check for changes to those properties, and could somehow else get `refreshPointPositions` to run at the end of a drag. That function can't really be called from `handleDragBinBoundaryEnd` though because the latter is defined first.

v3/src/components/graph/models/graph-data-configuration-model.ts
```diff
@@ -397,26 +397,18 @@ export const GraphDataConfigurationModel = DataConfigurationModel
...
-          binWidth: 1,
-          minBinEdge: 0,
-          maxBinEdge: 0,
-          minValue: 0,
-          maxValue: 0,
-          totalNumberOfBins: 0 
-        }
+        return { binAlignment: 0, binWidth: 1, minBinEdge: 0, maxBinEdge: 0, minValue: 0, maxValue: 0,
+                 totalNumberOfBins: 0 }
       }
```
bfinzer, line 413: When minValue === maxValue bad things happen, including that binWidth gets a value that the JSON parser chokes on during a restore. It's not clear what the best way to handle it is though.

bfinzer (changes requested):
I encountered a bug in which a saved document wouldn't restore. The JSON parser was croaking on the value of binWidth which was set to e-354 or something like that because, in the particular setting I was in, minValue === maxValue. I commented on this in binSettings.

v3/src/components/graph/models/graph-data-configuration-model.ts
```diff
@@ -397,26 +397,18 @@ export const GraphDataConfigurationModel = DataConfigurationModel
         }
       })
     },
-    binDetails(_binAlignment?: number, _binWidth?: number, reset = false) {
+    binDetails(_binAlignment?: number, _binWidth?: number) {
       const kDefaultNumberOfBins = 4
```
bfinzer, line 401: I wonder if this method belongs better in `GraphContentModel` where the `binWidth` and `binAlignment` are defined?

emcelroy: Yeah, I think that makes sense. I originally added it to the data configuration model because it uses a bunch of data configuration model properties, but I just realized we can easily get those in the graph model via its `dataConfiguration` property. So I've moved it there in the latest commit.

emcelroy:
> I encountered a bug in which a saved document wouldn't restore. The JSON parser was croaking on the value of binWidth which was set to e-354 or something like that because, in the particular setting I was in, minValue === maxValue. I commented on this in binSettings.

@bfinzer  That `5e-324` number is equal to `Number.MIN_VALUE` which is used in `binDetails` [here](https://github.com/concord-consortium/codap/blob/340f4bdc9d1505a41ec708fe2359589b3629a9df/v3/src/components/graph/models/graph-content-model.ts#L182).  I essentially copied that from [this line](https://github.com/concord-consortium/codap/blob/6f2377ba073d06ab1ad8f81b1d857c388a852eb2/apps/dg/utilities/math_utilities.js#L552) of V2's `goodTickValue` utility function. If I understand correctly, the purpose of using it there is to make sure not to return 0 or less than 0. Previously in `binDetails`, when min and max values were equal, you would end up with 0 as the value of `binRange`, which led to `binWidth` being set to `Number.MIN_VALUE` (` 5e-324`).

My first take on how to fix that is to just check if the min value is equal to the max and handle that case in a special way, like so:

```
const binRange = maxValue !== minValue
  ? (maxValue - minValue) / kDefaultNumberOfBins
  : 1 / kDefaultNumberOfBins
```

I made that change in the latest commit. It seems to work correctly to me. Does that make sense and work correctly for you, too?

kswenson (approved):
👍 Looks good -- just a couple remaining issues to consider.

bfinzer (approved):
👍 LGTM

v3/src/components/graph/models/graph-content-model.ts
```diff
@@ -139,6 +154,40 @@ export const GraphContentModel = DataDisplayContentModel
         xScale,
         yScale
       }
+    },
+    binDetails(_binAlignment?: number, _binWidth?: number) {
```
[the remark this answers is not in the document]
emcelroy: @kswenson I ended up refactoring `binDetails` following your (off-Github) suggestion of adding an option for specifying if the width and alignment should be initialized from the data. I did still split out the computation of bin width since that isn't always needed and moving the code for it helped simplify binDetails a little. What do you think of that?

v3/src/components/graph/models/graph-content-model.ts
```diff
@@ -139,6 +169,35 @@ export const GraphContentModel = DataDisplayContentModel
         xScale,
         yScale
       }
+    },
+    binDetails(options?: { shouldComputeInitial?: boolean }) {
```
kswenson, line 173:
```suggestion
    binDetails(options?: { initialize?: boolean }) {
```
Seems just as clear and more concise.

kswenson (approved):
👍 Changes look good -- just one naming suggestion

PR #1178, opened 2024-03-22
staging build 3.0.0-pre.1519

nstclair-cc:
staging build 3.0.0-pre.1519

### ✨ Features & Improvements:
- **PT-186957025:** Automation for Undo/Redo: legend
- **PT-187125519:** When Group into Bins is chosen, the menu changes to allow the user to change bin width and alignment.
- **PT-187259952:** Respond to get dataContextList requests
- **PT-181962713:** Map points can be displayed as a grid
- **PT-187230453:** Respond to attribute create and delete requests

### 🐞 Bug Fixes:
- **PT-187134147:** V2 map components and graph point properties are not being imported
- **PT-187128104:** Changes in point size aren't discrete wrt undo/redo
- **PT-187213848:** CODAP table attribute creation with out-of-range color value displays blank cell
- **PT-187286536:** Failure to plot multiple y-attributes

### DIFF
--- a/v3/CHANGELOG.md
+++ b/v3/CHANGELOG.md
@@ -1,4 +1,25 @@
 # Changelog
+## Version 3.0.0-pre.1519 - March 22, 2024
+
+### ✨ Features & Improvements:
+- **PT-186957025:** Automation for Undo/Redo: legend
+- **PT-187125519:** When Group into Bins is chosen, the menu changes to allow the user to change bin width and alignment.
+- **PT-187259952:** Respond to get dataContextList requests
+- **PT-181962713:** Map points can be displayed as a grid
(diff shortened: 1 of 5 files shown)


nstclair-cc:
FYI toolbar.spec.ts failed on my computer locally. I'll take a look in the morning but would like to move forward with the pull request for now.

nstclair-cc:
> 👍 LGTM -- I generally don't expect to see changes to `package-lock.json` in these PRs that don't otherwise change dependencies, but in this case they seem harmless.

Good catch on the `package-lock.json` file. I was indeed seeing the same issues we looked at together on Monday. I think next time it comes up I'll see if the tactic Teale suggested in Slack will work (deleting the `v3/node_modules` folder, then running `npm install`).

PR #1179, opened 2024-03-22
187263361 v3 Plugin Menu

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187263361

This PR adds the plugin button and menu to v3. It should look and work like in v2.

Note that some of the plugins aren't working yet, likely because many API handlers are not yet set up.

Additional improvements:
- The styles of the Tables and Options buttons have been improved.


kswenson (approved):
👍 LGTM -- just one note to check the optionality of properties of the `PluginData` interface.

PR #1180, opened 2024-03-22
Scatterplot extraneous points with multiple y-attributes

bfinzer:
[#187289494] Bug fix: In a scatterplot with multiple y-attributes, there are extraneous points at the top
* We need to override `DataConfigurationModel:filterCase` in `GraphDataConfigurationModel` so that we can properly handle the situation in which there are multiple y-attributes. As it was we were only testing the zeroth y-attribute.
* Paying attention to the `casesArrayNumber` in the `filterCase` method exposed a latent bug in which a crash occurs when removing one of multiple left y-axis attributes. We have to make sure each of the `GraphDataConfigurationModel.filteredCases` elements has the correct `casesArrayNumber`


kswenson (approved):
👍 Looks good -- I added a commit to tweak some of the types. Take a look and see if everything still makes sense.

PR #1181, opened 2024-03-24
Fix bug whereby only the points in the zeroth plot in a scatterplot could be dragged

bfinzer:
[#187289428] Bug fix: When dragging a scatterplot point with multiple y attributes, points not moving as they should

* The problem was that `ScatterDots` `refreshPointPositionsPerfMode`, when updating for just selected cases, was only updating the points for the zeroth plot. We fix by iterating through all the plots for each of the selected case
* Also a couple of lint fixes


kswenson (approved):
👍 LGTM

PR #1182, opened 2024-03-25
187308636 v3 DI update interativeFrame dimensions

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187308636

This PR allows plugins to update their interactiveFrame size through update interactiveFrame requests. Some built-in plugins used this feature to set their correct size, rather than including the correct size in the their plugin information.

This PR also makes CODAP return accurate dimensions in response to get interactiveFrame requests.

One question I have about this PR is where to put `getFreeTileLayout`. I originally put this function in `free-tile-row.ts`, but that caused an import cycle with `appState`. So I moved it into a new file, `document-utils.ts` in `models/document`, but this doesn't seem like the best place for it. Is there a better option?


kswenson (changes requested):
👍 It's nice that it's reasonably straightforward. As for the architectural question, I think my inclination would be to add `getTileDimensions()` and `setTileDimensions()` methods to the `DocumentContent` where (in theory) other layouts could be supported as well, although only the free layout need be supported at the moment.

kswenson (approved):
👍 Changes look good

PR #1183, opened 2024-03-26
187308660 v3 Correct plugin urls

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187308660

This PR appends `index.html` onto the paths to plugins that just direct CODAP to their source folder. It seems like several plugins were relying on the `index.html` being the default file loaded when visiting a directory, which is no longer the case on S3 (I guess?).

The majority of this PR actually hides the web view background, which shows through if the displayed web page/plugin has no background. Multiple plugins do not have a background and look bad with this message showing behind them.

### DIFF
--- a/v3/src/components/tool-shelf/plugins-button.tsx
+++ b/v3/src/components/tool-shelf/plugins-button.tsx
@@ -37,6 +37,9 @@ function PluginSelection({ pluginData }: IPluginSelectionProps) {
     documentContent?.applyUndoableAction(
       () => {
         const baseUrl = `${rootPluginUrl}${pluginData.path}`
+        // Some plugins relied on index.html being the default file loaded when pointing to a directory.
+        // This is no longer the case with our S3 hosted plugins, so we have to modify the path to point to the
+        // correct file in this case.
         const url = baseUrl.endsWith("/") ? `${baseUrl}index.html` : baseUrl
         const options = { height: pluginData.height, width: pluginData.width }
         const tile = documentContent?.createOrShowTile?.(kWebViewTileType, options)
--- a/v3/src/components/tool-shelf/plugins-button.tsx
+++ b/v3/src/components/tool-shelf/plugins-button.tsx
@@ -36,7 +36,8 @@ function PluginSelection({ pluginData }: IPluginSelectionProps) {
   function handleClick() {
     documentContent?.applyUndoableAction(
       () => {
-        const url = `${rootPluginUrl}${pluginData.path}`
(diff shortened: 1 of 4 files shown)


kswenson:
> > This PR appends `index.html` onto the paths to plugins that just direct CODAP to their source folder. It seems like several plugins were relying on the `index.html` being the default file loaded when visiting a directory, which is no longer the case on S3 (I guess?).
> 
> This is a non-trivial issue -- I talked to @scytacki about it recently. The redirect from a folder reference to `index.html` is a server-configuration issue. The `codap-resources` s3 bucket is apparently configured differently than `codap-server.concord.org` was. The fix proposed here works for the initial creation of the plugins, but we also need to make sure that we support saved documents with serialized plugin urls. Ultimately, we will need to decide how much of this to handle via client-side code, changing the urls in the `published-plugins.json`, and server configuration in combination.

Note that this doesn't mean that all of this has to be addressed in this PR -- just that there's more thinking to be done here.

v3/src/components/tool-shelf/plugins-button.tsx
```diff
@@ -36,7 +36,8 @@ function PluginSelection({ pluginData }: IPluginSelectionProps) {
   function handleClick() {
     documentContent?.applyUndoableAction(
       () => {
-        const url = `${rootPluginUrl}${pluginData.path}`
+        const baseUrl = `${rootPluginUrl}${pluginData.path}`
+        const url = baseUrl.endsWith("/") ? `${baseUrl}index.html` : baseUrl
```
kswenson, line 40: Probably worth a comment here. There may be more transformations we want to do to support legacy documents, etc.

v3/src/components/web-view/web-view-model.ts
```diff
@@ -11,12 +11,16 @@ export const WebViewModel = TileContentModel
...
+    dataInteractiveController: undefined as iframePhone.IframePhoneRpcEndpoint | undefined,
+    isPlugin: false
   }))
   .actions(self => ({
     setDataInteractiveController(controller?: iframePhone.IframePhoneRpcEndpoint) {
       self.dataInteractiveController = controller
     },
+    setIsPlugin(_isPlugin: boolean) {
+      self.isPlugin = _isPlugin
+    },
```
kswenson, line 23:
```suggestion
    setIsPlugin(isPlugin: boolean) {
      self.isPlugin = isPlugin
    },
```
Shouldn't need the underscores here.

kswenson (approved):
👍 Looks good -- just a couple stylistic suggestions

PR #1184, opened 2024-03-26
fix: Graph axis not split when categorical attribute added (PT-187314590)

emcelroy:
[[PT-187314590]](https://www.pivotaltracker.com/story/show/187314590)

When `renderNumericAxis` in `useSubAxis` is called, it needs to update the ~~axis'~~ axis copy's range. Otherwise the axis elements won't be sized appropriately when a categorical attribute is added to the opposite split.

### DIFF
--- a/v3/src/components/axis/hooks/use-sub-axis.ts
+++ b/v3/src/components/axis/hooks/use-sub-axis.ts
@@ -73,8 +73,8 @@ export const useSubAxis = ({
         axis = axisPlaceToAxisFn(axisPlace),
         type = _axisModel?.type,
         axisBounds = layout.getComputedBounds(axisPlace) as AxisBounds,
-        d3Scale: AxisScaleType = multiScale.scale.copy()
-          .range(axisIsVertical ? [rangeMax, rangeMin] : [rangeMin, rangeMax]) as AxisScaleType,
+        newRange = axisIsVertical ? [rangeMax, rangeMin] : [rangeMin, rangeMax],
+        d3Scale: AxisScaleType = multiScale.scale.copy().range(newRange) as AxisScaleType,
         initialTransform = (axisPlace === 'left')
           ? `translate(${axisBounds.left + axisBounds.width}, ${axisBounds.top})`
           : (axisPlace === 'top')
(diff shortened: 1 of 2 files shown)


v3/src/components/axis/hooks/use-sub-axis.ts
```diff
@@ -95,8 +95,8 @@ export const useSubAxis = ({
         },
         renderNumericAxis = () => {
           const numericScale = multiScale.scaleType === "linear"
-                                ? multiScale.numericScale as ScaleLinear<number, number>
-                                : undefined
+                                 ? multiScale.numericScale?.range(newRange) as ScaleLinear<number, number>
```
kswenson, line 98: This seems wrong:
1. It shouldn't be the job of the `renderNumericAxis` function of the `useSubAxis` hook to update the range of the scale owned/maintained by the `MultiScale`.
2. My recollection of the way this was supposed to work originally is that the scale owned/maintained by the `MultiScale` was responsible for mappings to the entire graph, while code responsible for subplots would make _a_ copy of that scale and adjust the range accordingly. We seem to be heading down a path in which there is no distinction between the scale owned/maintained by the `MultiScale` and the copies used by the subplots, but the code for subplots requires a blizzard of copies of that scale anyway.

I think there needs to be a larger consideration of the roles, responsibilities, and use of scales in the graph code.

v3/src/components/axis/hooks/use-sub-axis.ts
```diff
@@ -214,7 +214,7 @@ export const useSubAxis = ({
             )
         }
 
-      d3Scale.range(axisIsVertical ? [rangeMax, rangeMin] : [rangeMin, rangeMax])
+      d3Scale.range(newRange)
```
kswenson, line 217: We set the range on this scale when we created it in line 77. Seems like it shouldn't be necessary to set it again here.

kswenson (changes requested):
As mentioned in the comments below, I think this requires further consideration.

v3/src/components/axis/hooks/use-sub-axis.ts
emcelroy: @kswenson Sorry. I thought this would be a quick fix for a bug that's making it difficult to continue the bar chart work. Before the changes made [here](https://github.com/concord-consortium/codap/pull/1166) the range was being updated [here](https://github.com/concord-consortium/codap/blob/b658aa22f65c2f23a7d0625bc1c748dd1beb062a/v3/src/components/axis/hooks/use-sub-axis.ts#L76) in `useSubAxis`, although now I see that was being done to a copy made using `copy()`. Would using a copy and setting the copy's range here go at least some way toward addressing the issues you're concerned with? If not, I'm not sure I should devote more time to this right now.

kswenson: Yes, restricting the scale changes to scales created/maintained by this code would ameliorate the architectural issues raised above. The seemingly excessive creation of copies is a potential performance issue rather than an architectural one.

emcelroy: OK, thanks. I've made that change and drafted a [new story](https://www.pivotaltracker.com/story/show/187317584) about further investigation into the use of copied scales.

kswenson (approved):
👍 LGTM

PR #1185, opened 2024-03-27
187308847 Prevent webView iframes from displaying over other elements

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187308847

This PR fixes a bug that caused webView iframes to display over most other CODAP elements, including other tiles and dropdown menus.

The fix is to remove `z-index` for the webView background and iframe. As far as I can tell, these weren't actually doing anything, and were most likely added in addition to other css that actually fixed a problem.


kswenson (approved):
👍 LGTM! Always happy to simplify our z-index situation rather than complicate it.

PR #1186, opened 2024-03-27
187263404 v3 di broadcast global nofications

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187263404

This PR makes CODAP v3 broadcast notifications to plugins about global value changes.

It also makes substantial changes to `applyUndoableAction`:
- Adds optional `options` argument.
- Moves the required `undoStringKey` and `redoStringKey` parameters into `options` and makes them optional.
- Applies `withoutUndo` if either `undoStringKey` or `redoStringKey` are missing from `options`.
- Will broadcast a notification included in `options` to all plugins.

Most of the modified lines in this PR are from this change.

There is one more related change I'd like to make, but I figure I should check before I make it. Given that `applyUndoableAction` no longer always performs an undoable action, and it also has the potential to perform other generally useful functions, I think the name should be changed to something like `applyAction` or `applyCodapAciton`.


v3/cypress/e2e/plugin.spec.ts
```diff
@@ -7,88 +8,100 @@ context("codap plugins", () => {
...
-      webView.enterUrl(url)
-      cy.wait(1000)
+  const openAPITester = () => {
+    const url='https://concord-consortium.github.io/codap-data-interactives/DataInteractiveAPITester/index.html?lang=en'
+    toolbar.getOptionsButton().click()
+    toolbar.getWebViewButton().click()
+    webView.enterUrl(url)
+    cy.wait(1000)
+  }
+  it.skip('will handle plugin requests', () => {
```
kswenson, line 18:
```suggestion
  it('will handle plugin requests', () => {
```

v3/src/components/slider/editable-slider-value.tsx
```diff
@@ -3,10 +3,11 @@ import { autorun } from "mobx"
 import {observer} from "mobx-react-lite"
 import { isAlive } from "mobx-state-tree"
 import React, {useState, useEffect} from "react"
-import {ISliderModel} from "./slider-model"
 import {MultiScale} from "../axis/models/multi-scale"
+import {ISliderModel} from "./slider-model"
 
 import './slider.scss'
+import { valueChangeNoficiation } from "./slider-utils"
```
kswenson, line 10:
```suggestion
import { valueChangeNotification } from "./slider-utils"
```
and move import ⬆️

v3/src/components/slider/editable-slider-value.tsx
```diff
@@ -45,7 +46,12 @@ export const EditableSliderValue = observer(function EditableSliderValue({ slide
           sliderModel.encompassValue(inputValue)
           sliderModel.setValue(inputValue)
         },
-        "DG.Undo.slider.change", "DG.Redo.slider.change")
+        {
+          notification: valueChangeNoficiation(inputValue, sliderModel.globalValue.name),
```
kswenson, line 50:
```suggestion
          notification: valueChangeNotification(inputValue, sliderModel.globalValue.name),
```

v3/src/components/slider/slider-utils.ts
```diff
@@ -0,0 +1,10 @@
+import { debugLog, DEBUG_PLUGINS } from "../../lib/debug"
+
+export function valueChangeNoficiation(value: number, name?: string) {
```
kswenson, line 3:
```suggestion
export function valueChangeNotification(value: number, name?: string) {
```

v3/src/data-interactive/iframe-phone-types.ts
```diff
@@ -0,0 +1,3 @@
+// TODO any type. This is a message sent via iframe-phone. iframe-phone says it should be a string, but that's
+// just not true. Hopefully iframe-phone will provide an accurate type in the future.
```
kswenson, line 2: We should add a PT story to the CODAP v3 PT board about updating types in iframe-phone. We're the ones running into it right now so that makes us the logical team to fix it.

v3/src/models/document/base-document-content.ts
```diff
@@ -146,6 +148,14 @@ export const BaseDocumentContentModel = types
...
       self.findRowContainingTile(tileId)?.setTileDimensions(tileId, dimensions)
+    },
+    broadcastMessage(message: DIMessage, callback: iframePhone.ListenerCallback) {
+      const tileIds = self.tileMap.keys()
+      if (tileIds) {
+        Array.from(tileIds).forEach(tileId => {
+          self.tileMap.get(tileId)?.content.broadcastMessage(message, callback)
+        })
+      }
     }
```
kswenson, line 159: I would move this up to `document-content.ts` where more client-specific behavior lives.

kswenson (changes requested):
👍 Looks good -- just a handful of small changes

tealefristoe:
@bfinzer and @emcelroy  You probably don't need to look over this entire PR, but I wanted to make sure you saw changes I'm making to `applyUndoableAction`. I'm also curious about your and @kswenson 's thoughts about renaming this function to `applyAction` or something similar.

kswenson:
> @bfinzer and @emcelroy You probably don't need to look over this entire PR, but I wanted to make sure you saw changes I'm making to `applyUndoableAction`. I'm also curious about your and @kswenson 's thoughts about renaming this function to `applyAction` or something similar.

I don't like `applyAction` because most MST actions don't require this function. It could be something like `applyRecordedAction` or `applyMonitoredAction` but I don't particularly like either of those. Even though it's not always used for actions that are actually undoable, `applyUndoableAction` at least has the advantage that it directly suggests that the undo history is implicated, even though one of its uses is to explicitly disable undoability. I could see an argument for `applyUserAction` given that for most use cases those are the actions we want to undo, log, notify, etc., although of course not all such actions are triggered by the user.

emcelroy:
> @bfinzer and @emcelroy You probably don't need to look over this entire PR, but I wanted to make sure you saw changes I'm making to `applyUndoableAction`. I'm also curious about your and @kswenson 's thoughts about renaming this function to `applyAction` or something similar.

@tealefristoe Thanks for keeping me in the loop. Changes make sense and look good (thought there's one tiny import order issue in v3/src/components/slider/editable-slider-value.tsx). I can see why you'd want to consider renaming the function but I tend to agree with @kswenson 's take on that. I can't think of a better name either. The only thing that comes to mind is including something like `possibly` in it. So something like `applyActionPossiblyWithUndo` but I'm not sure that'd be much of an improvement.

kswenson:
> > @bfinzer and @emcelroy You probably don't need to look over this entire PR, but I wanted to make sure you saw changes I'm making to `applyUndoableAction`. I'm also curious about your and @kswenson 's thoughts about renaming this function to `applyAction` or something similar.
> 
> @tealefristoe Thanks for keeping me in the loop. Changes make sense and look good (thought there's one tiny import order issue in v3/src/components/slider/editable-slider-value.tsx). I can see why you'd want to consider renaming the function but I tend to agree with @kswenson 's take on that. I can't think of a better name either. The only thing that comes to mind is including something like `possibly` in it. So something like `applyActionPossiblyWithUndo` but I'm not sure that'd be much of an improvement.

How about `applyModelChange`?

v3/src/components/slider/editable-slider-value.tsx
```diff
@@ -3,10 +3,11 @@ import { autorun } from "mobx"
 import {observer} from "mobx-react-lite"
 import { isAlive } from "mobx-state-tree"
 import React, {useState, useEffect} from "react"
-import {ISliderModel} from "./slider-model"
 import {MultiScale} from "../axis/models/multi-scale"
+import {ISliderModel} from "./slider-model"
 
 import './slider.scss'
+import { valueChangeNotification } from "./slider-utils"
```
kswenson, line 10: Move import ⬆️

kswenson (approved):
👍 LGTM -- one suggestion to move an import statement and one thought about a name change to `applyModelChange` (which need not hold up the PR while we're ruminating on the name).

bfinzer (approved):
👍🏻LGTM

PR #1189, opened 2024-03-28
Staging build 3.0.0-pre.1520

nstclair-cc:
[Staging build 3.0.0-pre.1520](https://www.pivotaltracker.com/story/show/187326452)

Note that package-lock.json _should_ just be an updated version number in the file, but if it's not we might want to take a closer look because something might be happening when I run the script `npm version --no-git-tag-version 3.0.0-pre.<new-version>` to update `package-lock.json`

### DIFF
--- a/v3/CHANGELOG.md
+++ b/v3/CHANGELOG.md
@@ -1,4 +1,27 @@
 # Changelog
+## Version 3.0.0-pre.1520 - March 28, 2024
+
+### ✨ Features & Improvements:
+- **PT-186957009:** Automation for Undo/Redo: calculator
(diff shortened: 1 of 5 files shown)


kswenson:
> Note that package-lock.json _should_ just be an updated version number in the file, but if it's not we might want to take a closer look because something might be happening when I run the script `npm version --no-git-tag-version 3.0.0-pre.<new-version>` to update `package-lock.json`

Yes, there are unrelated changes in `package-lock.json`. If you discard your local changes to `package-lock.json` and then run `npm ci` followed by `npm version` it should eliminate them.

PR #1190, opened 2024-03-29
feat: compatibility with v2 name/title usage

kswenson:
- `V2Model` implements id, name, and title logic and is base model for `Attribute`, `Collection`, `DataSet`
- `matchNameOrId` method for use by plugins to match on name or any id
- Utility functions: `isV2ModelSnapshot`, `v2NameTitleToV3Title`, `v2ModelSnapshotFromV2ModelStorage`

Attributes, collections, and data contexts imported from v2 documents now preserve their v2 ids and names/titles so plugins that rely on v2 ids and names should continue to work. Other objects (e.g. global values and components) will require similar treatment.

Update: after further investigation of v2 behavior, v3 should now work like v2 in that:
- attributes and collections have title properties, but only ever get/set their `name` property.
- data sets _do_ maintain separate `name`/`_title` properties and user edits update the `_title` property.
- case table tiles treat their data set title as their tile title, ignoring the `title` property in `TileModel`.

@tealefristoe Requesting re-review, but note that you only need to consider the second commit, i.e. changes since the first review.

### DIFF
--- a/v3/src/components/case-table/case-table-title-bar.tsx
+++ b/v3/src/components/case-table/case-table-title-bar.tsx
@@ -12,8 +12,8 @@ import "./case-table-title-bar.scss"
 
 export const CaseTableTitleBar = observer(function CaseTableTitleBar({tile, ...others}: ITileTitleBarProps) {
   const data = isCaseTableModel(tile?.content) ? tile?.content.data : undefined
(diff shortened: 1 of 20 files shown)


tealefristoe (approved):
Looks good 👍

tealefristoe (approved):
Changes look good to me 👍

PR #1194, opened 2024-04-02
fix: import of attribute precision; fixes import of v2 Roller Coasters example document

kswenson:
[[PT-187296100]](https://www.pivotaltracker.com/story/show/187296100) "Roller Coasters" file fails to open in Example Docs menu

The root of the problem is that the `precision` property is declared to be an optional numeric value, but apparently sometimes a string (e.g. "" or "2") is written out 🤷. As mentioned previously, MST enforces types more stringently than v2 did, so it's important that we identify the _actual_ contents of v2 documents and make sure we support them.

### DIFF
--- a/v3/src/data-interactive/handlers/attribute-handler.ts
+++ b/v3/src/data-interactive/handlers/attribute-handler.ts
@@ -2,6 +2,7 @@ import { IAttribute, IAttributeSnapshot, isAttributeType } from "../../models/da
 import { IDataSet } from "../../models/data/data-set"
 import { v2ModelSnapshotFromV2ModelStorage } from "../../models/data/v2-model"
 import { getSharedCaseMetadataFromDataset } from "../../models/shared/shared-data-utils"
+import { hasOwnProperty } from "../../utilities/js-utils"
 import { t } from "../../utilities/translation/translate"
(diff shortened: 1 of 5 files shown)


v3/src/data-interactive/handlers/attribute-handler.ts
```diff
@@ -125,7 +126,9 @@ export const diAttributeHandler: DIHandler = {
       if (values?.editable != null) attribute.setEditable(!!values.editable)
       if (values?.formula != null) attribute.setDisplayExpression(values.formula)
       if (values?.name != null) attribute.setName(values.name)
-      if (values?.precision != null) attribute.setPrecision(values.precision)
+      if (hasOwnProperty(values, "precision")) {
+        attribute.setPrecision(values.precision == null || values.precision === "" ? undefined : values.precision)
+      }
```
kswenson, line 131: @tealefristoe Note that I only changed this for `precision`, but the `!= null` test is inappropriate for any optional property because it prevents the client from unsetting such optional properties. There should be another PR to address this issue for the other optional properties.

tealefristoe (approved):
Looks good! Always a little sad to see a clean implementation get a bit messier because of legacy concerns, but that's the way things go sometimes.

PR #1195, opened 2024-04-02
Make undo/redo of point size change discrete

bfinzer:
[#187128104] Bug fix: Changes in point size aren't discrete wrt undo/redo

* Followed the usual pattern of introducing a volatile dynamic property in `DisplayItemDescriptionModel` so that `DisplayItemFormatControl` can set the dynamic property during a drag and wait until the drag ends to set the non-volatile property with an undoable action.
* Graph and map V2 importers needed a tweak to accommodate the change.


emcelroy (approved):
Looks good 👍

PR #1196, opened 2024-04-02
feat: dot chart can be converted to bar chart (PT-181915881)

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/181915881


kswenson (changes requested):
👍 Functionality is impressive -- as you suggested there are some architectural considerations. I made some suggestions in comments below. See what you think and let me know if you want to touch base to discuss further.

kswenson (approved):
👍 Looks good -- nice refactor on the data tip code! I added a commit to tweak some of the type definitions. Otherwise, everything looks good.

PR #1197, opened 2024-04-03
187354366 v3 DI Global Requests

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187354366

This PR makes v3 handle `get`, `create`, and `update` `global`, as well as `get` `globalList`, requests from plugins.

These requests are handled a bit differently from v2 in some ways.
- `create global` requests will fail for duplicate global names. The error message in this case is "globals must have unique names".
- Non-string names included in `create global` requests are converted into strings.
- `create` and `update global` will fail for non-number values. The error message in this case is "global values must be numbers".
  - We try to convert the value into a number before failing here, which makes values like `"2"` and `[2]` work, but `"a"` and `[2,3]` fail.
- `get globalList` seems to be broken in v2, always returning `{ success: false }` with no error message! I've set it up to work as it seems intended in v3, returning a list of all global values.


bfinzer (approved):
👍🏻LGTM

v3/src/data-interactive/handlers/global-handler.ts
```diff
@@ -0,0 +1,61 @@
...
+  create(_resources: DIResources, values?: DIValues) {
+    const { document } = appState
+    const globalManager = getGlobalValueManager(getSharedModelManager(document))
+    const { name, value } = values as DIGlobal
+
+    const _value = Number(value ?? kDefaultSliderValue)
+    if (isNaN(_value)) return illegalValueResult
+
+    const globalSnapshot = {
+      name: (name ?? globalManager?.uniqueName() ?? kDefaultSliderName).toString(),
```
kswenson, line 21:
```suggestion
      name: (name || globalManager?.uniqueName() || kDefaultSliderName).toString(),
```
We shouldn't allow empty strings, either.

v3/src/data-interactive/handlers/global-handler.ts
```diff
@@ -0,0 +1,61 @@
...
+const illegalValueResult = { success: false, values: { error: t("V3.DI.Error.globalIllegalValue") } } as const
+
+export const diGlobalHandler: DIHandler = {
+  create(_resources: DIResources, values?: DIValues) {
+    const { document } = appState
+    const globalManager = getGlobalValueManager(getSharedModelManager(document))
+    const { name, value } = values as DIGlobal
+
+    const _value = Number(value ?? kDefaultSliderValue)
+    if (isNaN(_value)) return illegalValueResult
```
kswenson, line 18:
```suggestion
    if (!isFinite(_value)) return illegalValueResult
```
I would use `isFinite` here to prevent infinities as well.

v3/src/data-interactive/handlers/global-handler.ts
```diff
@@ -0,0 +1,61 @@
...
+    }
+  },
+  update(resources: DIResources, values?: DIValues) {
+    const { document } = appState
+    const { global } = resources
+    const { value } = values as DIGlobal
+    if (!global || !value) return { success: false, values: { error: t("V3.DI.Error.globalNotFound") } }
+
+    const _value = Number(value)
+    if (isNaN(_value)) return illegalValueResult
```
kswenson, line 52:
```suggestion
    if (!isFinite(_value)) return illegalValueResult
```
I would use `isFinite` here to prevent infinities as well.

kswenson (approved):
👍 Looks good -- just a few nits to pick 😉

PR #1198, opened 2024-04-03
feat: count adornment shows count per bin (PT-187233618)

emcelroy:
[[#187233618](https://www.pivotaltracker.com/story/show/187233618)]

These changes ensure that when points are grouped into bins and the Count/Percent adornment is activated, there is an instance of the adornment per region defined by the bin boundaries.

**Note:** The updated comments mention that it _should not_ be possible to have movable values present when points are grouped into bins, but that currently _is_ possible in V3. I believe that's essentially a bug. When points are binned in V2, Count/Percent is the only adornment available. In V3 you have the option to activate all the univariate plot adornments. I think fixing that should be a separate piece of work. If you agree, I will create a separate story for that.


v3/src/components/graph/adornments/count/count-adornment-component.tsx
```diff
@@ -35,13 +36,24 @@ export const CountAdornment = observer(function CountAdornment(props: IAdornment
...
   const [displayCount, setDisplayCount] = useState(<div>{textContent}</div>)
 
+  const subPlotRegionBoundaries = useCallback(() => {
+      // Sub plot regions can be defined by either bin boundaries when points are grouped into bins, or by
+      // instances of the movable value adornment. It should not be possible to have both bin boundaries and
+      // movable values present at the same time.
+      if (graphModel.pointDisplayType === "bins") {
+        const { binWidth, minBinEdge, maxBinEdge, totalNumberOfBins } = graphModel.binDetails()
+        return [
+          ...Array.from({ length: totalNumberOfBins }, (_, i) => minBinEdge + i * binWidth),
```
kswenson, line 48: Clever -- probably worth a comment, however, since the syntax is not immediately obvious.

v3/src/components/graph/adornments/count/count-adornment-component.tsx
```diff
@@ -109,21 +130,22 @@ export const CountAdornment = observer(function CountAdornment(props: IAdornment
...
         // set showPercent to false if attributes change to a configuration that doesn't support percent
-        const shouldShowPercentOption = !!dataConfig?.categoricalAttrCount || adornmentsStore.subPlotsHaveRegions
+        const shouldShowPercentOption = !!dataConfig?.categoricalAttrCount || adornmentsStore.subPlotsHaveRegions ||
+                                        graphModel.pointDisplayType === "bins"
         if (!shouldShowPercentOption && model?.showPercent) {
           model.setShowPercent(false)
         }
      }, { name: "CountAdornment.refreshPercentOption"}, model)
-  }, [adornmentsStore.subPlotsHaveRegions, dataConfig, model])
+  }, [adornmentsStore.subPlotsHaveRegions, dataConfig, graphModel.pointDisplayType, model])
```
kswenson, line 148:
```suggestion
  }, [adornmentsStore, dataConfig, graphModel, model])
```
If the `subPlotsHaveRegions` or `pointDisplayType` properties change we want the `autorun` to trigger rather than being reinstalled.

v3/src/components/graph/models/graph-data-configuration-model.ts
```diff
@@ -414,10 +414,11 @@ export const GraphDataConfigurationModel = DataConfigurationModel
     })
   }))
   .views(self => ({
-    casesInRange(min: number, max: number, attrId: string, cellKey: Record<string, string>) {
+    casesInRange(min: number, max: number, attrId: string, cellKey: Record<string, string>, inclusiveMax = true) {
       return self.subPlotCases(cellKey)?.filter(caseId => {
         const caseValue = self.dataset?.getNumeric(caseId, attrId)
-        if (isFiniteNumber(caseValue) && caseValue >= min && caseValue <= max) {
+        const upperRangeBoundary = inclusiveMax ? max : max - 0.000001
```
kswenson, line 420: These sorts of hard-coded offsets always make me nervous. Seems like it would be better as a percentage, e.g. max * 0.999999 because then at least it would be less resistant to magnitude effects.

v3/src/components/graph/adornments/count/count-adornment-component.tsx
```diff
@@ -60,27 +72,36 @@ export const CountAdornment = observer(function CountAdornment(props: IAdornment
...
-      // If there are no movable values present, we just show a single case count.
+      // If there are no bin boundaries or movable values present, we just show a single case count.
       setDisplayCount(<div>{textContent}</div>)
       return
     }
 
     const regionCountParams: IRegionCountParams = {
       cellKey,
       dataConfig,
+      inclusiveMax: graphModel.pointDisplayType !== "bins",
```
kswenson, line 93: Worth a comment here. I take it the idea is that movable value counts are inclusive while bin counts are exclusive? I trust that's the v2 behavior? A priori I might have expected both to be exclusive.

kswenson (changes requested):
👍 Generally looks good -- just a few suggestions.

v3/src/components/graph/adornments/count/count-adornment-component.tsx
emcelroy: @kswenson I think it's true only regions defined by movable values are inclusive. Maybe I'm thinking about it wrong and/or using the wrong terminology, though. Below is a screenshot from V2. With movable values, if a point has the same value as one of the movable values, it's included in the count for the region whose upper boundary is defined by that movable value. With bins, if a point's value matches a bin boundary, it's placed in the bin whose lower boundary is defined by that bin. Does that all seem to make sense with the use of `inclusiveMax` here? Or do you think it needs to be adjusted somehow?

image
(My mouse cursor is hovering over the point in the "50,60" bin in the rightmost graph to show that its value is 50.)

kswenson: I was just suggesting a comment, which you've added, and making sure you'd checked the v2 behavior, which it seems you have.

kswenson (approved):
👍 Changes LGTM

v3/src/components/graph/models/graph-data-configuration-model.ts
```diff
@@ -414,10 +414,11 @@ export const GraphDataConfigurationModel = DataConfigurationModel
     })
   }))
   .views(self => ({
-    casesInRange(min: number, max: number, attrId: string, cellKey: Record<string, string>) {
+    casesInRange(min: number, max: number, attrId: string, cellKey: Record<string, string>, inclusiveMax = true) {
       return self.subPlotCases(cellKey)?.filter(caseId => {
         const caseValue = self.dataset?.getNumeric(caseId, attrId)
-        if (isFiniteNumber(caseValue) && caseValue >= min && caseValue <= max) {
+        const upperRangeBoundary = inclusiveMax ? max : max * 0.999999
```
bfinzer, line 420: I think that rather than choosing an arbitrary multiplier (0.999999), it would be better to change the comparison.
const isLessThan = inclusiveMax ? caseValue <= upperRangeBoundary : caseValue < upperRangeBoundary
That way you don't have problems when values are very close to an upper boundary.

bfinzer (approved):
👍🏻LGTM with one small suggestion

emcelroy:
The latest changes implement Bill's suggestion for changing the comparison in `casesInRange`, and ensure the counts' font size auto-adjusts to fit smaller bin sizes. Regarding the latter, there appears to be a bug with the bin calculation that can sometimes make it seem like the count adornments aren't placed correctly. There is [a separate PT story](https://www.pivotaltracker.com/story/show/187398205) about fixing the bin calculations.

bfinzer:
I agree that getting rid of the non-applicable univariate plot options in the presence of a binned plot should be a separate story.

bfinzer (approved):
👍🏻LGTM

kswenson (approved):
👍 Changes look good

PR #1199, opened 2024-04-03
Fix slider axis missing numeric scale bug

bfinzer:
[#187333857] Bug fix: Missing numbers on Slider axis

* In `useSubAxis` the displayModel must return `true` for `hasDraggableNumericAxis` in the slider context. We fix this by having the default data display model defer to a passed in axis model.


emcelroy (approved):
Looks good -- thanks!

PR #1200, opened 2024-04-04
Use y attribute name for function plot label

bfinzer:
[#187152587] Feature: The label for a plotted function in a scatterplot uses the name of the attribute on the y-axis

* In `PlottedFunctionAdornmentBanner` we pull out the name of the first attribute on the y-axis to use as a label.
* We format so that the label is limited to just a portion of the function banner space and gets elided as needed.


emcelroy (approved):
Looks good 👍

PR #1201, opened 2024-04-04
187367138 v3 DI selectionList

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187367138

This PR makes CODAP v3 handle basic create, get, and update selectionList requests.

Collection related features are not implemented in this PR. That includes:
- Only true cases are handled, not pseudo-cases.
- `collectionName` and `collectionID` are not included in `get selectionList` responses.
- Cases are not filtered by the collection included in the API request.


v3/src/data-interactive/handlers/selection-list-handler.ts
```diff
@@ -0,0 +1,50 @@
+import { IDataSet } from "../../models/data/data-set"
+import { t } from "../../utilities/translation/translate"
+import { registerDIHandler } from "../data-interactive-handler"
+import { DIHandler, DIResources, DIValues } from "../data-interactive-types"
+
+const unknownDataContextResult =
+  { success: false, values: { error: t("V3.DI.Error.dataContextUnknown")} } as const
```
kswenson, line 7: Couldn't we just use `V3.DI.Error.dataContextNotFound` here?

kswenson (approved):
👍 LGTM -- just one thought/suggestion

v3/src/data-interactive/handlers/selection-list-handler.ts
tealefristoe: I was staying true to v2 here, but I agree that it would be better to be consistent with error messages. I'll go ahead and switch to `dataContextNotFound`.

PR #1202, opened 2024-04-04
Fix V2 map import bug affecting adding a legend

bfinzer:
[#187374319] Bug fix: Imported V2 maps have a problem when adding a legend

* In v2-map-importer.ts for point and polygon layers pull out the respective lat/long or polygon attribute IDs and assign them to the data configuration model for the layer
* v2MapImporter now properly imports visibility of grid, points and connecting lines. gridMultiplier is properly imported
* Also fixed a bug in MapPointLayer whereby if a map layer gets imported with points not visible, the toggle to make them visible was not working. The fix is to always call `refreshPoints` when point visibility becomes true. Though this was detected by importing from V2, it was also true of V3 restore.

### DIFF
--- a/v3/src/components/map/v2-map-importer.ts
+++ b/v3/src/components/map/v2-map-importer.ts
@@ -44,6 +44,7 @@ export function v2MapImporter({v2Component, v2Document, insertTile}: V2TileImpor
 */
       } = v2LayerModel,
       v3LegendType = v3TypeFromV2TypeIndex[legendAttributeType]
+    if (!data?.dataSet) return
 
     if (legendAttributeId) {
       _attributeDescriptions.legend = {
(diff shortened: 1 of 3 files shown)


v3/src/components/map/v2-map-importer.ts
```diff
@@ -52,12 +55,15 @@ export function v2MapImporter({v2Component, v2Document, insertTile}: V2TileImpor
...
+        grid, pointsShouldBeVisible, connectingLines
         /* Present in v2 layer model but not yet used in V3 layer model:
-        transparency, strokeTransparency, pointsShouldBeVisible,
-        grid, connectingLines,
+        transparency, strokeTransparency
         */
       } = v2LayerModel
-
+      // V2 point layers don't store their lat/long attributes, so we need to find them in the dataset
+      const {latId, longId} = latLongAttributesFromDataSet(data?.dataSet as IDataSet)
```
kswenson, line 64: Rather than casting here, I would add a test for `data?.dataSet` up above, i.e. something like:
```typescript
    // if we don't have a data set, then we can't add the layer
    if (!data?.dataSet) return
```
and then you shouldn't need the cast here or in line 100 below.

kswenson (changes requested):
👍 Looks good -- just one suggested change.

v3/src/components/map/v2-map-importer.ts
bfinzer: Good idea!

kswenson (approved):
👍 Changes look good

PR #1203, opened 2024-04-05
fix: graph update on replace of data set

kswenson:
[[PT-187348852]](https://www.pivotaltracker.com/story/show/187348852)

The `Graph` component was calling `GraphController.callMatchCirclesToData` on changes to the _number_ of `filteredCases`, but not when the contents of the `filteredCases` array changed without a corresponding change to its length. With this PR we correct that oversight.


bfinzer (approved):
👍🏻LGTM
Nice fix! And now I know how to detect a change in the contents of an array.

PR #1204, opened 2024-04-05
Fix map grid bug

bfinzer:
[#187381311] Bug fix: Map grid unable to handle us-cities dataset

* The problem is that in this dataset the "westernmost" city is across the international date line and thus has a positive longitude which interferes with computation of the "rectangle" that contains the points.
* This bug does not appear in V2. Careful comparison of the map-utils `getLatLongBounds` reveals that in V2 we pass a comparison routine to the sort function when sorting longitudes. Doing the same in V3 fixes the bug.


v3/src/components/map/utilities/map-utils.ts
```diff
@@ -70,7 +70,7 @@ export const boundaryAttributeFromDataSet = (dataSet: IDataSet) => {
  */
 export const getLatLongBounds = (dataConfiguration: IDataConfigurationModel) => {
   const getRationalLongitudeBounds = (longs: number[]) => {
-      longs.sort()
+      longs.sort((v1:number, v2:number) => (v1 - v2))
```
kswenson, line 73:
```suggestion
      longs.sort((v1: number, v2: number) => (v1 - v2))
```
would be the more conventional formatting. I had to look it up to remind myself why this fixes the bug -- the default behavior of `sort()` is to convert the contents to strings and sort them alphabetically. 🤷

kswenson (approved):
👍 LGTM -- just a minor formatting suggestion

PR #1206, opened 2024-04-08
chore: split ChartDots into bar and points components (PT-181915881)

emcelroy:
[[#181915881](https://www.pivotaltracker.com/story/show/181915881)]

As a follow up to [feat: dot chart can be converted to bar chart (PT-181915881)](https://github.com/concord-consortium/codap/pull/1196), these changes split the original `ChartDots` component into separate versions for bars and points. I moved portions of the original component to a `useChartDots` hook and to two new views on `GraphContentModel` to minimize duplicate code. One thing I'm especially wondering is if it'd be more appropriate to put one or both of those on `GraphDataConfigurationModel` instead? I'm also not quite sure about the names I chose for the views. Are they too generic in the context of the graph model? Similarly, the names of the types I added to graphing-types.ts may be too generic.


v3/src/components/graph/hooks/use-plot.ts
```diff
@@ -98,7 +98,7 @@ export const usePlotResponders = (props: IPlotResponderProps) => {
       () => GraphAttrRoles.map((aRole) => dataConfiguration?.attributeID(aRole)),
       () => {
         // if plot is not univariate and the attribute type changes, we need to update the pointConfig
-        if (graphModel?.plotType !== "dotPlot") {
+        if (graphModel?.plotType !== "dotPlot" && !graphModel?.pointsFusedIntoBars) {
```
emcelroy, line 101: This is a bug fix. Without this change, points fused into bars would automatically unfuse back to points if you added a categorical split to the graph.

bfinzer:
I notice one bug: Hover tips are not hiding when the mouse is no longer over a bar.

v3/src/components/graph/components/barchartdots.tsx
```diff
@@ -0,0 +1,184 @@
...
+    .data(barCovers)
+    .join((enter) => enter.append("rect")
+      .attr("class", (d) => d.class)
+      .attr("data-testid", "bar-cover")
+      .attr("x", (d) => d.x)
+      .attr("y", (d) => d.y)
+      .attr("width", (d) => d.width)
+      .attr("height", (d) => d.height)
+      .on("mouseover", function() { select(this).classed("active", true) })
+      .on("mouseout", function() { select(this).classed("active", false) })
```
bfinzer, line 47: For some reason this isn't causing the datatip to disappear.

bfinzer:
Both `cellParams` and `mapOfIndicesByCase` seem out of place to me in the GraphContentModel. I think of the models as knowing nothing about display level stuff. How about moving them to a utility file, perhaps named graph-display-utils?

bfinzer (approved):
👍🏻LGTM
Just one bug that could become a new story plus a suggestion about moving display level code out of GraphContentModel.

emcelroy:
> Both `cellParams` and `mapOfIndicesByCase` seem out of place to me in the GraphContentModel. I think of the models as knowing nothing about display level stuff. How about moving them to a utility file, perhaps named graph-display-utils?

Or maybe they could go in the `useChartDots` hook? They were originally part of the `ChartDots` component's `refreshPointPositions` method which is now in `useChartDots`. They would need to be separate from the latter since they're used directly by `BarChartDots`. But I think it may work to have them defined separately within the hook.

v3/src/components/graph/components/barchartdots.tsx
emcelroy: I meant to change this but didn't: currently the individual PixiPoint bars are still handling the showing and hiding of data tips. It should be the bar covers that do so. I'll need to rework things a little. Talya also pointed out that the click handling wasn't working right when there's a legend present and the bars are segmented by legend values. So I'll also have to make it so there's a bar cover per bar segment when there's a legend.

v3/src/components/data-display/models/data-configuration-model.ts
```diff
@@ -338,7 +338,7 @@ export const DataConfigurationModel = types
 
       selectCasesForCategoryValues(
         primaryAttrRole: AttrRole, primaryValue: string, secondaryValue?: string, extraPrimaryValue?: string,
-        extraSecondaryValue?: string, extend = false
+        extraSecondaryValue?: string, legendCat?: string, extend = false
```
kswenson, line 341: Could rename these to `primarySplitValue` and `secondarySplitValue` if you felt motivated but not required. 😉

v3/src/components/data-display/models/data-configuration-model.ts
```diff
@@ -357,7 +357,9 @@ export const DataConfigurationModel = types
               (extraPrimaryValue === "__main__" ||
                 dataset?.getStrValue(aCaseData.caseID, extraPrimaryAttrID) === extraPrimaryValue) &&
               (extraSecondaryValue === "__main__" ||
-                dataset?.getStrValue(aCaseData.caseID, extraSecondaryAttrID) === extraSecondaryValue)
+                dataset?.getStrValue(aCaseData.caseID, extraSecondaryAttrID) === extraSecondaryValue) &&
+              (!legendCat || legendCat === "" ||
+                dataset?.getStrValue(aCaseData.caseID, self.attributeID("legend")) === legendCat)
```
kswenson, line 362:
```suggestion
              (!legendCat || dataset?.getStrValue(aCaseData.caseID, self.attributeID("legend")) === legendCat)
```
Explicit test for empty string is redundant.

v3/src/components/graph/models/graph-content-model.ts
```diff
@@ -351,15 +350,108 @@ export const GraphContentModel = DataDisplayContentModel
         : casePrimaryValue
       const firstCount = legendAttrID ? totalCases : casesInSubPlot
       const secondCount = legendAttrID ? casesInSubPlot : totalCases
-
+      const percent = float(firstCount / secondCount * 100)
```
kswenson, line 353: I would write this as:
```suggestion
      const percent = float(100 * firstCount / secondCount)
```
to avoid the reader mistakenly thinking that the 100 is applied to the denominator (which it isn't, but visually it looks like it might be).

v3/src/components/graph/models/graph-content-model.ts
```diff
@@ -351,15 +350,108 @@ export const GraphContentModel = DataDisplayContentModel
...
       // <n> of <m> <category> (<p>%) are <legend category>
       const attrArray = [
         firstCount, secondCount, caseCategoryString, percent, caseLegendCategoryString
       ]
-
       return t("DG.BarChartModel.cellTipPlural", {vars: attrArray})
     }
   }))
+  .views(self => ({
+    cellParams(primaryCellWidth: number, primaryHeight: number) {
```
kswenson, line 362: In an earlier comment, the question was asked about where methods like `cellParams` and `mapOfIndicesByCase` should live. Perhaps they would make sense in the `SubPlotCells` class? Not a requirement for this PR, but something to consider.

v3/src/components/graph/components/dot-chart-bars.tsx
```diff
@@ -0,0 +1,234 @@
...
+  const {
+    primaryAxisPlace, primaryIsBottom, primarySplitAxisPlace, primarySplitCellWidth,
+    secondaryAxisPlace, secondaryCellHeight, numSecondarySplitBands
+  } = subPlotCells
+  const { p: primeCatIndex, ep: primeSplitCatIndex, es: secSplitCatIndex } = cellIndices
+  const primarySplitAxisScale = layout.getAxisScale(primarySplitAxisPlace) as ScaleBand<string>
+  const secondaryAxisScale = layout.getAxisScale(secondaryAxisPlace) as ScaleLinear<number, number>
+  const numPrimarySplitBands = Math.max(1, primarySplitAxisScale?.domain().length ?? 1)
+  const primaryCellWidth = layout.getAxisLength(primaryAxisPlace) / (primCatsCount ?? 1)
+  const primarySubCellWidth = primaryCellWidth / numPrimarySplitBands
```
kswenson, line 62: Some of these are either already available in the `SubPlotCells` or could be made available there. Again, not required for this PR but something to consider moving forward.

kswenson (approved):
👍 Looks good -- a few small suggestions and a couple of potential further small refactors to consider but that needn't be required at this point.

v3/src/components/graph/components/barchartdots.tsx
emcelroy: As noted offline, there is now a [separate PT story for this bug](https://www.pivotaltracker.com/story/show/187409553).

v3/src/components/graph/models/graph-content-model.ts
emcelroy: I tried moving those but ran into some confusing problems. It's not clear that they'd be easy to work out, so I'm going to opt out of that for this PR.

v3/src/components/graph/components/dot-chart-bars.tsx
emcelroy: Thanks for the reminder about that. I've cleaned this part up a bit as suggested.

PR #1207, opened 2024-04-09
187263404 v3 DI Notify Slider Animation

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187263404

This PR adds notifications for changing a global's value via the slider play animation button.

Additionally it allows the notification message to be determined with a callback function, rather than being specified at the time `applyUndoableAction` is called. This is sometimes necessary for an accurate message, and is almost always a more natural way of writing the message, so the client can access the state of the model after the action has been performed rather than before.


v3/src/components/slider/use-slider-animation.tsx
```diff
@@ -75,7 +95,8 @@ export const useSliderAnimation = ({sliderModel, running, setRunning}: IUseSlide
     const belowMin = (val: number) => {
       return axisMin
     }
-    sliderModel?.setValue(sliderModel.validateValue(newValue, belowMin, aboveMax))
+    updateSlider(newValue, belowMin, aboveMax)
+    // sliderModel?.setValue(sliderModel.validateValue(newValue, belowMin, aboveMax))
```
kswenson, line 99:
```suggestion
```
You can remove the commented out code now, here and below.

kswenson (approved):
👍 Looks good -- just some commented out code to remove.

PR #1210, opened 2024-04-10
Save/restore calculator position between invocations

bfinzer:
[#187391638] Feature: Calculator position is saved and restored on close and reopen

Note that this PR introduces a bug whereby once the calculator has been closed and opened again it can no longer be repositioned.

* Pass an `isHidden` property to `CodapComponent` so that singleton components and the case table/card pair can be hidden but retain their state.
* Include `isHidden` as a property of `FreeTileLayout`
* In `DocumentContent` when toggling visibility of a singleton, don't delete it, just mark it as hidden
* Added getFreeTileLayoutById to `BaseDocumentContent`
* In `CalculatorTitleBar` override the defaault `onCloseTile` function to instead call toggleSingletonTileVisibility

### DIFF
--- a/v3/src/components/calculator/calculator-title-bar.tsx
+++ b/v3/src/components/calculator/calculator-title-bar.tsx
@@ -1,12 +1,19 @@
-import React from "react"
-import { observer } from "mobx-react-lite"
-import { ComponentTitleBar  } from "../component-title-bar"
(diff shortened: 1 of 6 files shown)


kswenson (approved):
👍 Looks good after I added a commit and sync review with @bfinzer.

PR #1212, opened 2024-04-11
Open/close case table preserves position & extent

bfinzer:
[#187185393] Feature: Case table position is saved and restored on close and reopen

**Chore**
* Move `columnWidths` from `SharedCaseMetadata` to `CaseTableModel`. (Not actually necessary for this feature, but it was bothering me.) This necessitated changes in case-table-registration.ts and codap-v2-document.ts to change import from v2. Then, of course, places where we were getting column widths from `SharedCaseMetadata` we now get from the `CaseTableModel`.
* I discovered that in the previous PR for the Calculator hide/show it was the `CodapComponent` whose rendering was being hidden, it should actually be the `FreeTileComponent` whose rendering should be determined.

**Feature**
* Added some undo/redo strings for hiding/showing the case table
* `CaseTableTitleBar` now has its own method for `onCloseTile`
* `SharedCaseMetadata` Now has properties `caseTableTileId` and `caseCardTileId`
* The Table menu in the toolshelf now does a bit more work to make sure any newly created dataset's case table is hooked up properly and, if a case table already exists make sure that it is visible and has focus.
* The `linkTileToDataSet` function now takes a `TileModel` instead of a `TileContentModel`.

**Note**: This PR breaks the DataSummary because `DataSummaryModel.inspect` calls `linkTileToDataSet` and doesn't have a TileModel to offer. A subsequent PR will remove the DataSummary entirely from the codebase.

### DIFF
--- a/v3/src/components/case-table/case-table-model.ts
+++ b/v3/src/components/case-table/case-table-model.ts
@@ -1,5 +1,4 @@
 import { Instance, SnapshotIn, types } from "mobx-state-tree"
-import { applyUndoableAction } from "../../models/history/apply-undoable-action"
 import { getTileCaseMetadata, getTileDataSet } from "../../models/shared/shared-data-utils"
 import { ISharedModel } from "../../models/shared/shared-model"
 import { ITileContentModel, TileContentModel } from "../../models/tiles/tile-content"
@@ -58,7 +57,6 @@ export const CaseTableModel = TileContentModel
       self.scrollLeft = scrollLeft
     }
   }))
-  .actions(applyUndoableAction)
 export interface ICaseTableModel extends Instance<typeof CaseTableModel> {}
 export interface ICaseTableSnapshot extends SnapshotIn<typeof CaseTableModel> {}
 
--- a/v3/src/components/case-table/case-table-model.ts
+++ b/v3/src/components/case-table/case-table-model.ts
@@ -1,4 +1,5 @@
 import { Instance, SnapshotIn, types } from "mobx-state-tree"
+import { applyUndoableAction } from "../../models/history/apply-undoable-action"
 import { getTileCaseMetadata, getTileDataSet } from "../../models/shared/shared-data-utils"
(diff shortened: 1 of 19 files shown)


v3/src/components/case-table/case-table-model.ts
```diff
@@ -37,13 +43,22 @@ export const CaseTableModel = TileContentModel
...
+      }
+    },
     updateAfterSharedModelChanges(sharedModel?: ISharedModel) {
       // TODO
     },
     setScrollLeft(scrollLeft: number) {
       self.scrollLeft = scrollLeft
     }
   }))
+  .actions(applyUndoableAction)
```
kswenson, line 61: This should be added to the `TileContentModel` (I'm surprised it hasn't already been added there) so that it's present for all tiles.
```suggestion
```

v3/src/utilities/translation/lang/en-US.json5
```diff
@@ -33,6 +33,10 @@
     "V3.Redo.caseTable.create": "Redo adding case table",
     "V3.Undo.caseTable.delete": "Undo deleting case table",
     "V3.Redo.caseTable.delete": "Redo deleting case table",
+    "V3.Undo.caseTable.show": "Undo showing case table",
+    "V3.Redo.caseTable.show": "Redo showing case table",
+    "V3.Undo.caseTable.hide": "Undo hiding case table",
+    "V3.Redo.caseTable.hide": "Redo hiding case table",
```
kswenson, line 39: V2 doesn't have strings for these?

kswenson (approved):
👍 Looks good -- just one suggestion to move `applyUndoableAction` to the `TileContentModel`.

v3/src/utilities/translation/lang/en-US.json5
bfinzer: Turns out it had the strings for showing but not for hiding.

PR #1213, opened 2024-04-12
feat: show color swatch when editing color values

kswenson:
- "Strict" color values (e.g. `#FF00FF` or `rgb(0, 255, 255)`) display as a color swatch in the case table and show a color swatch when editing _unless_ the user has assigned a non-color type to the attribute.
- Color names (e.g. `red`, `green`, `blue`) display as a color swatch in the case table if the user has assigned a color type  to the attribute.
- During editing, color swatch updates color dynamically and becomes blank if the value is not a valid color.
- Decision to show the swatch or not during editing is made based on the initial value at the start of editing, i.e. the space allocated for the swatch does not appear or disappear during editing.

### DIFF
--- a/v3/src/components/case-table/case-table.scss
+++ b/v3/src/components/case-table/case-table.scss
@@ -271,6 +271,23 @@ $table-body-font-size: 8pt;
         }
       }
 
+      .color-cell-text-editor {
+        width: 100%;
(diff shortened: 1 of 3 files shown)


bfinzer (approved):
👍🏻LGTM

PR #1214, opened 2024-04-12
187371422 v3 DI Create and Update Case

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187371422

This PR adds support for `create` and `update case` requests from plugins in v3.

`create case` is not completely implemented in this PR, because the collection system in v3 is undergoing a fairly major overhaul. So the returned values from `create case` lack case ids for now (these should be included as `id`). Item ids are returned as expected.


kswenson:
Having looked at the code and thought about it some more, I think you should treat this PR as just handling flat data sets so you don’t need to deal with collections or hierarchy at all. I’m going to work on a PR to clean up some of the handling of collections and cases but you shouldn’t get bogged down.

kswenson (approved):
👍 LGTM

PR #1215, opened 2024-04-12
fix: GraphContentModel shouldn't save/restore NaN values for _binAlignment and _binWidth (PT-187423678)

kswenson:
[[PT-187423678]](https://www.pivotaltracker.com/story/show/187423678)

The original bug report attributed the problem to the hierarchical data but hierarchy is not implicated. The problem is that the `GraphContentModel` occasionally stored `NaN` in its `_binAlignment` and `_binWidth` properties. `NaN`s are turned into `null`s by `JSON.stringify` which then fail to load because `null` is not a valid value for a property typed as `types.number` in MST. The fix is to prevent `NaN`s from being stored in the first place and to add a `preProcessSnapshot` handler to deal with any existing documents.


emcelroy (approved):
Looks good 👍

PR #1216, opened 2024-04-13
chore: add `attributesMap` to `DataSet`

kswenson:
[[PT-187425753]](https://www.pivotaltracker.com/story/show/187425753)

Step one of a set of stories that are necessary to improve the `DataSet`'s handling of hierarchical collections. One of the requirements of handling hierarchical collections is that ordering of attributes be synchronized between the collections and the `DataSet` itself. This is currently made more complex than necessary due to the fact that `DataSet`s store their attribute models in an MST array. The usual process for changing the order of items in an array is to remove the item from its current location and then insert it in its new location. Unfortunately, when MST modes are removed from an array MST destroys the model. Previously, the workaround for this was to use `sort()` to reorder the attributes in the array and then use a custom undo/redo implementation.

With this PR, attribute models are stored in an MST map, and the `attributes` array, which previously stored the attribute models, now is simply an array of MST `reference`s. This means that reordering items in the array can use the usual remove-then-insert mechanism because removing a reference doesn't affect the underlying model. Furthermore, the standard MST patch-generation mechanism generates efficient patches in this case, so our custom undo/redo implementation is no longer required.


v3/src/models/data/data-set.ts
```diff
@@ -696,7 +711,7 @@ export const DataSet = V2Model.named("DataSet").props({
     if (index == null) { return }
     for (const key in caseValues) {
       if (key !== "__id__") {
-        const attribute = self.attrIDMap.get(key)
+        const attribute = self.attributesMap.get(key)
```
tealefristoe, line 714:
```suggestion
        const attribute = self.getAttribute(key)
```

v3/src/models/data/data-set.ts
```diff
@@ -79,7 +84,7 @@ export function fromCanonicalCase(ds: IDataSet, canonical: ICase | ICaseCreation
   for (const id in canonical) {
     if (id !== "__id__") {
       // if we can't find a name, just use the id
-      const name = ds.attrFromID(id)?.name || id
+      const name = ds.attributesMap.get(id)?.name || id
```
tealefristoe, line 87:
```suggestion
      const name = ds.getAttribute(id)?.name || id
```

v3/src/models/data/data-set.ts
```diff
@@ -710,12 +725,14 @@ export const DataSet = V2Model.named("DataSet").props({
      * public views
      */
     views: {
+      // [DEPRECATED] use getAttribute() instead
       attrFromID(id: string) {
-        return self.attrIDMap.get(id)
+        return self.attributesMap.get(id)
```
tealefristoe, line 730:
```suggestion
        return self.getAttribute(id)
```

v3/src/models/data/data-set.ts
```diff
@@ -710,12 +725,14 @@ export const DataSet = V2Model.named("DataSet").props({
...
       },
+      // [DEPRECATED] use getAttributeByName() instead
       attrFromName(name: string) {
         const id = self.attrNameMap[name]
-        return id ? self.attrIDMap.get(id) : undefined
+        return id ? self.attributesMap.get(id) : undefined
```
tealefristoe, line 735:
```suggestion
        return id ? self.getAttribute(id) : undefined
```

v3/src/models/data/data-set.ts
```diff
@@ -747,7 +764,7 @@ export const DataSet = V2Model.named("DataSet").props({
               return cachedCase[attributeID]
             }
           }
-          const attr = self.attrIDMap.get(attributeID)
+          const attr = self.attributesMap.get(attributeID)
```
tealefristoe, line 767:
```suggestion
          const attr = self.getAttribute(attributeID)
```

v3/src/models/data/data-set.ts
```diff
@@ -767,7 +784,7 @@ export const DataSet = V2Model.named("DataSet").props({
             return cachedCase[attributeID]?.toString()
           }
         }
-        const attr = self.attrIDMap.get(attributeID)
+        const attr = self.attributesMap.get(attributeID)
```
tealefristoe, line 787:
```suggestion
        const attr = self.getAttribute(attributeID)
```

v3/src/models/data/data-set.ts
```diff
@@ -787,7 +804,7 @@ export const DataSet = V2Model.named("DataSet").props({
             return Number(cachedCase[attributeID])
           }
         }
-        const attr = self.attrIDMap.get(attributeID)
+        const attr = self.attributesMap.get(attributeID)
```
tealefristoe, line 807:
```suggestion
        const attr = self.getAttribute(attributeID)
```

v3/src/models/data/data-set.ts
```diff
@@ -907,18 +931,17 @@ export const DataSet = V2Model.named("DataSet").props({
       },
 
       setAttributeName(attributeID: string, name: string | (() => string)) {
-        const attribute = attributeID && self.attrIDMap.get(attributeID)
+        const attribute = attributeID && self.attributesMap.get(attributeID)
```
tealefristoe, line 934:
```suggestion
        const attribute = attributeID && self.getAttribute(attributeID)
```

v3/src/models/data/data-set.ts
```diff
@@ -907,18 +931,17 @@ export const DataSet = V2Model.named("DataSet").props({
...
         if (attribute) {
           const nameStr = typeof name === "string" ? name : name()
           attribute.setName(nameStr)
         }
       },
 
       removeAttribute(attributeID: string) {
-        const attrIndex = self.attrIndexFromID(attributeID),
-              attribute = attributeID ? self.attrIDMap.get(attributeID) : undefined
+        const attribute = self.attributesMap.get(attributeID)
```
tealefristoe, line 942:
```suggestion
        const attribute = self.getAttribute(attributeID)
```

tealefristoe (approved):
Looks great! I just suggest using `dataset.getAttribute()` instead of digging into the `attributeMap` in a bunch of places.

PR #1217, opened 2024-04-16
Remove DataSummary component

bfinzer:
[#187444366] Feature: Remove the data summary component

Because we can save and restore documents, ~~we no longer need the dashboard capability~~, and, it turns out that the DataSummary component is getting in the way of implementing changes to the way case tables are shown and hidden.

* ~~Remove reference to a "dashboard" url parameter~~
* Remove the `DataSummary` component entirely
* ~~Remove `createStarterDataset` which was only used for the dashboard scenario~~


kswenson (approved):
👍 Looks good -- it turns out that the dashboard mode is used by cypress tests, so I added a commit which restored the `dashboard` mode but rearranged the components now that the data summary is gone.

PR #1218, opened 2024-04-17
187440921 v3 DI dataContext

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187440921

This PR adds support for `create`, `delete`, `get` and `update dataContext` requests from plugins in v3.


kswenson (commented):
👍 Looks good -- I added a commit to improve some of the TypeScript types.

PR #1220, opened 2024-04-18
Empty case card appears when toggled from table

bfinzer:
[#187381638] Feature: A basic Case Card component can be made to appear

* Enough of CaseCard component implemented to show an empty frame with the appropriate title bar
* Mechanisms put in place to keep track of the IDs of case card/table tiles and which one was last shown
* Title bar component for both card and table in new case-table-card-title-bar.tsx
* `DocumentContent:toggleCardTable` is new method to handle moving back and forth


kswenson (approved):
👍 LGTM -- I added a commit to clean up a bit.

kswenson:
Merging after adding a cleanup commit on review.

PR #1221, opened 2024-04-18
fix: % values in data tip incorrect when legend present (PT-181915881)

emcelroy:
[[#181915881](https://www.pivotaltracker.com/story/show/181915881)]

Previously when a legend was present and there was a split of the primary axis of a bar chart, the percent values shown in the bar segment data tips were incorrect. They would have values like `86 of 61 are sitdown (141%)`, for example. That was because the total number of cases matching a given legend value across the _entire_ graph plot was being used instead of just the matches inside the subplot. These changes correct that.


bfinzer (approved):
👍🏻LGTM

kswenson (approved):
👍 LGTM

PR #1222, opened 2024-04-18
Plotted function modal uses y-attribute name

bfinzer:
[#187401360] Bug fix: Function Label Does Not Update to Variable Name in Graph Plot

* Pass y-attribute name to `EditFormulaModal`


emcelroy (approved):
Looks good 👍

PR #1223, opened 2024-04-19
feat: when editing color value in case table, click on swatch invokes color palette

kswenson:
[[PT-187020075]](https://www.pivotaltracker.com/story/show/187020075)

- Clicking on swatch brings up color picker.
- Color picker allows selection of color, including opacity.
- Changes can be accepted by `enter` key, clicking `Set Color` button, or clicking outside color picker.
- Changes can be rejected by `escape` key or clicking `Cancel` button.
- Named colors are converted to hex colors when edited.

### DIFF
--- a/v3/cypress/e2e/table.spec.ts
+++ b/v3/cypress/e2e/table.spec.ts
@@ -641,4 +641,44 @@ context("case table ui", () => {
       })
     })
   })
+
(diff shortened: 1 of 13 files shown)


emcelroy (approved):
Looks good 👍

PR #1224, opened 2024-04-20
185315409 automation add new attribute

nstclair-cc:
[PT 185315409](https://www.pivotaltracker.com/story/show/185315409)

This is a simple automation test to check if a table creates a new attribute with undo/redo.


kswenson (commented):
👍 LGTM

PR #1225, opened 2024-04-22
187458550 v3 DI Initial Component Requests

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187458550

This PR makes CODAP v3 handle several API requests related to components.
- `get componentList` returns a list of information about all components in the document.
    - Currently, names are not included in this information, because components don’t have names in v3. It would be possible to return the id as the name, and I think this would allow plugins to reference components in the same way. However, this wouldn’t work if names are specified by plugins when they create components.
- `delete component` deletes a specified component (currently an id must be used to identify the component; names are not supported).
- `notify component` works for the `request: select`. This allows a plugin to make a component focused.
    - `request: autoScale` has not be implemented, because I don’t think functions to scale components exist currently.

I was hoping to complete all component API requests in this PR, but it quickly became clear that `create`, `get`, and `update` are quite complex with special code required for each component type. Since none of these requests are required for Choosy, I decided to skip them for now. It also seemed that even if I had started working on any of these requests, I wouldn’t have been able to finish them because some components haven’t been implemented yet, and it would be better to leave them unstarted rather than incomplete.


kswenson (approved):
👍 Looks good -- I rebased to deal with conflicts I introduced with my additional commit to the previous PR. I'll merge once the tests run.

PR #1226, opened 2024-04-22
feat: color table rows if first attribute of collection is color attribute

kswenson:
[[PT-187020220]](https://www.pivotaltracker.com/story/show/187020220)

If the first/left-most attribute of a collection is user-assigned the "color" type, then color the background of each row that color. For legibility, we use an alpha value of 10%.


emcelroy (approved):
Looks good 👍

PR #1231, opened 2024-04-25
187485557 v3 DI Select Case and Update DataContext Nofications

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187485557

This PR adds `updateDataContext` and `updateSelection` notifications to CODAP v3. These are the final two notifications the Choosy plugin expects to receive from CODAP.

`updateDataContext` is broadcast when the user changes a table title or when a dataset's properties are changed using the table's i button modal.
- Note that the table title UI has the same issue as the collection title UI, namely that the model is changed on every keystroke. This means that you'll currently see notifications broadcast on every keystroke until this bug is fixed.

`updateSelection` proved to be one of the more difficult notifications to broadcast, because it's possible to select cases in many different ways scattered all over the codebase. Much of this PR involves replacing these with utility function calls which broadcast a notification in addition to modifying the selection. These changes sometimes required refactoring code, for example, to move selection modification out of models and into components.

Problems
- There are many ways to modify the selection, but I only added a short cypress test for clicking on table cells.
- Adding notifications to selection modification in `categorical-legend.tsx` and `data-display/components/background.tsx` was causing the `graph-legend` cypress test to fail. I couldn't figure out why, so I left comments saying "This is breaking the graph-legend cypress test" with the correct code commented out. So these places are currently _not_ broadcasting notifications like they should.
- Additionally, I believe there was a change that was breaking the `map` cypress test, but that seems to be working now. However, I think it's still broken, but the problem test has been turned off for now. So don't be surprised if there is an issue related to broadcasting notifications when those tests are turned back on.
- Should `clearGridCells` in `map-grid-model.ts` be an action or a view? It modifies a non-observable object saved as a model's volatile property.


kswenson (approved):
👍 Looks good! I like the new utility functions and the separation of responsibility for selection out of the models.

>Should `clearGridCells` in `map-grid-model.ts` be an action or a view? It modifies a non-observable object saved as a model's volatile property.

There isn't actually a `clearGridCells` method, but I assume you're referring to methods like `selectCasesInRect` that modify the contents of the `latLngGrid` property. A couple of points:
- `volatile` properties _are_ actually observable. They're just not serialized.
- MST will actually warn if you assign to `volatile` properties outside an action.
- For complex properties like objects and arrays, MST doesn't track their internals, so it doesn't warn if you only modify the internals of `volatile` properties.

In this particular case, I'm fine with them being `views` because they're not changing serialized state and they're not undoable, so there's really no reason to spin up the action-tracking machinery.

PR #1232, opened 2024-04-25
187458550 v3 DI Delete Component Hides Singletons

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187458550

@nstclair-cc pointed out that `delete component` on singletons (like the calculator) should not actually delete the tile, but instead should be hide it, so it reappears in the same location when recreated. This PR fixes that issue by hiding singletons rather than deleting them from the document.

It's worth noting that this is not how v2 works. But v3 also differs from v2 in that closing a singleton will actually delete it in v2, while it just hides the singleton in v3.

This PR also adds the `hidden` field to the response to `get componentList` requests.

### DIFF
--- a/v3/src/models/document/document-content.ts
+++ b/v3/src/models/document/document-content.ts
@@ -185,20 +185,18 @@ export const DocumentContentModel = BaseDocumentContentModel
         }
       }
     },
+    // Hide the tile if it's a singleton and can be hidden. Otherwise, delete it.
     deleteOrHideTile(tileId: string) {
       const tile = self.getTile(tileId)
-      if (tile) {
-        const tileType = tile.content.type
-        const tileInfo = getTileContentInfo(tileType)
-        if (tileInfo?.isSingleton) {
-          const tileLayout = self.getTileLayoutById(tileId)
-          if (isFreeTileLayout(tileLayout)) {
-            tileLayout.setHidden(true)
-          }
-        } else {
-          self.deleteTile(tileId)
+      const tileInfo = getTileContentInfo(tile?.content.type)
+      if (tileInfo?.isSingleton) {
+        const tileLayout = self.getTileLayoutById(tileId)
+        if (isFreeTileLayout(tileLayout)) {
+          tileLayout.setHidden(true)
+          return
         }
       }
+      self.deleteTile(tileId)
     }
   }))
   .actions(self => ({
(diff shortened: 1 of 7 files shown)


v3/src/models/document/document-content.ts
```diff
@@ -184,6 +184,19 @@ export const DocumentContentModel = BaseDocumentContentModel
           return self.createTile(tileType, options)
         }
       }
+    },
+    // Hide the tile if it's a singleton and can be hidden. Otherwise, delete it.
+    deleteOrHideTile(tileId: string) {
+      const tile = self.getTile(tileId)
+      const tileInfo = getTileContentInfo(tile?.content.type)
+      if (tileInfo?.isSingleton) {
```
kswenson, line 192: I think we need another property here because the case table and data card tiles should also hide, but they're not technically singletons. I'm thinking there should be a `hideOnClose` property on `TileContentInfo` and then this would be
```suggestion
      if (tileInfo?.isSingleton || tileInfo?.hideOnClose) {
```

kswenson (changes requested):
👍 Looks good -- one suggestion to add an additional property (`hideOnClose`) to handle tiles that hide/show but aren't singletons.

v3/src/models/document/document-content.ts
tealefristoe: Good call 👍  I added the `hideOnClose` property and set it to true for the table and data card. I thought about going through the code and checking this flag whenever `toggleNonDestroyableTileVisibility()` is called, but that seemed outside the scope of this story.

This generally works, but it does seem to fail when the table is part of the loaded document. For example, deleting the table in mammals will hide it, but then showing the table will create a new one rather than showing the hidden one. If you then "delete" the new one and show the table, it will correctly show the 2nd table. I haven't dug into why the first table is being ignored after it has been hidden. Do you think I should do that?

kswenson: >I haven't dug into why the first table is being ignored after it has been hidden. Do you think I should do that?

I think it's worth a quick look because it suggests there's something we don't understand here that might affect the implementation. Once we know what's going on, I suspect the fix will be straightforward.

tealefristoe: Alright, looking into this more, I think the issue is that I was using an older document I had saved a couple of weeks ago. It looks like support for showing hidden tables/cards wasn't added until about half a week ago, so my document was just lacking the infrastructure needed to associate my "deleted" table with the dataset when I was showing the table again. Using a fresh document, and even saving and loading it again, results in the correct behavior.

kswenson (approved):
👍 Changes look good

PR #1233, opened 2024-04-25
staging version 3.0.0-pre.1522

nstclair-cc:
[Build 3.0.0-pre.1522](https://www.pivotaltracker.com/story/show/187502436)

Note that I "discarded" some of the changes that were in `package-lock.json` as a result of running `npm run build` locally. If this was not the correct approach, please let me know and we'll make staging changes to `package-lock.json`. 

Thanks!!

### DIFF
--- a/v3/CHANGELOG.md
+++ b/v3/CHANGELOG.md
@@ -1,4 +1,29 @@
 # Changelog
+### ✨ Features & Improvements:
+- **PT-187425753:** DataSet: add attributesMap to simplify attribute ordering
+- **PT-187391005:** In case table, show swatch when editing a color value
+- **PT-187444366:** Dashboard no longer displays DataSummary component
+- **PT-187381638:** A basic Case Card component can be made to appear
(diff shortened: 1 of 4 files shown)


kswenson (changes requested):
The build number on main is at 1620 right now. Not sure how you ended up with 1522 but that's not the right number.

nstclair-cc:
> The build number on main is at 1620 right now. Not sure how you ended up with 1522 but that's not the right number.

Hi @kswenson - how strange! I usually get the build number from the top right corner of the software and then increment by one. I also usually go back to Pivotal Tracker tags to make sure I've got the build number correct. Where do you see the build number on main at 1620?
Screenshot 2024-04-26 at 9 18 46 AM

That said, I forgot to include the build number in the file `CHANGELOG.md`, so I went ahead and made that commit.

nstclair-cc:
As discussed on Slack, I think we'll live with the incorrect build number for now and catch it on the next build.

PR #1235, opened 2024-04-27
Count adornment calls applyModelChange for setCount and setPercent

bfinzer:
[#187510616] Chore: The count adornment relies on `withUndoRedoStrings` instead of going through `applyModelChange`

* Removed dependency on `withUndoRedoStrings` in `CountAdornmentModel`
* Use `applyModelChange` when calling `setCount` and `setPercent` in count-adornment-registration.tsx and count-adornment-component.tsx

### DIFF
--- a/v3/src/components/graph/adornments/count/count-adornment-registration.tsx
+++ b/v3/src/components/graph/adornments/count/count-adornment-registration.tsx
@@ -38,7 +38,15 @@ const Controls = () => {
     const redoRemoveKey = checkBoxType === "count" ? "DG.Redo.graph.hideCount" : "DG.Redo.graph.hidePercent"
 
     const setShowAdornment = checkBoxType === "count"
-      ? () => adornment.setShowCount(checked)
+      ? () => {
+        graphModel.applyModelChange(
+          () => {
+            adornment.setShowCount(checked)
+          }, {
+            undoStringKey: undoAddKey,
+            redoStringKey: redoAddKey
+          })
+      }
       : () => handleShowPercent(adornment, checked)
 
     if (checked) {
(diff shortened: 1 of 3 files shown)


v3/src/components/graph/adornments/count/count-adornment-registration.tsx
```diff
@@ -38,7 +38,15 @@ const Controls = () => {
...
+      ? () => {
+        graphModel.applyModelChange(
+          () => {
+            adornment.setShowCount(checked)
+          }, {
+            undoStringKey: undoAddKey,
+            redoStringKey: redoAddKey
+          })
+      }
       : () => handleShowPercent(adornment, checked)
```
kswenson, line 50: I think you need to handle this case as well. Something like
```suggestion
      : () => () => {
        graphModel.applyModelChange(
          () => {
            adornment.setShowPercent(checked)
            setEnablePercentOptions(checked)
          }, {
            undoStringKey: undoAddKey,
            redoStringKey: redoAddKey
          })
      }
```

kswenson (changes requested):
👍 Looks good, but I think there's a case not handled.

v3/src/components/graph/adornments/count/count-adornment-registration.tsx
bfinzer: As it turned out I had forgotten to remove the applyModelChange for the "count" option.

kswenson (approved):
👍 LGTM -- it took me a moment to realize that removing the call to `applyModelChange` as you did made sense because the function in question is already being called from inside `applyModelChange`.

As a side note, the comment on the second commit seems to be related to some other body of work unrelated to this PR. Not worth doing anything about it here but something to keep an eye on for future efforts, perhaps when switching branches.

PR #1237, opened 2024-04-30
String function implementation

bfinzer:
[#187526964] Implement V2 string functions.

* Added two new "utils" files, one for data and one for dates. These are utilities needed for implementing string functions
* Implemented V2 string functions except for `combine` and `wordListMatches`
* Provided string-functions.test.ts

### DIFF
--- a/v3/src/models/formula/functions/string-functions.ts
+++ b/v3/src/models/formula/functions/string-functions.ts
@@ -1,0 +1,210 @@
+import { valueToString } from "../../../utilities/data-utils"
+import { FValue } from "../formula-types"
+
+export const stringFunctions = {
+  // beginsWith(text, prefix) Returns true if text begins with prefix, otherwise false.
+  beginsWith: {
+    numOfRequiredArguments: 2,
+    evaluate: (...args: FValue[]) => {
+      const text = valueToString(args[0]), prefix = valueToString(args[1])
+      if (!prefix) return true
+      if (!text) return false
+      return text.indexOf(prefix) === 0
+    }
+  },
+  // charAt(text, index) Returns the character at the specified 1-based index in the text.
+  charAt: {
+    numOfRequiredArguments: 2,
+    evaluate: (...args: FValue[]) => {
+      const text = valueToString(args[0]), index = Number(args[1])
+      return text.charAt(index - 1) // 1-based index for CODAP string functions
+    }
+  },
+  // combine(text) Returns the concatenation of text all the values of the attribute. This is an aggregate function.
+  combine: {
+    numOfRequiredArguments: 1,
+    evaluate: (...args: FValue[]) => {
+      // todo: Implement this function
+      return "NYI"
+    }
+  },
+  // concat(string1, string2, …) Returns a string that is the concatenation of the string arguments.
+  concat: {
+    numOfRequiredArguments: 2,
+    evaluate: (...args: FValue[]) => {
+      return args.map(valueToString).join("")
+    }
+  },
+  // endsWith(text, suffix) Returns true if text ends with suffix, otherwise false.
+  endsWith: {
+    numOfRequiredArguments: 2,
+    evaluate: (...args: FValue[]) => {
+      const text = valueToString(args[0]), suffix = valueToString(args[1])
+      if (!suffix) return true
+      if (!text) return false
+      return text.endsWith(suffix)
+    }
+  },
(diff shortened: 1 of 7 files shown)


v3/src/models/formula/functions/string-functions.ts
```diff
@@ -0,0 +1,210 @@
...
+    numOfRequiredArguments: 2,
+    evaluate: (...args: FValue[]) => {
+      const text = valueToString(args[0]), suffix = valueToString(args[1])
+      if (!suffix) return true
+      if (!text) return false
+      return text.endsWith(suffix)
+    }
+  },
+  // findString(stringToLookIn, stringToFind, start) Takes three arguments, returning the position of
+  // stringToFind in stringToFind starting from start. The first character of stringToLookIn is numbered 1.
```
kswenson, line 49:
```suggestion
  // stringToFind in stringToLookIn starting from start. The first character of stringToLookIn is numbered 1.
```

v3/src/models/formula/functions/string-functions.ts
```diff
@@ -0,0 +1,210 @@
...
+    }
+  },
+  // patternMatches(stringToLookIn, pattern) Returns the number of times in string_to_look_in the string or
+  // regular expression represented by pattern occurs. The search is case-insensitive.
+  patternMatches: {
+    numOfRequiredArguments: 2,
+    evaluate: (...args: FValue[]) => {
+      const stringToLookIn = valueToString(args[0]),
+        pattern = valueToString(args[1])
+      return (stringToLookIn.match(new RegExp(pattern, "gmi")) || []).length
```
kswenson, line 86: Strings, particularly user-entered strings, should always be escaped before using them as regular expressions.
```suggestion
      return (stringToLookIn.match(new RegExp(escapeStringRegexp(pattern), "gmi")) || []).length
```

v3/src/models/formula/functions/string-functions.ts
```diff
@@ -0,0 +1,210 @@
...
+  },
+  // replaceString(aString, stringToFind, substituteString) Takes three string arguments and substitutes the third
+  // for all occurrences of the second in the first.
+  replaceString: {
+    numOfRequiredArguments: 3,
+    evaluate: (...args: FValue[]) => {
+      const aString = valueToString(args[0]),
+        stringToFind = valueToString(args[1]),
+        substituteString = valueToString(args[2])
+      return aString.replace(new RegExp(stringToFind, "g"), substituteString)
```
kswenson, line 121: Strings, particularly user-entered strings, should always be escaped before using them as regular expressions.
```suggestion
      return aString.replace(new RegExp(escapeStringRegexp(stringToFind), "g"), substituteString)
```

v3/src/utilities/data-utils.ts
```diff
@@ -0,0 +1,28 @@
+import { isDate } from "mathjs"
+
+export const valueToString = function (iValue: any): string {
+  let valType:string = typeof iValue,
```
kswenson, line 4:
```suggestion
  let valType: string = typeof iValue,
```

v3/src/utilities/data-utils.ts
```diff
@@ -0,0 +1,28 @@
...
+    value = iValue
+  if (isDate(iValue)) {
+    // treat dates as strings
+    // todo: value = formatDate(iValue)
+    value = String(iValue)
+    valType = "string"
+  } else if (iValue instanceof Error) {
+    value = `${iValue.name  } ${  iValue.message}`
```
kswenson, line 12:
```suggestion
    value = `${iValue.name} ${iValue.message}`
```

kswenson (changes requested):
👍 Generally looks good. One suggestion to use `escapeStringRegexp` when converting strings to regular expressions. Beyond that, I think we'll need some additional argument validation for many of these functions. What happens when arguments that should be numbers are `NaN`s or other non-numeric values? What happens when real numbers or negative numbers are passed when indices are expected?

This would be a good opportunity to try some test-driven development (TDD), .e.g.:
- write a test for what should happen when 1.5 is passed as an index and then update the code (if necessary) to pass the test.
- write a test to search for a `.` in a string that doesn't have one. (I would expect the test to fail because the search should succeed since `.` is a wildcard in regular expressions.) Then add the call to `escapeStringRegexp` I recommended above which should allow the test to pass. 

The tests for each function could be expanded to check these edge cases and validate the arguments appropriately.

kswenson (approved):
👍 LGTM

PR #1239, opened 2024-05-02
fix: restore attributes with formulas properly

kswenson:
[[PT-187527563]](https://www.pivotaltracker.com/story/show/187527563)

We don't serialize values for attributes with formulas, instead choosing to recompute the values on restore. The result is that on restore an attribute with a formula is initialized to an empty array of values. Before the formula evaluation can update the values, we need to extend the arrays so there's a place to store the computed values.

Note that I also had to disable some unrelated map cypress tests that have proven to be unnecessarily finicky. I added a PT story [[PT-187534790]](https://www.pivotaltracker.com/story/show/187534790) to fix them at some point but didn't want to hold up this PR further.

### DIFF
--- a/v3/cypress/e2e/map.spec.ts
+++ b/v3/cypress/e2e/map.spec.ts
@@ -73,7 +73,7 @@ context("Map UI", () => {
       c.checkToolTip($element, c.tooltips.mapCameraButton)
     })
   })
-  it("checks numerical and categorical attributes for map legend", () => {
+  it.skip("checks numerical and categorical attributes for map legend", () => {
     cfm.openLocalDoc(filename2)
     c.getIconFromToolshelf("map").click()
     cy.dragAttributeToTarget("attribute", arrayOfAttributes[0], "map")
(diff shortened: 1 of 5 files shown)


v3/cypress/e2e/map.spec.ts
```diff
@@ -73,7 +73,7 @@ context("Map UI", () => {
       c.checkToolTip($element, c.tooltips.mapCameraButton)
     })
   })
-  it("checks numerical and categorical attributes for map legend", () => {
+  it.skip("checks numerical and categorical attributes for map legend", () => {
```
tealefristoe, line 76: Did you mean to skip several tests in this file?

tealefristoe (approved):
Looks good 👍 Just want to double check that you intended to skip several tests.

v3/cypress/e2e/map.spec.ts
kswenson: From the PR description:

>Note that I also had to disable some unrelated map cypress tests that have proven to be unnecessarily finicky. I added a PT story to fix them at some point but didn't want to hold up this PR further.

But your question is a good argument that I should put a comment in the code.

bfinzer (approved):
👍🏻 LGTM
Perhaps in a standup you can explain the difference between "production" and development modes?

PR #1240, opened 2024-05-02
fix: addCases performance

kswenson:
[[PT-187533902]](https://www.pivotaltracker.com/story/show/187533902)

The `insertCases` implementation was adding/inserting cases/values one at a time to potentially support more complex use cases than we actually need to support, e.g. adding n separate cases after n uniquely identified other cases. Here we simplify the implementation to add the new cases in a block.


bfinzer (approved):
👍🏻LGTM

PR #1241, opened 2024-05-02
187527496 v3 DI Choosy Case and Collection Notifications

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187527496

This PR introduces four notifications used by Choosy to CODAP v3:

`updateCases`
- This is currently broadcast when an attribute's formula is modified or when a value in a table is changed.
  - Are there any other circumstances when this should be broadcast???
- Because pseudo case ids aren't persistent, I wasn't able to return the updated information in the notification. This isn't an issue for Choosy, which is just watching for the operation and ignores the results. I left a comment in the code explaining how to fix the problem once pseudo cases get persistent ids.
- I noticed one bug related to this notification. Pushing enter after modifying a pseudo case's value in the table results in two notifications being broadcast. I didn't dig into why this was happening.

`createCollection`
- This is broadcast when an attribute is dragged into a new collection zone.

`deleteCollection`
- This is broadcast when a collection runs out of attributes, either because the last one is deleted or is moved into another (possibly new) collection.
- To determine if the collection was deleted, `dataSet.removeAttribute` and `dataSet.setCollectionForAttribute` now return a result object that has an optional `removedCollection` boolean.
- For the case where the last attribute in a collection is dragged to create a new collection, I'm still not sure how the old collection is destroyed, but it seems to be. In this case, I'm jumping through some hoops in the `handleNewCollectionDrop` event handler to determine if the old collection has only a single attribute in it, and if so the `deleteCollection` notification is broadcast in addition to `createCollection`.

`updateCollection`
- This is broadcast when the collection's name is changed.
  - Are there any other circumstances when this should be broadcast???
- The UI to update a collection's name changes the model on every keystroke, so this constantly fires notifications as the user types. This also makes it impossible to fully delete a name to start writing a new one, since an empty name is illegal.

Additional changes:
- Notification functions have been streamlined and moved into `data-set-notifications.ts`.
- The order of imports in several files have been corrected.
- Undo strings have been added to some actions that previously weren't wrapped in `applyModelChange`.
- I was able to add tests for most cases where these notifications are broadcast, except for deleting the last attribute in a collection. Using the attribute dropdown menu doesn't seem to work with the `mouseSensor` url parameter, but `mouseSensor` is required for dragging to work, which is needed to create collections.


v3/src/components/case-table/case-table.tsx
```diff
@@ -63,12 +66,32 @@ export const CaseTable = observer(function CaseTable({ setNodeRef }: IProps) {
...
+      const oldCollectionId = dataSet.getCollectionForAttribute(attrId)?.id
+      let removedOldCollection = false
+      if (oldCollectionId) {
+        if (oldCollectionId === dataSet.ungrouped.id) {
+          if (dataSet.ungroupedAttributes.length <= 1) removedOldCollection = true
+        } else {
+          const oldCollectionLength = dataSet.getGroupedCollection(oldCollectionId)?.attributes.length
+          if (oldCollectionLength && oldCollectionLength <= 1) removedOldCollection = true
+        }
+      }
```
kswenson, line 81: Instead of duplicating the logic from the `DataSet`, how about calling `getCollection(oldCollectionId)` after calling `moveAttributeToNewCollection` and if it's not there you know it's been removed?

v3/src/models/data/data-set-types.ts
```diff
@@ -58,6 +58,10 @@ export interface IMoveAttributeCollectionOptions extends IMoveAttributeOptions {
   collection?: string // id of destination collection; undefined => no collection (ungrouped)
 }
 
+export interface IAttributeChangeResult {
+  removedCollection?: boolean // True if changing the attribute resulted in its collection being removed
```
kswenson, line 62: How about making this a string and returning the id of the removed collection (if any)?

v3/src/data-interactive/data-interactive-utils.ts
```diff
@@ -13,3 +16,19 @@ export function canonicalizeAttributeName(name: string, iCanonicalize = true) {
   }
   return tName
 }
+
+export function getCaseValues(caseId: string, collectionId: string, dataSet: IDataSet) {
+  const attributes = dataSet.getGroupedCollection(collectionId)?.attributes ??
+    dataSet.ungroupedAttributes
+
+  const values: DICaseValues = {}
+  const actualCaseIndex = dataSet.cases.findIndex(actualCase => actualCase.__id__ === caseId)
```
kswenson, line 25:
```suggestion
  const actualCaseIndex = dataSet.caseIDMap.get(caseId)
```
Avoids potential O(n^2) lookup.

kswenson (approved):
👍 Looks good -- just a handful of suggestions to address.

v3/src/components/case-table/case-table.tsx
tealefristoe: I definitely like this idea, but it isn't working for the ungrouped collection. I added a TODO about revisiting this after the collection code gets shuffled.

v3/cypress/e2e/plugin.spec.ts
```diff
@@ -188,8 +188,25 @@ context("codap plugins", () => {
...
+    webView.clearAPITesterResponses()
+
+    cy.log("Broadcast updateCollection notifications")
+    table.renameCollection("c1", "Mammals")
+    webView.confirmAPITesterResponseContains(/"operation":\s"updateCollection/)
+    webView.clearAPITesterResponses()
+
+    // TODO Check for deleteCollection notifications when deleting the last attribute
+    // in a grouped or ungrouped collection. I couldn't figure out how to do this because
+    // attribute dropdown menus don't work with mouseSensor but dragging won't work without it
```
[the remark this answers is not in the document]
tealefristoe: It took some doing, but I was able to figure out a test for deleting the last attribute in a grouped collection. I did it by opening a file that already contains a dataset with multiple collections (Four Seals) and deleting all the attributes in the grouped collection.

However, in trying to get this to work for the upgrouped collection, I discovered that deleting the last attribute in the ungrouped collection doesn't actually remove the ungrouped collection... it continues to exist as an empty collection. That's because `dataSet.removeAttribute` lacks a clause like `dataSet.setCollectionForAttribute` has, where `collection` is handled when `isCollectionModel(collection)` is false. Given that this is all going to get refactored soon, I'm not going to fix it in this PR. I'll create a PT bug story so we don't forget about it, though.

PR #1245, opened 2024-05-06
feat: numeric legend sorts values (PT-187499873)

emcelroy:
[#187499873](https://www.pivotaltracker.com/story/show/187499873)

These changes update `getCaseDataArray` to sort cases by their numeric value for the legend attribute when a numeric legend is present. They also modify `DotChartBars` to use `caseDataArray` instead of `dataset.cases`, and sort its `legendCats` array to match the sorted order so the bars and bar covers are placed in the correct order.

### DIFF
--- a/v3/src/components/graph/components/dot-chart-bars.tsx
+++ b/v3/src/components/graph/components/dot-chart-bars.tsx
@@ -11,6 +11,7 @@ import { GraphLayout } from "../models/graph-layout"
 import { SubPlotCells } from "../models/sub-plot-cells"
 import { mstAutorun } from "../../../utilities/mst-autorun"
 import { useChartDots } from "../hooks/use-chart-dots"
+import { numericSortComparator } from "../../../utilities/data-utils"
 
 interface IRenderBarCoverProps {
   barCovers: IBarCover[]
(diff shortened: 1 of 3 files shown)


v3/src/components/graph/components/dot-chart-bars.tsx
```diff
@@ -135,20 +133,33 @@ export const DotChartBars = observer(function DotChartBars({ abovePointsGroupRef
...
                   if (!caseGroups.has(caseGroupKey)) {
                     caseGroups.set(caseGroupKey, [])
                   }
                   caseGroups.get(caseGroupKey).push(aCase)
                 })
+
+                // If the legend attribute is numeric, sort legendCats in descending order making sure to handle any NaN
+                // values for cases that don't have a numeric value for the legend attribute.
+                if (dataConfig.attributeType("legend") === "numeric") {
+                  legendCats.sort((cat1: string, cat2: string) => {
```
bfinzer, line 154: With the changes to DataConfigurationModel, this sorting is redundant, right? (I tried commenting it out and bar chart with numeric legend still was sorted.)

bfinzer (approved):
👍🏻LGTM
==> Just the one place to remove redundant code if I'm correct.

v3/src/components/graph/components/dot-chart-bars.tsx
emcelroy: @bfinzer I wish it was redundant, but it's not, at least not with things as they are. This part of the code is putting together the bar _covers_. Even though the actual bars would be sorted correctly without this, their associated covers would not be. We're iterating over `dataConfig.categorySetForAttrRole("legend")?.values` here which won't have the same order as `caseDataArray`. I thought maybe we could iterate over `caseGroups` instead and avoid the additional sort, but I couldn't get that to work. Maybe there's another option I'm overlooking?

kswenson: If nothing else, perhaps the compare function could be shared between the two implementations?

emcelroy: @bfinzer @kswenson Does the latest commit seem reasonable?

emcelroy: A detail I keep wondering about is the compare function name. Originally, I made it `compareNum` but that seemed too generic. I'm not sure `numericSortComparator` is quite right, though. Maybe something like `compareNumForSort` would be better?

kswenson: `numericSortComparator` seems fine.

kswenson (approved):
👍 Changes look good

PR #1246, opened 2024-05-06
fix: Legend compresses on tile resize (PT-187450934)

emcelroy:
[#187450934](https://www.pivotaltracker.com/story/show/187450934)

Previously, when a graph tile containing a categorical legend was sized down to the smallest possible width, the calculation to determine the number of columns in the legend area could return a value of `0`. This change ensures the value is always at least `1`.


kswenson (approved):
👍 LGTM

PR #1247, opened 2024-05-07
187548327 v3 Improve Collection Title UI

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187548327

This PR makes the collection name only change when the user confirms a change, rather than on every keystroke. This was previously causing problems with the undo and notification system. The PR also fixes a bug where it was impossible to delete the last character of a collection's name.


kswenson (approved):
👍 LGTM

PR #1248, opened 2024-05-07
187033159 v3 Tile Title UI

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187033159

This PR changes how tile titles are updated. Previously, they were changed on every keystroke. Now they are only changed when a user confirms a change (according to chackra's `onSubmit`). This fixes issues with the undo stack and notifications.


v3/src/components/component-title-bar.tsx
```diff
@@ -30,16 +31,36 @@ export const ComponentTitleBar = observer(function ComponentTitleBar(
     }
   }
```
kswenson, line 32: Shouldn't this call to `tile.setTitle` be wrapped in `applyModelChange`?

kswenson (approved):
👍 Looks good -- just one question about wrapping the default title change in `applyModelChange`.

v3/src/components/component-title-bar.tsx
tealefristoe: Good call 👍

PR #1250, opened 2024-05-09
feat: use numeric ids for plugin compatibility

kswenson:
[[PT-187573595]](https://www.pivotaltracker.com/story/show/187573595)

The proposed solution is as follows:
- CODAP v3 continues to use string ids for all models.
- For models trafficked by the plugin API (tiles, collections, attributes, cases, etc.), auto-generated ids will be large numeric integers that are then stringified.
- CODAP v2 documents imported into v3 will have their ids stringified on import.
- The plugin API is responsible for converting between the string ids used internally and the numeric ids used by the plugin API.

### DIFF
--- a/v3/package-lock.json
+++ b/v3/package-lock.json
@@ -46,7 +46,6 @@
         "rtree": "^1.4.2",
         "tslib": "^2.6.2",
         "type-fest": "^4.10.2",
-        "ulid": "^2.3.0",
         "use-debounce": "^10.0.0",
         "use-deep-compare": "^1.2.1",
         "use-memo-one": "^1.1.3"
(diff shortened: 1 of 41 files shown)


tealefristoe (approved):
This all looks good to me. It's too bad we're losing the prefixes, but the API is requiring some really annoying constraints. My only question with this work is will we have problems with loaded documents that have old style ids in them?

kswenson:
>My only question with this work is will we have problems with loaded documents that have old style ids in them?

The idea is that those ids should get converted to strings on import, but now that I look at the import code more closely it looks like we're dropping some of those ids on the floor rather than converting them, so I'll take another look. Good thought.

kswenson:
> > My only question with this work is will we have problems with loaded documents that have old style ids in them?
> 
> The idea is that those ids should get converted to strings on import, but now that I look at the import code more closely it looks like we're dropping some of those ids on the floor rather than converting them, so I'll take another look. Good thought.

Changes to support the correct behavior were fairly extensive, so requesting re-review.

kswenson:
> This all looks good to me. It's too bad we're losing the prefixes, but the API is requiring some really annoying constraints. My only question with this work is will we have problems with loaded documents that have old style ids in them?

If we _really_ wanted to maintain the prefixes, we could do that  -- we'd just have to strip them when converting from v3 to v2. I thought about that but wasn't convinced it was worth it, given that the conversion code is dead simple this way.

tealefristoe (approved):
This looks good to me 👍

kswenson:
Adding @bfinzer as a reviewer since he was one of the original proponents of the prefixes.

tealefristoe:
While the prefixes are nice, I personally vote to drop them to keep the conversion simple.

kswenson:
@tealefristoe Requesting re-re-review since I went ahead and implemented support for prefixes. It has the ancillary benefit of making the conversion calls more explicit.

PR #1251, opened 2024-05-09
187573958 v3 Fix Sampler

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187573958

This PR adds handlers for several API requests required for the Sampler plugin:
- `get collection`
- `create item`
- `get attributeList`
- `create component` (only for `type: caseTable`)
  - For this request, I ended up returning far more fail results than occur in v2. For example, if the client requests creating a table without a dataContext specified, or with a dataContext that doesn't exist, a fail result is returned.
  - I also made it so this will show the table if it is hidden, keeping in line with a broad change we're making in v3.

These are the API requests needed for basic functionality of the Sampler plugin, but it's very possible there are more. If any bugs with Sampler are discovered, please let me know so I can look into which API requests aren't being handled yet. But I think those shouldn't hold up this PR, which is already about three times the size it should be.

In addition to the above changes, there were a few other noteworthy changes in this PR:
- Common code from DI handler tests has been consolidated, which required tweaking many tests.
- Functions for creating or showing tables have been moved from a button component to a utility file where DI handlers can access them.


bfinzer (approved):
👍🏻LGTM
It was helpful to dive into your implementation and get a sense of where things are at. Looks good!

PR #1253, opened 2024-05-10
fix: legend issues with Display Only Selected Cases (PT-187458777)

emcelroy:
[#187458777](https://www.pivotaltracker.com/story/show/187458777)

Fixes two bugs related to the graph's Display Only Selected Cases option when a categorical legend is present:

1) When no cases are selected, the legend would show `__main__` as the only category

2) If no cases are selected and then a case is selected, the legend would update appropriately to show that case's value for the attribute. But if you then selected another single case that had a different value for the attribute, the legend would not update to show that difference. (See [this screencast](https://somup.com/cZhj3r53e9).)

### DIFF
--- a/v3/src/components/data-display/models/data-configuration-model.ts
+++ b/v3/src/components/data-display/models/data-configuration-model.ts
@@ -627,7 +627,13 @@ export const DataConfigurationModel = types
       // Invalidate cache when selection changes.
       addDisposer(self, reaction(
         () => self.dataset?.selection.values(),
-        () => self.clearCasesCache(),
+        () => {
+          if (self.displayOnlySelectedCases) {
+            self.clearCasesCache()
+          } else {
+            self.allCasesForCategoryAreSelected.invalidateAll()
+          }
+        },
         {
           name: "DataConfigurationModel.afterCreate.reaction [allCasesForCategoryAreSelected invalidate cache]",
           equals: comparer.structural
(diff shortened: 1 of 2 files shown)


v3/src/components/data-display/models/data-configuration-model.ts
```diff
@@ -627,7 +627,7 @@ export const DataConfigurationModel = types
       // Invalidate cache when selection changes.
       addDisposer(self, reaction(
         () => self.dataset?.selection.values(),
-        () =>  self.allCasesForCategoryAreSelected.invalidateAll(),
+        () => self.clearCasesCache(),
```
kswenson, line 630: The problem with this solution is that `clearCasesCache` means that absolutely everything has to be recomputed from scratch, which is expensive, particularly given that selection changes are very frequent. At the very least, we should only do this when `displayOnlySelectedCases` is true. The goal is to invalidate what we need to invalidate for the current situation and no more.

kswenson (changes requested):
👍 Looks good, but need to avoid expensive invalidation when it's not necessary.

v3/src/components/data-display/models/data-configuration-model.ts
emcelroy: I've added a check for whether `displayOnlySelectedCases` is true. I also tried limiting the invalidations to some subset of what's in `clearCasesCache`, but the bug isn't fixed unless all of them are called. Does the current state seem good? Or should I spend more time trying to limit the invalidation somehow?

kswenson (approved):
👍 Changes look good

PR #1254, opened 2024-05-10
fix: Dragging categories on axis (PT-187539044)

emcelroy:
[#187539044](https://www.pivotaltracker.com/story/show/187539044)

Although changes in [another PR](https://github.com/concord-consortium/codap/pull/1227) mostly fixed the problem identified in [this story](https://www.pivotaltracker.com/story/show/187539044), there was a remaining bug with dragging categories on an axis. In the previously-mentioned PR, we mistakenly changed the calculation of the index for the category _before which_ the dragged category should be moved if it was moved up from its initial position in the array.

### DIFF
--- a/v3/src/components/axis/hooks/use-sub-axis.ts
+++ b/v3/src/components/axis/hooks/use-sub-axis.ts
@@ -267,7 +267,7 @@ export const useSubAxis = ({
           // Figure out the label of the category before which the dragged category should be placed
           const moveToGreater = newCatIndex > dI.indexOfCategory,
             catToMoveBefore = moveToGreater
-              ? (newCatIndex === numCategories - 1 ? '' : dI.categories[newCatIndex])
+              ? (newCatIndex === numCategories - 1 ? '' : dI.categories[newCatIndex + 1])
               : dI.categories[newCatIndex]
           dI.indexOfCategory = newCatIndex
           dI.categorySet?.move(dI.catName, catToMoveBefore)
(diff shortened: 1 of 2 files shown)


kswenson (changes requested):
The change I made was to fix a different bug which this re-introduces. I suggest taking this opportunity to fix it properly. The now-reintroduced bug is this:
- start with mammals sample document (http://localhost:8080/#file=examples:Mammals)
- in the bottom right graph with Diet on the x axis:
  - drag `both` to the left
  - it switches with `meat` correctly
  - it switches with `plants` correctly
  - without releasing the mouse drag it back to the right
  - when it's supposed to switch with `plants` to the middle it jumps all the way to the right
 
So dragging to the right works differently depending on whether you started on the left or started on the right and are now moving back to the right.

emcelroy:
I think we didn't grasp the true nature of the bug and that's _probably_ because we were using Mammals as a test case. Since Mammals' categorical attributes only have three categories, the full extent of the problem was being masked. If you open [Roller Coasters on main](https://codap3.concord.org/branch/main/#file=examples:Roller%20Coasters) and put `Design` on the x axis, for example, and then try dragging the third category one spot to the right, it won't work.

Before the latest change to the `move` method in `CategorySet`, if `fromIndex` was less than `toIndex` there, nothing was done to modify `toIndex` if the item was being moved from left to right. And since `toIndex` is the index of the item before which the item is being moved, `toIndex` should be decremented by one if the item is being moved from left to right.

The latest commit seems to get dragging of categories to work correctly in all instances. What do you think of that solution, @kswenson?

For what it's worth, I did try adding some Cypress tests for reordering categories on the axis, but Cypress kept crashing when mousing down on the category labels. It wasn't clear to me why. The same thing happened when I tried adding Cypress tests for reordering categories in a legend. It'd be great if we could add some tests for these features, of course, but I don't know if it's worth spending a lot of time on right now.

kswenson (approved):
👍 LGTM

PR #1255, opened 2024-05-10
fix: Unable to reorder categories in legend (PT-187392421)

emcelroy:
[#187392421](https://www.pivotaltracker.com/story/show/187392421)

The `categoryData` ref used throughout the component, but perhaps most notably in the D3 code that constructs the legend element, was not being updated when the category order changed. Also, the graph data configuration model's `getAllCategoriesForRoles` view didn't include "legend". So the cached `categoryArrayForAttrRole` wasn't being invalidated in reaction to reordering the legend items.

I noticed the categorical legend reorder problem also exists in the map. So I added a `getAllCategoriesForRoles` view to `DataConfigurationModel` and moved the new reaction that uses `getAllCategoriesForRoles` as its data function from `GraphDataConfigurationModel` to `DataConfigurationModel`. I think some of the duplicate code introduced by that change could be reduced, but wanted to get thoughts on the general idea before devoting more time to that.

### DIFF
--- a/v3/src/components/data-display/components/legend/categorical-legend.tsx
+++ b/v3/src/components/data-display/components/legend/categorical-legend.tsx
@@ -88,6 +88,19 @@ export const CategoricalLegend = observer(
     const
       keysElt = useRef(null)
 
+    const setCategoryData = useCallback(() => {
+      if (categoriesRef.current) {
+        const newCategoryData = categoriesRef.current.map((cat: string, index) => ({
+          category: cat,
+          color: dataConfiguration?.getLegendColorForCategory(cat) || missingColor,
+          column: index % layoutData.current.numColumns,
+          index,
+          row: Math.floor(index / layoutData.current.numColumns)
+        }))
+        categoryData.current = newCategoryData
+      }
+    }, [dataConfiguration])
+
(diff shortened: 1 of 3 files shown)


v3/src/components/data-display/components/legend/categorical-legend.tsx
```diff
@@ -204,16 +208,17 @@ export const CategoricalLegend = observer(
             newCatIndex = coordinatesToCatIndex(lod, numCategories, newDragPosition)
           if (newCatIndex >= 0 && newCatIndex !== dI.indexOfCategory) {
             // swap the two categories
-            duration.current = transitionDuration / 2
```
emcelroy, line 207: This may seem counterintuitive, but I found removing this line made the user interaction much smoother.

kswenson (approved):
👍 Looks good -- I think the way to eliminate the redundancy would be to have a method in `DataConfigurationModel` called something like `potentiallyCategoricalRoles` which just returns the array of roles and could be overridden by the `GraphConfigurationModel`, and then the `getAllCategoriesForRoles` method would just call `potentiallyCategoricalRoles` and build up the categories.

kswenson (approved):
👍 Changes look good

PR #1258, opened 2024-05-14
fix: Graph has option to Display Only Selected Cases (PT-187458777)

emcelroy:
[#187458777](https://www.pivotaltracker.com/story/show/187458777)

These changes fix two bugs:

1. In graphs with a numeric legend, if Display Only Selected Cases is activated while there are no cases selected, the legend shows `NaN` for the values. There was some offline discussion about this behavior since there is a similar bug in V2. When Display Only Selected Cases is activated, the desired behavior is that numeric legends should always reflect the full range of all case values regardless of whether some or all are displayed. The only exception is when there is either one case displayed or two or more cases displayed that all have the same value for the legend attribute. In that case, the legend should only reflect that value. The changes here do not support that exception. Adding that will be done in a [separate story](https://www.pivotaltracker.com/story/show/187605700). Note that when cases are otherwise hidden by the user (using the Hide Unselected Cases option), a numeric legend should always display the range of values for only the cases that are not hidden (i.e. not the full range).

2. In a graph tile with no attributes added, if Display Only Selected Cases is activated while there are no cases selected, and a numeric attribute is added to the x or y axis, and then a case is selected in the table cell, the axis domain does not get updated and the case dot that appears remains in the upper left corner of the plot.

[Description edited after a [comment from bfinzer](https://github.com/concord-consortium/codap/pull/1258#issuecomment-2113013945).]

### DIFF
--- a/v3/src/components/data-display/components/legend/numeric-legend.tsx
+++ b/v3/src/components/data-display/components/legend/numeric-legend.tsx
@@ -45,13 +45,19 @@ export const NumericLegend =
       }
 
       if (choroplethElt) {
-        // If some or all cases are hidden, the legend should still reflect the full range of values for both hidden
-        // and visible cases.
-        // TODO: When all visible cases have the exact same value for the legend attribute, the legend should only
-        // reflect the values of the cases shown.
+        /**
+         *  Adjust the value range displayed by the legend based on the data configuration model's properties:
+         *  1. If `displayOnlySelectedCases` is true and not all cases are visible, the legend displays the range of all
+         *     cases, both hidden and visible.
+         *  2. If `displayOnlySelectedCases` is false, and some cases are explicitly hidden by the user, the legend
+         *     displays the range of only the visible cases.
+         *  3. If all cases are hidden, the legend displays no range.
+         */
         const allCasesCount = dataConfiguration?.dataset?.cases.length ?? 0
-        let values = dataConfiguration?.numericValuesForAttrRole("legend") ?? []
-        if (values.length < allCasesCount) {
+        const hiddenCasesCount = dataConfiguration?.hiddenCases.length ?? 0
+        const allCasesHidden = hiddenCasesCount === allCasesCount
+        let values = allCasesHidden ? [] : dataConfiguration?.numericValuesForAttrRole("legend") ?? []
+        if (dataConfiguration?.displayOnlySelectedCases && !allCasesHidden && values.length < allCasesCount) {
           const attribute = dataConfiguration?.dataset?.attrFromID(dataConfiguration?.attributeID("legend"))
           values = attribute?.numValues ?? []
         }
--- a/v3/src/components/data-display/components/legend/numeric-legend.tsx
+++ b/v3/src/components/data-display/components/legend/numeric-legend.tsx
@@ -2,6 +2,7 @@ import {ScaleQuantile, scaleQuantile, schemeBlues} from "d3"
 import {reaction} from "mobx"
 import {observer} from "mobx-react-lite"
 import React, {useCallback, useEffect, useRef, useState} from "react"
+import { mstReaction } from "../../../../utilities/mst-reaction"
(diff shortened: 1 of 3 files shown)


tealefristoe (approved):
This looks good to me 👍

bfinzer:
I think there is a difference between "hidden" cases and cases that are not showing because "Display only selected cases" is chosen wrt a numeric legend.
1. For "hidden" cases in a graph, a numeric legend should always display the range of values for the cases that are *not* hidden; i.e. not the full range.
2. But in the situation in which some cases are not showing because "Display only selected cases" is chosen, the full range of the "not hidden" cases should be reflected in the numeric legend scale.
I'm not sure how difficult this would be to bring about.

emcelroy:
> I think there is a difference between "hidden" cases and cases that are not showing because "Display only selected cases" is chosen wrt a numeric legend.
> 
> 1. For "hidden" cases in a graph, a numeric legend should always display the range of values for the cases that are _not_ hidden; i.e. not the full range.
> 2. But in the situation in which some cases are not showing because "Display only selected cases" is chosen, the full range of the "not hidden" cases should be reflected in the numeric legend scale.
> 
> We discussed and Ethan will attempt to bring this about and I'll complete my review after that's done.

@bfinzer @tealefristoe There's a [new commit](https://github.com/concord-consortium/codap/pull/1258/commits/9c3b539d676d78172fd1672b96d91ae37a2c36c2) with changes to `NumericLegend` that take the above into account. Please take a look and let me know what you think.

bfinzer (approved):
👍🏻LGTM

v3/src/components/data-display/components/legend/numeric-legend.tsx
```diff
@@ -44,7 +45,23 @@ export const NumericLegend =
...
+         */
+        const allCasesCount = dataConfiguration?.dataset?.cases.length ?? 0
+        const hiddenCasesCount = dataConfiguration?.hiddenCases.length ?? 0
+        const allCasesHidden = hiddenCasesCount === allCasesCount
+        let values = allCasesHidden ? [] : dataConfiguration?.numericValuesForAttrRole("legend") ?? []
+        if (dataConfiguration?.displayOnlySelectedCases && !allCasesHidden && values.length < allCasesCount) {
+          const attribute = dataConfiguration?.dataset?.attrFromID(dataConfiguration?.attributeID("legend"))
+          values = attribute?.numValues ?? []
+        }
+        valuesRef.current = values
```
tealefristoe, line 64: I was having a pretty tough time understanding the logic here, even with your comments above. I wonder if something like this would be more clear? I also think you should put the cases in the comment in the same order they appear in the code, and you could even move each comment case to just above the conditional block to which it's referring.
```suggestion
        if (allCasesHidden) {
          valuesRef.current = []
        } else if (dataConfiguration?.displayOnlySelectedCases && values.length < allCasesCount) {
          const attribute = dataConfiguration?.dataset?.attrFromID(dataConfiguration?.attributeID("legend"))
          valuesRef.current = attribute?.numValues ?? []
        } else {
          valuesRef.current = dataConfiguration?.numericValuesForAttrRole("legend") ?? []
        }
```

tealefristoe (approved):
Looks good! I did leave a comment about reformatting some of the code and the related comment, but it's up to you if you want to follow that advice, switch the structure some other way, or just keep it as is.

v3/src/components/data-display/components/legend/numeric-legend.tsx
emcelroy: @tealefristoe Good points, thanks. I've updated things following your suggestions with just some very minor tweaks. I brought back a todo item to the comment block. I removed it because I thought it would add confusion, but I think it's probably best to leave it since it'll make it more likely the information in the comment gets updated when that part is added. 

Do you think that all looks good?

kswenson (commented):
👍 LGTM

PR #1259, opened 2024-05-15
fix: Show hide parent visibility toggles (PT-187458779)

emcelroy:
[#187458779](https://www.pivotaltracker.com/story/show/187458779)

Two bug fixes:

1. Moves calls to the graph model's `setShowOnlyLastCase` method inside `applyModelChange` so the toggling of the last case option is properly handled on undo/redo.

2. Optimizes `handleCaseButtonClick` to prevent crashes when working with large datasets.


tealefristoe (approved):
Looks good 👍

kswenson (commented):
👍 LGTM

PR #1262, opened 2024-05-16
fix: replace autorun with mstAutorun to eliminate MST warning

kswenson:
[[PT-187615908]](https://www.pivotaltracker.com/story/show/187615908)

Rather than trying to recall the steps, I followed them to make sure I got them right and have documented them here and in the corresponding PT story for future reference:

1. The first console warning occurs in `graph-attribute-label.tsx` line 45 in my version (47 in yours above), which is the `getClickHereCue` method.
2. I tried setting a breakpoint there, but it triggers too often, so instead I added this line at the top of the function
  - `if (!isAlive(graphModel)) debugger`
3. When the breakpoint is hit, scan the stack trace to find the autorun or reaction that is triggering after the model has been destroyed.
4. In this case, it’s the one named `AttributeLabel.autorun [refreshLabel]` in `attribute-label.tsx`.
5. We need to use `mstAutorun` so that the disposer is called when the model goes away.
6. That component doesn’t currently make use of the content model, but can get it with `useDataDisplayModelContext`.
7. Switching to `mstAutorun` and passing in the `contentModel` eliminates the warning.

### DIFF
--- a/v3/src/components/data-display/components/attribute-label.tsx
+++ b/v3/src/components/data-display/components/attribute-label.tsx
@@ -1,12 +1,13 @@
-import { autorun } from "mobx"
 import React, { ForwardedRef, forwardRef, useEffect, useState } from "react"
 import { createPortal } from "react-dom"
 import { IDataSet } from "../../../models/data/data-set"
 import {kPortalClassSelector} from "../data-display-types"
 import { GraphPlace } from "../../axis-graph-shared"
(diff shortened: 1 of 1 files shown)


emcelroy (approved):
Looks good, and thanks for the detailed explanation 👍

bfinzer (approved):
👍🏻 LGTM
Excellent documentation of a hard-to-track-down problem!

PR #1264, opened 2024-05-16
187564414 v3 DI Fix Choosy Plugin

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187564414

This PR updates the v3 API handling system to get the choosy plugin working. Two changes were required to make this happen:
- The correct `id` needed to be included in the response to `get dataContextList` requests.
- Despite the documentation claiming that notifications are not broadcast from CODAP in response to API requests, this does happen in at least some cases, and Choosy was using these notifications (rather than the response to its request) to update its state. So I added these notifications in two places, `create attribute` and `update attribute`.
  - That said, it feels like we should either always broadcast notifications in response to API requests that modify the document, or never do it. Always doing it will take some work and I can imagine could make some plugins go into infinite loops. Never doing it will require Choosy to be updated so it stops relying on these notifications.

Additionally, several other changes were made in this PR:
- I turned off toast notifications when the API handler receives a request. This was becoming quite cumbersome now that we're dealing with actual plugins, which sometimes rapid-fire half a dozen requests in response to one user action. It's still possible to see what requests the API system is receiving in the console by setting the local storage `debug` key to `plugins`.
- There were many places where v3 ids were not being converted to v2 ids when sending responses and notifications to plugins. I went through and tried to fix this, which is what caused most files in this PR to change.
- Ids weren't being converted from v2 to v3 in response to component API requests. I fixed this. But it was a little tricky because each tile type has its own prefix, and I'm not sure if my solution is the best one.
- I moved `valuesFromGlobal` into `data-interactive-type-utils.ts` so I could delete the redundant `di-conversion-utils.ts` file.
- I expanded the `resource-parser` jest test quite a bit. There are still more cases to cover, but it's a lot better than it was.


v3/src/data-interactive/resource-parser.ts
```diff
@@ -98,7 +101,13 @@ export function resolveResources(
...
   if (resourceSelector.component) {
     // TODO Get tile by name?
-    result.component = document.content?.getTile(resourceSelector.component)
+    const { component } = resourceSelector
+    // We look for every possible v3 id the component might have (because each tile type has a different prefix).
+    // Is there a better way to do this?
+    const possibleIds =
+      [component, toV3TileId(component), ...getTilePrefixes().map(prefix => toV3Id(prefix, component))]
+    const componentId = possibleIds.find(id => document.content?.getTile(id))
+    if (componentId) result.component = document.content?.getTile(componentId)
```
kswenson, line 110: I'm inclined to say that this is fine for now. If we wanted to do better we could put the responsibility on the `DocumentContent`, which could maintain a mapping from prefix-less id to tile, for instance, but that can be left as a potential optimization for another day.

kswenson (approved):
👍 Looks good

PR #1266, opened 2024-05-17
fix: Map zoom range not framed around dataset (PT-187620092)

emcelroy:
[#187620092](https://www.pivotaltracker.com/story/show/187620092)

### DIFF
--- a/v3/src/components/map/components/map-interior.tsx
+++ b/v3/src/components/map/components/map-interior.tsx
@@ -1,5 +1,5 @@
 import {observer} from "mobx-react-lite"
-import React, {useCallback} from "react"
+import React, {useCallback, useEffect} from "react"
 import {PixiPoints} from "../../data-display/pixi/pixi-points"
 import {useMapModelContext} from "../hooks/use-map-model-context"
 import {useMapModel} from "../hooks/use-map-model"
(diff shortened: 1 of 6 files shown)


v3/src/components/map/components/map-interior.tsx
```diff
@@ -9,20 +9,31 @@ import {MapPointLayer} from "./map-point-layer"
...
 
+
+  // Ensure the map rescales when its tile's CSS transition is complete to compensate for any changes to the tile's
+  // size that may have occurred after the map was initially rendered.
+  useEffect(function rescaleOnTileTransitionEnd() {
+    if (tileTransitionComplete) {
+      mapModel.rescale()
+    }
+  }, [tileTransitionComplete, mapModel])
+
```
emcelroy, line 36: I'm not sure this is the best way to do this since it will result in two calls to `rescale` on the map model, the other being in `useMapModel`. I'm not sure there's a way around that, though. Unless it would be OK to remove the call to `rescale` in `useMapModel`? It's not clear to me whether that may be required for other things.

bfinzer: I removed the call in useMapModel as this call handles all the relevant situations.

bfinzer (approved):
👍🏻LGTM
Truly elegant solution!

PR #1268, opened 2024-05-17
Bug fix - Coerce arguments passed to random, randomNormal and randomBinomial to numbers

bfinzer:
[#187531276] Bug fix: randomNormal function bug fix when encountering string arguments

* We coerce the arguments passed to all the randomGen functions to numbers since that is what they expect


tealefristoe (approved):
Looks good! 👍

PR #1269, opened 2024-05-18
fix: MST warning on closing map component

kswenson:
[[PT-187628522]](https://www.pivotaltracker.com/story/show/187628522)

1. The warning indicates that the error occurs in the `selection` method at `DataConfigurationModel` line 222.
2. Adding `if (!isAlive(self)) debugger` at that line yields the stack trace in the enclosed screen shot.
3. Clicking on the `runReaction_` line in the debugger results in the following `scope` section of the debugger, as seen in the next screen shot.
4. From this, we can see that the reaction in question is named `observerMapPointLayer`.
5. This string doesn't appear in the code because that's the name given to `observer` components, so the reaction in question is the `MapPointLayer` component which is an `observer`.
6. The fix is to test `isAlive` before dereferencing the `dataConfiguration` because there isn't a good way to tell an `observer` component to stop observing.

MSTWarning

MSTWarningScope

### DIFF
--- a/v3/src/components/map/components/map-point-layer.tsx
+++ b/v3/src/components/map/components/map-point-layer.tsx
@@ -1,6 +1,7 @@
 import React, {useCallback, useEffect, useRef} from "react"
 import {comparer, reaction} from "mobx"
 import { observer } from "mobx-react-lite"
+import { isAlive } from "mobx-state-tree"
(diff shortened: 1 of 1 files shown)


bfinzer (approved):
👍🏻LGTM
I find this to be a subtle bug! Thanks for tracking it down.

PR #1273, opened 2024-05-20
User can specify number of standard errors

bfinzer:
[#187626777] Feature: User can specify how many standard errors are shown in the standard error bar

* We add a NumberInput to the Standard Error checkbox item in standard-error-adornment-registration.tsx
* We do a considerable amount of "futzing" to make sure undo/redo works properly and the user can't get in trouble by deleting the numeric value
* Note that the formatting of this input is not good. I'm adding a PT story about it.
* Also note that the jest tests produce a console.warn because "MockAdornment" is an unknown adornment type. I wasn't sure whether to modify adornment-types.ts to account for it, to remove the console warning or whether there might be some other way to deal with the problem.

### DIFF
--- a/v3/src/components/graph/adornments/univariate-measures/standard-error/standard-error-adornment-model.ts
+++ b/v3/src/components/graph/adornments/univariate-measures/standard-error/standard-error-adornment-model.ts
@@ -31,12 +31,14 @@ export const StandardErrorAdornmentModel = UnivariateMeasureAdornmentModel
     get hasRange() {
       return true
     },
+  }))
+  .views(self => ({
     computeMeasureValue(attrId: string, cellKey: Record<string, string>, dataConfig: IGraphDataConfigurationModel) {
       // The measure value is the standard error of the mean.
       const caseValues = self.getCaseValues(attrId, cellKey, dataConfig)
       // If there are less than two values, the adornment should not render.
       if (caseValues.length < 2) return
-      return this.numStErrs * Number(std(caseValues)) / Math.sqrt(caseValues.length)
+      return self.numStErrs * Number(std(caseValues)) / Math.sqrt(caseValues.length)
     }
   }))
   .views(self => ({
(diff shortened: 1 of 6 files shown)


v3/src/components/graph/adornments/univariate-measures/standard-error/standard-error-adornment-model.ts
```diff
@@ -26,7 +36,7 @@ export const StandardErrorAdornmentModel = UnivariateMeasureAdornmentModel
       const caseValues = self.getCaseValues(attrId, cellKey, dataConfig)
       // If there are less than two values, the adornment should not render.
       if (caseValues.length < 2) return
-      return self.numStErrs * Number(std(caseValues)) / Math.sqrt(caseValues.length)
+      return this.numStErrs * Number(std(caseValues)) / Math.sqrt(caseValues.length)
```
emcelroy, line 39: I think it's preferred and would keep things more consistent to use `self` instead of `this` here, and move `computeMeasureValue` to a separate `views` block. Using `this` should work OK for now, but it could potentially cause problems in the future as things get changed and possibly shifted around. I don't think it's a huge problem, though, if you really want to keep this grouped with the other methods here.

emcelroy (approved):
Looks good 👍 I left a non-blocking comment about `self` versus `this` in the MST model.

For the Jest test warning, I think it'd be best to update `adornments-store.test.ts` to use `"Count"` (or better yet `kCountType`) in place of `"Mock Adornment"` everywhere. Since the mock adornment purports to be a Count adornment, that would make sense. My original choice of using `"Mock Adornment"` doesn't serve any purpose really. Using `kCountType` works fine and is probably better since its an actual type.

bfinzer:
Unfortunately, using kCountType produces another error because in adornments-base-store we push the MockAdornment and there's already one of kCountType in the array. I guess I'll skip the offending test and put a comment about why.

PR #1275, opened 2024-05-21
feat: Parent Visibility Toggles should scroll nicely (PT-187578056)

emcelroy:
[#187578056](https://www.pivotaltracker.com/story/show/187578056)

When the list of toggle buttons is wider than the available space, these changes ensure that:

- The last button in the list is visible when the Parent Visibility Toggles option is activated
- The left and right arrows appear only when the list is scrollable in their associated direction
- All currently visible buttons are _fully_ visible with no cropping by the boundaries of the button list container


bfinzer (approved):
👍🏻LGTM

kswenson (approved):
👍 Looks pretty subtle and I can't say I followed it completely but seems reasonable to me!

PR #1276, opened 2024-05-22
187637447 v3 DI caseByID and caseByIndex Requests

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187637447

This PR adds handlers for `caseByID` and `caseByIndex` API requests. Both support `delete`, `get`, and `update` and are identical except for how the relevant case is referred, which takes place in `resourceParser`. Since the handlers act the same way, their functions are defined in a third file, `case-by-handler-functions.ts`, with the actual handlers just calling these functions with the proper case.

One thing worth bringing up is how similar the response to `get caseByID/Index` is to `updateCase` and `selectCases` notifications broadcast by CODAP. I originally thought I could reuse the same formatting function before realizing they are slightly different. The result is one that's extremely similar with a lot of reused code. Both functions are in `data-interactive-type-utils.ts`, with `getCaseRequestResultValues` added in this PR and `convertCaseToV2FullCase` left as is. This is the sort of inconsistency that would be great to remove from CODAP. I think it would be safe to combine the information returned, though that would result in some awkwardness, like `id`, `caseId`, and `itemId` all being included in the final result. It might also be safe to totally change what's returned in the notifications, since at least Choosy ignores the details of these notifications (perhaps other plugins care about them, though). In any case, v3 is a great opportunity to try to clean up some of these inconsistencies that most likely arose in v2 organically without much thought to the bigger picture.

Another weak point of the API is the expected `values` for `update caseByID/Index`. The expected structure is like:
```
values: {
  values: {
    attr1: val1,
    ...
  }
}
```
Having two layers of `values` is just confusing. However, I don't think this is something we can change at this point.

Other changes in this PR:
- I streamlined the standard error results for API requests. This reduced duplicate code and also made error messages more consistent across different handlers. This does make some error responses different from v2, but in most if not all cases it makes the error messages more detailed and accurate.
- I made formatting changes in some old handler files to clean them up a bit.

### DIFF
--- a/v3/src/data-interactive/data-interactive-type-utils.ts
+++ b/v3/src/data-interactive/data-interactive-type-utils.ts
@@ -9,7 +9,7 @@ import { kAttrIdPrefix, maybeToV2Id, toV2Id } from "../utilities/codap-utils"
 import {
   ICodapV2AttributeV3, ICodapV2CollectionV3, ICodapV2DataContextV3, v3TypeFromV2TypeString
 } from "../v2/codap-v2-types"
-import { DIAttribute, DIResources, DISingleValues } from "./data-interactive-types"
+import { DIAttribute, DIGetCaseResult, DIResources, DISingleValues } from "./data-interactive-types"
 import { getCaseValues } from "./data-interactive-utils"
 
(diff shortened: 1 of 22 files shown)


kswenson (approved):
👍 Looks good!

PR #1277, opened 2024-05-23
187651946 v3 DI Get Multidata Plugin Running

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187651946

This PR adds the necessary API handlers to get the Multidata plugin to work in CODAP v3. The following handlers were implemented:
- `get collectionList`
- `update attributeLocation`
  - Implementing this handler required moving code from `column-header-divider.tsx` into `data-set-utils.ts` so it could be reused in `attribute-location-handler.ts`.
  - Though it's not mentioned in the documentation, the API tester plugin has an example of this request where the target collection is specified as "parent". This doesn't actually work in v2. However, it was easy enough to implement it, so I just went ahead and did it in v3.
  - v2 was very open to how the `position` was specified, allowing for fractions and snapping to the end of of the attribute list even when given negative or very large numbers. I've replicated this in v3.
- `create collection`
  - This is a very complex request which has to handle many options and fail cases. I tried to stay true to the v2 implementation, including places where it veered from the documentation, while also catching errors that v2 missed.
  - `name` is now required (in v2 you could create a collection with no name).
  - The ungrouped collection is ignored when determining where to place the new collection in the hierarchy. That means that making a new collection the right-most (not including a `parent`) will put it 2nd right-most. This is incorrect behavior, but the underlying infrastructure will change soon and it should then work properly.
    - One quirk to point out is that the documentation says to use "root" as the `parent` to make the new collection the left-most. However, this isn't quite right. Any `parent` specified that doesn't resolve into an actual collection will result in the new collection being added as the left-most, not just "root". And if there happens to be a collection named "root", using "root" as the `parent` will NOT make the new collection the left most, it will put it right of "root". I made v3 work the same way as v2.
  - v2 allows `attrs` and/or `attributes` to specify attributes to add to the new collection. I've made v3 work the same way as v2.
    - It's possible to create a collection with no attributes. In v2, these collections disappear immediately. In v3, they hang around, even though they're empty. It seems like these should be automatically cleaned up in v3. But possibly complicating this, Multidata will create an empty new collection and then move an existing attribute into it in a separate API request, so it would be important to not remove the collection between these two actions.
  - Even though the v2 documentation says that API requests do not result in CODAP broadcasting notifications, v2 actually does broadcast notifications in at least some circumstances, and some plugins rely on these notifications. Multidata is one of these plugins, relying on a notification after creating a new collection to trigger an update, so `create collection` broadcasts notifications.

One bug fix was also required to get Multidata to work properly: `add attribute` will add the new attribute to a specified collection. Before it would only add it to the ungrouped collection.

These changes got Multidata working pretty well, but I know of one bug and I wouldn't be surprised if there were more lurking somewhere.


v3/src/data-interactive/resource-parser.ts
```diff
@@ -121,9 +121,14 @@ export function resolveResources(
       getSharedDataSets(document).map(sharedDataSet => sharedDataSet.dataSet)
   }
 
+  if ("collectionList" in resourceSelector) {
+    if (dataContext) {
+      result.collectionList = [...Array.from(dataContext.collections), dataContext.ungrouped]
```
kswenson, line 126: Can also use:
```suggestion
      result.collectionList = [...dataContext.collectionModels]
```

kswenson (approved):
👍 Looks good -- just a couple of lint warnings to fix (some of which may require running `lint:build` to see) and one minor simplification.

PR #1279, opened 2024-05-24
187654045 csv import performance

kswenson:
Implements several changes to the way csv files are imported to improve performance. Also fixes our code coverage reports. The performance-related changes for csv imports were:
- Re-implement `DataSet`'s `attrNameMap`.
  - At some point, this was changed to be an MST `view` method under the assumption that the result would be cached. We've learned that the circumstances under which this caching occurs are not always obvious, and this seems to have been one of the cases where caching didn't work as expected. The result was that every attribute name lookup required looping through all attributes, building up the attribute name map, and then returning the desired result from the map. Since this lookup is required for every value of every attribute of every case, the result was an `O(An^2 * Cn)` algorithm with `An` being the number of attributes and `Cn` being the number of cases. (Dan's test document had a large number of attributes as well as a large number of cases.)
  - The fix is to implement our own caching for the `attrNameMap`. Because some clients (e.g. Piotr's formula manager code) rely on being able to observe attribute name changes, the new `attrNameMap` is implemented as a volatile `observable.map`.
- Add a `noInvalidate` option to the `Attribute.setValue` method.
  - Each affected attribute needs to be invalidated when the import occurs, but it doesn't need to be invalidated for each case. The `noInvalidate` options allows the client to handle invalidation once per attribute.
- Add a `canonicalize` option to the `DataSet.addCases` method.
  - Previously, the parsed csv objects were all being converted to canonicalized case objects before being imported. This allows the canonicalization to occur within the `addCases` method obviating the need for the temporary canonicalized representation.
- in `DataSet.addCases`, use `for..in` rather than `Object.entries` to iterate over the cases during import to avoid the creation of the intermediate temporary array of entries.

The attribute name map issue was the major one. The others were just smaller enhancements. The code coverage fix was simply to remove the specification of the `working-directory` from our `codecov-action` configuration. @scytacki had speculated that it was a path-matching issue about whether to prefix src paths with `v3` or not and it seems that the `working-directory` specification was enough to confuse it.

### DIFF
--- a/v3/src/components/case-card/case-card.v2.test.tsx
+++ b/v3/src/components/case-card/case-card.v2.test.tsx
@@ -10,6 +10,7 @@ import "./case-card.v2"
 const { CaseCard } = DG.React as any
 
 describe("CaseCard component", () => {
+  jest.setTimeout(10000)
   const user = userEvent.setup()
 
   // https://github.com/jsdom/jsdom/issues/1590#issuecomment-1379728739
--- a/.github/workflows/v3.yml
+++ b/.github/workflows/v3.yml
@@ -36,7 +36,6 @@ jobs:
         uses: codecov/codecov-action@v4
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
-          working-directory: v3
           flags: jest
   cypress:
     runs-on: ubuntu-latest
@@ -88,7 +87,6 @@ jobs:
         uses: codecov/codecov-action@v4
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
-          working-directory: v3
           flags: cypress
   s3-deploy:
     name: S3 Deploy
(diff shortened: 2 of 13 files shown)


v3/src/components/case-card/case-card.v2.test.tsx
```diff
@@ -10,6 +10,7 @@ import "./case-card.v2"
 const { CaseCard } = DG.React as any
 
 describe("CaseCard component", () => {
+  jest.setTimeout(10000)
```
tealefristoe, line 13: Is this just adding a 10 second wait to our jest tests? 😢

tealefristoe (approved):
Looks good! I'm very happy that patch coverage will be returning 👍

v3/src/components/case-card/case-card.v2.test.tsx
kswenson: It's raising the failure timeout, so if it succeeds in less time it will proceed without waiting. It just means that it will wait 10 sec (rather than 5) before deciding to fail the test based on lack of response. While testing locally, I had it fail the 5-sec timeout a few times so I thought it prudent to extend it. In other words, it's not a `wait` in the cypress sense.

PR #1280, opened 2024-05-28
187690238 v3 DI Collection Requests

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187690238

This PR implements two new request handlers:
`delete collection` - This handler works as expected, except that the ungrouped collection is ignored for the time being, since accommodating it would have been time consuming and it should be removed from v3 in the near future.
`update collection` - This handler allows a plugin to change a collection's `title` or one of its `labels`.
- Currently, labels aren't used for anything in v3, but they do exist in the `CollectionPropsModel`. I had to fill this model out with extra actions to modify the labels to make this handler work. Tests for these additional actions are included in this PR.
- Should `update` be able to change anything in addition to `title` and `labels`? I didn't see anything about it in the documentation, API tester examples, or in the code.

Additionally, `get collection` now returns information about the collection's labels.


kswenson (approved):
👍 Looks good

PR #1281, opened 2024-05-28
chore: update mobx-state-tree to v6.0.0

kswenson:
- Updated [our fork of mobx-state-tree](https://github.com/concord-consortium/mobx-state-tree) to the recently released v6.0.0 from upstream and published a new `6.0.0-cc.1` version.
  - I didn't create a separate PR there, but see the [v6.0.0-cc.x branch](https://github.com/concord-consortium/mobx-state-tree/tree/v6.0.0-cc.x) where I cherry-picked our local changes.
- Updated CODAP v3 to use the newly published version. Most of the code changes required resulted from these two changes in MST v6.0.0:
  - `getEnv()` now throws an exception if called in the absence of an environment (rather than returning an empty object). Requires use of `hasEnv()` to pre-flight calls to `getEnv()`.
  - Changes to union typing which required some changes to adornment clients.

### DIFF
--- a/v3/src/components/graph/adornments/adornments-base-store.ts
+++ b/v3/src/components/graph/adornments/adornments-base-store.ts
@@ -2,7 +2,9 @@ import {Instance, types} from "mobx-state-tree"
 import {getAdornmentComponentInfo} from "./adornment-component-info"
 import {AdornmentModelUnion, kDefaultFontSize} from "./adornment-types"
 import {IAdornmentModel, IUpdateCategoriesOptions} from "./adornment-models"
+import {IMovableValueAdornmentModel} from "./movable-value/movable-value-adornment-model"
 import {kMovableValueType} from "./movable-value/movable-value-adornment-types"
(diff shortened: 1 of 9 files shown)


v3/src/components/graph/adornments/adornments-base-store.ts
```diff
@@ -17,6 +19,18 @@ export const AdornmentsBaseStore = types.model("AdornmentsBaseStore", {
...
+    return !!self.adornments.find(a => a.type === type)?.isVisible
+  },
+  get activeUnivariateMeasures() {
+    return self.adornments.filter(adornment => adornment.isUnivariateMeasure && adornment.isVisible) as
+            IUnivariateMeasureAdornmentModel[]
+  },
+  findAdornmentOfType<T extends IAdornmentModel = IAdornmentModel>(type: string): T | undefined {
+    return self.adornments.find(adornment => adornment.type === type) as T
+  }
+}))
```
kswenson, line 33: Note that these are not new methods. They've just moved so that they can be more easily called internally.

tealefristoe (approved):
Looks good 👍

PR #1282, opened 2024-05-29
187651946 v3 DI Fix Multidata

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187651946

This PR starts to address some issues with the Multidata plugin in v3 discovered during testing.
- Notifications are now broadcast when an attribute is moved in response to an API request. Multidata relies on these notifications to update its representation of the dataset when a user moves an attribute in the plugin.
- Infrastructure has been added to handle `update dataContext` requests with `sort attribute`. Since sorting attributes hasn't been implemented in v3 yet, the request isn't fully handled, but it should be easy to finish once it is possible to sort attributes.


kswenson (approved):
👍 LGTM

PR #1284, opened 2024-05-30
187692388 v3 DI Item Requests

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187692388

This PR implements handlers for several similar item related API requests:
`delete`, `get`, and `update item`
`delete`, `get`, and `update itemByID`
`delete`, `get`, and `update itemByCaseID`

These requests mostly work the same aside from how they reference the target item, so most of the work here happens in common utility functions and the `resourceParser`. The main exception to this rule is `update item`, which allows multiple items to be updated if no index is specified (`item` requests generally reference the target item by index).

One major departure to note is with the responses for all of the `update` requests. In v2, responses to most `get` requests look like this:
```
{
  "success": true,
  "values": {
    "createdCases": [],
    "deletedCases": []
  }
}
```
That's right, no `updatedCases`, which is what the request actually does. And there are always no created or deleted cases, because the request never creates or deletes cases. And these are _item_ handlers, not _case_ handlers, so it's a little strange that we get `createdCases` and not `createdItems`.

Meanwhile, `update item` when updating multiple items gives a response like this:
```
{
  "success": true,
  "changedCases": [
    13,
    14
  ],
  "deletedCases": [],
  "createCases": []
}
```
Here you can see that `changedCases` is included (not sure why it isn't called `updatedCases`), but note that these fields are not wrapped in a `values` object. Also, there's `createCases` instead of `createdCases` (missing "d").

I couldn't bring myself to replicate these inconsistencies. So instead, I made all successful `update` requests give a response like this:
```
{
  "success": true,
  "values": {
    "changedCases": [13],
    "createdCases": [],
    "deletedCases": []
  }
}
```
There are still some major quirks with this, but at least it's consistent within `update item` requests, and general `update` requests now contain some information. Given that the general `update` request responses didn't contain any relevant information previously, I doubt any plugins will care that it has changed.


v3/src/data-interactive/handlers/handler-functions.ts
```diff
@@ -0,0 +1,130 @@
...
+  if (!aCase) return caseNotFoundResult
+
+  const pseudoCase = dataContext.pseudoCaseMap.get(aCase.__id__)
+  const cases = pseudoCase?.childCaseIds ?? [aCase.__id__]
+
+  dataContext.applyModelChange(() => {
+    dataContext.removeCases(cases)
+  })
+
+  return { success: true }
```
kswenson, line 23: `deleteCaseBy` doesn't return the deleted ids?

kswenson (approved):
👍 Looks good

v3/src/data-interactive/handlers/handler-functions.ts
tealefristoe: It just returns `{ success: true }` in v2. It would be easy to add the deleted ids, and it's very unlikely that would break any plugins.

kswenson: Seems like we generally do elsewhere, so consistency would argue that we should do so here as well.

PR #1285, opened 2024-05-30
Fix for bug found in prior PR for specifying num stdErrors

bfinzer:
[#187626777] Feature: User can specify how many standard errors are shown in the standard error bar

* This is a fix for a bug caught in a prior PR whereby the label and text tip for standard error was not being updated when the number of standard errors was changed. The fix is to make it possible to invalidate measures maintained by UnivariateMeasureAdornmentModels and to do this invalidatation in the StandardErrorAdornmentModel when the number of standard errors is changed.


emcelroy (approved):
Looks good to me 👍

PR #1286, opened 2024-05-30
feat: Graph histogram (PT-186948982)

emcelroy:
[#186948982](https://www.pivotaltracker.com/story/show/186948982)

When points are grouped into bins in a univariate plot, the plot can be converted to a histogram by checking the "Fuse Dots into Bars" checkbox.

These changes add a new `Histogram` component that is a sibling of `BinnedDotPlotDots` and `FreeDotPlotDots`. The three components share to some extent two new hooks: `useDotPlot` and `useDotPlotResponders`, and the new dot-plot-utils.ts. The `Histogram` component also shares the new bar-utils.ts file with the `DotChartBars` component. (An earlier version of these changes had the new `Histogram` component using `useChartDots` but that is no longer necessary. )


v3/src/components/data-display/data-display-types.ts
```diff
@@ -7,7 +7,7 @@ export type Point = { x: number, y: number }
 export type CPLine = { slope: number, intercept: number, pivot1?: Point, pivot2?: Point }
 export const kNullPoint = {x: -999, y: -999}
 
-export const PointDisplayTypes = ["points", "bars", "bins"] as const
+export const PointDisplayTypes = ["points", "bars", "bins", "histogram"] as const
```
emcelroy, line 10: One thing I'm especially looking for thoughts on is the choice to add `histogram` as an option here. Does that seem reasonable? It seemed to provide the most straightforward way to check when the histogram display is active. In theory, you could instead use a combination of the `bins` option and the separate `graphModel.pointsFusedIntoBars`, but I found that more difficult to work with.

v3/src/components/data-display/models/data-display-content-model.ts
```diff
@@ -43,6 +43,14 @@ export const DataDisplayContentModel = TileContentModel
...
       return isNumericAxisModel(axisModel) && self.pointDisplayType !== "bins"
     },
+    hasHistogramAxis() {
+      // derived models should override
+      return false
+    },
+    histogramAxisTicks(): { tickValues: number[], tickLabels: string[] } {
+      // derived models should override
+      return {tickValues: [], tickLabels: []}
+    },
```
emcelroy, line 53: Another thing I'm especially wondering about is the names for these methods. It seems to me that the histogram does require a special numeric axis where the ticks are equivalent to the bin boundaries in the binned plot. They are different from the binned plot's axis (which uses `nonDraggableAxisTicks`) in that the axis should be draggable and have different tick values. These methods get used in the Axis world, though, so should probably have less graph-centric names? I haven't been able to think of anything better, though.

bfinzer:
Some bugs (observed with `Top_Speed` on x-axis:
1. I Group into Bins and then fuse the dots, but the Count axis is not correctly scaled.
2. The histogram doesn't respond to changes I make to the count axis through dragging.
3. Pressing the graph's "Rescale" causes the Count axis numbers to disappear and the histogram bars do not change to fit within the plot area.
4. I should be able to rescale the x-axis by dragging, but I'm unable to do so.
5. As I hover and move over the histogram bars, white horizontal lines appear between the rectangles for the cases.

v3/src/components/data-display/models/data-display-content-model.ts
bfinzer: As we discussed on Slack, the histogram's numeric axis is perfectly ordinary, just providing a mapping from "world" values to "pixel" values.

v3/src/components/data-display/data-display-types.ts
bfinzer: I agree that it could be helpful to add "histogram" as a PointDisplayType

bfinzer (changes requested):
I can do a more thorough code review once the bugs I commented on are fixed.

emcelroy:
> I can do a more thorough code review once the bugs I commented on are fixed.

@bfinzer Thanks. I have fixes for #1, #4, and #5, and I'm halfway through fixing #2. I'm going to have to stop for now, though, and continue on Monday.

emcelroy:
@bfinzer @kswenson I've made some changes that I believe fix most of the reported bugs. Some notes/caveats below.

> Some bugs (observed with `Top_Speed` on x-axis:
> 
> 1. I Group into Bins and then fuse the dots, but the Count axis is not correctly scaled.

Fixed.

> 2. The histogram doesn't respond to changes I make to the count axis through dragging.

Fixed.

> 3. Pressing the graph's "Rescale" causes the Count axis numbers to disappear and the histogram bars do not change to fit within the plot area.

I think this should be a separate story since the problem seems to already exist. For example, try rescaling after you use the "Fuse Dots into Bars" option with a categorical attribute on the x axis. The same thing happens. It would be nice to fix that, of course, but since I'm running low on time I think we should skip that for now.

> 4. I should be able to rescale the x-axis by dragging, but I'm unable to do so.

Fixed.

> 5. As I hover and move over the histogram bars, white horizontal lines appear between the rectangles for the cases.

Fixed.

v3/src/components/graph/components/graph.scss
```diff
@@ -171,10 +171,10 @@
...
 .bar-cover {
   cursor: pointer;
-  fill: white;
-  opacity: 0;
+  fill: rgba(255, 255, 255, 0);
+  transition: 1s;
 
   &.active {
-    opacity: 0.3;
+    fill: rgba(255, 255, 255, 0.3);
```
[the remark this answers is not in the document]
emcelroy: That would be good, but it's unfortunately not as easy as you might think. There's [a PT story about data tips and bars](https://www.pivotaltracker.com/story/show/187409553) that describes an issue that complicates things. It's in the "Background & Notes" section and starts with "Modify DataTips so its `showTip` and `hideTip` methods are accessible from other components..." 

So I think that should be done, but as a separate piece of work from this PR.

v3/src/components/graph/components/inspector-panel/display-config-palette.tsx
```diff
@@ -62,12 +64,20 @@ export const DisplayConfigPalette = observer(function DisplayConfigPanel(props:
...
-    if (fuseIntoBars) {
-      graphModel?.pointDescription.setPointStrokeSameAsFill(true)
-    } else {
-      graphModel?.pointDescription.setPointStrokeSameAsFill(false)
-    }
+    const [undoStringKey, redoStringKey] = fuseIntoBars
+      ? ["DG.Undo.graph.fuseDotsToRectangles", "DG.Redo.graph.fuseDotsToRectangles"]
+      : ["DG.Undo.graph.dissolveRectanglesToDots", "DG.Redo.graph.dissolveRectanglesToDots"]
+
+    graphModel?.applyModelChange(() => {
```
emcelroy, line 71: While testing things I realized I had left out adding undo/redo support for this handler. I tried to likewise add it for `handleSelection` above, but ran into some weird problems and decided to skip that for this PR.

v3/src/components/graph/components/graph.scss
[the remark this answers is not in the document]
bfinzer: Understood

bfinzer:
A bug also for data tips: With `Top_Speed` on the x-axis and `Type` as the legend attribute, hovering over a bin shows the correct count but incorrect bounds. For example, hovering over the bin from 80 to 100 shows "8 of 146 5.5 (80%) are ≥ 100 and <" In V2 we see "8 of 146 (5.5%) are ≥ 80 and < 100". Easy fix?

bfinzer (approved):
👍🏻LGTM
* In playing around, I encountered quite a few bugs. Given that you have to move on to another project, I approve this PR and I'll log bugs against the histogram.
* I did comment on a couple having to do with data tips that *might* be easy to fix. Use your discretion.

v3/src/components/graph/components/histogram.tsx
```diff
@@ -0,0 +1,143 @@
...
+import { useDotPlotResponders } from "../hooks/use-dot-plot-responders"
+import { renderBarCovers } from "../utilities/bar-utils"
+import { SubPlotCells } from "../models/sub-plot-cells"
+
+export const Histogram = observer(function Histogram({ abovePointsGroupRef, pixiPoints }: PlotProps) {
+  const { dataset, dataConfig, graphModel, isAnimating, layout, getPrimaryScreenCoord, getSecondaryScreenCoord,
+          pointColor, pointStrokeColor, primaryAttrRole, primaryAxisScale, primaryIsBottom, primaryPlace,
+          refreshPointSelection, secondaryAttrRole } = useDotPlot(pixiPoints)
+  const barCoversRef = useRef<SVGGElement>(null)
+  const pointDisplayType = "bars"
```
kswenson, line 20: Not "histogram"? Seems like a comment is warranted here.

v3/src/components/graph/components/histogram.tsx
```diff
@@ -0,0 +1,143 @@
...
+      primaryIsBottom, primaryPlace, secondaryAttrRole])
+
+  usePlotResponders({pixiPoints, refreshPointPositions, refreshPointSelection})
+  useDotPlotResponders(refreshPointPositions)
+
+  useEffect(() => {
+    if (pixiPoints) {
+      pixiPoints.pointsFusedIntoBars = graphModel.pointsFusedIntoBars
+    }
+  }, [pixiPoints, graphModel.pointsFusedIntoBars])
```
kswenson, line 119: Seems like you could handle setting `pixiPoints.pointsFusedIntoBars` in the `handleFuseIntoBars` `mstAutorun` below and eliminate this `useEffect` entirely.

v3/src/components/graph/components/histogram.tsx
```diff
@@ -0,0 +1,143 @@
...
+    setPointCoordinates({
+      pointRadius: graphModel.getPointRadius(),
+      selectedPointRadius: graphModel.getPointRadius("select"),
+      pixiPoints, selectedOnly, pointColor, pointStrokeColor, getWidth, getHeight,
+      getScreenX, getScreenY, getLegendColor, getAnimationEnabled: isAnimating,
+      pointDisplayType, dataset, pointsFusedIntoBars: true
+    })
+  }, [abovePointsGroupRef, dataConfig, dataset, getPrimaryScreenCoord, getSecondaryScreenCoord, graphModel,
+      isAnimating, layout, pixiPoints, pointColor, pointStrokeColor, primaryAttrRole, primaryAxisScale,
+      primaryIsBottom, primaryPlace, secondaryAttrRole])
```
kswenson, line 110: Because `getPrimaryScreenCoord` and `getSecondaryScreenCoord` are unstable (as pointed out elsewhere), `refreshPointPositions` will be re-instantiated on every render as well.

v3/src/components/graph/hooks/use-dot-plot-responders.ts
```diff
@@ -0,0 +1,36 @@
...
+  // Respond to binAlignment and binWidth changes. We include both the volatile and non-volatile versions of these
+  // properties. Changes to the volatile versions occur during bin boundary dragging and result in the appropriate
+  // behavior during a drag. Changes to the non-volatile versions occur when a drag ends (or the user sets the bin
+  // and alignment values via the form fields) and result in the behavior required when bin boundary dragging ends.
+  useEffect(function respondToGraphBinSettings() {
+    return mstReaction(
+      () => [graphModel._binAlignment, graphModel._binWidth, graphModel.binAlignment, graphModel.binWidth],
+      () => refreshPointPositions(false),
+      {name: "respondToGraphBinSettings", equals: comparer.structural}, graphModel)
+  }, [dataset, graphModel, refreshPointPositions])
```
kswenson, line 22: As mentioned elsewhere, `refreshPointPositions` changes on every render, so this reaction will be destroyed and re-installed on every render.

v3/src/components/graph/hooks/use-dot-plot-responders.ts
```diff
@@ -0,0 +1,36 @@
...
+  useEffect(function setInitialBinSettings() {
+    if (!dataConfig) return
+    if (graphModel.binWidth === undefined || graphModel.binAlignment === undefined) {
+      const { binAlignment, binWidth } = graphModel.binDetails({ initialize: true })
+      graphModel.applyModelChange(() => {
+        graphModel.setBinWidth(binWidth)
+        graphModel.setBinAlignment(binAlignment)
+      })
+    }
+  })
```
kswenson, line 35: Probably doesn't matter in this case, but was leaving off the dependencies intentional? Because it bails quickly it should be fine if it runs after every render, but you could minimize the number of times it runs with something like:
```suggestion
  }, [dataConfig, graphModel, graphModel.binAlignment, graphModel.binWidth])
```

kswenson (approved):
👍 LGTM -- I made a number of comments/suggestions, some of which should be simple to address, others of which may not be. Given that you're moving on to another project, you may leave non-straightforward issues for another day, although perhaps a TODO comment could be added in appropriate places.

v3/src/components/graph/hooks/use-dot-plot-responders.ts
emcelroy: @kswenson Yes, I left off the dependencies because I believe this should only ever have to run once when a document is imported. There's a note about that in the comment that precedes the `useEffect`. Does that seem reasonable?

v3/src/components/graph/components/inspector-panel/display-config-palette.tsx
```diff
@@ -62,12 +64,20 @@ export const DisplayConfigPalette = observer(function DisplayConfigPanel(props:
...
+        graphModel?.setPointsFusedIntoBars(fuseIntoBars)
+        if (fuseIntoBars) {
+          graphModel?.pointDescription.setPointStrokeSameAsFill(true)
+        } else {
+          graphModel?.pointDescription.setPointStrokeSameAsFill(false)
+        }
```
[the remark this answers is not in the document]
emcelroy: Yes, thanks for catching that. I'll fix it.

v3/src/components/graph/hooks/use-dot-plot.ts
```diff
@@ -0,0 +1,128 @@
...
+  const extraPrimaryAxisScale = layout.getAxisScale(extraPrimaryPlace) as ScaleBand<string>
+  const numExtraPrimaryBands = Math.max(1, extraPrimaryAxisScale?.domain().length ?? 1)
+  const extraSecondaryRole = primaryIsBottom ? "rightSplit" : "topSplit"
+  const extraSecondaryAttrID = dataConfig?.attributeID(extraSecondaryRole) ?? ""
+  const primaryAttrID = dataConfig?.attributeID(primaryAttrRole) ?? ""
+  const primaryPlace: AxisPlace = primaryIsBottom ? "bottom" : "left"
+  const primaryAxisScale = layout.getAxisScale(primaryPlace) as ScaleLinear<number, number>
+  const pointDiameter = 2 * graphModel.getPointRadius()
+  const secondaryAttrID = dataConfig?.attributeID(secondaryAttrRole) ?? ""
+  const secondaryAxisScale = layout.getAxisScale(secondaryPlace) as ScaleBand<string>
```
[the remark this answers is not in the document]
emcelroy: That's a very sensible suggestion but I tried it and it causes things to break for somewhat mysterious reasons. I'm going to skip this change, but will add a comment about how we should be using `getBandScale` here instead of `getAxisScale`, why we're not yet, etc.

PR #1289, opened 2024-05-31
187709040 v3 DI Item Search Requests

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187709040

This PR adds handlers for `delete` and `get itemSearch` API requests. The majority of the work took place in the `ResourceParser` and related utility functions, which parse the search query and find the appropriate items.

### DIFF
--- a/v3/src/data-interactive/resource-parser-utils.ts
+++ b/v3/src/data-interactive/resource-parser-utils.ts
@@ -11,7 +11,7 @@ export function parseSearchQuery(query: string, dataContext?: IDataSet): DIParse
   
   const parseOperand = (_rawValue: string) => {
     // Trim whitespace
-    const rawValue = _rawValue.replace(/^\s+|\s+$/, '')
+    const rawValue = _rawValue.replace(/^\s+|\s+$/g, '')
 
     const numberValue = Number(rawValue)
     const value = rawValue === "true" ? true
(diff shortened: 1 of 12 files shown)


v3/src/data-interactive/resource-parser-utils.ts
```diff
@@ -0,0 +1,41 @@
+import { IDataSet } from "../models/data/data-set"
+import { DIParsedQuery, DIQueryFunction } from "./data-interactive-types"
+
+export function parseSearchQuery(query: string, dataContext?: IDataSet): DIParsedQuery {
+  if (query === "*") {
+    return { valid: true, func: () => true }
+  }
+
+  const matches = query.match(/([^=!<>]+)(==|!=|<=|<|>=|>)([^=!<>]+)/)
```
kswenson, line 9: I assume this parsing code is taken from v2?

kswenson (approved):
👍 Looks good

v3/src/data-interactive/resource-parser-utils.ts
tealefristoe: Yes. Also on line 14, though there was a bug in that code in v2 that chatGPT helped me fix 👍  Should I add a comment mentioning where this code came from?

kswenson: Seems like it could be useful.

PR #1293, opened 2024-06-03
fix: handling of hidden attributes in case table

kswenson:
[[PT-187637670]](https://www.pivotaltracker.com/story/show/187637670)


tealefristoe (approved):
Looks good 👍

PR #1299, opened 2024-06-06
Fix marquee selection in subplot bug

bfinzer:
[#187728058] Bug fix: Marquee selection shows up in wrong subplot

* This bug was introduced when we switched to PixiJS. This fix is in `onDrag` in `Background` where we now simply use the boundingClientRect to compute the marquee coordinates from the mouse event coordinates.


kswenson (approved):
👍 LGTM

PR #1301, opened 2024-06-07
Normal curve for histograms

bfinzer:
This PR covers most situations for plotting normal curves on histograms, but there are probably some wrinkles to be worked out when categorical attributes are on the top and right. These are difficult to diagnose until histograms behave better when split on top and right.

[#187751261] Feature: User can plot a normal curve on top of a histogram

Fortunately, it just takes a few tweaks to normal-curve-adornment-component.tsx to get the curve to use the count axis and histogram bin width in place of the dots and overlap to get this to work.
* Also fixed a dependency error in background.tsx


kswenson (approved):
👍 LGTM

PR #1302, opened 2024-06-07
Fix missing attribute label bug

bfinzer:
When adding an attribute to the y-axis of a graph that has an attribute on the x-axis, the y-attribute label was not showing.

[#187744843] Bug fix: Axis attribute label not showing

* The GraphAttributeLabel component was unnecessarily complicated. Simplifying it by getting rid of the `setupTitle` useEffect solved the problems

It's worth noting that the d3 'join' pattern is used here, but is really unnecessarily complicated since there is exactly one element that needs to be created and maintained. But, in the interest of fixing the bug and moving on, I didn't change/simplify the existing architecture.


kswenson (approved):
👍 LGTM

PR #1303, opened 2024-06-08
Fix crashing bug on graph x-y switch

bfinzer:
[#187757747] Bug fix: Drag-switching x-y attributes is broken

* This bug was caused by an inconsistency in the attribute descriptions were being returned from the graph's data configuration. The claim in `attributeDescriptionForRole` was that we would return a snapshot of an `AttributeDescription`. But for y attributes, which are stored in an array, we were actually returning the `AttributeDescription` model and this was causing problems for MST in the circumstance in which the user is switching attributes between x and y. We fix this by converting the y-attribute description to a snapshot before we return it.
* As a convenience we also introduce `safeGetSnapshot` which checks for `undefined` before converting its argument to a snapshot.


kswenson (approved):
👍 LGTM

PR #1304, opened 2024-06-10
fix: z-index for resizing components

kswenson:
[[PT-187751757]](https://www.pivotaltracker.com/story/show/187751757)

The bug was introduced with the recent z-index changes to fix the iframe reload bug. The z-index wasn't being propagated correctly for the resizing component.


tealefristoe (approved):
👍

PR #1305, opened 2024-06-11
Fix for graph rehydration bug

bfinzer:
[#187757811] Bug fix: Rehydrating graph problem

* The symptom was an axis that retained multiple sub-axes when a document was loaded that had a variation on the instance of the graph tile present. The fix is to reset any `GraphLayout` axis scales that aren't being set up by an axis model that is brought in.


kswenson (approved):
👍 LGTM

PR #1306, opened 2024-06-13
187737737 v3 DI caseSearch

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187737737

This PR adds a handler for `get caseSearch` API requests.


v3/src/data-interactive/data-interactive-types.ts
```diff
@@ -136,7 +136,7 @@ export interface DIResources {
   caseByID?: ICase
   caseByIndex?: ICase
   caseFormulaSearch?: DICase[]
-  caseSearch?: DICase[]
+  caseSearch?: CaseGroup[]
```
kswenson, line 139: `CaseGroup` is more of an internal implementation type that is going to change significantly as part of the refactor I'm in the middle of. I think this should use a custom DI type that defines the information required by the API instead. It appears that the case search handler only uses the `pseudoCase` property of the `CaseGroup`, and then the `getCaseRequestResultValues` function only uses the `__id__` of the case, so it would seem that a much simpler type would suffice here -- perhaps even the `DICase` type that was here before.

kswenson (changes requested):
👍 Implementation is fine -- I made a suggestion about the typing of the `caseSearch` property.

kswenson (approved):
👍 LGTM

PR #1307, opened 2024-06-13
Improve No Cases Message

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187309840

This PR improves the "No Cases" temporary message in a case table.
- It no longer overlaps with attribute names.
- It is now reactive, so it will immediately disappear when cases get added to the dataset.


kswenson (approved):
👍 LGTM -- This message will go away once the input row is implemented, but until then it might as well behave itself. 😉

eireland (approved):
👍 LGTM

PR #1308, opened 2024-06-13
Ensure New Collections Have Unique Names

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187638145

This PR ensures new collections have unique names by appending a number to the end of a new collection's name if it shares a name with an existing collection. This is the same way v2 handles this problem.

Allowing duplicate names was causing a bug in the Choosy plugin, which was the motivation for fixing this issue.

I made it so whenever an attribute is added to a dataset, its name is ensured to be unique. But I wasn't sure if this was the correct place to make this intervention, so something to think about when looking over this PR.


kswenson (approved):
👍 LGTM

PR #1309, opened 2024-06-13
187793171 Prevent Automatic Action from being Undoable

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187793171

This PR prevents an automatic `setTransitionComplete` call from being undoable. This was causing a problem because the automatic call was happening after undoing an attribute assignment, wiping out the undo stack moving forward, preventing the action from being redoable.

### DIFF
--- a/v3/src/components/container/free-tile-component.tsx
+++ b/v3/src/components/container/free-tile-component.tsx
@@ -139,7 +139,8 @@ export const FreeTileComponent = observer(function FreeTileComponent({ row, tile
   // the tile's dimensions. To prevent this, we add a transitionend handler that sets a flag on the tile model when the
   // transition completes. Child components can check this flag to avoid or counteract premature application of effects.
   useEffect(function addTransitionEndHandler() {
-    const handleTransitionEnd = () => tile.setTransitionComplete(true)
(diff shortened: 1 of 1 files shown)


emcelroy (approved):
Looks good to me 👍

kswenson:
FYI: I would prefer a solution that made the `transitionComplete` property `volatile`, since it doesn't need to be serialized, and then this wouldn't be necessary.

tealefristoe:
@kswenson You can't set a volatile property in the snapshot, can you? I'm setting `transitionComplete` in a snapshot in https://github.com/concord-consortium/codap/pull/1294 to make a new map component zoom and center properly. Maybe there's a better solution there, but something to consider when changing the status of `transitionComplete`.

kswenson:
> @kswenson You can't set a volatile property in the snapshot, can you? I'm setting `transitionComplete` in a snapshot in #1294 to make a new map component zoom and center properly. Maybe there's a better solution there, but something to consider when changing the status of `transitionComplete`.

No, you can't, so you're right that that will be a wrinkle to consider in my plan to make it volatile.

PR #1312, opened 2024-06-20
187737791 DI Handle Special Collaborative Requests

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187737791

This PR implements a couple of API requests specifically required by the Collaborative plugin. The API requests are simple, but the underlying functionality had not yet been added to v3, so this PR adds these features to the models and UI, which constitutes the bulk of this PR.

`update dataContext handlingController`
This command links a `dataSet` to a `webView`, allowing the following settings of the `webView` apply to the `dataSet`.

`update interactiveFrame preventAttributeDeletion`
Setting this to `true` will make a table with a `dataSet` associated with the `interactiveFrame` have its Delete Attribute attribute menu item disabled.

`update interactiveFrame respectEditableItemAttribute`
Setting this to `true` will modify the behavior of any `dataSet` associated with the `interactiveFrame` when a special attribute named `__editable__` has a truthy value for an item:
- The item cannot be deleted with the Delete Selected Cases, Delete Unselected Cases, or Delete All Cases trash menu options.
- Cells for this item cannot be modified in the table.
- NOTE: v2 has special code to handle an `__editable__` attribute with a formula. I didn't implement anything special for this in v3. Does anything special need to happen here?

General notes:
- There are several important functions in this PR that are included in `collaborator-utils.ts`. It would have felt more natural to make these views on the `DataSet` model, but this wasn't possible without adding import cycles.
- The table trash menu now has its Delete Selected/Unselected/All Cases options disabled when there are no cases that can be deleted with that option. For example, Delete Selected Cases is disabled if there are no selected cases. This happens all the time, even if the collaborative specific features are not being used.


v3/src/components/web-view/collaborator-utils.ts
```diff
@@ -0,0 +1,45 @@
+// Functions used by the Collaborative plugin
+
```
bfinzer, line 2: I think this file deserves a comment, especially about the __editable__ attribute functionality.

v3/src/components/web-view/collaborator-utils.ts
```diff
@@ -0,0 +1,45 @@
...
+import { appState } from "../../models/app-state"
+import { IDataSet } from "../../models/data/data-set"
+import {
+  isWebViewModel, kDefaultPreventAttributeDeletion, kDefaultRespectEditableItemAttribute
+} from "./web-view-model"
+
+function getManagingController(dataset: IDataSet) {
+  const { content } = appState.document
+  const tile = content?.getTile(dataset.managingControllerId)
```
bfinzer, line 11: Comment about managingControllerId? What is it?

v3/src/data-interactive/resource-parser-utils.ts
```diff
@@ -41,3 +44,13 @@ export function parseSearchQuery(query: string, dataContextOrCollection?: IDataS
   
   return { valid, left, right, func }
 }
+
+export function findTileFromV2Id(v2Id: string) {
+  const { document } = appState
+  // We look for every possible v3 id the component might have (because each tile type has a different prefix).
+  // Is there a better way to do this?
```
bfinzer, line 51: I don't know if there's a better way. Probably worth asking Kirk to weigh in when he returns.

bfinzer (approved):
👍🏻LGTM
I don't _think_ anything special is need for the __editable__ attribute with a formula but you might have Kirk weigh in, as with the question about findTileFromV2Id left over from previous work. You don't need to hold up a merge; just queue questions up for Kirk.

v3/src/data-interactive/resource-parser-utils.ts
tealefristoe: This function was introduced in a previous PR, so Kirk has seen it and gave it a 👍 previously. It's not the most elegant function, but since the number of tile types is so low at least it's not terribly inefficient.

PR #1315, opened 2024-06-21
Gaussian Fit

bfinzer:
[#187776062] Feature: CODAP has URL parameters for gaussianFit and displays it for histogram

* We store 'gaussianFitEnabled' as a volatile property of `DocumentContentModel` and return true if it is set as a url parameter or if the volatile property has been set through the plugin API
* We pass a boolean parameter `useGaussianOptions` down to `getAdornmentsMenuItemsFromTheStore` so we can modify the ruler menu as needed to indicate gaussian fit instead of normal curve
* We bring `fitGaussianGradientDescent` from V2 code to math-utils to fit a mean and standard deviation to histogram bars
* In `NormalCurveAdornmentComponent` we detect the condition under which we should use the mean and standard deviation from a gaussian fit rather the sample mean and sample standard deviation as parameters for the normal curve
* In `NormalCurveAdornmentModel` we add `computeHistogram` to find the center and count of each histogram bin
* Created use-graph-options.ts as a place for a hook that enables normal-curve-adornment-registration.tsx to decide whether to use gaussian fit
* The label and data tip for a gaussian fit includes a title indicating that it is such
* Laying out the measure labels in the presence of a gaussian fit adornment requires some special handling and addition of `labelLines` to `AdornmentModel`


eireland (approved):
👍 LGTM

PR #1316, opened 2024-06-22
Regain ability to plot 2 y attributes in scatterplot

bfinzer:
[#187841073] Bug fix: Unable to add second y-attribute to scatterplot

* The recent refactoring of `GraphController` left out a call to `CallMatchCirclesToData` in `handleAttributeAssignment`

This definitely fixes the bug. Is it the best way?


tealefristoe (approved):
Looks good 👍

tealefristoe (changes requested):
Actually, it looks like a jest test broke with this change. I suspect it will be a quick fix.

tealefristoe (approved):
👍

PR #1317, opened 2024-06-24
187718437 v3 DI Create Component Fixes

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187718437

This PR fixes bugs with the original `create component` PR.
- New graphs now have their `primaryRoles` set properly.
- Attempting to assign an attribute to a role it can't fulfill (like a categorical attribute to `rightNumeric`) will now result in a `create component graph` request failing.
  - Note that there are many similar cases where attempts at illegal assignments are simply ignored. It would be a major undertaking to check every possible option in the API.
- Singleton tiles have their `titles` updated when they are displayed.


bfinzer (approved):
👍🏻LGTM

PR #1318, opened 2024-06-25
187849639 v3 DI Multidata Case Cycles

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187849639

This PR fixes an error introduced with the recent changes to cases and items which resulted in an infinite loop in the Multidata plugin.

The immediate issue was that cases in the child-most collection included themselves as children. This was causing Multidata to eternally request information about these cases, which caused the plugin to hang up and remain blank.

Ultimately, the problem is that cases in the child-most collection are treated differently than other cases. In the codebase, they are items, not cases, which means that any time a developer is working with a case, they must check to see if they're working with an actual case or an item. I suspect there are other bugs related to this issue, and I have no doubt it will cause confusion in the future.

I also took this opportunity to correct the names of a few `DataSet` views. We should strive to use "case" and "item" consistently throughout the code. There are still many places where "case" is incorrectly used instead of "item".

### DIFF
--- a/v3/src/components/data-display/models/data-configuration-model.ts
+++ b/v3/src/components/data-display/models/data-configuration-model.ts
@@ -562,7 +562,7 @@ export const DataConfigurationModel = types
       // whether the value changes result in adding/removing any cases from the filtered set
       // a single call to setCaseValues can result in up to three calls to the handlers
       if (cases.added.length) {
-        const newCases = self.dataset?.getCases(cases.added)
+        const newCases = self.dataset?.getItems(cases.added)
         self.handlers.forEach(handler => handler({name: "addCases", args: [newCases]}))
       }
       if (cases.removed.length) {
--- a/v3/src/components/graph/components/scatter-plot-utils.ts
+++ b/v3/src/components/graph/components/scatter-plot-utils.ts
@@ -50,7 +50,7 @@ export function scatterPlotFuncs(layout: GraphLayout, dataConfiguration?: IGraph
     const xValue = getXCoord(caseID)
     const yValue = getYCoord(caseID, plotNum)
     if (isFinite(xValue) && isFinite(yValue)) {
-      const caseData = dataset?.getCase(caseID, { numeric: false })
+      const caseData = dataset?.getItem(caseID, { numeric: false })
       if (caseData) {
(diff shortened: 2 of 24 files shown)


bfinzer (approved):
👍🏻LGTM
We talked about this PR consisting of a bug fix plus the rename of three functions. The renaming exposes ambiguities where the entity being computed is clearly an item or array of items, but the name assigned to it makes it seem like a case. And Teale will write a story about going through all the code to make things clearer.

kswenson:
> The immediate issue was that cases in the child-most collection included themselves as children. This was causing Multidata to eternally request information about these cases, which caused the plugin to hang up and remain blank.

It is factually correct that the cases of the child-most collection should contain their underlying items as children in the `childItemIds`, but _not_ in `childCaseIds`. If the sharing of ids between cases and items is causing issues here then we may need to reconsider that.

> Ultimately, the problem is that cases in the child-most collection are treated differently than other cases. In the codebase, they are items, not cases, which means that any time a developer is working with a case, they must check to see if they're working with an actual case or an item. I suspect there are other bugs related to this issue, and I have no doubt it will cause confusion in the future.
> 
> I also took this opportunity to correct the names of a few `DataSet` views. We should strive to use "case" and "item" consistently throughout the code. There are still many places where "case" is incorrectly used instead of "item".

It is certainly true that we should be clear in our use of "case" and "item" and that that work was not complete as of my latest PR on the subject. I also haven't thought deeply about the changes in this PR but I'm not sure they're based on a full appreciation of these relationships.

1. The cases in the child-most collection _are_ cases and they are separate from items. That was a mistake in the prior implementation that my latest PR was trying to fix.
2. Cases in the child-most collection should reference their underlying items in the same way all other cases in other collections do, with the proviso that for the child-most collection the relationship is one-to-one.
3. Internal clients of the DataSet, e.g. graphs, tables, etc. are generally interested in cases, not items.
4. Only the DI API deals directly with items.
5. For implementation convenience, cases of the child-most collection and items are sometimes interchangeable, e.g. they share the same id, but that does not mean they are always interchangeable.
6. My expectation was that the full set of case-related API would remain in tact and that additional item-specific API might be necessary.

PR #1319, opened 2024-06-25
187849622 v3 DI Dataset Notifications

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187849622

This PR fixes a couple of issues with the Multidata plugin:
- It broadcasts `dataContextCountChanged` and `dataContextDeleted` notifications, which allows the Multidata plugin to keep its dataset list up to date.
- It handles `update interactiveFrame name` API requests.
  - Technically, the plugin should probably use `update interactiveFrame title`, as `name` is not included in the documentation, but it was an easy fix and felt like less trouble than asking for a change to the plugin.

I'm not sure why these unrelated bugs were included in the same PT story, but making a second story probably would have required more time than just fixing them.

One additional minor change is removing a redundant cypress function.

### DIFF
--- a/v3/cypress/e2e/plugin.spec.ts
+++ b/v3/cypress/e2e/plugin.spec.ts
@@ -162,6 +162,17 @@ context("codap plugins", () => {
     openAPITester()
     webView.toggleAPITesterFilter()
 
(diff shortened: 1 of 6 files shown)


bfinzer (approved):
👍🏻LGTM

PR #1320, opened 2024-06-25
Fix of bar chart/histogram restore bug

bfinzer:
[#187751419] Bug fix: Restore of bar chart and histogram doesn't show bars

* The initial call to pixiPoints.matchPointsToData was bailing because the pointDisplayType was different than the default type ("points") and thus regarded as a transition. But there were not points to transition to rectangles and thus no points got constructed. Fixed by not bailing if there are not yet any points.


tealefristoe (approved):
Looks good 👍

PR #1321, opened 2024-06-26
187849648 v3 DI Create Attribute with Duplicate Names

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187849648

This PR fixes an issue with the Sensor Interactive plugin. The symptom was that Sensor Interactive was creating new attributes with every run.

The underlying problem was that Sensor Interactive was making the same `create attribute` requests with every run. In v2, requesting to create an attribute with a name that already exists in the specified collection resulted in the existing attribute getting updated, but in v3 it was just creating new attributes every time.

This led me to discover some other issues, like `create attribute` requiring a collection in v2, but not in v3. So I tried to get v3 more in line with v2. I also added tests for the changes.

I do have one remaining question. v2 will not allow an attribute with a duplicate name to be created within a collection, but it doesn't check between collections. So it's possible to create two attributes with the same name in the same dataset, as long as they're in different collections. I've mimicked the v2 behavior in v3, but this feels like a bug to me, so please let me know if I should change how this is handled in v3.

### DIFF
--- a/v3/cypress/e2e/plugin.spec.ts
+++ b/v3/cypress/e2e/plugin.spec.ts
@@ -27,7 +27,7 @@ context("codap plugins", () => {
     cy.log("Handle get attribute request")
     const cmd1 = `{
       "action": "get",
-      "resource": "dataContext[Mammals].collection[Mammals].attribute[Order]"
+      "resource": "dataContext[Mammals].collection[Cases].attribute[Order]"
     }`
     webView.sendAPITesterCommand(cmd1)
     webView.confirmAPITesterResponseContains(/"success":\s*true/)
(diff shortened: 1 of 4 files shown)


bfinzer:
It's definitely a bug to allow duplicate attribute names as apparently v2 does when the attributes are in different collections. Up to you whether to fix in this PR or make a PT story about it.

bfinzer (approved):
👍🏻LGTM
Note response to duplicate attribute question

tealefristoe:
@bfinzer I'm going to make a separate bug about this. It would actually be an easy single line change to check all attributes in the dataset rather than just in the collection for a duplicate name, but I'm not sure the best way to deal with a duplicate in a separate collection. It seems like more of a stretch to just update that attribute and be done with it. Should we also move the attribute to the specified collection? Should we give up and respond with an error?

kswenson:
> @bfinzer I'm going to make a separate bug about this. It would actually be an easy single line change to check all attributes in the dataset rather than just in the collection for a duplicate name, but I'm not sure the best way to deal with a duplicate in a separate collection. It seems like more of a stretch to just update that attribute and be done with it. Should we also move the attribute to the specified collection? Should we give up and respond with an error?

My inclination is to respond with an error.

PR #1322, opened 2024-06-27
187849622 v3 DI Better Tracking of Datasets in Multidata

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187849622

A recent addition of `dataContextCountChanged` and `dataContextDeleted` notifications missed a few places:
- `dataContextCountChanged` notifications need to be broadcast when deleting datasets, in addition to the `dataContextDeleted` notifications that were already being broadcast.
- `dataContextCountChanged` and `dataContextDeleted` notifications need to be broadcast when datasets are created or deleted via the plugin API.
- `dataContextCountChanged` notifications need to be broadcast when importing datasets.

### DIFF
--- a/v3/src/components/app.tsx
+++ b/v3/src/components/app.tsx
@@ -28,6 +28,7 @@ import "../models/shared/shared-case-metadata-registration"
 import "../models/shared/shared-data-set-registration"
 
 import "./app.scss"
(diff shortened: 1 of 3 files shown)


bfinzer (approved):
👍🏻LGTM

PR #1323, opened 2024-06-28
Display standard error in gaussian fit

bfinzer:
[#187850012] Feature: When the std err and gaussian fit adornments are both showing the gaussian fit label should show the std err computed from estimate of standard deviation

* Place creation of text tip and label for standard error in `UnivariateMeasureAdornmentHelper.computeTextContentForStdErr` so it can be used in both standard error and gaussianFit contexts.
* normal-curve-adornment-component.tsx detect when standard error should be drawn as part of a gaussianFit curve and do so
* The normal curve adornment is the one adornment whose label takes more than one line to display and, furthermore, the number of lines depends both on whether it is a gaussian fit curve and whether the standard error adornment is also being display. We handle this special case in AdornmentsBaseStore.getLabelLinesAboveAdornment.
* Added the hook `useGraphOptions` to enable us to detect the presence of a gaussian fit situation without introducing Lakosian violations. This hook could be expanded as needed to detect other special conditions in adornment land.
* In using the NumberInput widget for number of standard errors to display we were not detecting the removal of the component and thus not properly storing the dynamic value. We fix this with a useEffect that calls `handleBlur` when the component is unmounted.


eireland (approved):
Synchronized review. Looks good.

PR #1324, opened 2024-06-29
Fix for LSRL's not properly split by categorical legend

bfinzer:
[#187879648] Bug fix: Least squares line is not properly split by categorical legend

* In `LSRLAdornmentModel.updateLines` there were some wrong assumptions being made about the array of lines for a given cell. In particular, a new line was being pushed into the array on the assumption that they are added sequentially (I guess) even though the desired index was given.
* The jest test for this model was *not* passing the desired index, so we fixed that

### DIFF
--- a/v3/src/components/graph/adornments/lsrl/lsrl-adornment-model.test.ts
+++ b/v3/src/components/graph/adornments/lsrl/lsrl-adornment-model.test.ts
@@ -67,7 +67,7 @@ describe("LSRLAdornmentModel", () => {
   })
   it("can have a line added to its lines property", () => {
     const lSRL = LSRLAdornmentModel.create()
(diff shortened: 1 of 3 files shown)


tealefristoe (approved):
Looks good to me 👍 

I'm a little surprised that it's ok to have a required parameter after an optional one (since `key` has a default value it's not required), but as long as you're not getting complaints from the linter I won't complain either.

I do notice that an unrelated line is over 120 characters (line 314 of `normal-curve-adornment-component.tsx`). Even though it's not related to this PR, you could sneak in a fix to that issue. Looks like it would be easy enough to remove the spaces around `${t(kGaussianFitLabelKey)}` to fix the line length.

PR #1325, opened 2024-07-02
187573237 case card attribute menu

eireland:
In the case card:
- Adds ability to change the properties of an attribute
- Adds ability to add a formula to an attribute
- Changed to use V3 code to add a new attribute in the case card.
- Adds ability to delete an attribute


v3/src/components/case-card/case-card.v2.js
```diff
@@ -501,30 +532,51 @@ iDataContext.doSelectCases({
...
-                  DG.DataContextUtilities.hideAttribute(iContext, iAttr.get('id'))
+                  const attrId = iAttr.get('id')
+                  iCaseMetadata?.applyModelChange(
+                    () => iCaseMetadata?.setIsHidden(attrId, true),
+                    {
+                      notifications: hideAttributeNotification([attrId], iContext.data),
+                      undoStringKey: "DG.Undo.caseTable.hideAttribute",
+                      redoStringKey: "DG.Redo.caseTable.hideAttribute"
+                    }
+                  )
```
kswenson, line 551: For some of these you could reduce the changes in this already large file by replacing the implementation of `DG.DataContextUtilities.hideAttribute` with the implementation here, and then we can eventually eliminate the unused portions of `DG.DataContextUtilities` so that what remains serves as an interface between v2 and v3.

v3/src/components/case-card/case-card.tsx
```diff
@@ -23,6 +24,7 @@ export const CaseCard = observer(function CaseCard({ setNodeRef }: IProps) {
   const instanceId = useInstanceIdContext() || "case-card"
 */
   const data = useDataSetContext()
+  const caseMetadata = useCaseMetadata()
```
kswenson, line 27: The `useDataSet` hook will return both the data and the metadata.

bfinzer (approved):
👍🏻LGTM

kswenson (approved):
👍 Looks good -- note that I added a commit which retrieves the case metadata from the data set when needed which avoids the need to pass it as a prop or argument in several places.

PR #1326, opened 2024-07-03
187864576 v3 DI Collaborative Plugin

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187864576

This PR makes Codap v3 support the Collaborative plugin (`codap-shared-table-plugin`). Many small bugs needed to be fixed and missing features needed to be implemented to get the plugin working:
- `itemSearch` and `caseSearch` requests needed to handle queries with empty right sides (like `__collaborator__==`). Collaborative uses these requests to find cases not yet associated with any user.
- When `respectEditableItemAttribute` is `true`, block editing the item when its `__editable__` value is "false" (in addition to falsy). This is necessary because Collaborative uses a formula to fill out this value, which results in "true" or "false".
- Adds the `deleteable` field to the `Attribute` model. So far this doesn't actually do anything in the UI, but Collaborative uses it to avoid deleting certain special attributes.
- Use the `cid` specified for collections and attributes in `create dataContext` API requests to specify ids for these models. This was necessary so attributes in shared datasets have the same id even when in different instances of Codap.
- `create item` API requests are now handled when the `values` are like `{ id: string, values: { ... } }`.
  - This allows an API request to specify the `id` of an `item`, which wasn't possible before, but is necessary for Collaborative to keep shared items up to date as they are changed.
  - This goes against the API documentation, which says `values` should be like `{ ... }`, but Collaborative uses this format and it seemed easier to accommodate it rather than force an update of Collaborative.
- `notify itemSearch itemOrder` API requests are now supported. Collaborative uses `itemOrder: last` to move the user's items to the bottom of the table, which puts them next to the new case row at the bottom of the table*.
  - This seems to be working correctly in the context of the Collaborative plugin now, but when making these API calls using the API tester plugin, the table doesn't update immediately. Is there a way to fix this?
  - `itemOrder` can be "first", "last", or a `number[]`. The first two cases are easy: they move the items that satisfy the search query to the top or bottom of a table. The `number[]` option is more confusing. It moves the items to the indexes specified in the array. I believe I implemented this in the same way as v2: the items are moved in the order they are found, and they shift each other around as they are moved (so an array like [1, 1] would actually end with the first item in the 2 position). I kind of doubt this quirky feature is used by any plugins, but it's possible it's not implemented quite right, so it would be worth testing if it actually is used.
- When a new collection is created with a `create collection` API request, the default collection is deleted if it exists and is empty. This is necessary because Collaborative creates an empty dataset and then adds collections to it, rather than specifying the dataset's collections in the `create dataContext` request.
  - It seems like this should be done in a more general way, like a reaction in response to any changes to the attributes, but I wasn't sure what was safe or not so handled this as a special case.
- Collections created with `create collection` API requests are placed `after` the right-most collection if no `before` is specified for them. This is necessary because the API specifies this behavior, but the `dataset` model places new collections with no `before` or `after` specified left of the right-most collection.
- New datasets created with `create dataContext` API requests are properly added to the `FormulaManager` now. Collaborative uses formulas in datasets it creates, so this was a necessary bug fix.
- `create item` API requests now broadcast `createCases` notifications. Collaborative relies on these notifications to keep its database synched with items it creates. As the case with all notifications broadcast from API requests, this goes against what the documentation states.

Other improvements that are side effects from this work or were very easy to implement:
- `create item` API request responses now include `caseIDs`.

\* There is still one major missing feature for Collaborative, which is the new case row at the bottom of the table. This should be handled in a different, single feature PR.


v3/src/data-interactive/handlers/collection-handler.ts
```diff
@@ -23,6 +23,14 @@ export const diCollectionHandler: DIHandler = {
 
     const newCollections: ICollectionModel[] = []
     dataContext.applyModelChange(() => {
+      // Remove the default collection if it exists and is empty
+      // This is a special case and it might make more sense to do this in a more generic way,
+      // like a reaction that removes empty collections whenever attributes change.
+      if (dataContext.collections.length === 1) {
+        const oldCollection = dataContext.collections[0]
+        if (oldCollection.attributes.length === 0) dataContext.removeCollection(oldCollection)
+      }
```
kswenson, line 32: It is possible that the loop below won't create any collections (e.g. if none of them have a `name`), in which case the `DataSet` would be left with no collections. You can rectify this by making it a local function and then calling it just before `dataContext.addCollection()` in line 69 below, so that empty collections are deleted only when adding new collections. I'm intrigued by the idea of automating this, but since users can't create collections on their own, I think handling this special case here seems acceptable.

kswenson (changes requested):
👍 Looks good -- just a handful of suggestions

tealefristoe:
@kswenson Thanks for all the feedback. I think I've addressed all your suggestions.

I did want to ask about the table not updating immediately in response to `notify itemSearch itemOrder` requests. Is there a way to force a table to update?

kswenson (approved):
👍 Changes look good

>I did want to ask about the table not updating immediately in response to notify itemSearch itemOrder requests. Is there a way to force a table to update?

Log a bug and we can look into it at some point. I don't think it should hold up the PR.

PR #1328, opened 2024-07-05
[#185315714] - focus on newly created attribute

eireland:
Scrolls to newly added attribute and goes into input mode.

### DIFF
--- a/v3/src/components/case-table/case-table-types.ts
+++ b/v3/src/components/case-table/case-table-types.ts
@@ -21,10 +21,7 @@ export interface TCalculatedColumn extends CalculatedColumn<TRow> {}
 export interface TRenderers extends Renderers<TRow, unknown> {}
 export interface TRenderEditCellProps extends RenderEditCellProps<TRow> {}
 export interface TRenderCellProps extends RenderCellProps<TRow> {}
-export interface TRenderHeaderCellProps extends RenderHeaderCellProps<TRow> {
-  attrIdToEdit?: string
-  setAttrIdToEdit?: (attrId: string | undefined) => void
-}
+export interface TRenderHeaderCellProps extends RenderHeaderCellProps<TRow> {}
 export interface TRenderRowProps extends RenderRowProps<TRow> {}
 export interface TCellClickArgs extends CellClickArgs<TRow> {}
 export type TColSpanArgs = ColSpanArgs<TRow, unknown>
--- a/v3/src/components/case-table/case-table-types.ts
+++ b/v3/src/components/case-table/case-table-types.ts
@@ -21,7 +21,10 @@ export interface TCalculatedColumn extends CalculatedColumn<TRow> {}
 export interface TRenderers extends Renderers<TRow, unknown> {}
 export interface TRenderEditCellProps extends RenderEditCellProps<TRow> {}
 export interface TRenderCellProps extends RenderCellProps<TRow> {}
-export interface TRenderHeaderCellProps extends RenderHeaderCellProps<TRow> {}
+export interface TRenderHeaderCellProps extends RenderHeaderCellProps<TRow> {
+  attrIdToEdit?: string
+  setAttrIdToEdit?: (attrId: string | undefined) => void
+}
 export interface TRenderRowProps extends RenderRowProps<TRow> {}
 export interface TCellClickArgs extends CellClickArgs<TRow> {}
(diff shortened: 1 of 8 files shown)


bfinzer:
One problem I found:
* When you make a new attribute, if you click to edit the selected default name, the default name gets accepted and the cell loses focus instead of initiating editing. Also, the header cell retains its blue border if you immediately go on to make *another* new attribute.

bfinzer (changes requested):
I think the bug I noted should be fixed, assuming it fits easily in this work.

bfinzer (approved):
👍🏻LGTM

v3/src/components/case-table/use-columns.tsx
```diff
@@ -109,8 +111,10 @@ export const useColumns = ({ data, indexColumn }: IUseColumnsProps) => {
                 // which leads to undesirable browser behavior.
                 width: kDefaultColumnWidth,
                 resizable: true,
-                headerCellClass: "codap-column-header",
-                renderHeaderCell: ColumnHeader,
+                headerCellClass: `codap-column-header ${attrIdToEdit === id ? "selected" : ""}`,
+                renderHeaderCell:
+                  (props) =>
+                    <ColumnHeader {...props} attrIdToEdit={attrIdToEdit} setAttrIdToEdit={setAttrIdToEdit} />,
```
kswenson, line 117: There's a potential performance issue here in that changes to the column definitions require re-rendering the entire table. If `attrIdToEdit` were managed by the `CollectionTableModel`, then `ColumnHeader` could access them directly and no change would be required here. If the `selected` class could be applied at the `ColumnHeader` level that would be preferable as well, but I'm not sure if that's a straightforward change.

v3/src/components/case-table/use-columns.tsx
```diff
@@ -125,7 +129,7 @@ export const useColumns = ({ data, indexColumn }: IUseColumnsProps) => {
       },
       { name: "useColumns [rebuild columns]", equals: comparer.structural, fireImmediately: true }, data
     )
-  }, [RenderCell, caseMetadata, collectionId, data, indexColumn, parentCollection])
+  }, [RenderCell, caseMetadata, collectionId, data, indexColumn, parentCollection, attrIdToEdit, setAttrIdToEdit])
```
kswenson, line 132: Changes to `attrIdToEdit` would result in reinstalling the MobX reaction rather than just having it fire.

v3/src/components/case-table/case-table-types.ts
```diff
@@ -21,7 +21,10 @@ export interface TCalculatedColumn extends CalculatedColumn<TRow> {}
 export interface TRenderers extends Renderers<TRow, unknown> {}
 export interface TRenderEditCellProps extends RenderEditCellProps<TRow> {}
 export interface TRenderCellProps extends RenderCellProps<TRow> {}
-export interface TRenderHeaderCellProps extends RenderHeaderCellProps<TRow> {}
+export interface TRenderHeaderCellProps extends RenderHeaderCellProps<TRow> {
+  attrIdToEdit?: string
+  setAttrIdToEdit?: (attrId: string | undefined) => void
+}
```
kswenson, line 27: My inclination would be to store this information in the `CollectionTableModel` rather than in React state, in which case this change wouldn't be required.

v3/src/components/case-table/column-header.tsx
```diff
@@ -12,17 +12,19 @@ import { kIndexColumnKey, TRenderHeaderCellProps } from "./case-table-types"
 import { ColumnHeaderDivider } from "./column-header-divider"
 import { useRdgCellFocus } from "./use-rdg-cell-focus"
 
-export function ColumnHeader({ column }: Pick<TRenderHeaderCellProps, "column">) {
+export function ColumnHeader({ column, attrIdToEdit, setAttrIdToEdit }: TRenderHeaderCellProps) {
```
kswenson, line 15: As mentioned elsewhere, my inclination would be to store this information in the `CollectionTableModel` rather than in React state and passed as props.

kswenson (changes requested):
👍 In general this looks good, but I think the edit state would be better stored in the `CollectionTableModel` (which is accessible via context/hook to any component) rather than in React state and passed as props. We can meet to discuss if it would help.

kswenson (approved):
👍 LGTM -- just a couple of remaining suggestions

PR #1332, opened 2024-07-06
[#187849761]Fixes styling of component title bar when dragging is initiated on top of the component title.

eireland:
Changed the component from a ChakraUI EditableInput to just ChakraUI Input.


tealefristoe (approved):
Looks good 👍

PR #1335, opened 2024-07-09
fix: handling of units in case card

kswenson:
[[PT-187906401]](https://www.pivotaltracker.com/story/show/187906401)


eireland (approved):
👍 LGTM

PR #1336, opened 2024-07-10
187810830 edit attr prop modal tuneup

eireland:
Makes Edit Attribute Property modal, and Formula Editor modal styles closer to the V2 version, with some improvements
- Input labels are now capitalized
- Modal background is gray, and inputs are white.
- Input fields except for Description (and formula text area in the Formula Editor) are left aligned and responsive to length of labels.
- Description text area is still expandable and modal expands as test area expands
- Apply button has default color.
- Typing Enter submits the changes to the modal and closes the modal
- Icon added to the modal header for the edit attribute properties modal

Note: This does not include the ability to drag and drop the modals anywhere in the window.


v3/src/components/case-table/attribute-menu/edit-attribute-properties-modal.tsx
```diff
@@ -78,90 +81,102 @@ export const EditAttributePropertiesModal = ({ attributeId, isOpen, onClose }: I
     onClose()
   }
 
+  const handleKeyDown = (e: React.KeyboardEvent) => {
+    e.stopPropagation()
+    if (e.key === "Enter") {
+      applyChanges()
+    }
```
kswenson, line 88: Should we handle the escape key here as well or is that handled elsewhere?

kswenson (changes requested):
👍 Looks good but I think you should handle the escape key to close the dialog as well.

kswenson (approved):
👍 LGTM

bfinzer (approved):
👍🏻 LGTM

PR #1340, opened 2024-07-11
fix: dot chart handling of added cases

kswenson:
[[PT-187878126]](https://www.pivotaltracker.com/story/show/187878126)

`CategorySet`s are created on demand when required, but do not need to be serialized unless the order of categories or assigned colors are user-modified. To accomplish this within the constraints of MST's `views`/`actions` architecture, `CategorySet`s are created in a provisional state where they are kept in volatile state and are only stored in the serialized models once user-modified. Previously, the `onAnyAction` code that listens for changes to the underlying attribute was only being installed for category sets attached to the model. This PR extends the observation to provisional `CategorySet`s.


bfinzer (approved):
👍🏻LGTM

PR #1342, opened 2024-07-11
181846592 v3 Input Row

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/181846592

This PR begins to implement the input row in case tables. The implementation is incomplete in two ways:
- Switching cells within the input row will add the new case. In v2, the new is only added to the dataset when the user navigates away from the input row, allowing a more complete case to be created.
- Editing a cell in the input row requires double clicking. In v2, clicking a cell in the input row would immediately allow editing. Additionally, the mouse cursor doesn't change when hovering over a cell in the input row, but it should.
- It's not currently possible to drag the input row to adjust where the new case will be added in the dataset.
  - While there is no UI to allow the user to change where to create the new case, the infrastructure has been put into place to allow the case to be created anywhere in the dataset. If a developer would like to test this, they could manually change `inputRowIndex` in `collection-table-model.ts`.

Note that code coverage is not quite up to snuff for this PR. That's mostly because a lot of code went into supporting the input row at locations other than at the bottom, but because there's no UI to move the input row, tests were not added for these cases. Once the UI is added, it should be relatively easy to implement tests that cover all of the code in this PR.

Some more general improvements made in this PR:
- Notifications from updating cases by modifying cells in the case table have been improved, with correct case and item ids.
- The `before` and `after` options for `DataSet.addCases()` can now be case ids rather than just item ids.
- A bug was fixed related to `caseIDs` in the response to `create item` API requests.


v3/src/components/case-table/use-rows.ts
```diff
@@ -222,18 +224,96 @@ export const useRows = () => {
...
+      () => {
+        // Update existing cases
+        if (casesToUpdate.length > 0) {
+          data.setCaseValues(casesToUpdate)
+          if (collection?.id === data.childCollection.id) {
+            // The child collection's case ids are persistent, so we can just use the casesToUpdate to
+            // determine which case ids to use in the updateCasesNotification
+            updatedCaseIds = casesToUpdate.map(aCase => aCase.__id__)
+          } else {
+            // Other collections have cases whose ids change when values change, so we have to check which case ids
```
kswenson, line 276:
```suggestion
            // Other collections have cases whose ids change when values change, so we have to check which case ids
```
Comment may be a bit mysterious to those unfamiliar with case grouping details. Perhaps add something like: "whose ids change when values change due to updated case grouping" or something like that.

v3/src/components/case-table/use-rows.ts
```diff
@@ -222,18 +224,96 @@ export const useRows = () => {
...
+          }
+          oldCaseIds = new Set(collection?.caseIds ?? [])
+        }
+
+        // Create new cases
+        if (creatingCases) {
+          const options: IAddCasesOptions = {}
+          if (collectionTableModel?.inputRowIndex != null && collectionTableModel.inputRowIndex >= 0) {
+            options.before = collection?.caseIds[collectionTableModel.inputRowIndex]
+            collectionTableModel.inputRowIndex += 1
```
kswenson, line 290: Should use an `@action` here rather than updating the model directly:
```suggestion
            collectionTableModel.setInputRowIndex(collectionTableModel.inputRowIndex + 1)
```

kswenson (approved):
👍 Looks good -- just a few minor suggestions.

PR #1343, opened 2024-07-12
fix: case table refresh issues [PT-187931119] [PT-187931309] [PT-187931070]

kswenson:
[[PT-187931119]](https://www.pivotaltracker.com/story/show/187931119)
[[PT-187931309]](https://www.pivotaltracker.com/story/show/187931309)
[[PT-187931070]](https://www.pivotaltracker.com/story/show/187931070)

There are two basic fixes here:
1. Eliminate the local cases cache in the `useRows` hook. It merely served as an opportunity to get out of sync.
2. Add a case grouping `validationCount` to the `DataSet`. There was a `reaction` in the `useRows` hooks that was listening to `isValidCaseGroups` that never fired because `isValidCaseGroups` would be set to false and then back to true before the reaction ran, so no change was detected. Adding the `validationCount` means that the reaction fires any time the grouping has been re-validated.

There are a few changes unrelated to the refresh bugs that were made along the way:
1. Rename `handleTopClick` function to `handleExpandCollapseAllClick`
2. A couple of unrelated lint warnings were fixed

### DIFF
--- a/v3/src/components/case-table/collection-table-spacer.tsx
+++ b/v3/src/components/case-table/collection-table-spacer.tsx
@@ -67,7 +67,7 @@ export const CollectionTableSpacer = observer(function CollectionTableSpacer({ o
   //   }
   // }
 
-  function handleTopClick() {
+  function handleExpandCollapseAllClick() {
     caseMetadata?.applyModelChange(() => {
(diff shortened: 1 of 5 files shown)


tealefristoe (approved):
Looks great 👍

PR #1347, opened 2024-07-12
fix: undo/redo of component creation [PT-187949749][PT-187948687]

kswenson:
[[PT-187949749]](https://www.pivotaltracker.com/story/show/187949749) -- Redo of component creation results in invisible components
[[PT-187948687]](https://www.pivotaltracker.com/story/show/187948687) -- MST warning on undo create case table

The MST warning was due to the `FormulaManager`'s use of a MobX `reaction` which accessed the `DataSet`. Undoing the creation of the case table destroys the underlying `DataSet`, and so the `DataSet` must be removed from the `FormulaManager` when it is destroyed.

The bug in redoing component creation occurs because of the way component animation was implemented. When the component is created, its pre-animation dimensions are stored in the model, i.e. `{ x: 0, y: 0, width: 0, height: 25 }`. Then in a separate step, the tile's dimensions are updated to their desired position and extent (without undo), which triggers the animation due to a CSS transition. The redo bug arises because the undo history captures the pre-animation dimensions, so redo instantiates a zero-width tile.

After considering several model-related approaches to solving the problem (`onPatch`?, MST middleware?), it occurred to me that the problem arises because the model is being manipulated to achieve a display-only result (animation). A better approach is to store the correct tile dimensions in the model from the start, and then handle animating to those dimensions in the React component when desired. This simplified the implementation substantially.

The solution involves several pieces:
- a new `animateCreation` option is added to `INewTileOptions`
- this option is used when creating tiles from the tool shelf (not when restoring documents, for instance)
- when the `animateCreation` option is specified, the id of the tile is stored in volatile state in the `FreeTileRow`
- when a `FreeTileComponent` is first rendered, if `animateCreation` was specified, then it is initially rendered in pre-animation dimensions (`{ x: 0, y: 0, width: 0, height: 25 }`) on the first render, and then a second render is triggered, after which the final dimensions of the tile are used.
- the preexisting CSS transition triggers the animation

As a result, the model always stores the correct dimensions for all tiles, undo/redo works as desired, and animation is handled entirely in volatile MST state and React state. The moral of the story is that it's better to keep model changes simple and handle complexities like animations in the React components.

### DIFF
--- a/v3/src/models/document/free-tile-row.ts
+++ b/v3/src/models/document/free-tile-row.ts
@@ -66,6 +66,7 @@ export interface IFreeTileInRowOptions extends ITileInRowOptions {
   width?: number
   height?: number
   zIndex?: number
+  animateCreation?: boolean
 }
 export const isFreeTileInRowOptions = (options?: ITileInRowOptions): options is IFreeTileInRowOptions =>
               !!options && ("x" in options && options.x != null) && ("y" in options && options.y != null)
@@ -81,6 +82,10 @@ export const FreeTileRow = TileRowModel
     tiles: types.map(FreeTileLayout), // tile id => layout
     maxZIndex: 0
   })
+  .volatile(self => ({
+    // tile ids of tiles created by user whose creation should be animated
+    animateCreationTiles: new Set<string>()
+  }))
   .views(self => ({
     get acceptDefaultInsert() {
       return true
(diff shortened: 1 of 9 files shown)


tealefristoe (approved):
Seems like a great change to me 👍

v3/src/models/document/free-tile-row.ts
```diff
@@ -131,9 +136,11 @@ export const FreeTileRow = TileRowModel
...
-      const { x = 50, y = 50, width = undefined, height = undefined, zIndex = this.nextZIndex() } =
-        isFreeTileInRowOptions(options) ? options : {}
+      const {
+        x = 50, y = 50, width = undefined, height = undefined, zIndex = this.nextZIndex(), animateCreation = false
+      } = isFreeTileInRowOptions(options) ? options : {}
       self.tiles.set(tileId, { tileId, x, y, width, height, zIndex })
+      animateCreation && self.animateCreationTiles.add(tileId)
     },
     removeTile(tileId: string) {
       self.tiles.delete(tileId)
```
bfinzer, line 146: Is it worth removing this tileId from animateCreationTiles?

bfinzer (approved):
👍🏻 LGTM
Just one niggly question/comment

v3/src/models/document/free-tile-row.ts
kswenson: No, because that would prevent animation on redo of creation. Essentially, the logic here is that if a tile was user-created from the tool shelf, it should animate any time its component mounts for the lifetime of this browser session. In practice, the only times that the tile mounts are when it's created initially and when the creation is re-done, but the fact that it is a tile that should be animated on creation is a persistent property within the session, not one that's tied to whether the tile is currently in the document.

PR #1348, opened 2024-07-13
chore: update dependencies

kswenson:
Something to do on a Friday afternoon when fewer brain cells are firing....

Update all dependencies except:
1. ESLint v9 uses a new configuration format which requires a migration. [[PT-187950465]](https://www.pivotaltracker.com/story/show/187950465)
2. MathJS v12.3.2 breaks our unit tests. (MathJS v13 requires ES2020 which would necessitate a minimum supported browser conversation). [[PT-187950451]](https://www.pivotaltracker.com/story/show/187950451)
3. PixiJS v8 contains API changes that require migration. [[PT-187950454]](https://www.pivotaltracker.com/story/show/187950454)

A couple of the updates required additional changes:
1. Eliminated unnecessary `as` casts after `filter()` calls due to TypeScript 5.5's better type inference.
2. `webpack-dev-server` required some adjustments in `package.json` and `webpack.config.js`.
3. `sass` required some rearrangements of CSS so that current scope declarations come before nested scope declarations.
2. `css-loader` required `namedExport: false` in `webpack.config.js` to retain current behavior.

Also, run `npm audit fix` to bring our known vulnerabilities to zero.


tealefristoe (approved):
Looks good 👍 

Seems like this added a bunch of lint warnings, though.

kswenson:
> Looks good 👍
> 
> Seems like this added a bunch of lint warnings, though.

That's really weird. I knew that TypeScript 5.5 added improved type inference around `filter()` and was expecting to see a number of cases where we could eliminate `as` casts after `filter()` calls but was disappointed when lint didn't report any such cases after the update. I run lint/build/test after every step of the process, so I definitely would have noticed if those lint warnings were reported. The only thing I can figure is that the lint cache doesn't get cleared when dependencies are updated, so the lesson is that I should make sure to clear the lint cache when linting after dependency changes. Fixed.

PR #1349, opened 2024-07-13
fix: redundant calls to handleRowsChange

kswenson:
The `ColorCellTextEditor` component (which is used in situations in which no attribute type has been assigned, since the user _could_ theoretically enter a color at any time) was handling key events like `enter`/`escape`, but RDG handles key events internally as well, resulting in redundant calls.


tealefristoe (approved):
Looks good to me, thanks for tracking that down 👍

PR #1350, opened 2024-07-14
fix: can't hide last attribute of collection

kswenson:
[[PT-187735631]](https://www.pivotaltracker.com/story/show/187735631)


eireland (approved):
👍 LGTM.

PR #1351, opened 2024-07-16
Add case table dates support

pjanik:
https://www.pivotaltracker.com/story/show/187965959

This PR adds basic date support to the case table. It also lays the foundation for further date-related work, as various utilities were ported from V2.

I split date-utils and date-parser, as the parser has quite a bit of code. I've also ported existing V2 tests to modern Jest and ensured they all pass, which should ensure backward compatibility.

If we want to improve date support in the future, I'd suggest leaving this code and adding a second pass to the `isDateString` and `parseDate` functions. This second pass would attempt to interpret the date using an external library that likely supports more formats.

I also had to add a small helper to the translation file.

Regarding case table logic, I based my implementation on my tests of the V2 behavior. It seems formatting happens only when the "date" type is explicitly selected (so `userType = "date"` is requried, it's not enough if the inferred `type` is equal to `"date"`). This is more strict than the graph, which assumes dates even if the type is inferred. I described all the rules in the PT story, but I'm pasting them here too:

> 1. Strings resembling dates are NOT formatted in any way unless the user explicitly selects the "Date" type in the Attribute Properties dialog (this is the CODAP v2 behavior).
> 2. When the "Date" type is selected by the user, CODAP tries to interpret strings and numbers as dates and formats them accordingly.
> 3. Output formatting should depend on the selected language (add `?lang=es` parameter in V3 or change it in V2 using the menu in the top-right corner).
> 4. Language does NOT affect how the string is interpreted (e.g., 2/1/2020 is always interpreted as February 1st, both in en-US and in Spanish).
> 5. Strings and numbers that are not recognized as dates are rendered in quotes (e.g., "ABC").
> 6. Formatting in V2 and V3 for various strings should be identical. Some test cases:
>  - 11/9/2016
>  - 2016-02-02
>  - April 4, 2016
>  - Thu Jul 11 09:12:47 PDT 2019
>  - 11/9/2016 7:18:02.234am

### DIFF
--- a/v3/src/components/case-table/use-columns.tsx
+++ b/v3/src/components/case-table/use-columns.tsx
@@ -15,6 +15,8 @@ import { kDefaultColumnWidth, symDom, TColumn, TRenderCellProps } from "./case-t
 import CellTextEditor from "./cell-text-editor"
 import ColorCellTextEditor from "./color-cell-text-editor"
 import { ColumnHeader } from "./column-header"
+import { parseDate } from "../../utilities/date-parser"
+import { DatePrecision, formatDate } from "../../utilities/date-utils"
 
 // cache d3 number formatters so we don't have to generate them on every render
 type TNumberFormatter = (n: number) => string
@@ -50,6 +52,24 @@ export function renderValue(str = "", num = NaN, attr?: IAttribute, key?: number
     const formatStr = attr?.format ?? kDefaultFormatStr
     const formatter = getNumFormatter(formatStr)
     if (formatter) str = formatter(num)
+  }
+
+  // dates
+  // Note that CODAP v2 formats dates in the case table ONLY if the user explicitly specified type as "date".
+  // Dates are not interpreted as dates and formatted by default.
+  if (userType === "date" && str !== "") {
+    const date = parseDate(str, true)
+    if (date) {
+      // TODO: add precision support for date formatting
+      const formattedDate = formatDate(date, DatePrecision.None)
+      return {
+        value: str,
+        content: <span className="cell-span" key={key}>{formattedDate || str}</span>
+      }
+    } else {
+      // If the date is not valid, wrap it in quotes (CODAP V2 behavior).
+      str = `"${str}"`
+    }
   }
 
   return {
--- a/v3/src/utilities/date-parser.test.ts
+++ b/v3/src/utilities/date-parser.test.ts
@@ -1,0 +1,80 @@
+import { isDateString, parseDate } from './date-parser'
+
+describe('Date Parser tests - V2 compatibility', () => {
+  // These tests are ported from V2 and should always pass unchanged as long as we want to maintain compatibility.
+  test('local date format', () => {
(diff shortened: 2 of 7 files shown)


v3/src/components/case-table/use-columns.tsx
```diff
@@ -52,6 +54,24 @@ export function renderValue(str = "", num = NaN, attr?: IAttribute, key?: number
...
+  // Note that CODAP v2 formats dates in the case table ONLY if the user explicitly specified type as "date".
+  // Dates are not interpreted as dates and formatted by default.
+  if (userType === "date" && str !== "") {
+    const date = parseDate(str, true)
+    if (date) {
+      // TODO: add precision support for date formatting
+      const formattedDate = formatDate(date, DatePrecision.None)
+      return {
+        value: str,
+        content: <span className="cell-span" key={key}>{formattedDate || str}</span>
```
kswenson, line 67: For consistency, shouldn't this be:
```suggestion
        content: <span className="cell-span" key={key}>{formattedDate || `"${str}"`}</span>
```

v3/src/utilities/date-parser.test.ts
kswenson, line 1: Given the issues with the ill-specified range tests in `isValidDateSpec()` in v2, we should make sure to have some out-of-range date testing here in v3. Also, there are a number of too-long lines that should be broken up or eslint-disabled.

kswenson (changes requested):
👍 Looks great! Just a few minor suggestions.

pjanik:
@kswenson, thanks! I believe I've addressed all the comments.

I think it might be worth reviewing the very last commit carefully, where I had to update `isFiniteNumber` so it could be actually reused. It accepted `any` as an argument but returned `true` only for proper numbers. Strings (e.g., "1", "2") were not supported before. Assuming we want to have one helper, I made it more flexible and added specs to cover the new functionality.

It also uncovered an interesting bug in date-utils. Previously, the formatting logic relied on the fact that `isNumeric(new Date())` was returning `true` (since `isNaN(new Date())` is `false` 🧐). :)

v3/src/utilities/date-parser.test.ts
kswenson: Additional tests look good. There are some stray semicolons that can be removed with a `lint:fix`.

kswenson (changes requested):
👍 Changes mostly look good -- I think my original suggestion to reuse `isFiniteNumber()` has turned out to be ill-advised. The original strict interpretation was intentional. If the date code needs a looser implementation then I think there should be a separate function for that. Sorry for the misdirection. 😬

pjanik:
@kswenson, I see. I've restored the original `isFiniteNumber` implementation. I had considered merging these two functionalities into a more generic version, but I understand your point.

Actually, the V2 code was relying on some unusual features of `DG.MathUtiltiies.isNumeric`, particularly the fact that it returned true for `Date` instances. 

However, V3 handles this differently - now, I explicitly check if the value is `isFiniteNumber` or `isDate`. So, I think we're fine and don't need string support. So, I'm not adding a new variant yet, as it would be unused.

kswenson (approved):
👍 Looks good!

PR #1352, opened 2024-07-16
Fix incorrect range comparison in V2 date_utilities

pjanik:
This PR fixes incorrect range comparisons that all looked like `if (1 < x < 2)`. Unfortunately, this syntax is not supported in JS/TS.

### DIFF
--- a/apps/dg/utilities/date_utilities.js
+++ b/apps/dg/utilities/date_utilities.js
@@ -404,7 +404,7 @@ DG.DateUtilities.dateParser = (function () {
         return null;
       }
       var newHr = Number(hr);
-      if (amPm != null && (0<newHr<=12)) {
+      if (amPm != null && (0<newHr && newHr<=12)) {
         newHr = newHr % 12;
         if (amPm && amPm.toLowerCase() === 'pm') {
           newHr += 12;
@@ -446,11 +446,11 @@ DG.DateUtilities.dateParser = (function () {
   function isValidDateSpec(dateSpec) {
     var isValid =
         !isNaN(dateSpec.year) &&
-        (!isNaN(dateSpec.month) && (1<=dateSpec.month<=12)) &&
-        (!isNaN(dateSpec.day) && (1<=dateSpec.day<=31)) &&
-        (!isNaN(dateSpec.hour) && (0<=dateSpec.hour<=23)) &&
-        (!isNaN(dateSpec.min) && (0<=dateSpec.min<=59)) &&
-        (!isNaN(dateSpec.sec) && (0<=dateSpec.sec<=59)) &&
+        (!isNaN(dateSpec.month) && (1<=dateSpec.month && dateSpec.month<=12)) &&
+        (!isNaN(dateSpec.day) && (1<=dateSpec.day && dateSpec.day<=31)) &&
+        (!isNaN(dateSpec.hour) && (0<=dateSpec.hour && dateSpec.hour<=23)) &&
+        (!isNaN(dateSpec.min) && (0<=dateSpec.min && dateSpec.min<=59)) &&
+        (!isNaN(dateSpec.sec) && (0<=dateSpec.sec && dateSpec.sec<=59)) &&
         !isNaN(dateSpec.subsec);
     if (isValid) { return dateSpec; }
   }


kswenson (approved):
👍 Looks good -- if it were a straightforward matter to add tests of out-of-range dates that would pass the prior implementation but fail now I would recommend it, but my sense is that the v2 tests are a pain to work with, so I'll limit the suggestion to the v3 context.

kswenson:
Out of curiosity I checked to see what ChatGPT thought of the original code:

Me: What is wrong with the following code and how should it be fixed?

```javascript
  function isValidDateSpec(dateSpec) {
    var isValid =
        !isNaN(dateSpec.year) &&
        (!isNaN(dateSpec.month) && (1<=dateSpec.month<=12)) &&
        (!isNaN(dateSpec.day) && (1<=dateSpec.day<=31)) &&
        (!isNaN(dateSpec.hour) && (0<=dateSpec.hour<=23)) &&
        (!isNaN(dateSpec.min) && (0<=dateSpec.min<=59)) &&
        (!isNaN(dateSpec.sec) && (0<=dateSpec.sec<=59)) &&
        !isNaN(dateSpec.subsec);
    if (isValid) { return dateSpec; }
  }
```

ChatGPT:
>There are a couple of issues with the given code:
>
>1. The expression `(1 <= dateSpec.month <= 12)` does not work as intended in JavaScript. It should be split into two comparisons: `1 <= dateSpec.month && dateSpec.month <= 12`. The same applies to the other range checks.
>2. The function does not return `false` if the date specification is invalid. It should explicitly return `false` when the validation fails.
>
> Here's the corrected version of the code:
>
```javascript
function isValidDateSpec(dateSpec) {
    var isValid =
        !isNaN(dateSpec.year) &&
        (!isNaN(dateSpec.month) && (1 <= dateSpec.month && dateSpec.month <= 12)) &&
        (!isNaN(dateSpec.day) && (1 <= dateSpec.day && dateSpec.day <= 31)) &&
        (!isNaN(dateSpec.hour) && (0 <= dateSpec.hour && dateSpec.hour <= 23)) &&
        (!isNaN(dateSpec.min) && (0 <= dateSpec.min && dateSpec.min <= 59)) &&
        (!isNaN(dateSpec.sec) && (0 <= dateSpec.sec && dateSpec.sec <= 59)) &&
        !isNaN(dateSpec.subsec);
    if (isValid) {
        return dateSpec;
    } else {
        return false;
    }
}
```
>
>This version properly checks the ranges and returns `false` if the date specification is invalid.

Me: Well done. What would the behavior of the original code have been, i.e. how would the bug have manifested itself as a result of the incorrect comparisons?

>In the original code, the bug with the incorrect comparisons `(1 <= dateSpec.month <= 12)` and similar comparisons for day, hour, minute, and second would have caused these conditions to always evaluate to `true`, regardless of the value of `dateSpec.month` (or other fields).
>
>Here's why: In JavaScript, the expression `(1 <= dateSpec.month <= 12)` is evaluated as `(1 <= dateSpec.month)` first, which results in a boolean (`true` or `false`). This boolean value is then compared to `12`, which is always `true` because both `true` and `false` are less than `12`.
>
>As a result, even if `dateSpec.month` was out of the valid range (e.g., 0 or 13), the condition would incorrectly evaluate to `true`, leading to an invalid date being considered valid.
>
>For example, with the following `dateSpec` object:
>
```javascript
var dateSpec = {
    year: 2023,
    month: 13,
    day: 31,
    hour: 10,
    min: 30,
    sec: 45,
    subsec: 123
};
```
>
>The incorrect comparison `(1 <= dateSpec.month <= 12)` would be evaluated as:
>
>1. `1 <= dateSpec.month` which is `1 <= 13`, resulting in `true`.
>2. `true <= 12`, which evaluates to `true`.
>
>Therefore, the `isValid` variable would incorrectly be set to `true`, causing the function to return `dateSpec` even though the month is invalid.
>
>The correct version of the code ensures that the comparisons are done properly and invalid dates are correctly identified:
>
```javascript
function isValidDateSpec(dateSpec) {
    var isValid =
        !isNaN(dateSpec.year) &&
        (!isNaN(dateSpec.month) && (1 <= dateSpec.month && dateSpec.month <= 12)) &&
        (!isNaN(dateSpec.day) && (1 <= dateSpec.day && dateSpec.day <= 31)) &&
        (!isNaN(dateSpec.hour) && (0 <= dateSpec.hour && dateSpec.hour <= 23)) &&
        (!isNaN(dateSpec.min) && (0 <= dateSpec.min && dateSpec.min <= 59)) &&
        (!isNaN(dateSpec.sec) && (0 <= dateSpec.sec && dateSpec.sec <= 59)) &&
        !isNaN(dateSpec.subsec);
    if (isValid) {
        return dateSpec;
    } else {
        return false;
    }
}
```

Which seems like a reasonable response.

pjanik:
Thanks, @kswenson. ChatGTP gave very reasonable answers. :) I've added tests to the V3 utilities. Also, I'm not risking modifying `isValidDateSpec` to return `false` instead of `undefined` not to break some V2 pieces that potentially could rely on this specific value.

However, in V3, I made `isValidDateSpec` return `false` following ChatGPT's suggestion, as it's new code there and doesn't bring any risk.

@kswenson, I'll let you merge the PR, as I'm not sure about the V2 release process.

kswenson:
> Thanks, @kswenson. ChatGTP gave very reasonable answers. :) I've added tests to the V3 utilities. Also, I'm not risking modifying `isValidDateSpec` to return `false` instead of `undefined` not to break some V2 pieces that potentially could rely on this specific value.
> 
> However, in V3, I made `isValidDateSpec` return `false` following ChatGPT's suggestion, as it's new code there and doesn't bring any risk.
> 
> @kswenson, I'll let you merge the PR, as I'm not sure about the V2 release process.

I don't feel strongly about ChatGPT's suggestion to return false in general and agree with your choosing not to change it here in any case.

bfinzer (approved):
👍🏻LGTM

kswenson:
One other thought on this before we merge it. The effect of the bug being fixed here is that specifying individual out-of-range values as part of a date would be allowed. The behavior of such out-of-range values when passed to the JavaScript `Date` object constructor is [documented](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date):

>If any parameter overflows its defined bounds, it "carries over". For example, if a `monthIndex` greater than `11` is passed in, those months will cause the year to increment; if a `minutes` greater than `59` is passed in, `hours` will increment accordingly, etc. Therefore, `new Date(1990, 12, 1)` will return January 1st, 1991; `new Date(2020, 5, 19, 25, 65)` will return 2:05 A.M. June 20th, 2020.
>
>Similarly, if any parameter underflows, it "borrows" from the higher positions. For example, `new Date(2020, 5, 0)` will return May 31st, 2020.

Thus, the fixed code will reject dates that the underlying `Date` object would interpret as valid dates. We should at least consider whether eliminating the validation altogether and preserving the original "buggy" behavior might make sense. In other words, perhaps we should eliminate the bogus validation comparisons rather than fixing them.

kswenson:
After discussion with @bfinzer, we've decided not to merge this into v2 because we don't want to change the current behavior of v2. There might be value in eliminating the incorrect code and explicitly allowing out-of-range values, but since that's the current behavior we're just going to let sleeping dogs lie.

PR #1353, opened 2024-07-16
187797451 v3 DI interactiveFrame Requests

tealefristoe:
PT Stories:
https://www.pivotaltracker.com/story/show/187797451
https://www.pivotaltracker.com/story/show/187985989
https://www.pivotaltracker.com/story/show/187792879

This PR adds support for a number of fields to `get` and `update interactiveFrame` API requests:
- `allowEmptyAttributeDeletion`
- `cannotClose`
- `preventBringToFront`
- `preventDataContextReorg`
- `preventTopLevelReorg`
- `version`

Most of these were not in the model previously, so I added them. Most of them were added to the `WebViewModel` as props so they will be saved and reloaded. I also made `preventAttributeDeletion` and `respectEditableItemAttribute` props instead of volatile because these should also be serializable.

I looked into implementing the behavior of all of these fields with mixed success.
- `allowEmptyAttributeDeletion` - This overrides `preventAttributeDeletion`, used by the Collaborative plugin, when an attribute has no values.
- `cannotClose` - Already implemented.
- `preventBringToFront` - When this is true, the plugin will not move to the top of the tile stack when it's focused, moved, etc. I believe this is the intended functionality of this flag. To accomplish this, I added a `getPreventBringToFront` view to the `TileContent` model, which gets overridden in the `WebViewContent` model.
- `preventDataContextReorg` - As far as I could tell, in v2 this flag prevents reordering attributes in case tables and also prevents title changes in case tables, so that's what I did in v3. In v2 it doesn't seem to prevent title changes in case cards, which seems like a bug, so I went ahead and disabled this in v3.
- `preventTopLevelReorg` - This one is pretty complicated, and if I'd realized that when I started working on it I would have done it in a separate PR. The high level goal is to prevent the user from modifying the parent-most collection, but this is accomplished by many tweaks throughout the code, described below. This feature is used by the Collaborative plugin. In any case, these are the changes made to accomplish this functionality, copied over from v2:
  - Case Table
    - Disable New Attribute in Collection ruler menu item.
    - Does not render an input row for the parent-most collection
    - Disables the menus in the index column for the parent-most collection
    - Disables the new attribute button in the collection title bar for the parent-most collection
      - Note: This actually doesn't work in v2, but looking at the code it seems like it's supposed to
    - Disables the Delete Attribute item in attribute menus for the parent-most collection
      - Note: This actually doesn't work in v2, but looking at the code it seems like it's supposed to
    - Disables the Delete Attribute item in the attribute menu for the last attribute in a child collection
    - Prevents the user from moving an attribute in the parent-most collection via drag-and-drop
    - Prevents the user from moving an attribute into the parent-most collection via drag-and-drop
    - Prevents the user from being able to create a new parent-most collection by dragging an attribute
  - Case Card
    - Disables the new attribute button for the parent-most collection
      - Note that the styling does not show this, but that's what it's like in v2 as well.
    - Also note that it's possible to perform a number of operations in the card card to modify this parent-most collection in v2, such as deleting attributes in the parent collection, or the last attribute in the child collections. I didn't expand what's prevented, even though this feels like a bug to me.
- `version` - The version is now displayed in left of the plugin's title bar. I'm pretty sure that's all the version is supposed to do. I made this a volatile property because it doesn't seem like it is serializable in v2.

Additional improvements:
- I updated the case table ruler menu to include an item to add an attribute to each collection. Previously, there was an item here to add an attribute to the dataset, but v2 has an item for each collection. This was necessary to get one of the restrictions for `preventTopLevelReorg` working.

Note: I originally included `externalUndoAvailable` and `standaloneUndoModeAvailable` in this PR, but I now see that, despite being reported to plugins via `get interactiveFrame` requests, they are not actually closely connected to plugins, and would be better addressed in a different PR.


kswenson:
> PT Stories: [pivotaltracker.com/story/show/187797451](https://www.pivotaltracker.com/story/show/187797451) [pivotaltracker.com/story/show/187792879](https://www.pivotaltracker.com/story/show/187792879)
> 
> This PR adds support for a number of fields to `get` and `update interactiveFrame` API requests:
> 
> * `cannotClose`
> * `preventBringToFront`
> * `preventDataContextReorg`
> * `version`
> 
> Most of these were not in the model previously, so I added them. Most of them were added to the `WebViewModel` as volatile state, but I'm not sure if they should be props instead.

Based on this code, it seems that some of these should be serializable props on the model:
```javascript
      /**
       * List of properties stored to/restored from component storage.
       */
      storedProperties: ['preventBringToFront', 'preventDataContextReorg', 'preventTopLevelReorg',
                          'preventAttributeDeletion', 'allowEmptyAttributeDeletion',
                          'respectEditableItemAttribute'],
```

> I looked into implementing the behavior of all of these fields with mixed success.
> 
> * `cannotClose` - Already implemented.
> * `preventBringToFront` - When this is true, the plugin will not move to the top of the tile stack when it's focused, moved, etc. I believe this is the intended functionality of this flag. To accomplish this, I actually made `preventBringToFront` a volatile property of the `TileModel`, so theoretically it could be true for any tile.

Based on the above, it seems that `preventBringToFront` should be a serializable property of the `WebView` specifically for v2 compatibility.

> * `preventDataContextReorg` - I couldn't figure out what this is supposed to do based on looking at v2 code. It seems like it prevents attributes from being moved, but I wasn't sure.

Yes, it seems that `preventDataContextReorg` prevents attributes from being moved at all. There's a separate property, `preventTopLevelReorg`, that prevents changes to the top-most collection but allows changes in child collections. I don't know whether that's meant to be controllable via the API.

> * `version` - The version is now displayed in left of the plugin's title bar. I'm pretty sure that's all the version is supposed to do.
> 
> Note: I originally included `externalUndoAvailable` and `standaloneUndoModeAvailable` in this PR, but I now see that, despite being reported to plugins via `get interactiveFrame` requests, they are not actually closely connected to plugins, and would be better addressed in a different PR.

tealefristoe:
@kswenson This PR is ready for a review when you have some time.

kswenson (approved):
👍 Looks great! You went above and beyond in terms of implementing the behaviors behind those properties even though you could easily have justified just implementing the existence of the properties and leaving their implementation for later. As we discussed, I pushed a couple of commits with a bunch of mostly cosmetic changes, e.g. naming, API ergonomics, etc. See what you think and let me know if you have thoughts on any of them. Details of some of the changes are listed below:

- Rename `components/web-view/collaborator-utils` => `utilities/plugin-utils`.
  - The functions in there are for more than just collaboration and components generally shouldn't be dependent on other components.
- Rename many `getPrevent...` and `getAllow...` functions to just `prevent...` and `allow...` for readability.
- Add `allowAttributeDeletion` utility function.
  - Fixed a preexisting bug in the `attributeIsEmpty` logic.
- Change the `allow...`/`prevent...` functions to return false/true for `undefined` data sets.
  - `undefined` data sets can't be modified; allows clients to avoid existence tests in many cases.
- Eliminate `preventParentCollectionReorg` and fold its functionality into `preventCollectionReorg`.
- Rename `preventAttributeReorg` to `preventAttributeMove`.
  - Collections are reorganized; attributes are just moved.
- In `CollectionTable`, invert logic from `excludeInputRow` to `showInputRow`.
- In `TileRow.moveTileToTop`, change second argument from `preventBringToFront` to `allowBringToFront`.
  - Avoids a naming conflict between the property and the view in `WebViewModel`.
- Changed the default version number string to the empty string.
  - We shouldn't be supplying bogus version numbers for plugins.

PR #1354, opened 2024-07-16
fix: undo/redo create/delete cases [PT-187127871][PT-187597588]

kswenson:
[[PT-187127871]](https://www.pivotaltracker.com/story/show/187127871)
[[PT-187597588]](https://www.pivotaltracker.com/story/show/187597588)

A number of things were necessary to get this to work as expected:
1. Consistent use of `applyModelChange` and passing appropriate undo/redo strings.
2. Undo/redo of inserting empty cases (e.g. from the index menu) is straightforward.
3. Undo/redo of deleting cases is more complicated
  - We need to use a custom undo/redo implementation to save the contents of the cases before they're deleted and then to restore them to their appropriate positions with appropriate contents, because the values aren't observable/auto-serializable all the way down.
  - The user can delete a discontiguous selection of cases, but the DataSet API only allows passing a single `before` id, so the custom undo/redo code breaks the cases down into contiguous batches, each of which can be inserted with a single call.
  - The code would be simpler if we simply restored the cases one at a time, but one-at-a-time case insertion has been a performance problem in the past that was addressed by making sure that cases were added in batches wherever possible.
4. Within the `DataSet` itself, we were relying on `onAnyAction` to invalidate our cached groupings, but `onAnyAction` isn't triggered on undo/redo. Therefore, I switched to using `onPatch`, which _is_ called on undo/redo, specifically for handling addition/removal of cases/items. We may eventually want to similarly transition other action handlers this way, but that's left as a project for another day.
5. The `useRows` hook, which is responsible for managing the row objects that are rendered by RDG, similarly switched to using `onPatch` for some of its notifications.
6. A downside of `onPatch`, however, is that it is called once for every added/removed case/item, so it can lead to performance issues if one is not careful. In the case of the `DataSet`, it only sets an invalid flag, but `useRows` was doing real work in response, so I debounced the callback so that, presumably, only the last call in the batch will trigger actual work.
7. Penultimately, there was a case table jest test that started failing when I added the debounced callback to `useRows`. I spent a little while trying to fix it (`jest.useFakeTimers`?) and eventually decided it wasn't worth the time and `skip`ped it.
8. Finally, a seemingly unrelated graph cypress test was failing in code that was testing the contents of actively edited cells in the table.(?) Simply accepting the cell edits and moving on to testing the graph functionality seemed the path of least resistance.

### DIFF
--- a/v3/cypress/e2e/graph.spec.ts
+++ b/v3/cypress/e2e/graph.spec.ts
@@ -143,22 +143,18 @@ context("Graph UI", () => {
       cy.dragAttributeToTarget("table", "LifeSpan", "bottom")
       cy.dragAttributeToTarget("table", "Height", "left")
 
-      // change calues in table so that points need rescale
+      // change values in table so that points need rescale
       table.getGridCell(2, 2).should("contain", "African Elephant")
       cy.log("double-clicking the cell")
       // double-click to initiate editing cell
       table.getGridCell(3, 4).dblclick()
-      table.getGridCell(3, 4).find("input").type("700")
-      cy.log("check the editing cell contents")
-      table.getGridCell(3, 4).find("input").should("have.value", "700")
+      table.getGridCell(3, 4).find("input").type("700{enter}")
 
       table.getGridCell(2, 2).should("contain", "African Elephant")
       cy.log("double-clicking the cell")
       // double-click to initiate editing cell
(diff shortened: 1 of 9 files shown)


tealefristoe (approved):
Very clever, looks great 👍

PR #1355, opened 2024-07-17
feat: case table enter key advances to next row [PT-187967544]

kswenson:
[[PT-187967544]](https://www.pivotaltracker.com/story/show/187967544)

From the PT story:
>When editing a cell in the case table, if the user types the enter key it should accept the cell edits and advance to the next row like CODAP v2. By default, RDG enters/exits edit mode without changing the selected cell when the enter key is pressed.

RDG makes it reasonably straightforward to intervene in the key-handling to achieve the desired behavior.


tealefristoe (approved):
Looks great 👍

PR #1356, opened 2024-07-17
fix: undo hierarchy change in Roller Coasters [PT-187967231]

kswenson:
[[PT-187967231]](https://www.pivotaltracker.com/story/show/187967231)

From the PT story:
>Roller Coasters fails in this case because of the hidden `Boundary` attribute. Moving the `State` attribute causes the `Boundary` attribute formula to fail, which results in a call to `setCaseValues()` to set the error message. `setCaseValues()` has a custom undo/redo implementation which interferes with the overall undo/redo of the attribute move.

The key realization here is that `setCaseValues()` calls from the formula engine need not be undoable at all. The same is true for `setCaseValues()` calls from the graph in response to dragging points. Therefore, we take this opportunity to eliminate undo/redo considerations from the `DataSet` itself and add a new `setCaseValuesWithCustomUndoRedo()` utility function for handling undo/redo. Clients for whom undo/redo support is important should then call the new utility function. Clients that call `DataSet.setCaseValues()` directly are on their own.

A consequence of this change is that the `DataSet` is no longer involved in undo/redo considerations, which allowed me to eliminate a dependency cycle between `data-set.ts` and `data-set-undo.ts`.

### DIFF
--- a/v3/src/models/data/data-set-undo.ts
+++ b/v3/src/models/data/data-set-undo.ts
@@ -34,21 +34,14 @@ const setCaseValuesCustomUndoRedoPatcher: ICustomUndoRedoPatcher = {
 }
 
 export function setCaseValuesWithCustomUndoRedo(data: IDataSet, cases: ICase[], affectedAttributes?: string[]) {
-  const items: IItem[] = []
-  cases.forEach(aCase => {
-    // convert each parent case change to a change to each underlying item
-    const caseGroup = data.caseGroupMap.get(aCase.__id__)
-    if (caseGroup?.childCaseIds?.length) {
-      items.push(...caseGroup.childItemIds.map(id => ({ ...aCase, __id__: id })))
-    }
-  })
-  const _cases = items.length > 0 ? items : cases
-  const before = data.getItems(_cases.map(({ __id__ }) => __id__))
+  const items = data.getItemsForCases(cases)
+  const itemIds = items.map(({ __id__ }) => __id__)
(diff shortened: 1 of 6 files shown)


v3/src/models/data/data-set-undo.ts
```diff
@@ -32,3 +32,25 @@ export const setCaseValuesCustomUndoRedo: ICustomUndoRedoPatcher = {
...
 }
+
+export function setCaseValuesWithCustomUndoRedo(data: IDataSet, cases: ICase[], affectedAttributes?: string[]) {
+  const items: IItem[] = []
+  cases.forEach(aCase => {
+    // convert each parent case change to a change to each underlying item
+    const caseGroup = data.caseGroupMap.get(aCase.__id__)
+    if (caseGroup?.childCaseIds?.length) {
+      items.push(...caseGroup.childItemIds.map(id => ({ ...aCase, __id__: id })))
+    }
```
tealefristoe, line 43:
```suggestion
    if (caseGroup?.childCaseIds?.length) {
      items.push(...caseGroup.childItemIds.map(id => ({ ...aCase, __id__: id })))
    } else {
      items.push(aCase)
    }
```
This would allow you to update both cases and items in the same call to `setCaseValuesWithCustomUndoRedo`. But if there's some reason to not do this, I think it would be worth including a comment that says that if you include any cases, items specified will not be updated.

tealefristoe (approved):
Looks good 👍

PR #1297, opened 2024-06-05
fix: iframe reload on component selection

kswenson:
[[PT-187281586]](https://www.pivotaltracker.com/story/show/187281586)

PR #1357, opened 2024-07-17
chore: eliminate no-longer-necessary case table scroll kludge

kswenson:
The original problem was that the `scrollTop` would be reset when the tile order was changed. This is reminiscent of the issue of the WebView iframe reloading when the tile order was changed, and the root cause of that bug -- re-ordering the array of tiles in the `FreeTileRow` model -- is presumably the root cause of this issue as well. When that issue was fixed in #1297, this code became unnecessary, and so we take this opportunity to eliminate it to prevent future confusion.


tealefristoe (approved):
Love it 👍

PR #1358, opened 2024-07-18
187932225 movable modals

eireland:
Makes modals (edit attribute props, insert cases, formula, etc) movable.

### DIFF
--- a/v3/src/components/codap-modal.tsx
+++ b/v3/src/components/codap-modal.tsx
@@ -70,7 +70,6 @@ const DraggableModalContent = ({children, modalWidth, modalHeight, onClick, fowa
 
     const onMouseDown = (e: MouseEvent) => {
       e.stopPropagation()
-      e.preventDefault() // Prevent the default behavior
       isDragging = true
       startX = e.clientX
       startY = e.clientY
--- a/v3/src/components/codap-modal.tsx
+++ b/v3/src/components/codap-modal.tsx
@@ -28,7 +28,7 @@ export const CodapModal = forwardRef(({
     >
       <ModalOverlay />
       <DraggableModalContent
-          fowardRef={ref}
+          fRef={ref}
           modalWidth={modalWidth}
           modalHeight={modalHeight}
           onClick={onClick}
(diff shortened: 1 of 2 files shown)


v3/src/components/codap-modal.tsx
```diff
@@ -15,28 +15,122 @@ interface IProps {
...
-        className="codap-modal-content"
-        h={modalHeight || "500px"}
-        onClick={onClick}
-        ref={ref}
-        w={modalWidth || "400px"}
-      >
-        {children}
-      </ModalContent>
+      <DraggableModalContent
+          fowardRef={ref}
```
kswenson, line 31:
```suggestion
          forwardRef={ref}
```

v3/src/components/codap-modal.tsx
```diff
@@ -15,28 +15,122 @@ interface IProps {
...
   )
 })
 CodapModal.displayName = "CodapModal"
+
+interface IDraggableModalContentProps {
+  children: ReactNode
+  modalWidth?: string
+  modalHeight?: string
+  onClick?: () => void
+  fowardRef: React.Ref<HTMLElement> | undefined
```
kswenson, line 49:
```suggestion
  forwardRef: React.Ref<HTMLElement> | undefined
```

v3/src/components/codap-modal.tsx
```diff
@@ -15,28 +15,122 @@ interface IProps {
...
+  isOpen: boolean
+}
+
+const DraggableModalContent = ({children, modalWidth, modalHeight, onClick, fowardRef, isOpen
```
kswenson, line 53:
```suggestion
const DraggableModalContent = ({children, modalWidth, modalHeight, onClick, forwardRef, isOpen
```

v3/src/components/codap-modal.tsx
```diff
@@ -15,28 +15,122 @@ interface IProps {
...
+    }: IDraggableModalContentProps) => {
+  const [modalPos, setModalPos] = useState({left: 350, top: 250})
+  const modalRef = useRef<HTMLElement | null>(null)
+  const mergedRef = useMergeRefs(fowardRef, modalRef)
```
kswenson, line 57:
```suggestion
  const mergedRef = useMergeRefs(forwardRef, modalRef)
```

kswenson (approved):
👍 Looks good -- just a few typos to fix. At some point we should restrict the movement so you can't drag it mostly offscreen, but since that's the v2 behavior it's fine to leave it as a separate story.

v3/src/components/codap-modal.tsx
```diff
@@ -15,28 +15,122 @@ interface IProps {
...
+      return
+    }
+
+    let isDragging = false
+    let startX = 0
+    let startY = 0
+    let initialX = 0
+    let initialY = 0
+
+    const onMouseDown = (e: MouseEvent) => {
```
kswenson, line 71: One thing I forgot to mention on my earlier review -- the general convention for event handler functions like this is that the implementing functions are named `handleEvent` and the props used to pass them to other components are named `onEvent`. So this function and its brethren would more properly be:
```suggestion
    const handleMouseDown = (e: MouseEvent) => {
```
and then down below when passed as a prop it would be `onMouseDown={handleMouseDown}`.

bfinzer (approved):
👍🏻 LGTM

PR #1359, opened 2024-07-18
187810613 ruler add attribute

eireland:
Adds ability to add a new attribute in a collection from the table ruler menu.


kswenson (changes requested):
Let's put this on hold until #1353 lands because @tealefristoe made changes in this vicinity as well.

v3/src/components/case-table/inspector-panel/ruler-menu-list.tsx
```diff
@@ -35,9 +42,14 @@ export const RulerMenuList = () => {
...
   return (
     <MenuList data-testid="ruler-menu-list">
-      <MenuItem
-        onClick={handleAddNewAttribute}>{t("DG.Inspector.newAttribute", { vars: [data?.name || ""] })}
-      </MenuItem>
+      {collections?.map(collection => {
+        return (
+          <MenuItem key={`${collection.id}`}
+            onClick={()=>handleAddNewAttribute(collection?.id)}>
+            {t("DG.Inspector.newAttribute", { vars: [collection?.name || ""] })}
```
tealefristoe, line 49:
```suggestion
            {t("DG.Inspector.newAttribute", { vars: [collection.title] })}
```

v3/src/components/case-table/collection-table.tsx
```diff
@@ -125,6 +125,12 @@ export const CollectionTable = observer(function CollectionTable(props: IProps)
       caseTableModel)
   }, [caseTableModel, columns, forceUpdate])
 
+  useEffect(() => {
+    if (collectionTableModel?.attrIdToEdit !== undefined) {
+      gridRef.current?.selectCell({idx: columns.length, rowIdx: -1})
```
tealefristoe, line 130: It would be better to figure out the index based on the id, rather than assuming it's the right-most column. That's a safe assumption right now but it could change in the future, in which case this would cause an obscure bug to solve.

tealefristoe (changes requested):
This looks pretty good, but it's going to require a merge with main. This implementation starts editing the attribute title immediately when you create a new one, which is a good touch my implementation didn't include. I think you should be able to mostly use this implementation over mine, but `isDisabled` will have to be brought over from mine. This is just a one line change.

One thing that confuses me about this PR is the change to case card. The PT story makes no mention of the case card. And looking through the code, it looks like the `caseMetaData` is being prop drilled here, but it isn't ever being used for anything. What am I missing?

nstclair-cc (approved):
I focused on reviewing `table.spec.ts`. LGTM! Thanks for updating the specs in the UI of the test. 👍🏼

v3/src/components/case-card/case-card.v2.js
```diff
@@ -16,6 +16,17 @@ import "./column-resize-handle.v2"
...
+  type: PropTypes.any,
+  data: PropTypes.any,
+  collections: PropTypes.object,
+  categories: PropTypes.object,
+  hidden: PropTypes.object,
+  caseTableTileId: PropTypes.string,
+  caseCardTileId: PropTypes.string,
+  lastShownTableOrCardTileId: PropTypes.string // used to restore the last shown tile both have been hidden
+})
+
```
tealefristoe, line 29:
```suggestion
```

v3/src/components/case-card/case-card.tsx
```diff
@@ -22,7 +22,7 @@ export const CaseCard = observer(function CaseCard({ setNodeRef }: IProps) {
   const { active } = useDndContext()
   const instanceId = useInstanceIdContext() || "case-card"
 */
-  const data = useDataSetContext()
+  const {data, metadata} = useDataSet()
```
tealefristoe, line 25:
```suggestion
  const { data } = useDataSet()
```
You could also revert back to `useDataSetContext()`... I'm not sure if there's an advantage either way.

v3/src/components/case-table/collection-table.tsx
```diff
@@ -126,6 +126,13 @@ export const CollectionTable = observer(function CollectionTable(props: IProps)
       caseTableModel)
   }, [caseTableModel, columns, forceUpdate])
 
+  useEffect(() => {
+    if (collectionTableModel?.attrIdToEdit !== undefined) {
+      const newAttrIdx = columns.findIndex(column => column.key === collectionTableModel?.attrIdToEdit)
```
tealefristoe, line 131:
```suggestion
      const newAttrIdx = columns.findIndex(column => column.key === collectionTableModel.attrIdToEdit)
```

v3/src/components/case-table/inspector-panel/ruler-menu-list.tsx
```diff
@@ -7,9 +7,11 @@ import { createAttributesNotification } from "../../../models/data/data-set-noti
 import { uniqueName } from "../../../utilities/js-utils"
 import { preventCollectionReorg } from "../../../utilities/plugin-utils"
 import { t } from "../../../utilities/translation/translate"
+import { useCaseTableModel } from "../use-case-table-model"
```
tealefristoe, line 10: Remove line 5, `import { ICollectionModel } from "../../../models/data/collection"`.

tealefristoe (approved):
Still a few more things to clean up, but after that you should be good to go 👍

v3/src/components/case-card/case-card.v2.test.tsx
```diff
@@ -7,6 +7,7 @@ import { createDataSet } from "../../models/data/data-set-conversion"
 import { DGDataContext } from "../../models/v2/dg-data-context"
 import { t } from "../../utilities/translation/translate"
 import "./case-card.v2"
+import { SharedCaseMetadata } from "../../models/shared/shared-case-metadata"
```
kswenson, line 10: These `metadata` imports and properties should be removed from the tests as well at some point.

PR #1361, opened 2024-07-19
Support dates in formulas with +/- operators, add date() implementation, tests

pjanik:
https://www.pivotaltracker.com/story/show/187799270

This PR adds support for dates in the formula context. I've tried to replicate V2 behavior, but hopefully, it's also improved in some cases:
- We always produce date output instead of epoch values.
- The `-` (subtract) operator is supported.
- The `date()` formula works better when referencing other attributes.

Also, V2 had essentially 3 ways of interpreting input as a date: `parseDate` helper, `createDate` helper, `date()` formula.
They're slightly overlapping, but not too much and they might produce different results. E.g. they'd all return different result for "25" input. So, I'm trying to cleanup that in V3 a bit, and just have two options:
- `parseDate`
- `date()` formula with all its quirks and implicit treatment of the year argument

So, I'm intentionally not bringing in `createDate` (I removed it from V3 helpers). Hopefully places where `createDate` used to be called can be replaced with parseDate or regular new Date().

So far I'm doing well without custom/modified Date class that has `.valueOf()` modified to return seconds instead of milliseconds. I don't think we need that, but let me know if you see a reason for that. The only place where that matters is the + and - operators, as that's probably the only place where a Date object might be mixed with a number that has to be interpreted as ms or seconds. But handling that doesn't require using a custom Date class and modifying the numeric value internally.

Another side effect of this PR is that users might now concatenate strings in formulas using the `+` operator.

### DIFF
--- a/v3/src/models/formula/functions/date-functions.test.ts
+++ b/v3/src/models/formula/functions/date-functions.test.ts
@@ -1,0 +1,56 @@
+import { formatDate } from "../../../utilities/date-utils"
+import { math } from "./math"
+
+describe("date", () => {
+  it("returns current date if no arguments are provided", () => {
+    const fn = math.compile("date()")
+    expect(fn.evaluate()).toEqual(formatDate(new Date()))
+  })
+
+  it("interprets a small number as 20xx year if it's smaller than current year + 10", () => {
+    const fn = math.compile("date(x)")
+    expect(fn.evaluate({ x: 1 })).toEqual(formatDate(new Date(2001, 0, 1)))
+    expect(fn.evaluate({ x: 10 })).toEqual(formatDate(new Date(2010, 0, 1)))
+    expect(fn.evaluate({ x: 20 })).toEqual(formatDate(new Date(2020, 0, 1)))
+    expect(fn.evaluate({ x: 30 })).toEqual(formatDate(new Date(2030, 0, 1)))
+  })
+
+  it("interprets a small number as 19xx year if it's bigger than current year + 10, but smaller than 100", () => {
+    const fn = math.compile("date(x)")
+    expect(fn.evaluate({ x: 50 })).toEqual(formatDate(new Date(1950, 0, 1)))
+    expect(fn.evaluate({ x: 60 })).toEqual(formatDate(new Date(1960, 0, 1)))
+    expect(fn.evaluate({ x: 70 })).toEqual(formatDate(new Date(1970, 0, 1)))
+    expect(fn.evaluate({ x: 99 })).toEqual(formatDate(new Date(1999, 0, 1)))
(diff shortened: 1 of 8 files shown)


v3/src/models/formula/functions/date-functions.test.ts
```diff
@@ -0,0 +1,56 @@
+import { formatDate } from "../../../utilities/date-utils"
+import { math } from "./math"
+
+describe("date", () => {
+  it("returns current date if no arguments are provided", () => {
+    const fn = math.compile("date()")
+    expect(fn.evaluate()).toEqual(formatDate(new Date()))
+  })
+
+  it("interprets a small number as 20xx year if it's smaller than current year + 10", () => {
```
kswenson, line 10: In the date-parser's `fixYear` function, we use:
```typescript
function fixYear(y: string) {
    if (y.length === 2) {
      const yNumber = Number(y)
      if (yNumber < 49) {
        return 2000 + yNumber
      } else {
        return 1900 + yNumber
      }
    }
```
Shouldn't these two assumptions be synchronized? Perhaps current year + 20 or 25?

v3/src/models/formula/functions/date-functions.ts
```diff
@@ -0,0 +1,61 @@
...
+      const monthIndex = args[1] != null ? Math.max(0, Number(args[1]) - 1) : 0
+      const day = args[2] != null ? Number(args[2]) : 1
+      const hours = args[3] != null ? Number(args[3]) : 0
+      const minutes = args[4] != null ? Number(args[4]) : 0
+      const seconds = args[5] != null ? Number(args[5]) : 0
+      const milliseconds = args[6] != null ? Number(args[6]) : 0
+
+      const currentYear = new Date().getFullYear()
+      // dividing line for two-digit years is 10 yrs from now
+      const cutoffYear = (currentYear + 10) % 100
```
kswenson, line 44: See comment elsewhere about synchronizing this assumption.

kswenson (approved):
👍 LGTM -- just one question about synchronizing assumptions about two-digit dates.

bfinzer (approved):
👍🏻 LGTM

pjanik:
@kswenson, I've just pushed commits that synchronize these two assumptions.

I've decided to go with constant ranges [0, 50) and [50, 100) as it's pretty close to the suggested current_year + 25, but it does not change over time. While I understand this way of thinking, I believe it might be useful to be consistent and not change the interpretation based on the current year. It could be confusing to have the same document break because now it's 2025 or 2026 instead of 2024 (and since the cutoff would be around 50, that's quite likely).

kswenson (approved):
👍 Looks good

PR #1362, opened 2024-07-19
fix: case table enter key advances to next row [PT-187967544]

kswenson:
[[PT-187967544]](https://www.pivotaltracker.com/story/show/187967544)

The [previous fix](https://github.com/concord-consortium/codap/pull/1355) worked some of the time but other times the internal event handling of RDG would exit edit mode immediately after we'd requested to enter it. This version adds a `setTimeout` so that edit mode is entered after RDG has completely finished processing the current event and no longer has the opportunity to summarily cancel the edit mode we've requested. This version also add the ability to advance (retreat?) to the previous row if the shift key is pressed along with the enter key.


tealefristoe (approved):
Looks good 👍

PR #1363, opened 2024-07-19
chore: reduce cypress waits

kswenson:
The `dragAttributeToTarget` command is used repeatedly throughout our cypress tests and specifies that it must wait for five seconds after each drop before continuing with the test. CODAP animations generally have two second durations, so even if a test requires waiting for animation to complete, 2.5 seconds should be plenty of time. This PR tests that hypothesis.

### DIFF
--- a/v3/cypress/support/commands.ts
+++ b/v3/cypress/support/commands.ts
@@ -130,7 +130,7 @@ Cypress.Commands.add("dragAttributeToTarget", (source, attribute, target, target
         })
       })
     })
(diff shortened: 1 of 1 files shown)


nstclair-cc:
lgtm, and I'm all about reducing Cypress waits. I see some conflicts in the branch but otherwise good to go!

nstclair-cc (commented):
Please resolve branch conflicts and then go ahead and merge!

PR #1364, opened 2024-07-20
fix: collapsed case rows shouldn't show index menu [PT-187986400]

kswenson:
[[PT-187986400]](https://www.pivotaltracker.com/story/show/187986400)

- clicking on collapsed case row no longer shows index menu
- clicking on collapsed case row selects child cases (and parent case)


tealefristoe (approved):
Pretty low test coverage, but otherwise looks good 👍

kswenson:
> Pretty low test coverage, but otherwise looks good 👍

I added a test.

PR #1365, opened 2024-07-20
fix: case table column header response to undo edit properties change [PT-187237814]

kswenson:
[[PT-187237814]](https://www.pivotaltracker.com/story/show/187237814)


bfinzer (approved):
👍🏻LGTM

PR #1367, opened 2024-07-22
make parseDate support more date formats (e.g. 3-digit years) 

pjanik:
While testing some edge cases, I've noticed that the existing V2 parser doesn't work well with years that have three digits. Since I rely on parseDate more than on V2, I wanted to fix that. Rather than modifying the existing regexp patterns, I've set things up in a way that allows us to gradually move to other date parsers in the future. Currently, I'm using the built-in Date.parse, but it's easy to replace it with something else.


kswenson (approved):
👍 Looks good

bfinzer (approved):
👍🏻 LGTM

PR #1368, opened 2024-07-22
Implement most of date functions, extract createDate/ and convertToDate, add tests

pjanik:
https://www.pivotaltracker.com/story/show/187799270

This PR adds most of the date-related formula implementations. I've also extracted `createDate` (essentially making the `date()` formula reuse it 100%, in contrast to the V2 pattern) and `convertToDate`.

One breaking change compared to V2 is that formulas returning the day or month name now always do so in English. V2 would return these in the local language. On one hand, that's useful. On the other hand, it prevents users from using this value in further calculations (e.g., `if (...)`). However, if necessary, users can always use formulas that return numeric values. So I wonder if I should actually change it or not.

Finally, the only formula not implemented yet is `seconds`. This one revealed an issue in the current approach, where dates are converted to strings too early, causing the seconds to be lost in the default date formatting. I'll work on that tomorrow. However, I'm not sure if this is going to be a bigger issue. `seconds(date(2020, 5, 4, 3, 2, 1))` should be relatively easy to fix. However, `seconds(other_attribute_with_formatted_date)` might be much harder (or even impossible) with the current approach where attribute values are always strings and formulas store their results in attribute values.

### DIFF
--- a/v3/src/models/formula/functions/date-functions.test.ts
+++ b/v3/src/models/formula/functions/date-functions.test.ts
@@ -1,4 +1,5 @@
 import { formatDate } from "../../../utilities/date-utils"
+import { UNDEF_RESULT } from "./function-utils"
 import { math } from "./math"
 
 describe("date", () => {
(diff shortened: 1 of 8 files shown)


kswenson (approved):
👍 Looks good -- thx for thinking through some of these consistency issues.

>One breaking change compared to V2 is that formulas returning the day or month name now always do so in English. V2 would return these in the local language. On one hand, that's useful. On the other hand, it prevents users from using this value in further calculations (e.g., `if (...)`). However, if necessary, users can always use formulas that return numeric values. So I wonder if I should actually change it or not.

I don't have a strong feeling for what makes sense _a priori_. I can see your argument that a consistent return value is better than a locale-dependent one in the world of formula evaluation, but the backward-compatibility argument also carries some weight. 🤷 @bfinzer?

bfinzer:
Regarding `monthName()` and `dayOfWeekName`: These are most often used to create a categorical attribute that can be displayed on a graph or map to help a user understand the data. Thus local language is most helpful for this purpose. Using the result in a formula would be rare, and, as you note, better done with numeric values. So I vote for keeping V2 behavior.

pjanik:
I just pushed a commit to revers my change and restores V2 behavior. So, I'm merging this PR.

PR #1369, opened 2024-07-22
187994058 v3 DI horizontalScrollOffset

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187994058

This PR adds support for `horizontalScrollOffset` for case tables.
- This field updates automatically as the user scrolls and is serializable, so will reload when a saved document is opened.
- The field can be specified with `create` and `update component caseTable` API requests.

I was also able to remove a hack that was necessary with the old method of specifying tile order.


v3/src/components/case-table/case-table.tsx
```diff
@@ -37,16 +35,26 @@ export const CaseTable = observer(function CaseTable({ setNodeRef }: IProps) {
...
+    // Initial scoll is delayed a frame to let RDG do its thing
+    setTimeout(() => updateScroll(tableModel?.horizontalScrollOffset))
+
+    // Handle changes to the model, such as via the API
+    return reaction(
+      () => tableModel?.horizontalScrollOffset,
+      horizontalScrollOffset => {
+        updateScroll(horizontalScrollOffset)
+      }
+    )
```
kswenson, line 56: Should use `mstReaction` here so that the reaction is disposed when the `tableModel` is destroyed. Also, if you pass `fireImmediately: true`, you could avoid the separate call to `updateScroll` on initialization, although you'd presumably have to use `setTimeout` inside the reaction in that case, so not clear whether that's a win or not. 🤷

v3/src/components/case-table/case-table-model.ts
```diff
@@ -11,11 +11,9 @@ export const CaseTableModel = TileContentModel
     type: types.optional(types.literal(kCaseTableTileType), kCaseTableTileType),
     // key is attribute id; value is width
     columnWidths: types.map(types.number),
-  })
-  .volatile(self => ({
     // entire hierarchical table scrolls as a unit horizontally
-    scrollLeft: 0
-  }))
+    horizontalScrollOffset: 0
+  })
```
kswenson, line 16: Rather than eliminating the `volatile` property, my inclination would be to follow the pattern we use elsewhere for such ephemeral properties that we actually want to store in the document of manipulating the `volatile` property during the running of the app, and then having a `prepareSnapshot` method in the `CaseTableModel` which copies the `volatile` property into the serialized property, and an `afterCreate` method which copies the serialized property into the `volatile` property. Then you don't need `applyModelChange` because you're only modifying a `volatile` property, and there's not as much MST machinery spun up during scrolling operations. Volatile properties are still observable, so your `reaction` would still work. There may be something I'm missing, however, so let me know if there's a reason to prefer this implementation.

kswenson (changes requested):
👍 In general this looks good. I made a couple of minor suggestions and one medium-sized one related to keeping the `volatile` property -- details below.

v3/src/components/case-table/case-table.tsx
tealefristoe: I was originally using `fireImmediately: true`, but switched it so the `setTimeout` is only for the initial load. My preference is to minimize `setTimeout` (I'm not even happy about the one!) but if you prefer the other way around I'll switch it.

kswenson: No, I think you're right that restricting the use of `setTimeout` to the situations we need it is probably a higher priority. I just wanted to point out the option so the options was considered. There is such thing as being too DRY.

kswenson (approved):
👍 Looks good!

PR #1370, opened 2024-07-23
Make Date a valid IValueType, refactor how dates are handled in formulas and case table

pjanik:
This PR aims to fix some design issues that were causing the following problem:

`seconds(date(2020, 5, 14, 5, 4, 3))` or `seconds(AnotherAttrWithFormattedDate)` was returning 0 because I used the `formatDate` utility function, which formats the date to a locale string often without seconds, too early in the whole processing chain. The problem was that Date was not a first-class citizen type and was always stringified, and some data was lost.

Here’s what I did to fix it:
- I made Date a valid type that can be passed directly in formulas. This fixed `seconds(date(2020, 5, 14, 5, 4, 3))` because now `seconds` receives a Date instance.
- I also made Date a valid IValueType so it can be passed around in data-set/attribute functions. I updated `importValueToString` to serialize Date to an explicit string format (I chose ISO string, as it's complete, can be easily parsed, and is relatively compact) because Date cannot be stored in any attribute storage (string values or numeric values). This also fixes `seconds(AnotherAttrWithFormattedDate)` because each attribute now stores complete date information (ISO strings are a complete representation of a date). This is better than a number representing epoch milliseconds, as ISO strings can be easily interpreted as dates, unlike milliseconds which can be mistaken for regular numbers.
- Finally, I updated `use-columns` to recognize this ISO string and treat it as if it were an instance of Date. I believe the comments in this file explain that clearly.

@kswenson, does this make sense to you? Another approach might be to have a "real" Date storage in the attribute, similar to strings and numbers. However, I'm not sure where and when even these numeric values are used, as often various helpers work with string values only (e.g. useColumns). And I'm concerned this would be quite a significant change. Also, you might have a better sense of how to implement this approach than I do. This is always an option for the future, as the current implementation does not lock us into it in any way.

Also, I did not update `attribute.toNumeric` to convert date strings to epoch seconds/milliseconds because I don't see a use case for it yet, and it would cause some performance overhead if we had to parse each string passed there, hoping it's a date.

### DIFF
--- a/v3/src/models/data/attribute.ts
+++ b/v3/src/models/data/attribute.ts
@@ -40,7 +40,7 @@ export const kDefaultFormatStr = ".3~f"
 const isDevelopment = () => process.env.NODE_ENV !== "production"
 const isProduction = () => process.env.NODE_ENV === "production"
 
-export type IValueType = string | number | boolean | undefined | Date
+export type IValueType = string | number | boolean | Date | undefined
 
 export interface ISetValueOptions {
   noInvalidate?: boolean
--- a/v3/src/models/data/attribute.ts
+++ b/v3/src/models/data/attribute.ts
@@ -55,7 +55,7 @@ export function importValueToString(value: IValueType): string {
   }
   if (value instanceof Date) {
     // Convert Date to ISO string format. It's a consistent format that can be parsed back into a Date object
-    // without loosing any information. Also, it's relatively compact and it can be easily recognized as a date string,
+    // without losing any information. Also, it's relatively compact and it can be easily recognized as a date string,
     // in contrast to storing the date as a number (e.g. milliseconds since epoch).
     return value.toISOString()
   }
--- a/v3/src/models/data/attribute.ts
+++ b/v3/src/models/data/attribute.ts
@@ -40,14 +40,26 @@ export const kDefaultFormatStr = ".3~f"
 const isDevelopment = () => process.env.NODE_ENV !== "production"
 const isProduction = () => process.env.NODE_ENV === "production"
 
-export type IValueType = string | number | boolean | undefined
+export type IValueType = string | number | boolean | undefined | Date
 
 export interface ISetValueOptions {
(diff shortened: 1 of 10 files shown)


v3/src/models/data/attribute.ts
```diff
@@ -40,14 +40,26 @@ export const kDefaultFormatStr = ".3~f"
 const isDevelopment = () => process.env.NODE_ENV !== "production"
 const isProduction = () => process.env.NODE_ENV === "production"
 
-export type IValueType = string | number | boolean | undefined
+export type IValueType = string | number | boolean | undefined | Date
```
kswenson, line 43: I would put the `undefined` last, so:
```suggestion
export type IValueType = string | number | boolean | Date | undefined
```

v3/src/utilities/date-parser.ts
```diff
@@ -302,3 +302,12 @@ export function isDateString(iValue: any, iLoose?: boolean) {
     return spec.regex.test(iValue)
   }) || (!!iLoose && parseDateV3(iValue) != null)
 }
+
+// Regular expression to match ISO 8601 date strings as produced by Date.toISOString.
+// Note that this regular expression is more strict than the one used in parseDate (isoDateTimeRE) which supports
+// additional formats.
+const browserIsoDatePattern = /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.\d{3}Z$/
```
kswenson, line 309: Seems like we should support the [27-char](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString) format as well. Something like:
```suggestion
const browserIsoDatePattern = /^([+-]\d{6}|\d{4})-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.\d{3}Z$/
```

kswenson (approved):
👍 Looks good -- I think this is the right approach. Just a handful of minor suggestions.

PR #1371, opened 2024-07-23
Update MathJS to v12.4.3 and handle the new approach to scope/partitioned map

pjanik:
https://www.pivotaltracker.com/story/show/187950451

This fix doesn't look pretty, but it seems to work. @kswenson, you're right that the problematic change on the MathJS side was introduced here: https://github.com/josdejong/mathjs/pull/3150.

### DIFF
--- a/v3/src/models/formula/functions/function-utils.ts
+++ b/v3/src/models/formula/functions/function-utils.ts
@@ -38,7 +38,7 @@ export const evaluateNode = (node: MathNode, scope?: FormulaMathJsScope) => {
 // This function is used to get the root scope (an instance of our FormulaMathJSScope) within custom MathJS functions.
 // When the formula expression is executed, the initially passed scope can be wrapped in MathJS's PartitionedMap, which
 // is used to store temporary values. This function retrieves the root scope from the PartitionedMap.
-// This approach has been recommended by a MathJS maintainer. For more details, see:
+// This approach has been recommended by the MathJS maintainer. For more details, see:
 // https://github.com/josdejong/mathjs/pull/3150#issuecomment-2248101774
 export const getRootScope = (currentScope: CurrentScope): FormulaMathJsScope => {
   return isPartitionedMap(currentScope) ? getRootScope(currentScope.a) : currentScope
--- a/v3/src/models/formula/functions/function-utils.ts
+++ b/v3/src/models/formula/functions/function-utils.ts
@@ -1,6 +1,7 @@
 import { MathNode } from "mathjs"
 import { FormulaMathJsScope } from "../formula-mathjs-scope"
 import { isValueNonEmpty } from "../../../utilities/math-utils"
+import { CurrentScope, MathJSPartitionedMap } from "../formula-types"
 export { isNumber, isValueNonEmpty } from "../../../utilities/math-utils"
 
 export const UNDEF_RESULT = ""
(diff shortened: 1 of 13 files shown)


kswenson (changes requested):
👍 Thx for doing this. Quick review for now -- I'll do a more thorough review on the next round. Two suggestions before that, however:
1. There's a failing jest test that appears related that should be fixed. (I suspect the failing cypress test is just one of our usual flaky ones.)
2. A comment that explains how this interacts with MathJS would be very helpful. I'm interested in the question of whether we're relying on documented aspects of MathJS, in which case the change that triggered this should have been semver-major, or perhaps more likely, we're relying on undocumented implementation details, in which case we should at least document that fact so that we'll know where to start looking the next time this comes up and you're not around. 😉

kswenson:
Also, since `PartitionedMap` has no meaning in CODAP, perhaps we can call it something like `scopeMap` when used as an argument that might give some indication of its purpose.

pjanik:
@kswenson, sorry, I actually forgot to push the last changes, that's why the test was broken. The last commit fixes that and reactors a few things a bit.

I've actually added a type called `MathJSPartitionedMap`. Not sure, but I'd actually suggest to keep it, as that's what might allow us to orient ourselves in the MathJS code. If it's just scope, it can be confusing. While paritionedMap doesn't say much in CODAP, it's something passed from MathJS, and at least it explains why we need to access `.a` property.

> A comment that explains how this interacts with MathJS would be very helpful. I'm interested in the question of whether we're relying on documented aspects of MathJS

I based that implementation on the fact that the official documentation described scope as an object that implements the Map interface, and also on this example (which seems to be outdated now, btw):
https://mathjs.org/examples/advanced/custom_scope_objects.js.html
But apparently, MathJS isn't that strict about that, and the the example is now outdated and nonfunctional.

My last commit adds a comment about all that where the MathJSPartitionedMap is defined.

I'm also talking directly to MathJS maintainer:
https://github.com/josdejong/mathjs/pull/3150
He's very responsive. Apparently, they might be considering a better solution for our use case. It would be good if someone else also follows that thread in the PR in case I'm working on another project when the fix arrives. I'll get notified anyway, but just in case.

pjanik:
Okay, I've just pushed the final commit. I followed the suggested approach:
https://github.com/josdejong/mathjs/pull/3150#issuecomment-2248101774
I could make checks more strict, but this would make testing more annoying, so I hope it's a reasonable compromise.

I hope that new types and variable names are also clearer.

v3/src/models/formula/functions/function-utils.ts
```diff
@@ -33,3 +34,23 @@ equal = (a: any, b: any): boolean => {
 export const evaluateNode = (node: MathNode, scope?: FormulaMathJsScope) => {
   return node.compile().evaluate(scope)
 }
+
+// This function is used to get the root scope (an instance of our FormulaMathJSScope) within custom MathJS functions.
+// When the formula expression is executed, the initially passed scope can be wrapped in MathJS's PartitionedMap, which
+// is used to store temporary values. This function retrieves the root scope from the PartitionedMap.
+// This approach has been recommended by a MathJS maintainer. For more details, see:
```
kswenson, line 41: I think it's safe to call him _the_ MathJS maintainer. 😉 
```suggestion
// This approach has been recommended by the MathJS maintainer. For more details, see:
```

v3/src/models/formula/functions/other-functions.ts
```diff
@@ -51,4 +51,6 @@ export const otherFunctions = {
     }
   },
 
+
+
```
kswenson, line 55: Probably don't need the extra blank lines here. 😉 
```suggestion
```

kswenson (approved):
👍 Looks great! Thx for the fix and for contacting the maintainer to get some additional guidance.

PR #1373, opened 2024-07-24
feat: implement `number` formula [PT-188007822]

pjanik:
https://www.pivotaltracker.com/story/show/188007822

The custom implementation of the `number` formula was missing, and it will be useful for Joe in his work on the CODAP plugins.


kswenson (approved):
👍 LGTM

PR #1374, opened 2024-07-24
188002783 investigate flaky table tests

nstclair-cc:
[PT-188002783](https://www.pivotaltracker.com/story/show/188002783)

**Description:**
This pull request addresses several flaky Cypress tests that needed investigation and fixes. The primary issues seemed to be related to finding target elements, likely due to changes in the elements for editing table cells and editing the slider value component.

**Summary of Changes:**
1. **Add Logs for CODAP Document Load:**
   - Added logs in each `beforeEach()` statement to ensure that the CODAP document loads correctly. This addition follows the discussion in the standup on 7/24.

2. **Fix Table Cell Editing Tests:**
   - Adjusted the selectors and interaction methods to ensure that the tests for editing table cells are more reliable.

3. **Enhance Color Spec Tests:**
   - The color spec is functioning correctly locally but fails remotely. I suggested some code but commented it out for now and will create a separate PT story to fix the flaky test. This test could be improved if we ensure our HTML includes the correct data-testid attributes.

**Details:**
- Logs have been added to provide better visibility into the loading process of CODAP documents. This will help in identifying any issues that occur before the tests run.
- For the table cell editing tests, updated selectors and ensured proper wait times to account for any changes in the DOM structure.
- The color spec test adjustments have been tested locally and show successful results. However, there are inconsistencies in remote environments that need to be addressed further.

**Next Steps:**
- Monitor the CI/CD pipeline to identify if the logs provide sufficient information for any further issues.
- Continue enhancing the color spec test to ensure it passes in both local and remote environments.
- Investigate any new flaky tests that might arise and apply similar debugging and fixing strategies.

---

Thank you!


kswenson (approved):
👍 Looks good

I added a couple of commits with some more changes:
1. Remove `&dashboard` from the visit url -- the table specs don't rely on any other components, so there's no point in loading them.
2. Add a `cy.intercept()` for the AWS request for the published plugins, which I believe to be the root of the failed fetch errors.
3. Reduced the wait after visit to one second, since we no longer have to wait for the AWS published plugins response.
4. Re-enabled some tests that just needed to have their logic tweaked.
5. Added an additional click to the color swatch test which improved its success rate when testing locally.

kswenson:
Also, moving forward we should:
1. Figure out a way to add that `cy.intercept()` line to all of the tests (preferably better than copy-paste).
2. Remove the logging around the initial load assuming this solves the problem.

nstclair-cc:
> * Figure out a way to add that `cy.intercept()` line to all of the tests (preferably better than copy-paste).
> * Remove the logging around the initial load assuming this solves the problem.

Thank you @kswenson! I think I'll merge the code for now so that we can fix the flaky tests issue. I also added a separate PT story here for the issues raised: https://www.pivotaltracker.com/story/show/188015800

PR #1377, opened 2024-07-26
fix: parsing of inequality operators in formulas

kswenson:
[[PT-188009575]](https://www.pivotaltracker.com/story/show/188009575)

We preprocess expressions to convert `=` to `==`, but the regular expression used to do so needed to be more restrictive.


tealefristoe (approved):
Looks good 👍

PR #1378, opened 2024-07-26
182089941 log events

eireland:
Adds storing document metadata in the `properties` property of the document model.
Adds logger library to send log messages to the log server (currently using concordqa server)
Adds some log messages to some actions for testing (adding a tile, changing a tile title, and closing a tile), though the actual language is still different from V2

Things not included:
1. Adding run_remote-endpoint to the log message
2. Adding parameters to the log message


kswenson (approved):
👍 Looks good -- as we discussed I added a couple of commits to tweak things a bit. Also, I feel like there must be more subtleties around the document name/title issue but we can figure that out as we go along.

PR #1380, opened 2024-07-27
fix: undo multiple y [PT-187833677]

kswenson:
[[PT-187833677]](https://www.pivotaltracker.com/story/show/187833677)

The crash occurred on undoing the drop of a second (or more) Y attribute because `GraphDataConfigurationModel.filterCase` can get called for the "missing" attribute and crashed when it encountered an `undefined` attribute description. Fixing that was straightforward, but revealed an underlying issue that the undo/redo of adding an additional y attribute still didn't work. The fix there was to add a MobX reaction to synchronize the `filteredCases` array with the attribute configuration. This meant that some bespoke code to manage that synchronization in response to particular events could be eliminated. The new reaction also relies on making `filteredCases` a MobX shallow-observable array so that additions/removals to/from it can trigger the necessary reaction(s).


bfinzer (approved):
👍🏻LGTM

PR #1381, opened 2024-07-27
187738952 v3 DI Get Component

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187738952

This PR adds support for most `get component` API requests. Types that remain unsupported have yet to be implemented in v3, most notably text and guide.

A few tiles/components return more information than they did in v2:
- Sliders include information about their globals, `globalValueName` and `value`.
- Graphs include information about more roles: `captionAttributeName`, `rightSplitAttributeName`, `topSplitAttributeName`.
- CaseCards and Maps include their dataset name under `dataContext`.

Additionally, this PR makes `TileModel` derive from `V2Model` so tiles/components now have names, which can be useful in referencing them. That said, following v2, default tile names have limited utility:
- Most tiles have no names at all.
- The calculator is called "Calculator".
- All maps are called "Map" (meaning that if you add multiple maps, they will all have the same name).
- In v2, webViews are all called "Web View", while games have no names. In v3, these tiles are the same under the hood, and it's not possible to tell if a particular tile will be a game or webView when it is created. So as of this PR, both tile types are called "Web View".


kswenson (commented):
👍 Looks good -- as we discussed, I added a couple commits which refactored things somewhat, notably introducing a component handler registration system so that component-specific code could live with the component. Take a look and fix anything that might have been left in a broken state, e.g. the couple of jest tests I commented out. I can re-review once everything is working again.

tealefristoe:
@kswenson I got the jest tests working again on this PR so it's ready for another review.

I did make one additional change, which was to get rid of the `setOptions` and `setCreateOrShow` arguments to `ComponentHandler.create()`, and instead allow a custom `createOrShow` function and/or `options` object to be included in a dictionary, along with the `content`.

I didn't carefully check all changes made from the original PR, so I hope no problems snuck through. I think it would have been easier to manage these changes if the original PR was accepted and merged, and then the `ComponentHandler` registration system was introduced in a separate PR. I think these changes were all improvements, it's just difficult to tell what changed when all the changes are stacked on top of each other.

PR #1382, opened 2024-07-29
Fix Flaky Cypress Test for Slider Component

nstclair-cc:
### Pull Request: Fix Flaky Cypress Test for Slider Component
[PT-188031351](https://www.pivotaltracker.com/story/show/188031351)

**Description:**
This PR addresses a flaky Cypress test related to the slider component. The test intermittently failed due to assertion errors caused by the slider component's shadow rendering with an effective width of 0. Changes include:

1. **Check Existence Before Visibility:** Ensuring the component exists in the DOM before checking visibility.
2. **Check Internal Elements:** Verifying visibility of internal elements within the slider component.

Please review and provide feedback. Thank you!

### DIFF
--- a/v3/cypress/support/elements/slider-tile.ts
+++ b/v3/cypress/support/elements/slider-tile.ts
@@ -2,7 +2,7 @@ import { ComponentElements as c } from "./component-elements"
 
 export const SliderTileElements = {
   getSliderTile(index = 0) {
-    return c.getComponentTile("slider", index)
+    return cy.get(`[data-testid="codap-slider"]`).eq(index)
   },
   getVariableName(index = 0) {
     return this.getSliderTile(index).find("[data-testid=slider-variable-name-text]")


johnTcrawford (approved):
Looks good to me.

v3/cypress/support/elements/slider-tile.ts
```diff
@@ -2,7 +2,7 @@ import { ComponentElements as c } from "./component-elements"
 
 export const SliderTileElements = {
   getSliderTile(index = 0) {
-    return c.getComponentTile("slider", index)
+    return cy.get(`[data-testid="codap-slider"]`).eq(index)
```
kswenson, line 5: Do you have any reason to believe that failure to find the element is the root of the problem? When I have seen it fail the message is:
```
expected '<div.codap-component.codap-slider.shadowed>' to be 'visible'

This element <div.codap-component.codap-slider.shadowed> is not visible because it has an effective width and height of: 0 x 98 pixels.
```
which indicates that finding the element isn't the problem -- there's something about the properties of the element that leads cypress to believe that it has zero width. I'm hesitant to introduce a code change which treats the slider differently from other components in cypress tests without at least a theory as to why that is appropriate.

kswenson (changes requested):
I'm not convinced that this changes the behavior of the test. See inline comment for details.

kswenson:
Superseded by #1404.

PR #1383, opened 2024-07-29
fix: disable vertical auto-scroll of table on attribute drag [PT-187949371]

kswenson:
[[PT-187949371]](https://www.pivotaltracker.com/story/show/187949371)

From the PT story:
>The bug occurs when dragging an attribute near the bottom of a table triggers auto-scroll, which allows the column headers to be scrolled out of view and leaves a gap at the bottom of the table. The table must support auto-scroll horizontally when dragging an attribute, but should not auto-scroll vertically.

Auto-scrolling during drag is the province of DndKit, which supports several auto-scroll configuration options described in [DndKit #140](https://github.com/clauderic/dnd-kit/pull/140). In particular, the `canScroll` callback function is used to indicate particular elements that should not be auto-scrolled. None of the existing configuration options make it possible to support auto-scroll in one dimension only rather than enabling/disabling scrolling for all dimensions.

Therefore, we patch `@dnd-kit/core` to pass `direction` information to the `canScroll` callback function so that clients can enable/disable scrolling in each dimension separately. In addition, we implement a registry of such callback functions so that individual tiles can register their individual `canScroll` callbacks. The table uses this to allow horizontal auto-scroll but prevent vertical auto-scroll of the case table component, and to disable auto-scroll of the table body (grid cells) in both directions since there's nothing droppable to scroll to in the grid.

The `@dndkit/core` patch changes from this:
```typescript
        if (canScroll?.(scrollContainer) === false) {  <= canScroll called with just an element
          continue;
        }

        const index = scrollableAncestors.indexOf(scrollContainer);
        const scrollContainerRect = scrollableAncestorRects[index];

        if (!scrollContainerRect) {
          continue;
        }

        const {direction, speed} = getScrollDirectionAndSpeed(
          scrollContainer,
          scrollContainerRect,
          rect,
          acceleration,
          threshold
        );

```

in which the `canScroll` function is only passed the element, to:
```typescript
        const index = scrollableAncestors.indexOf(scrollContainer);
        const scrollContainerRect = scrollableAncestorRects[index];

        if (!scrollContainerRect) {
          continue;
        }

        const {direction, speed} = getScrollDirectionAndSpeed(
          scrollContainer,
          scrollContainerRect,
          rect,
          acceleration,
          threshold
        );

        if (canScroll?.(scrollContainer, direction) === false) {  <= canScroll() is passed a direction as well
          continue;
        }
```


eireland (approved):
👍 LGTM

PR #1385, opened 2024-07-31
182089941 add log strings (Part 1)

eireland:
Adds logging to calculator
Adds logg
Adds logging for the case table actions:
- Resize columns
- Edit cell value
- Insert cases 
- Sort column
- Fit column width
- Expand/Collapse cases
- Table/card toggle
Adds logging to a couple of graph actions:
- Background change
- Point color change
- Hide/Show parent toggle and measures for selection


v3/src/components/graph/components/inspector-panel/hide-show-menu-list.tsx
```diff
@@ -75,7 +75,9 @@ export const HideShowMenuList = observer(function HideShowMenuList({tile}: IProp
 
     dataConfig?.applyModelChange(
       () => graphModel?.setShowParentToggles(!graphModel?.showParentToggles),
-      { undoStringKey, redoStringKey }
+      { undoStringKey, redoStringKey,
+        log: graphModel?.showParentToggles ? "Enable NumberToggle" : "Disable NumberToggle"
```
kswenson, line 79: Logic appears to be reversed.

v3/src/components/graph/components/inspector-panel/hide-show-menu-list.tsx
```diff
@@ -85,7 +87,9 @@ export const HideShowMenuList = observer(function HideShowMenuList({tile}: IProp
       : ["DG.Undo.enableMeasuresForSelection", "DG.Redo.enableMeasuresForSelection"]
     dataConfig?.applyModelChange(
       () => graphModel?.setShowMeasuresForSelection(!graphModel?.showMeasuresForSelection),
-      { undoStringKey, redoStringKey }
+      { undoStringKey, redoStringKey,
+        log: graphModel?.showMeasuresForSelection ? "Enable MeasuresForSelection" : "Disable MeasuresForSelection"
```
kswenson, line 91: Logic appears to be reversed.

v3/src/lib/logger.ts
```diff
@@ -15,6 +15,7 @@ export interface LogMessage {
   application: string
   activity?: string
   event: string
+  event_value?: Record<string, unknown>
```
kswenson, line 18: It's not clear to me why we need another `Record<string, unknown>` property (`event_value`) when we already have one (`parameters`) that is essentially (or completely?) unused.

kswenson (changes requested):
A bunch of suggestions, but some broad categories:
- Old cruft -- the `uuid` dependency and the `logger-types.ts` file were removed in a previous PR. They should be removed and not return.
- Logging should be tied to model changes via `applyModelChange` wherever possible. Adding logging without tying it to `applyModelChange` just means we'll have to come back and fix it later. If the feature hasn't been implemented yet, then a solo log may make sense as an interim step, but otherwise they should be tied together.
- Logic is reversed for logging of some of the toggle values.
- I don't see the need for the `event_value` argument, when we already have `parameters` for catch-all properties. We can get together to discuss at some point if that would help.

v3/src/lib/logger.ts
eireland: In V2, it looks like `event_value` were variable values passed in to the log function, and `parameters` is `event_value` as an object. https://github.com/concord-consortium/codap/blob/cf79d90f7058fa9f913658de80d58bef04362a4f/apps/dg/controllers/authorization.js#L73.
And then the body has both: https://github.com/concord-consortium/codap/blob/cf79d90f7058fa9f913658de80d58bef04362a4f/apps/dg/controllers/authorization.js#L121

If we know nobody is using one over the other, we can remove event_value.

bfinzer (approved):
👍🏻LGTM
I've done a superficial review and everything looked good. I leave it to Kirk review more deeply.

kswenson (approved):
👍 Looks good -- I added a commit to tweak a few things. Take a look and merge if you agree.

PR #1386, opened 2024-08-01
188045173 DI Fix Collaborative Input Row

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/188045173

This PR addresses a bug that prevented new cases from being broadcast to other users when a new case was added to a shared table using the input row. The issue is that a notification was only being sent out when new case ids were encountered, but this wasn't being updated in some circumstances because `dataset.validateCases()` was not being called. I'm still not sure why this wasn't being called in the context of a Collaborative shared table, but was in most other cases.

The solution is to call `validateCases()` in `dataset.addCases()` after adding cases to the dataset. I'm not totally sure this is the right place to do it, since this function is called in many other places, but it seems to make sense to me that you'd want to update the cases after a new item is added. If we wanted to be precise about this call, it could be made in `handleRowsChange` in `use-rows.ts` after calling `data.addCases()`.

### DIFF
--- a/v3/src/components/case-table/use-rows.ts
+++ b/v3/src/components/case-table/use-rows.ts
@@ -302,6 +302,9 @@ export const useRows = () => {
             collectionTableModel.setInputRowIndex(collectionTableModel.inputRowIndex + 1)
           }
           data.addCases(casesToCreate, options)
+          // Make sure things are updated since adding cases
+          // TODO Would it be better to make collection.caseIds a getter that automatically validates the cases?
+          data.validateCases()
           // We look for case ids that weren't present before adding the new cases to determine which case ids
           // should be included in the createCasesNotification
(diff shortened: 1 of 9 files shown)


v3/src/components/case-table/use-rows.ts
```diff
@@ -302,6 +302,9 @@ export const useRows = () => {
             collectionTableModel.setInputRowIndex(collectionTableModel.inputRowIndex + 1)
           }
           data.addCases(casesToCreate, options)
+          // Make sure things are updated since adding cases
```
kswenson, line 305:
```suggestion
          // Make sure things are updated since adding cases invalidates grouping
```

kswenson (approved):
👍 Looks good

PR #1387, opened 2024-08-01
188050066 v3 Collaborative Premade Tables

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/188050066

This PR fixes a bug where sharing premade tables using the Collaborative plugin would result in tables losing all data. This was the result of the `cid` property of attributes not being implemented in v3. Collaborative relies on matching `cids`to connect attributes between different users, setting the `cid` of attributes with matching `names` when first connecting. V3 was not updating the `cid` of these attributes properly, so the attributes were later not being matched, and Collaborative was ruthlessly deleting them as a result.

The fix is to better support attribute `cids` in v3. The id is still used as the `cid` in most cases, unless:
- A plugin requests that a `cid` be set.
- A `cid` is set in a v2 document that is imported.


kswenson (approved):
👍 Looks good -- one suggestion that you might as well take care of as part of this PR: there are a couple of references to `cid` in `v2-model.ts` from back when I thought it might be used more broadly. I think those references can just be removed at this point so that the only references to `cid` are in the functional code in this PR.

PR #1388, opened 2024-08-05
fix: handling of fetch errors when loading plugin config

kswenson:
While testing v3 with a spotty/nonexistent internet connection recently it became apparent that we weren't handling errors that occurred when loading the default plugins configuration using `fetch` correctly. This came to our attention recently as a source of flaky cypress tests as well. My first approach was a fairly small change to the `PluginsButton` code to improve handling of `fetch` errors. This wasn't easy to write a jest test for, however, so I split out the fetching code to a new `useStandardPlugins()` custom hook and wrote a jest test for that. To test various possible `fetch` responses I installed `jest-fetch-mock` (also used by CLUE). Then for good measure, I wrote a jest test for the `PluginsButton` component as well.

### DIFF
--- a/v3/package-lock.json
+++ b/v3/package-lock.json
@@ -104,6 +104,7 @@
         "jest": "^29.7.0",
         "jest-canvas-mock": "^2.5.2",
         "jest-environment-jsdom": "^29.7.0",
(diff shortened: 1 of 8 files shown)


eireland (approved):
👍 LGTM

PR #1389, opened 2024-08-06
chore: simplify ComponentTitleBar

kswenson:
After the last round of `ComponentTitleBar` and `getTitle` refactoring, we ended up in a situation in which:
1. Every component provides a `CustomTitleBar` which passes a `getTitle` prop to `ComponentTitleBar`.
2. Every `CustomTitleBar` component passes the same global `getTitle` function as its `getTitle` prop.

The purpose of passing the `getTitle` function to the `ComponentTitleBar` originally was so that `CustomTitleBar` implementations could provide their own component-specific means of retrieving the title in a way that would allow the access of the title to occur in the `ComponentTitleBar` so that only it would re-render when the title changed. Once the `getTitle` customization moved from the component to the registration, there's no need for the `getTitle` prop at all. This allows components that don't require title bar customization, e.g. graph, map, and the upcoming text component (which is what got me here in the first place) to use the default `ComponentTitleBar` without customization.


eireland (approved):
👍 LGTM

PR #1390, opened 2024-08-06
chore: simplify ToolShelf by moving undo/redo strings into component registration

kswenson:
Moving the undo/redo strings into the component registration allows us to eliminate component-specific code in the `ToolShelf` component.

Note: This PR is rooted on #1389, which should be merged first.


eireland (approved):
👍 LGTM

PR #1391, opened 2024-08-06
188045959 fix sampler output translation key

eireland:
This updates the existing corresponding Sampler translation for device name to "output"


bfinzer (approved):
👍🏻LGTM

PR #1392, opened 2024-08-07
feat: text component

kswenson:
Adds basic text component functionality:
- basic editing works
- toolbar works
- serialization works
- import from v2 works
- basic jest and cypress tests
- adds a text component to the `dashboard` to facilitate testing

Still to do:
- increase/decrease font size
- publish version of slate-editor library updated to React 18
- update CODAP to the updated slate-editor library

Some independent changes in service of this PR were submitted separately as #1389, #1390. This PR is rooted on those and should not be merged until those have been merged.

Along the way, several utilities were developed that may have use elsewhere. Some of them were developed for use in positioning the toolbar alongside the component but I ended up solving that problem another way, so in some cases these utilities are not used in this PR but I'm checking them in anyway on the grounds that they seem likely to be useful at some point.
- `useCodapComponentContext` -- provides access to the `.codap-component` via React context. There are a number of places in the code where we use DOM mechanisms to find the `.codap-component`. This custom hook provides a ref via context which should be preferable.
- `useComponentWrapperContext` -- provides access to the `.free-tile-component` (parent of the `.codap-component`) using the same mechanism.
- `useDocumentContainerContext` -- provides access to the `.codap-container` using the same mechanism.
- `useMutationObserver` -- custom hook to provide access to DOM mutation observer functionality.


tealefristoe (approved):
Looks good! 👍

PR #1394, opened 2024-08-11
feat: delete formula retaining values

kswenson:
[[PT-181889964]](https://www.pivotaltracker.com/story/show/181889964)

@pjanik actually did the heavy lifting here, having implemented the desired behavior of preserving values on removing a formula in the engine. This PR just enables it in the UI. While in the code for handling the column header menu items I made a few other changes:
- Refactored to keep code specific to a given menu item co-located.
- Unimplemented menu items are now disabled with a 🚧 (under construction) icon appended rather than the previous behavior of being enabled and then putting up a "toast" when selected. The "toast" behavior made sense when the menu was first implemented but has outlived its usefulness.


bfinzer (approved):
👍🏻LGTM

PR #1395, opened 2024-08-11
fix: recalculate formulas on initial load of document

kswenson:
[[PT-188095695]](https://www.pivotaltracker.com/story/show/188095695)

The fix is to initialize the formula adapters later in the document creation process.


bfinzer (approved):
👍🏻LGTM

PR #1396, opened 2024-08-12
187178550 v3 Improve Item Sort

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187178550

This PR improves the efficiency of the `notify itemSearch itemOrder` API request by adding a new `dataset.moveItems` action that rearranges the order of items without removing them and then re-adding them.

As a step towards this solution, I also improved the efficiency of `dataset.removeCases`, changing it from O(n^2) to O(n*log(n)). Even though the API request no longer uses this action, it still seemed worth keeping the improvement for other cases where it's used.

I tested all three variants with tables shared between three users with approximately 5000 items. Here are the results:
- Original (less efficient `removeCases`) - 1200-1500 ms
- More efficient `removeCases` - 850-1000 ms
- `moveItems` - 650-800 ms

### DIFF
--- a/v3/src/models/data/data-set-types.ts
+++ b/v3/src/models/data/data-set-types.ts
@@ -42,6 +42,13 @@ export interface IAddCasesOptions {
   canonicalize?: boolean;
 }
 
+export interface IMoveItemsOptions {
+  // id of item before/after which to move items
+  // if not specified, items are moved to end
+  before?: string;
+  after?: string;
+}
+
 export interface IAddAttributeOptions {
   // id of attribute before which to insert new cases
   // if not specified, new attribute is appended
--- a/v3/src/models/data/data-set.ts
+++ b/v3/src/models/data/data-set.ts
@@ -919,8 +919,12 @@ export const DataSet = V2Model.named("DataSet").props({
       },
 
       removeCases(caseIDs: string[]) {
-        caseIDs.forEach((caseID) => {
-          const index = self.getItemIndex(caseID)
+        // Remove the items last -> first, so we only have to update itemInfo once
+        const items = caseIDs.map(id => ({ id, index: self.getItemIndex(id) }))
+          .filter(info => info.index != null) as { id: string, index: number }[]
+        items.sort((a, b) => b.index - a.index)
+        const firstIndex = items[items.length - 1].index
+        items.forEach(({ id: caseID, index }) => {
           if (index != null) {
             self.itemIds.splice(index, 1)
             self.attributes.forEach((attr) => {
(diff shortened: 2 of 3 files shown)


v3/src/models/data/data-set.ts
```diff
@@ -919,22 +919,74 @@ export const DataSet = V2Model.named("DataSet").props({
...
       removeCases(caseIDs: string[]) {
-        caseIDs.forEach((caseID) => {
-          const index = self.getItemIndex(caseID)
+        // Remove the items last -> first, so we only have to update itemInfo once
+        const items = caseIDs.map(id => ({ id, index: self.getItemIndex(id) }))
+          .filter(info => info.index != null) as { id: string, index: number }[]
+        items.sort((a, b) => b.index - a.index)
+        const firstIndex = items[items.length - 1]?.index ?? -1
+        items.forEach(({ id: caseID, index }) => {
           if (index != null) {
```
kswenson, line 928: Having filtered out nullish `index` values above, I think you can remove the `if` here.

v3/src/models/data/data-set.ts
```diff
@@ -919,22 +919,74 @@ export const DataSet = V2Model.named("DataSet").props({
...
+          } else {
+            self.itemIds.push(item.__id__)
+            self.attributes.forEach((attr, index) => {
+              attr.strValues.push(values[index].strValue)
+              attr.numValues.push(values[index].numValue)
+            })
+          }
+        })
+
+        // Fix indecies
```
kswenson, line 984:
```suggestion
        // Fix indices
```

v3/src/models/data/data-set.ts
```diff
@@ -919,22 +919,74 @@ export const DataSet = V2Model.named("DataSet").props({
...
+          for (let i = firstIndex; i < self.itemIds.length; ++i) {
+            const itemId = self.itemIds[i]
+            const itemInfo = self.itemInfoMap.get(itemId)
+            if (itemInfo) itemInfo.index = i
+          }
+        }
+      },
+
+      // If order is "first", items will be moved to the front. Otherwise, they are moved to the end.
+      moveItems(itemIds: string[], order?: string) {
```
kswenson, line 947: It would be more useful as well as more consistent with the existing API for this to take a `beforeId` for the destination, and to append if no `beforeId` is provided. I don't think it complicates the implementation much -- remove the items being moved, check the index of `beforeId` in the remaining cases, and append if it's not found.

v3/src/models/data/data-set.ts
```diff
@@ -919,22 +919,74 @@ export const DataSet = V2Model.named("DataSet").props({
...
+              attr.numValues.unshift(values[index].numValue)
+            })
+          } else {
+            self.itemIds.push(item.__id__)
+            self.attributes.forEach((attr, index) => {
+              attr.strValues.push(values[index].strValue)
+              attr.numValues.push(values[index].numValue)
+            })
+          }
+        })
```
kswenson, line 982: Seems like you should be able to use `splice()` to insert all of the cases at once rather than having to loop through and add them one at a time. If you make the API change I suggested above, then there would be a `beforeIndex` determined from the `beforeId` and then each array can be updated with `array.splice(beforeIndex, 0, ...movedValues)`.

kswenson (changes requested):
👍 Generally looks good -- see inline suggestions.

>As a step towards this solution, I also improved the efficiency of `dataset.removeCases`, changing it from O(n^2) to O(n*log(n)). Even though the API request no longer uses this action, it still seemed worth keeping the improvement for other cases where it's used.

In my defense, the removal of the cases was/is O(n) -- it was only updating the indices that was O(n<sup>2</sup>), but even so eliminating it is worthwhile. 😉

v3/src/models/data/data-set-types.ts
```diff
@@ -42,6 +42,13 @@ export interface IAddCasesOptions {
   canonicalize?: boolean;
 }
 
+export interface IMoveItemsOptions {
+  // id of item before/after which to move items
+  // if not specified, items are moved to end
+  before?: string;
+  after?: string;
```
kswenson, line 49: It would have been fine to just support `before`, but nice to have `after` as well.

kswenson (approved):
👍 Looks good!

PR #1397, opened 2024-08-12
187178550 v3 DI Delete Improvements

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187178550

This PR makes several improvements related to deleting items.
- Deleting items now broadcasts a notification like it does in v2. The lack of notifications was causing a bug in the Collaborative plugin.
  - In order to include the deleted cases in the notification, a new volatile property has been added, `DataSet.itemIdChildCaseMap`. This links each item to the child case that contains it, and is computed in `validateCases`.
- It's now possible to batch items in a `delete item` API request. This is much more efficient than sending a separate API request for each item to be deleted. An API request that deletes multiple items looks like this:
```
{
  "action": "delete",
  "resource": "dataContext[Mammals].item",
  "values": [
    { "id": 1 },
    { "id": 2 },
    ...
  ]
}
```
- The delete case(s) option is now disabled in the case menu (comes up when you click on the index cell) when the plugin prevents the case from being edited. Now this menu item is consistent with the delete all/selected/unselected cases menu items in the trash menu, and protects other users' data in shared tables.
  - I also disabled the insert case(s) options here at the request of Dan.

### DIFF
--- a/v3/cypress/e2e/plugin.spec.ts
+++ b/v3/cypress/e2e/plugin.spec.ts
@@ -1,4 +1,3 @@
-/// <reference types="cypress" />
 import { CfmElements as cfm } from "../support/elements/cfm"
 import { ComponentElements as c } from "../support/elements/component-elements"
 import { SliderTileElements as slider } from "../support/elements/slider-tile"
--- a/v3/cypress/e2e/plugin.spec.ts
+++ b/v3/cypress/e2e/plugin.spec.ts
@@ -1,3 +1,4 @@
+/// <reference types="cypress" />
 import { CfmElements as cfm } from "../support/elements/cfm"
 import { ComponentElements as c } from "../support/elements/component-elements"
 import { SliderTileElements as slider } from "../support/elements/slider-tile"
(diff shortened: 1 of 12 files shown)


v3/cypress/e2e/plugin.spec.ts
```diff
@@ -1,3 +1,4 @@
+/// <reference types="cypress" />
```
kswenson, line 1: I found myself annoyed by this as well and submitted a PR to fix it, so you shouldn't need this line any more.

v3/src/data-interactive/handlers/interactive-frame-handler.ts
```diff
@@ -20,6 +20,7 @@ export const diInteractiveFrameHandler: DIHandler = {
     } = webViewContent ?? {}
     const values: DIInteractiveFrame = {
       allowEmptyAttributeDeletion,
+      codapVersion: "v3", // TODO Correct version
```
kswenson, line 23: See `handle-cfm-event.ts` for a means of retrieving the version number. Seems like there should be a means of retrieving it once and caching it.

kswenson (approved):
👍 Looks good -- see inline comment about retrieving the version number.

PR #1398, opened 2024-08-12
Basic date-time axis

bfinzer:
[#181971361] Feature: Date attributes can be plotted on a date-time axis

This is a "first pass" implementation of a date-time axis. Missing features and bugs should be logged in separate PT stories unless they break non-date-time axes or otherwise disrupt normal V3 behavior.

* Detect "date" as a valid attribute type when setting primary type in setPrimaryRoleAndPlotType
* Define `AbstractNumericAxisModel` with descendant `NumericAxisModel`
* Add DateAxisModel as a descendant of `AbstractNumericAxisModel`
* Allow for creation of DatAxisModel in `setupAxes`
* Defined AxisHelper and subclasses for empty, numeric and categorical. Create these in useSubAxis.
* DateAxisHelper is responsible for rendering the data axis scale
* Both NumericAxisModel and DateAxisModel return true for the new test `isAbstractNumeric`
* Added data-display-value-utils.ts with dataDisplayGetNumericValue as its only member (so far) to lessen likelihood of Lakosian violations
* To get drag rescale working, define `isAbstractNumericAxisModel` and use it, where appropriate, in place of `isNumericAxisModel`


v3/src/components/axis/helper-models/axis-helper.ts
```diff
@@ -0,0 +1,76 @@
...
+  multiScale: MultiScale | undefined
+
+  constructor(
+    protected displayModel: IDataDisplayContentModel,
+    protected subAxisIndex: number,
+    protected axisPlace: AxisPlace,
+    protected subAxisElt: SVGGElement | null,
+    protected axisModel: IAxisModel,
+    protected layout: IAxisLayout,
+    protected isAnimating: () => boolean
```
kswenson, line 19: Given how complicated the derived class constructors become, I'm inclined to think that the constructor shorthand syntax is not worth it in this case. I'd recommend just going back to taking an args object like:
```typescript
export interface IAxisHelperArgs {
  displayModel: IDataDisplayContentModel
  subAxisIndex: number
  subAxisElt: SVGGElement | null
  axisModel: IAxisModel
  layout: IAxisLayout
  isAnimating: () => boolean
}
```
and then the derived classes simplify to something like:
```typescript
interface IDateAxisHelperArgs extends IAxisHelperArgs {
  subAxisSelectionRef: MutableRefObject<Selection<SVGGElement, any, any, any> | undefined>
}

export class DateAxisHelper extends AxisHelper {
  subAxisSelectionRef: MutableRefObject<Selection<SVGGElement, any, any, any> | undefined>
  maxNumberExtent: number = kDefaultFontHeight

  constructor(props: IDateAxisHelperArgs) {
    super(props)
    this.subAxisSelectionRef = props.subAxisSelectionRef
  }
```
It's true that the redundancy that led us to consider the constructor shorthand syntax is still present, but in this case I think the redundancy is preferable to the alternative.

v3/src/components/axis/helper-models/axis-helper.ts
```diff
@@ -0,0 +1,76 @@
...
+      .style("stroke", "darkgrey")
+      .style("stroke-opacity", "0.7")
+  }
+
+}
+
+export class EmptyAxisHelper extends AxisHelper {
+  constructor(...args: ConstructorParameters<typeof AxisHelper>) {
+    super(...args)
+  }
```
kswenson, line 71: I think you can just eliminate the constructor in this case.
```suggestion
```

v3/src/components/axis/helper-models/date-axis-helper.ts
```diff
@@ -0,0 +1,447 @@
...
+  constructor(...args: [...ConstructorParameters<typeof AxisHelper>, IDateAxisHelperProps]) {
+    const dateAxisProps = args[args.length - 1] as IDateAxisHelperProps,
+      axisHelperProps = args.slice(0, args.length - 1) as ConstructorParameters<typeof AxisHelper>
+    super(...axisHelperProps)
+    this.subAxisSelectionRef = dateAxisProps.subAxisSelectionRef
+  }
+
+  get dataConfig() {
+    return this.displayModel.dataConfiguration
+  }
```
kswenson, line 196: This can move to the base class.

v3/src/components/axis/helper-models/date-axis-helper.ts
```diff
@@ -0,0 +1,447 @@
...
+import {convertToDate, createDate, determineLevels, EDateTimeLevel, shortMonthNames}
+          from "../../../utilities/date-utils"
+import { measureTextExtent } from "../../../hooks/use-measure-text"
+
+type ILabelDateAndString = {
+  labelDate: Date
+  labelString: string
+}
+
+const getLevelLabelForValue = (level: EDateTimeLevel, date: Date | null): ILabelDateAndString => {
```
kswenson, line 15: We haven't generally used `null` values in CODAP, but I see I approved it in Piotr's date utilities code so let's not worry about it for the purpose of this PR.

v3/src/components/axis/helper-models/date-axis-helper.ts
```diff
@@ -0,0 +1,447 @@
...
+      lowerBoundsDate = createDate(lowerBoundsSeconds) as Date,
+      upperBoundsMS = 1000 * upperBoundsSeconds,
+      upperBoundsDate = createDate(upperBoundsSeconds) as Date,
+      sAS = this.subAxisSelectionRef.current,
+      {rangeMin, rangeMax} = this
+
+    sAS.selectAll('*').remove()
+
+    this.renderAxisLine()
+    drawTicks()
```
kswenson, line 445: Can date axes show gridlines? Seems like it could be desirable.

v3/src/components/axis/helper-models/categorical-axis-helper.ts
```diff
@@ -0,0 +1,128 @@
...
+    this.subAxisSelectionRef = categoricalAxisProps.subAxisSelectionRef
+    this.categoriesSelectionRef = categoricalAxisProps.categoriesSelectionRef
+    this.swapInProgress = categoricalAxisProps.swapInProgress
+    this.centerCategoryLabels = categoricalAxisProps.centerCategoryLabels
+    this.dragInfo = categoricalAxisProps.dragInfo
+  }
+
+  get dataConfig() {
+    return this.displayModel.dataConfiguration
+  }
```
kswenson, line 47: Move to base class.

v3/src/components/axis/helper-models/axis-helper.ts
```diff
@@ -0,0 +1,76 @@
...
+  renderAxisLine() {
+    select(this.subAxisElt).selectAll('*').remove()
+    select(this.subAxisElt)
+      .attr("transform", this.initialTransform)
+      .append('line')
+      .attr('x1', 0)
+      .attr('x2', this.isVertical ? 0 : this.subAxisLength)
+      .attr('y1', 0)
+      .attr('y2', this.isVertical ? this.subAxisLength : 0)
+      .style("stroke", "darkgrey")
```
kswenson, line 62: It seems like the empty axis used to use `lightgrey` and now it'll be `darkgrey`. Two questions:
1. British spelling? 😉 
2. Was this change intentional?

kswenson (changes requested):
👍 Looks good -- a bunch of suggestions, hopefully not too onerous.

kswenson:
The other thing I forgot to mention in my review, is that we're going to want some tests. If we don't want to hold up this PR, then we should plan to come back to them. There are a number of utility functions that should be straightforward to test with jest tests and then we should have a cypress test that brings up a date axis.

v3/src/components/axis/hooks/use-sub-axis.ts
```diff
@@ -58,6 +54,38 @@ export const useSubAxis = ({
...
+          break
+        case 'date':
+          subAxisSelectionRef.current = subAxisElt ? select(subAxisElt) : undefined
+          helper = new DateAxisHelper(
+            displayModel, subAxisIndex, axisPlace, subAxisElt,
+            axisModel, layout, isAnimating, {subAxisSelectionRef})
+      }
+      return helper
+    }, [axisModel, axisPlace, displayModel, layout, subAxisElt, subAxisIndex,
+      showScatterPlotGridLines, isAnimating, centerCategoryLabels]),
```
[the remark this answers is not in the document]
bfinzer: Actually none of the properties will change, including the two booleans.

v3/src/components/axis/models/axis-model.ts
```diff
@@ -9,7 +9,7 @@ export const AxisModel = types.model("AxisModel", {
   place: types.enumeration([...AxisPlaces]),
   scale: types.optional(types.enumeration([...ScaleTypes]), "ordinal"),
 })
-  .volatile(self => ({
+  .volatile(_self => ({
```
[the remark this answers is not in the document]
bfinzer: If you're wondering about the underscore, it prevents eslint from complaining about an unused parameter

v3/src/components/case-table/case-table.tsx
```diff
@@ -34,7 +34,7 @@ export const CaseTable = observer(function CaseTable({ setNodeRef }: IProps) {
   useEffect(() => {
     // disable vertical auto-scroll of table (column headers can't scroll out of view)
     return registerCanAutoScrollCallback((element, direction) => {
-      return element !== contentRef.current || direction.y === 0
+      return element !== contentRef.current || (direction && direction.y === 0)
```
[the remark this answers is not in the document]
bfinzer: I did encounter that. Sorry, but I don't recall the circumstances.

v3/src/components/axis/models/axis-model.ts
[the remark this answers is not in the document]
kswenson: That's not a project-level eslint setting, i.e. no one else sees it. Perhaps an IntelliJ-specific one? In any case, I'm hesitant to endorse individual editor-specific changes.

v3/src/components/graph/models/graph-model-utils.ts
```diff
@@ -31,9 +32,12 @@ function setPrimaryRoleAndPlotType(graphModel: IGraphContentModel) {
       : attributeType !== 'empty' ? oldPrimaryRole : otherAttrRole
   dataConfig?.setPrimaryRole(primaryRole)
   // TODO COLOR: treat color like categorical for now
-  const primaryType = attributeType === 'color' ? 'categorical' : attributeType
+  const primaryType = attributeType === 'color' ? 'categorical'
+    : attributeType === 'date' ? 'numeric' : attributeType,
+    secondaryType = otherAttributeType === 'color' ? 'categorical'
+      : otherAttributeType === 'date' ? 'numeric' : otherAttributeType
```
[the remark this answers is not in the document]
bfinzer: Certainly easier to interpret! And I am now reminded that such a thing is possible.

v3/src/components/graph/models/graph-model-utils.ts
```diff
@@ -74,6 +78,17 @@ function setupAxes(graphModel: IGraphContentModel, layout: GraphLayout) {
...
         break
+      case 'date': {
+        if (!currAxisModel || !isDateAxisModel(currAxisModel)) {
+          const newAxisModel = DateAxisModel.create({place, min: 0, max: 1})
+          graphModel?.setAxis(place, newAxisModel)
+          dataConfig?.setAttributeType(attrRole, 'date')
+          layout.setAxisScaleType(place, 'linear')
+          const valuesInSeconds = stringValuesToDateSeconds(attr?.strValues || [])
+          setNiceDomain(valuesInSeconds, newAxisModel, graphModel?.axisDomainOptions)
+        }
```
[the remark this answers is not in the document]
bfinzer: Yes, we do need to worry about it. Good catch!

v3/src/components/axis/models/axis-model.ts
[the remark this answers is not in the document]
bfinzer: Interesting. OK, I'll forgo the underscore.

v3/src/components/axis/helper-models/axis-helper.ts
```diff
@@ -0,0 +1,92 @@
...
+
+export class AxisHelper {
+  displayModel: IDataDisplayContentModel
+  subAxisIndex: number
+  subAxisElt: SVGGElement | null
+  axisModel: IAxisModel
+  layout: IAxisLayout
+  isAnimating: () => boolean
+  multiScale: MultiScale | undefined
+  axisPlace: AxisPlace
```
kswenson, line 26: Instead of a property, you could also just make this an accessor function which forwards to the model.

v3/src/components/axis/models/axis-model.ts
```diff
@@ -23,6 +23,12 @@ export const AxisModel = types.model("AxisModel", {
     get isCategorical() {
       return self.type === "categorical"
     },
+    get isDate() {
+      return self.type === "date"
+    },
+    get isAbstractNumeric() {
```
kswenson, line 29:
```suggestion
    get isBaseNumeric() {
```

kswenson (approved):
👍 Looks good -- just a couple of nits.

kswenson:
@bfinzer I see that there's a failing jest test which should be addressed as well.

v3/src/components/axis/helper-models/axis-helper.ts
bfinzer: Done

PR #1400, opened 2024-08-13
182089941 add log strings graph

eireland:
Adds logging for graph events including:
- axes changes
- toggling adornments
- legend changes
- hide/show points
- point color and size changes


kswenson (approved):
👍 Good start -- your examples convinced me that we need some utility functions that establish patterns to be followed so that the code doesn't seem so ad hoc at every call site and so that the results are more consistent, so I added a commit which adds several utility functions and demonstrates their use. Please make use of these functions moving forward and add new ones as the need arises. A couple of general principles:
- The translation function (`t`) can be used for generic parameter replacement.
- The utility functions generally eliminate the need to pass the same information multiple times, e.g. as replacement strings and as values in `args`.
- There's an important distinction between passing a structure (or string) as the `log` property to `applyModelChange` and passing a function: the arguments are evaluated _before_ `applyModelChange` makes the change, but the functional form is called _after_, which can be useful for logging the result of making a change.
- Function closures can be used to store temporary values instead of `useRef`s. See the `logModelChangeFn` utility function for an example.
- As mentioned previously, logging should really be associated with the actual model change. Some of the places where logging is added here, notably in cases where dragging is involved, the model change occurs in the `drag` function but the logging occurs in the `endDrag` function. This code will have to be revisited so that it doesn't generate excessive undoable events and so the final model change occurs in the `endDrag`, but that's outside the scope of this PR.

PR #1401, opened 2024-08-13
chore: fix cypress TypeScript types

kswenson:
Since v3 development began, the TypeScript configuration for our cypress tests has been mis-configured, resulting in a blizzard of red squiggles (compile errors) in every cypress test file. I (finally) reached the point of being sufficiently annoyed by this to do something about it. Running the TypeScript compiler on the cypress sources before this PR resulted in 1,917 errors. This PR reduces that number to zero by:
- Making cypress types available by adding `"types": ["cypress"]` to the `cypress/tsconfig.json` file.
- Adding a `build:cypress` command to `package.json` which "compiles" (i.e. type-checks) the cypress source files.
- Adding `build:cypress` to the set of tasks triggered by `npm run build`.
  - This means that errors in the cypress test files will cause the build to fail.
- Adding ambient type declarations for our cypress extensions, e.g. the commands defined in `commands.ts`.
- Fixing the remaining blizzard of type errors from untyped function arguments and local variables, etc.
- Reducing the `wait` in the `dragAttributeToTarget` command from 2500 to 1000.
- Fixing a type error in `axis-spec.ts` that resulted in code for checking axis ticks not running.
- Eliminating dozens of `wait`s from the `axis-spec.ts` test.
- Fixing a smattering of other small issues that I encountered along the way.


v3/cypress/e2e/axis.spec.ts
```diff
@@ -15,18 +15,16 @@ const arrayOfValues = [
   { attribute: "Mass", values: [ "0", "1000", "2000", "3000", "4000", "5000", "6000", "7000" ]},  // Y
   { attribute: "Sleep", values: [...Array(12).keys()].map(i => `${2 * i}`)},  // Y
   { attribute: "Speed", values: [ ]},
-  { attribute: "Habitat", values: [ "both", "land", "water" ]},
-  { attribute: "Diet", values: [ "both", "meat", "plants" ]},
+  { attribute: "Habitat", values: [ "land", "water", "both" ]},
+  { attribute: "Diet", values: [ "plants", "meat", "both" ]},
```
kswenson, line 19: We apparently didn't notice these obvious errors because the code for testing these values wasn't running. See comment in `axis-helper.ts` for details.

v3/cypress/support/helpers/axis-helper.ts
```diff
@@ -63,38 +63,39 @@ export const AxisHelper = {
       expect($length).to.be.greaterThan(0)
     })
   },
-  verifyAxisTickLabels(axis, attributeValues, categorical = false) {
+  verifyAxisTickLabels(axis: string, attributeValues: string[], categorical = false) {
     ae.getAxisTickLabels(axis, categorical).should('have.length', attributeValues.length)
-    for (let index = 0; index < attributeValues; index++) {
+    for (let index = 0; index < attributeValues.length; index++) {
```
kswenson, line 68: Apparently, this comparison always failed, so the loop never ran.

tealefristoe (approved):
Looks great!

Not sure which I'm more excited about, having cypress keywords recognized, or removing a minute or two of waits 👍

PR #1402, opened 2024-08-14
chore: move lint cache files into `node_modules/.cache` folder

kswenson:
Recently, @nstclair-cc encountered an issue whereby `lint:build` failed with a module resolution error but all other commands (`lint`, `build:webpack`, `test`) succeeded. Seemed likely to be a caching issue but `clean:cache` didn't fix it. 🤔 Only later did I realize that the lint caches aren't cleared by `clean:cache` because they aren't stored in the `node_modules/.cache` folder with the other caches (e.g. webpack's). This PR fixes that so that the lint caches are in `node_modules/.cache` along with our other development-related caches, which means that the `clean:cache` command clears the lint caches as well.

Note that I've removed the old files from the `.gitignore` file, so that developers with existing repositories may encounter unversioned files named `.eslintcache` and `.eslintcache.build` in their workspace. These files can safely be deleted.


tealefristoe (approved):
👍

PR #1403, opened 2024-08-14
188104017 v3 Input Row Tab

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/188104017

In a shared table, pressing the tab key in the input row would result in two cases being added to the table. This didn't happen in normal tables. This PR fixes that issue by removing `onBlur` handlers from `CellTextEditor` and `ColorCellTextEditor`, which seem to be unnecessary.


kswenson (approved):
👍 Gotta love fixing bugs by removing code!

PR #1405, opened 2024-08-14
Graph responds to change in attribute's collection

bfinzer:
[#187798725] Bug fix: Graph turns into scatterplot when creating new collections in Mammals dataset

This fixes the graph bugs related to move an attribute within a dataset. But a bug occurs when the attribute is moved to a parent collection and then the user presses Undo. See #188117637.

* Remove `attribute.completeSnapshot` from `attribute-menu-list.ts handleClick` so that MST won't complain.
* `RemoveAttributeAction` in `data-set-actions` removed because no longer called.
* Changed the `onAnyAction` in `handleDeleteAttribute` to an mstReaction in what is now `handleAttributeCollectionMapChange`. This detects both deletion and moving of attributes to different collections.

### DIFF
--- a/v3/src/components/graph/components/graph.tsx
+++ b/v3/src/components/graph/components/graph.tsx
@@ -1,6 +1,6 @@
 import {comparer} from "mobx"
 import {observer} from "mobx-react-lite"
-import {addDisposer, isAlive} from "mobx-state-tree"
+import {isAlive} from "mobx-state-tree"
 import React, {MutableRefObject, useCallback, useEffect, useMemo, useRef} from "react"
 import {select} from "d3"
 import {clsx} from "clsx"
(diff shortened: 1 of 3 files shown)


v3/src/components/graph/components/graph.tsx
```diff
@@ -122,20 +122,46 @@ export const Graph = observer(function Graph({graphController, graphRef, pixiPoi
...
-          } else {
-            graphModel.setAttributeID(role as GraphAttrRole, "", "")
+  useEffect(function handleAttributeCollectionMapChange() {
+
+    const constructAttrCollections = () => {
+      const graphAttrs = graphModel.dataConfiguration.attributes
+      const attrCollections: Record<string, string> = {}
+      graphAttrs.forEach(attrId => {
+        const collection = dataset?.getCollectionForAttribute(attrId)?.id
+        collection && (attrCollections[attrId] = collection)
```
kswenson, line 132:
```suggestion
        const collectionId = dataset?.getCollectionForAttribute(attrId)?.id
        collectionId && (attrCollections[attrId] = collectionId)
```
for naming consistency.

kswenson (approved):
👍 LGTM!

PR #1406, opened 2024-08-15
188104291 v3 DI Return Table to Normal after Sharing

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/188104291

Previously, a shared table was not disconnecting from its managing controller after sharing stopped. This PR fixes that bug.

### DIFF
--- a/v3/src/data-interactive/handlers/data-context-handler.ts
+++ b/v3/src/data-interactive/handlers/data-context-handler.ts
@@ -88,11 +88,7 @@ export const diDataContextHandler: DIHandler = {
 
         if (managingController) {
           const tile = findTileFromNameOrId(managingController)
-          if (tile) {
-            dataContext.setManagingControllerId(tile.id)
-          } else {
-            dataContext.setManagingControllerId("")
-          }
+          dataContext.setManagingControllerId(tile?.id)
         }
 
         if (sort?.attr) {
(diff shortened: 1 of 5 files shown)


v3/src/data-interactive/handlers/data-context-handler.ts
```diff
@@ -88,7 +88,11 @@ export const diDataContextHandler: DIHandler = {
 
         if (managingController) {
           const tile = findTileFromNameOrId(managingController)
-          if (tile) dataContext.setManagingControllerId(tile.id)
+          if (tile) {
+            dataContext.setManagingControllerId(tile.id)
+          } else {
+            dataContext.setManagingControllerId("")
+          }
```
kswenson, line 95: Could also do:
```suggestion
          dataContext.setManagingControllerId(tile?.id ?? "")
```
or just
```typescript
          dataContext.setManagingControllerId(tile?.id)
```
if `undefined` were interpreted as the empty string.

kswenson (approved):
👍 LGTM

v3/src/data-interactive/handlers/data-context-handler.ts
tealefristoe: Much cleaner, thanks for the suggestion 👍

PR #1407, opened 2024-08-15
182089941 add log strings maps

eireland:
Add logging to map actions:
- Change grid size
- Marquee select toggle
- Toggle map base
- Toggle show/hide data set 
- Map base selection
- Change point size, stroke, and color


kswenson (changes requested):
I added a commit which:
1. Removed `LatLng` references from common code. Tile-specific types should not be imported into common code like `log-message.ts` and `translate.ts`.
2. Added a `logStringifiedMapMessage()` function in the map code to handle `LatLng`.
3. Fixed the handling of the prev value in `map-grid-slider.tsx`.
4. Fixed the logging of pan and zoom mediated by `LeafletMapState`.

I also left some suggestions for things like adding `%@` -- `logStringifiedObjectMessage()` doesn't log the object if there isn't a `%@` in the message.

kswenson (approved):
👍 LGTM

PR #1408, opened 2024-08-15
188118418 v3 DI Update Component webView

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/188118418

This PR adds support for generic tile and webView features when handling `update component` API requests.


v3/src/data-interactive/handlers/component-handler-web-view.test.ts
```diff
@@ -1,19 +1,20 @@
+import { kTitleBarHeight } from "../../components/constants"
 import { kV2GameType } from "../../components/web-view/web-view-defs"
 import { IWebViewModel, isWebViewModel } from "../../components/web-view/web-view-model"
-import { kWebViewIdPrefix } from "../../components/web-view/web-view-registration"
+import { kDefaultWebViewHeight, kDefaultWebViewWidth, kWebViewIdPrefix } from "../../components/web-view/web-view-registration"
```
kswenson, line 4: Line length warning

v3/src/data-interactive/handlers/component-handler-web-view.test.ts
kswenson, line 1: I didn't do this when I did my refactor a while back, but since most of the code being tested here is in `web-view`, seems like this and other analogous test files should really move into the corresponding component folder.

v3/src/data-interactive/handlers/component-handler.ts
```diff
@@ -150,14 +150,30 @@ export const diComponentHandler: DIHandler = {
...
-      return handler.update?.(content, values) ?? errorResult(t("V3.DI.Error.notFound"))
-    }
+    let result: DIHandlerFnResult | undefined
+    component.applyModelChange(() => {
+      // Handle updating generic component features
+      const { cannotClose, dimensions, position, title } = values as V2Component
+      if (cannotClose != null) component.setCannotClose(cannotClose)
+      if (title) component.setTitle(title)
+      // TODO Handle string positions?
+      const _position = position && typeof position !== "string" ? position : undefined
```
kswenson, line 160: How about:
```suggestion
      const _position = typeof position === "string" ? +position : position
```
which will attempt to convert the string to a number, if possible.

v3/src/models/document/free-tile-row.ts
```diff
@@ -158,6 +158,14 @@ export const FreeTileRow = TileRowModel
     setTileDimensions(tileId: string, dimensions: { width?: number, height?: number }) {
       const freeTileLayout = self.getNode(tileId)
       freeTileLayout?.setSize(dimensions.width, dimensions.height)
+    },
+    setTilePosition(tileId: string, position: { x?: number, y?: number }) {
+      const freeTileLayout = self.getNode(tileId)
+      if (freeTileLayout) {
```
kswenson, line 164:
```suggestion
      if (freeTileLayout && (x != null || y != null)) {
```
so we skip the whole thing if they're both nullish.

v3/src/models/document/free-tile-row.ts
```diff
@@ -158,6 +158,14 @@ export const FreeTileRow = TileRowModel
...
+        const x = position.x ?? freeTileLayout.x
+        const y = position.y ?? freeTileLayout.y
```
kswenson, line 166: FYI, I think you could also do:
```suggestion
        const { x = freeTileLayout.x, y = freeTileLayout.y } = position
```

kswenson (approved):
👍 Looks good -- just a few suggestions to consider, including moving the test file.

v3/src/data-interactive/handlers/component-handler.ts
tealefristoe: The strings I was thinking of here are "top" and "bottom", which are legal options for `position` when creating components (though not yet implemented in v3). However, it seems like v2 ignores these `position` options for `update component`, so I think it should be alright to just ignore all strings here.

PR #1409, opened 2024-08-16
188125035 Fix selectCasesNotifications

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/188125035

There was confusion between items and cases in the `selectCasesNotification`, which was resulting in notifications with almost all undefined values being included. This PR fixes the issue by finding the actual cases selected/deselected, rather than the items. Unfortunately, this is O(items * collections), rather than O(items) as before.


kswenson (approved):
👍 Looks good -- if it becomes a problem we can cache selected cases information in the `DataSet` at some point, but this seems good enough for now.

bacalj:
Thanks, this really improves things for the day-length plugin!

PR #1411, opened 2024-08-17
Scatterplot grid lines with date axis

bfinzer:
[#188109499] Bug fix: Date axis grid lines missing for scatterplot

* We pass showScatterPlotGridLines into DateAxisHelper just as we do for NumericAxisHelper so we can know when to draw grid lines.
* We refactor a bit to get access to what we need to be able to routines we need to visit each location for a grid line.


kswenson (approved):
👍 LGTM

PR #1412, opened 2024-08-18
fix: MST warning on date axis

kswenson:
[[PT-188130687]](https://www.pivotaltracker.com/story/show/188130687)

The `useSubAxis` hook makes use of an `AxisHelper` instance, which was being created/stored in a local variable using `useMemo()`. When transitioning from an empty axis to a date axis, a MobX reaction within `useSubAxis` was being triggered before the `useSubAxis` hook was re-rendered to update the axis helper instance, which meant that a stale axis helper connected to a defunct axis model was being called.

Instead of storing the axis helper instance in a local variable we now store it in a [WeakMap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap) keyed by the `AxisModel`, and look it up when needed so that the axis helper used will always be the one associated with the correct axis model. Because it's a `WeakMap`, axis helpers associated with defunct axis models will automatically be garbage-collected, so we don't need to worry about cleanup.

Separately, there was another reaction in the `useAxis` hook which was turned into a `mstReaction` to avoid another MST warning on destruction of the axis model.

### DIFF
--- a/v3/src/components/axis/helper-models/axis-helper.ts
+++ b/v3/src/components/axis/helper-models/axis-helper.ts
@@ -84,6 +84,10 @@ export class AxisHelper {
       .style("stroke-opacity", "0.7")
   }
 
+  render() {
+    /* istanbul ignore next */
+    throw new Error("Subclass should override")
+  }
 }
 
 export class EmptyAxisHelper extends AxisHelper {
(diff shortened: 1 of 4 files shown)


bfinzer (approved):
👍🏻LGTM
This PR appears not only to have fixed the MST warnings issue but also to have fixed the problem I observed when switching attributes in a scatterplot. Cool. (I'll be asking you about that!)

kswenson:
> 👍🏻LGTM This PR appears not only to have fixed the MST warnings issue but also to have fixed the problem I observed when switching attributes in a scatterplot. Cool. (I'll be asking you about that!)

I didn't see the infinite loop behavior you described after I fixed the first issue with the helper confusion, so I assume it was another symptom of using a stale helper. 🤷

PR #1417, opened 2024-08-20
chore: eliminate `baseUrl` from tsconfig.json

kswenson:
As @scytacki pointed out on slack, [`baseUrl` is no longer required and not recommended](https://www.typescriptlang.org/docs/handbook/modules/reference.html#baseurl).


tealefristoe (approved):
A little outside my expertise, but if it's good enough for Scott, it's good enough for me 👍

PR #1418, opened 2024-08-20
fix: graph response to collection hierarchy change

kswenson:
[[PT-188117637]](https://www.pivotaltracker.com/story/show/188117637) Undo of graph response to moving attribute from child to parent collection not working

A couple of issues:
- collection parent/child links weren't being cleared in some cases when hierarchy changes occurred
- collection cleared its case id map cache unnecessarily in some cases
- graph was responding too early to hierarchy change, before `DataSet` had completed its response

Here we fix the collection issues and add a mechanism for clients to listen for `DataSet` response to hierarchy changes.


bfinzer (approved):
👍🏻LGTM
Strikes me that this was a subtle bug. I still don't understand what caused the case ids to change.

PR #1419, opened 2024-08-20
188131728 v3 DI Url Parameter

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/188131728

This PR adds support for an optional `di` url parameter. When present, Codap will add a plugin with the specified url to the document on load.

Note: Code coverage is reportedly low for this PR, but I think it's actually 100% covered. Not sure why it's reported at only 22.22% right now.

### DIFF
--- a/v3/src/components/app.tsx
+++ b/v3/src/components/app.tsx
@@ -9,7 +9,7 @@ import { MenuBar, kMenuBarElementId } from "./menu-bar/menu-bar"
 import { useCloudFileManager } from "../lib/use-cloud-file-manager"
 import { Logger } from "../lib/logger"
 import { appState } from "../models/app-state"
-import { addDefaultComponents } from "../models/codap/add-default-content"
+import { addDefaultComponents, createDefaultTileOfType } from "../models/codap/add-default-content"
 import {gDataBroker} from "../models/data/data-broker"
 import {IDataSet} from "../models/data/data-set"
 import { dataContextCountChangedNotification } from "../models/data/data-set-notifications"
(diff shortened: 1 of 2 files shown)


v3/src/components/app.tsx
```diff
@@ -100,6 +101,17 @@ export const App = observer(function App() {
...
         }
       }
+
+      if (typeof urlParams.di === "string") {
+        const plugin = createDefaultTileOfType(kWebViewTileType)
+        const row = appState.document.content?.getRowByIndex(0)
+        if (isWebViewModel(plugin?.content) && row) {
+          plugin.content.setUrl(urlParams.di)
+          const pluginOptions = isFreeTileRow(row) ? { x: 5, y: 5 } : undefined
+          appState.document.content?.insertTileInRow(plugin, row, pluginOptions)
```
kswenson, line 111: The `di` parameter can be added when launching with existing documents as well, so my inclination would be to use something like `DocumentContent.createTile()` which will make use of the auto-placement code.

kswenson (approved):
👍 Looks good -- I suggested making use of the tile auto-placement code rather than hard-coding the position.

v3/src/components/app.tsx
tealefristoe: I was able to get this to work, but I had to wrap it in a `setTimeout` because elements must be rendered for `getPositionOfNewComponent` to work. Is that alright? Is there another place it would make sense to add the plugin instead, so a `setTimeout` isn't necessary?

kswenson: I think the `setTimeout` is okay for now.

v3/src/components/app.tsx
```diff
@@ -100,6 +100,19 @@ export const App = observer(function App() {
...
       }
+
+      const { di } = urlParams
+      if (typeof di === "string") {
+        // setTimeout ensures that other components have been rendered,
+        // which is necessary to properly position the plugin.
+        setTimeout(() => {
+          appState.document.content?.applyModelChange(() => {
+            const plugin = appState.document.content?.createTile?.(kWebViewTileType)
+            if (isWebViewModel(plugin?.content)) plugin.content.setUrl(di)
```
kswenson, line 111: FYI: You can pass a content snapshot to `createTile` which would allow you to eliminate the second line, if this code is ever revisited.

PR #1420, opened 2024-08-21
fix: insert cases in hierarchical collection; insert cases undo/redo

kswenson:
[[PT-188139181]](https://www.pivotaltracker.com/story/show/188139181)


tealefristoe (approved):
Looks good 👍 As always, love the reduced waits in the cypress tests.

kswenson:
>As always, love the reduced waits in the cypress tests.

Whenever I'm working on cypress tests these days I try to reduce the `wait`s.

PR #1421, opened 2024-08-21
188150453 v3 DI Handle Slate Objects in API Requests

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/188150453

This PR updates the text request handler.
- Previously, only strings were supported for the `text` field. Now, Slate style objects are handled as well.
- Text components now update when their contents are changed via API requests.

### DIFF
--- a/v3/src/components/text/text-model.ts
+++ b/v3/src/components/text/text-model.ts
@@ -35,6 +35,7 @@ export const TextModel = TileContentModel
   }))
   .actions(self => ({
     setValue(value: EditorValue) {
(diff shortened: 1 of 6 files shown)


kswenson (approved):
👍 Looks good -- I added a commit to tweak a few things:
1. The document type is `SlateExchangeValue`, so you don't need to define `SlateDocument`.
2. The `objTypes` field was part of the document format in the pre-0.5 slate days, but is no longer required. It should be harmless if passed from the plugin API, however.

PR #1422, opened 2024-08-21
188150453 DI Fix Get dataContextList

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/188150453

This PR fixes the outdated `get dataContextList` handler, which was written at a time before datasets had titles.


kswenson (approved):
👍 LGTM

PR #1423, opened 2024-08-22
chore: enable eslint-plugin-mocha; warn/error on use of `.only` in cypress tests

kswenson:
After tweaking some cypress tests recently I was disappointed to see that the code coverage was lower than I expected and that some lines I knew I had written tests for were not covered. 😕 It took me a few minutes to realize that I had inadvertently committed a cypress test with `.only()`, so that other tests in the file were being skipped. This confused me further, because I thought we had a lint rule to catch such problems. 😕 Further investigation revealed that we have a lint rule for use of `.only()` in jest tests, but _not_ in cypress tests. This PR fixes that by installing/configuring `eslint-plugin-mocha` and tweaking the configuration so that `.only()` warns during development and errors on build, i.e. it breaks the build.

### DIFF
--- a/v3/.eslintrc.build.cjs
+++ b/v3/.eslintrc.build.cjs
@@ -10,7 +10,13 @@ module.exports = {
     { // rules specific to Jest tests
       files: ["src/**/*.test.*", "src/test/**"],
       rules: {
-        "jest/no-focused-tests": "warn"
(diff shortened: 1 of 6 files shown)


tealefristoe (approved):
Great improvement 👍

PR #1427, opened 2024-08-23
187936148 map layer palette tuneup

eireland:
Restyles the map tool palettes.


v3/src/components/inspector-panel.scss
```diff
@@ -7,7 +7,7 @@ $inspector-panel-width: 50px;
   top: 0;
   width: $inspector-panel-width;
   min-height: 50px;
-  z-index: 10;  // Will need to be higher to accommodate Leaflet
+  z-index: 50;
```
kswenson, line 10: Out of curiosity, what was the symptom that led you to bump up the `z-index` here? I'd like to consider whether there might be a better solution than `z-index` inflation.

eireland: This was mainly for the window expand/contract icon to not appear on top of the palette.

kswenson: There may be a better solution lurking here, but for now make it `15` and we can put off a deeper analysis for later.
```suggestion
  z-index: 15;
```

kswenson (approved):
👍 LGTM -- just one suggestion on `z-index`.

PR #1430, opened 2024-08-24
fix: Formula compilation adds extraneous actions to undo history [PT-188166140]

kswenson:
[[PT-188166140]](https://www.pivotaltracker.com/story/show/188166140)

Previously, the canonical representation of a formula was stored in the model along with the user-entered display representation. This meant that by default, compiling/canonicalizing a formula would result in model changes that were undoable by default. But canonicalization occurs at times when we don't want the document to be dirtied, e.g. on initial load of a document. This PR moves the canonical representation of a formula into `volatile` storage so that changes to it do not affect the undo history.

The primary use of the canonical representation of a formula is to handle name changes so that, for instance, when an attribute's name is changed, formulas that reference that attribute can be updated to refer to the new name. This fix assumes that there is no additional value in serializing the canonical representation of a formula, but that assumption should be considered carefully. The first commit in this PR implements a more elaborate fix (which I implemented first) which moves the canonical representation of a formula into `volatile` storage during the running of the application, but then copies it back into the model for serialization purposes. If it is determined that there are scenarios in which it is necessary/important to serialize the canonical form of formulas, we can adopt that implementation instead.

### DIFF
--- a/v3/src/models/data/attribute.ts
+++ b/v3/src/models/data/attribute.ts
@@ -207,6 +207,7 @@ export const Attribute = V2Model.named("Attribute").props({
       withoutUndo({ suppressWarning: true })
       self.values = undefined
     }
+    self.formula?.prepareSnapshot()
   },
   // should be called after retrieving snapshot (i.e. after serialization)
   completeSnapshot() {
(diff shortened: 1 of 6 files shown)


pjanik (approved):
Looks good 👍 

> The primary use of the canonical representation of a formula is to handle name changes so that, for instance, when an attribute's name is changed, formulas that reference that attribute can be updated to refer to the new name. This fix assumes that there is no additional value in serializing the canonical representation of a formula, but that assumption should be considered carefully. 

It has been a while since I worked on that, but I believe this is correct.

PR #1431, opened 2024-08-24
fix: display of lists in text component [PT-188103062]

kswenson:
There was a style conflict between the slate-editor library and Chakra which resulted in list item markers being displayed outside the bounds of the element, i.e. they were not visible.


tealefristoe (approved):
👍

PR #1433, opened 2024-08-26
chore: conditionalize logging to server when debugging

kswenson:
Adds a separate `DEBUG_LOG_TO_SERVER` flag to enable/disable logging to the log server. This flag defaults to `!DEBUG_LOGGER`, i.e. if you are debugging/observing logs in the browser the assumption is that logging to an actual server is not necessary. If a developer is actually debugging the server communication, of course, the flag can be enabled locally.


eireland (approved):
👍 LGTM

PR #1434, opened 2024-08-27
Fix bar charts

bfinzer:
[#188144634] Bug fix: Bar charts are broken

* Centralize the response to the change in `graphModel.pointsFusedIntoBars` as an mstReaction in graph.tsx so that the response doesn't rely on order of operations.


kswenson (approved):
👍 Looks good!

PR #1435, opened 2024-08-27
fix: fix blank url in sharing dialog by updating to latest CFM version

kswenson:
[[PT-188104620]](https://www.pivotaltracker.com/story/show/188104620)

- Update to latest CFM version which fixed the dialog display issue [CFM PR #358](https://github.com/concord-consortium/cloud-file-manager/pull/358).
- Use `cloneDeep()` when serializing document snapshots to avoid MST warning on use of frozen snapshots.


tealefristoe (approved):
Looks good 👍

PR #1436, opened 2024-08-27
188150453 v3 DI Get StoryQ Working

tealefristoe:
PT story: https://www.pivotaltracker.com/story/show/188150453

This PR makes the necessary modifications to v3 to get the StoryQ plugin working.
- `id` (case id) is now included in response to `create case` API requests.
- Responses to `get selectionList` API requests have been improved.
  - Case ids are now included instead of item ids.
  - `collectionID` and `collectionName` are now included.
  - Included cases are now filtered by the collection specified in the API request, if there is one.
- Fixes the `patternMatches` function by calling `unescapeBacktickString` on the pattern, rather than `escapeStringRegexp`. (This makes the function work with patterns specified by StoryQ, and it didn't seem like anything required `escapeStirngRegexp`, but I could be wrong about that.)
- Implements the `wordListMatches` function.
- Adds support for `notify dataContext` requests, which allow plugins to control `setAsides`.
- Broadcasts notifications after responding to `create` and `update case` API requests.


v3/src/models/formula/functions/string-functions.ts
```diff
@@ -229,9 +233,51 @@ export const stringFunctions = {
...
+      const ratingAttributeName = ratingAttributeNameArg?.value
+      const ratingAttribute = ratingAttributeName ? dataSet.getAttributeByName(ratingAttributeName) : undefined
+
+      const wordRatingMap: Record<string, number> = {}
+      wordAttribute.strValues.forEach((word, index) => wordRatingMap[word] = ratingAttribute?.numeric(index) ?? 1)
+
+      let result = 0
+      wordAttribute.strValues.forEach(word => {
+        const isRegEx = word.startsWith("/") && word.endsWith("/")
+        const pattern = isRegEx ? word.substring(1, word.length - 1) : `\\b${word}\\b`
```
kswenson, line 273:
```suggestion
        const pattern = isRegEx ? word.substring(1, word.length - 1) : `\\b${escapeStringRegexp(word)}\\b`
```

v3/src/models/formula/functions/string-functions.ts
```diff
@@ -92,7 +96,7 @@ export const stringFunctions = {
     evaluate: (...args: FValue[]) => {
       const stringToLookIn = valueToString(args[0]),
         pattern = valueToString(args[1])
-      return (stringToLookIn.match(new RegExp(escapeStringRegexp(pattern), "gmi")) || []).length
+      return (stringToLookIn.match(new RegExp(unescapeBacktickString(pattern), "gmi")) || []).length
```
kswenson, line 99: Still need to sanitize the input, so it would have to be something like:
```suggestion
      return (stringToLookIn.match(new RegExp(escapeStringRegexp(unescapeBacktickString(pattern)), "gmi")) || []).length
```

v3/src/data-interactive/handlers/selection-list-handler.ts
```diff
@@ -43,13 +43,19 @@ export const diSelectionListHandler: DIHandler = {
...
     if (!dataContext) return dataContextNotFoundResult
 
-    const caseIds = Array.from(dataContext.selection)
-    // TODO Include collectionID and collectionName
-    const values = caseIds.map(caseID => ({ caseID: toV2Id(caseID) }))
-    // TODO Filter based on collection
+    const itemIds = new Set(dataContext.selection)
+    const _values = Array.from(dataContext.caseInfoMap.values())
+      // Only include cases where all child items are selected
+      .filter(aCase => !aCase.childItemIds.some(itemId => !itemIds.has(itemId)))
```
kswenson, line 53: That's the behavior of `DataSet.isCaseSelected()`, so you should be able to use that here.

v3/src/models/formula/formula-mathjs-scope.ts
```diff
@@ -217,6 +217,10 @@ export class FormulaMathJsScope {
     return this.context.dataSets.get(dataSetId)
   }
 
+  getDataSetByTitle(dataSetTitle: string) {
+    return Array.from(this.context.dataSets.values()).find(dataSet => dataSet._title === dataSetTitle)
```
kswenson, line 221: I would think we'd want 
```suggestion
    return Array.from(this.context.dataSets.values()).find(dataSet => dataSet.title === dataSetTitle)
```

v3/src/models/formula/functions/string-functions.ts
```diff
@@ -229,9 +233,51 @@ export const stringFunctions = {
...
+        throw new Error(t("DG.Formula.FuncArgsErrorPlural.description", { vars: [ functionName, numOfReqArgs ] }))
+      }
+      [1, 2].forEach(i => {
+        if (!isConstantStringNode(args[i])) {
+          throw new Error(t("V3.formula.error.stringConstantArg", { vars: [ functionName, i + 1 ] }))
+        }
+      })
+      const [_textRefArg, dataSetTitleArg, wordAttributeNameArg, ratingAttributeNameArg] =
+        args as LookupStringConstantArg[]
+      const text = evaluateNode(args[0], scope) as string
```
kswenson, line 250:
```suggestion
      const text = evaluateNode(_textRefArg, scope) as string
```

v3/src/models/formula/functions/string-functions.ts
```diff
@@ -229,9 +233,51 @@ export const stringFunctions = {
...
+      wordAttribute.strValues.forEach((word, index) => wordRatingMap[word] = ratingAttribute?.numeric(index) ?? 1)
+
+      let result = 0
+      wordAttribute.strValues.forEach(word => {
+        const isRegEx = word.startsWith("/") && word.endsWith("/")
+        const pattern = isRegEx ? word.substring(1, word.length - 1) : `\\b${word}\\b`
+        const flags = isRegEx ? "g" : "gi"
+        const regEx = new RegExp(pattern, flags)
+        const match = text.match(regEx)
+        result += wordRatingMap[word] * (match ? match.length : 0)
```
kswenson, line 277: Could also use:
```suggestion
        result += wordRatingMap[word] * (match?.length ?? 0)
```

kswenson (changes requested):
👍 Looks good -- beyond a couple of cosmetic issues, the main suggestion is to always use `escapeStringRegexp` both for functionality, to deal with the case of users searching for strings that include regex symbols (`.`, `?`, `\`, etc.), but also for security to protected against regex DOS issues.

v3/src/models/formula/functions/string-functions.ts
tealefristoe: Unfortunately, I don't think this will work for the StoryQ plugin. It sends patterns like `\\\\bword\\\\b`, which reach the function as `\\bword\\b`. `unescapeBacktickString` translates this to `\bword\b`, which is required for the function to work properly. I don't think there's any way to call `escapeStringRegexp` on this string without adding extra `\`, breaking the function.

kswenson: We can put this off to later if we don't want to deal with it now, but the right thing to do in that case would be to extract `word`, escape it, and then wrap it in `\b` again. Passing external input directly to a regex is like passing user input directly to SQL -- never a good/safe idea -- just ask [Little Bobby Tables](https://xkcd.com/327/). 😜

v3/src/models/formula/functions/string-functions.ts
tealefristoe: This was giving me type problems because `_textRefArg` is cast as a `LookupStringConstantArg`.

kswenson: In that case, it's probably not worth worrying about. You could, however, type it as something like `as [any/node/whatever, LookupStringConstantArg, LookupStringConstantArg. LookupStringConstantArg]`.

v3/src/models/formula/functions/string-functions.ts
tealefristoe: I can understand that this is important for security, but it's going to be tricky to get it to work with old plugins. v2 feeds patterns directly into RegExps without any processing. Playing around with StoryQ, I've already found one other example of patterns with `\` (`\\\\!` for a punctuation feature), and I wouldn't be surprised if there were more. Will we be able to catch all of these cases with individual special rules?

The alternative would be to change the plugins, which we might want to do. But what could StoryQ pass to Codap to get the `\bword\b` pattern? `escapeStringRegexp` will not add a `\` to `b`, and if StoryQ includes any `\`s, they will become `\\` when we run `escapeStringRegexp` on the pattern.

kswenson: So stick with v2 behavior for now and add a TODO to return to it at some point.

kswenson: I was thinking about this more recently and another approach would be to do the regex matching in a WebWorker which could be canceled if it took too long. In any case, we can certainly put this off for another day.

PR #1437, opened 2024-08-28
chore: "Rename" => "Rename..." in CFM menu

kswenson:
[[PT-181840559]](https://www.pivotaltracker.com/story/show/181840559)


kswenson:
Merging without review as it's a trivial string change.

PR #1438, opened 2024-08-28
feat: Representation of dataset in case card is hierarchical (PT-188173137)

emcelroy:
[#188173137](https://www.pivotaltracker.com/story/show/188173137)

### DIFF
--- a/v3/cypress/e2e/case-card.spec.ts
+++ b/v3/cypress/e2e/case-card.spec.ts
@@ -1,22 +1,79 @@
+import { TableTileElements as table } from "../support/elements/table-tile"
 
 context("case card", () => {
   beforeEach(() => {
     // cy.scrollTo() doesn't work as expected with `scroll-behavior: smooth`
-    const queryParams = "?sample=mammals&scrollBehavior=auto"
+    const queryParams = "?sample=mammals&mouseSensor&scrollBehavior=auto"
     const url = `${Cypress.config("index")}${queryParams}`
     cy.visit(url)
     cy.wait(2000)
   })
 
+  const tableHeaderLeftSelector = ".codap-component.codap-case-table .component-title-bar .header-left"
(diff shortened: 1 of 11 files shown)


kswenson (approved):
👍 Looks good -- I added a commit which fixes a couple of things, including a latent `DataSet` bug (#1441) which I discovered while reviewing it. There's also a few minor suggestions. Unfortunately, there are now failing cypress tests which seem unrelated to this PR but are now occurring consistently after my commit, so I'm not sure what's going on there, but we'll need to figure it out before merging. 🤔

emcelroy:
> Unfortunately, there are now failing cypress tests which seem unrelated to this PR but are now occurring consistently after my commit, so I'm not sure what's going on there, but we'll need to figure it out before merging. 🤔

The newly added `overflow: hidden` line in free-tile-row.scss was preventing the inspector menu to be hidden in at least some cases, hence the failing tests. Is that style rule important? If so, would something like `overflow-y: hidden` be an option?

kswenson:
> > Unfortunately, there are now failing cypress tests which seem unrelated to this PR but are now occurring consistently after my commit, so I'm not sure what's going on there, but we'll need to figure it out before merging. 🤔
> 
> @kswenson The newly added `overflow: hidden` line in free-tile-row.scss was preventing the inspector menu to be hidden in at least some cases, hence the failing tests. Is that style rule important? If so, would something like `overflow-y: hidden` be an option?

🤦 That makes sense. That should have occurred to me last night. I was trying to fix the problem of the case card overflowing its container, so `overflow-y: hidden` should be fine for now. I suspect I'm going to have to revisit that at some point, but  so be it.

emcelroy:
> 🤦 That makes sense. That should have occurred to me last night. I was trying to fix the problem of the case card overflowing its container, so `overflow-y: hidden` should be fine for now. I suspect I'm going to have to revisit that at some point, but so be it.

I tried `overflow-y` but the tests failed for that, too. So I'm going to leave out any overflow rules for now.

PR #1439, opened 2024-08-28
Fix date axis restore bug

bfinzer:
[#188188327] Bug fix: Date axis not restored properly

* In the restore process the date axis' multiscale is not getting 'linear' as its scaleType. We fix this in use-sub-axis.ts `installDomainSync` by guaranteeing that if we're dealing with BaseNumericAxisModel, the multiscale has the right scaleType.


kswenson (approved):
👍 LGTM

PR #1441, opened 2024-08-29
fix: reset `caseIds` in `itemInfoMap` when appropriate

kswenson:
While reviewing #1438, I noticed some redundant values in the `caseIds` field of the `itemInfoMap`. Turns out the `caseIds` were never being cleared, so the array just kept getting longer.


tealefristoe (approved):
Good catch 👍

PR #1443, opened 2024-08-29
182089941 add log strings plugins

eireland:
Adds logging sent by a plugin. 
Tested with Sampler, and SageModeler

### DIFF
--- a/v3/src/components/tool-shelf/plugins-button.tsx
+++ b/v3/src/components/tool-shelf/plugins-button.tsx
@@ -36,7 +36,8 @@ function PluginItem({ pluginData }: IPluginItemProps) {
         if (isWebViewModel(tile?.content)) tile.content.setUrl(url)
       }, {
         undoStringKey: t("V3.Undo.plugin.create", { vars: [pluginData.title] }),
-        redoStringKey: t("V3.Redo.plugin.create", { vars: [pluginData.title] })
+        redoStringKey: t("V3.Redo.plugin.create", { vars: [pluginData.title] }),
+        log: {message: "Add Plugin", args: { name: pluginData.title, url: pluginData.path } }
       }
     )
   }
--- a/v3/src/components/tool-shelf/plugins-button.tsx
+++ b/v3/src/components/tool-shelf/plugins-button.tsx
@@ -17,6 +17,7 @@ const standardPlugins = _standardPlugins as PluginMenuConfig
 const combinedPlugins = [...standardPlugins, ...debugPlugins]
 
 import "./plugins-button.scss"
+import { logMessageWithReplacement } from "../../lib/log-message"
 
 interface IPluginItemProps {
   pluginData: PluginData | null
@@ -37,7 +38,8 @@ function PluginItem({ pluginData }: IPluginItemProps) {
       }, {
         undoStringKey: t("V3.Undo.plugin.create", { vars: [pluginData.title] }),
         redoStringKey: t("V3.Redo.plugin.create", { vars: [pluginData.title] }),
-        log: {message: "Add Plugin", args: { name: pluginData.title, url: pluginData.path } }
+        log: logMessageWithReplacement("Add Plugin: %@", { name: pluginData.title, url: pluginData.path })
+
       }
     )
   }
--- a/v3/src/components/tool-shelf/plugins-button.tsx
+++ b/v3/src/components/tool-shelf/plugins-button.tsx
@@ -10,6 +10,7 @@ import { isWebViewModel } from "../web-view/web-view-model"
 import { kRootPluginUrl, processPluginUrl } from "../web-view/web-view-utils"
 import { ToolShelfButtonTag } from "./tool-shelf-button"
 import { PluginData, PluginMenuConfig } from "./plugin-config-types"
+import { logMessageWithReplacement } from "../../lib/log-message"
 import _debugPlugins from "./debug-plugins.json"
 import _standardPlugins from "./standard-plugins.json"
 const debugPlugins = DEBUG_PLUGINS ? _debugPlugins as PluginMenuConfig : []
@@ -17,7 +18,6 @@ const standardPlugins = _standardPlugins as PluginMenuConfig
 const combinedPlugins = [...standardPlugins, ...debugPlugins]
 
 import "./plugins-button.scss"
-import { logMessageWithReplacement } from "../../lib/log-message"
 
 interface IPluginItemProps {
   pluginData: PluginData | null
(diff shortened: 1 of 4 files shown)


v3/src/data-interactive/handlers/log-message-handler.ts
```diff
@@ -1,8 +1,24 @@
...
-  notify: diNotImplementedYet
+  notify(resources: DIResources, values?: DIValues) {
+    const webviewModel = resources.interactiveFrame?.content
+    const {formatStr, replaceArgs} = values as DILogMessage
+    const message = formatStr ?? ""
+    const argsArray = replaceArgs != null
+      ? Array.isArray(replaceArgs) ? replaceArgs : [replaceArgs]
+      : undefined
+    const args: LoggableObject = {}
+    argsArray?.forEach((value: any, index: number) => {
```
tealefristoe, line 14:
```suggestion
    argsArray?.forEach((value: LoggableObject, index: number) => {
```

v3/src/data-interactive/handlers/log-message-handler.ts
```diff
@@ -1,8 +1,24 @@
...
+      args[index.toString()] = value
+    })
+    webviewModel?.applyModelChange(()=>{},
+      { log: logMessageWithReplacement(message, args)}
```
tealefristoe, line 18:
```suggestion
      { log: logMessageWithReplacement(message, args) }
```

tealefristoe (changes requested):
Looks great, but I think you should add a `log-message-handler.test.ts` test. Hopefully this will be easy... there are examples for almost all of the other handlers, so the only potentially tricky thing would be making sure the logging actually occurs. I hope you can mimic how you do that for tests for core logging functionality?

v3/src/components/tool-shelf/plugins-button.tsx
```diff
@@ -36,7 +36,8 @@ function PluginItem({ pluginData }: IPluginItemProps) {
         if (isWebViewModel(tile?.content)) tile.content.setUrl(url)
       }, {
         undoStringKey: t("V3.Undo.plugin.create", { vars: [pluginData.title] }),
-        redoStringKey: t("V3.Redo.plugin.create", { vars: [pluginData.title] })
+        redoStringKey: t("V3.Redo.plugin.create", { vars: [pluginData.title] }),
+        log: {message: "Add Plugin", args: { name: pluginData.title, url: pluginData.path } }
```
kswenson, line 40: Is there a reason not to do:
```suggestion
        log: logStringifiedObjectMessage("Add Plugin: %@", { name: pluginData.title, url: pluginData.path })
```
?

kswenson (commented):
👍 One suggestion and otherwise I concur with @tealefristoe's suggestions.

v3/src/data-interactive/handlers/log-message-handler.ts
```diff
@@ -1,8 +1,28 @@
+import { LoggableObject, LoggableValue, logMessageWithReplacement } from "../../lib/log-message"
 import { registerDIHandler } from "../data-interactive-handler"
-import { DIHandler, diNotImplementedYet } from "../data-interactive-types"
+import { DIHandler, DILogMessage, DIResources, DIValues,  } from "../data-interactive-types"
 
 export const diLogMessageHandler: DIHandler = {
-  notify: diNotImplementedYet
+  notify(resources: DIResources, values?: DIValues) {
+    const webviewModel = resources.interactiveFrame?.content
+    if (values) {
```
tealefristoe, line 8: Did you double check how this works in v2? If not, you should definitely try using the API tester plugin to try this out. I seem to remember the API handler sending an empty log message when no values are included, but maybe I'm misremembering. If you do want to send something even if no values are specified, you could replace the following line with this:
`const { formatStr, replaceArgs } = (values ?? {}) as DILogMessage`

v3/src/data-interactive/handlers/log-message-handler.ts
```diff
@@ -1,8 +1,28 @@
...
+      const args: LoggableObject = {}
+      argsArray?.forEach((value: LoggableValue, index: number) => {
+        args[index.toString()] = value
+      })
+      webviewModel?.applyModelChange(()=>{},
+        { log: logMessageWithReplacement(message, args) }
+      )
+      return {success: true}
+    } else {
+      return {success: false, error: "No values provided"}
```
tealefristoe, line 23: If we want to keep this result (see last comment), it should be a translatable string.

tealefristoe (changes requested):
Super close! The main thing to deal with is making sure the behavior of log requests with no values are handled properly, and if so using a translatable string rather than a hard coded one.

v3/src/data-interactive/handlers/log-message-handler.ts
kswenson: > If we want to keep this result (see last comment), it should be a translatable string.

I'm not sure that's right. Where logs and the plugin API are concerned, these strings are generally being handled by code (i.e. scripts or plugins) where consistency is likely to be more useful than localizability. That said, if we imagine a particular string being provided for the specific purpose of showing to the user, then maybe a localizable string would make sense, but I'm not sure we have a way of indicating that distinction in the API.

tealefristoe: This line is an API request result, not a logged message, so I think it makes sense to make it translatable. At least, that's what I've done for all other request messages.

kswenson: 👍 Consistency is a virtue at this point.

v3/src/data-interactive/handlers/log-message-handler.ts
eireland: We tried no `replaceArgs`, and that worked.
V2 returns this error when there is no `values`:
```{
  "success": false,
  "values": {
    "error": "TypeError: Cannot read properties of undefined (reading 'topic')"
  }
}```

File: v3/src/data-interactive/handlers/log-message-handler.ts [hunk 6]
```diff
@@ -1,8 +1,29 @@
...
+      const args: LoggableObject = {}
+      argsArray?.forEach((value: LoggableValue, index: number) => {
+        args[index.toString()] = value
+      })
+      webviewModel?.applyModelChange(()=>{},
+        { log: logMessageWithReplacement(message, args) }
+      )
+      return {success: true}
+    } else {
+      return {success: false, error: t("V3.DI.Error.valuesRequired")}
```
tealefristoe, line 24:
```suggestion
      return valuesRequiredResult
```
I should have thought of this before. You can import `valuesRequiredResult` from `di-results.ts`.

tealefristoe (approved):
Looks good, just one more small change I should have recommended before.

kswenson (approved):
👍 LGTM

PR #1444, opened 2024-08-29
Fix date axis dynamic resize

bfinzer:
[#188109309] Bug fix: Graph doesn't respond dynamically to change in extent of date axis

* Includes a couple of lint fixes

* In use-axis.ts we weren't doing a test for numeric axis that includes a date axis. And in computeDesiredExtent for the date axis we weren't getting the dynamic values.


kswenson (approved):
👍 LGTM

PR #1445, opened 2024-08-29
fix flaky Cypress test (text tile undo/redo)

nstclair-cc:
### Fixes Medium Flakiness in "updates text title with undo/redo" Test

This PR addresses a small issue in the "updates text title with undo/redo" test, which has recently become "medium" flaky (see [Flaky Test Analytics](https://cloud.cypress.io/projects/msrfxa/analytics/flaky-tests)).

**Issue**: During the test run, it seems that the Redo button is momentarily disabled right as Cypress attempts to press it, leading to test failures.

**Solution**: I've added a small wait before Cypress presses the Redo button as a minor tweak to address this issue. I'm hopeful this will resolve the flakiness, but I’ll keep an eye on the test over the next few runs to confirm.

Please review and feel free to merge if it looks good. I’ll update if any further adjustments are needed.

### DIFF
--- a/v3/cypress/e2e/text.spec.ts
+++ b/v3/cypress/e2e/text.spec.ts
@@ -24,6 +24,9 @@ context("Text tile", () => {
 
     // Redo title change
     toolbar.getRedoTool().click()
+    cy.wait(300) // this test has become flaky.
+    // Added a small wait here to see if this fixes
+    // the disabled redo button issue
     c.getComponentTitle(kTextTileTestId).should("have.text", newTextTileName)
   })
   it("close text tile from close button with undo/redo", () => {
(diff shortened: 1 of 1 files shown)


v3/cypress/e2e/text.spec.ts
```diff
@@ -23,6 +23,9 @@ context("Text tile", () => {
     c.getComponentTitle(kTextTileTestId).should("have.text", textDefaultTitle)
 
     // Redo title change
+    cy.wait(300) // this test has become flaky.
+    // Added a small wait here to see if this fixes
+    // the disabled redo button issue
```
kswenson, line 28: Rather than adding an arbitrary timed wait, it would be preferable to wait on what we actually care about:
```suggestion
    toolbar.getRedoTool().should("be.enabled")
```

kswenson (changes requested):
See inline suggestion for avoiding arbitrary wait.

nstclair-cc:
> See inline suggestion for avoiding arbitrary wait.

Thanks, @kswenson! I probably should have meditated on it more—something like that occurred to me last night after work. I appreciate you making the fix to the code!

kswenson (approved):
👍 LGTM

PR #1447, opened 2024-08-30
fix: parent cases maintain persistent ids when possible

kswenson:
[[PT-188003195]](https://www.pivotaltracker.com/story/show/188003195)

Parent collections group items into cases based on the attribute values in those collections (and their parents). For tracking purposes, the case id generated for each case is associated with the specific set of values that defines that case. By default, therefore, changing the values of a parent case or adding new attributes to a parent collection necessarily results in new case ids being generated, because the set of values being used as a key has changed. We would prefer, however, that changes that _don't materially affect the case/item groups_ should _not_ change the case ids. Therefore, this PR adds a post-processing step to the case/item grouping algorithm which detects when new case ids were generated for a case whose constituent items didn't change and instead persists the prior case id. It has to be done as a post-processing step because we don't know whether the resulting case/item groups have changed until after we've completed the first pass.

### DIFF
--- a/v3/src/models/data/collection.test.ts
+++ b/v3/src/models/data/collection.test.ts
@@ -357,6 +357,20 @@ describe("CollectionModel", () => {
     expect(c1.caseIds).toEqual(originalParentCaseIds)
     expect(c2.caseIds).toEqual(originalChildCaseIds)
 
+    // removing attr1 from the parent collection invalidates grouping and changes parent case ids
+    c1.removeAttribute(a1.id)
+    expect(itemData.invalidate).toHaveBeenCalledTimes(2)
+    validateCases()
+    expect(c1.caseIds).not.toEqual(originalParentCaseIds)
+    expect([...c2.caseIds].sort()).toEqual([...originalChildCaseIds].sort())
+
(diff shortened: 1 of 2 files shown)


tealefristoe (approved):
Looks good 👍 

I'm hopeful that this will fix the issue with StoryQ, but there are still situations where v3 will not match v2 behavior when it comes to persistent case ids. One that comes to mind is when a change to a case combines it with another case. A quick experiment indicates that in this case the first id is kept for the now larger single case. There might be other situations where v2 maintains case ids as well.

kswenson:
> I'm hopeful that this will fix the issue with StoryQ, but there are still situations where v3 will not match v2 behavior when it comes to persistent case ids. One that comes to mind is when a change to a case combines it with another case. A quick experiment indicates that in this case the first id is kept for the now larger single case. There might be other situations where v2 maintains case ids as well.

I'd like to dig into this further. If I'm understanding the scenario you're describing I think it would be handled correctly in v3. That said, while playing around I did discover some situations around response to hierarchy changes that were clearing the group key => case id map that upon further consideration probably shouldn't be, so I added a commit to this PR along with an additional test that fails without this change but passes with it.

tealefristoe (approved):
👍

PR #1449, opened 2024-08-30
feat: highlight grouping lines in case table based on selection

kswenson:
While working on #1447 I found myself annoyed that the grouping/relationship lines in the case table did not indicate selection. It was straightforward to color the area as selected or not depending on whether the parent case (and hence all the child cases) were selected. I subsequently found [this PT story](https://www.pivotaltracker.com/story/show/185226959) which describes a more elaborate version of this feature in which the area between the grouping/relationship lines would be colored proportionally to the percentage of child cases selected. That PT story, however, also states that

>The first pass of implementation should be quick and dirty to determine whether the effect is actually helpful.

so this can be viewed as the initial quick and dirty implementation.

CaseTableSelectedGroupingLines

### DIFF
--- a/v3/src/components/case-table/collection-table-spacer.tsx
+++ b/v3/src/components/case-table/collection-table-spacer.tsx
@@ -17,9 +17,10 @@ import { CurvedSpline } from "./curved-spline"
 import { useCollectionTableModel } from "./use-collection-table-model"
 
 interface IProps {
+  selectedFillColor?: string
   onDrop?: (dataSet: IDataSet, attrId: string) => void
 }
-export const CollectionTableSpacer = observer(function CollectionTableSpacer({ onDrop }: IProps) {
+export const CollectionTableSpacer = observer(function CollectionTableSpacer({ selectedFillColor, onDrop }: IProps) {
   const data = useDataSetContext()
   const caseMetadata = useCaseMetadata()
   const parentCollectionId = useParentCollectionContext()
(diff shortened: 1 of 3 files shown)


bfinzer (approved):
👍🏻LGTM
I really like the quick and dirty experiment. But are we OK with the conceptual implications? Is selecting all the children of a parent case really the same as also selecting the parent? In V2, of course, it isn't. It also means that selecting child cases in a graph can change parent case selection, also not true in V2. Might be worth discussion with Tim?

kswenson:
> I really like the quick and dirty experiment. But are we OK with the conceptual implications? Is selecting all the children of a parent case really the same as also selecting the parent? In V2, of course, it isn't. It also means that selecting child cases in a graph can change parent case selection, also not true in V2. Might be worth discussion with Tim?

🤦 I'd forgotten that. This is already true in v3 at the DataSet level, where the implementation of `isCaseSelected` is that a case is selected if all of its items are selected. So if we're sticking with the v2 behavior, then there's a non-trivial bug to fix in v3 and _then_ potentially the rule here would change to showing the relationship lines as selected if the parent _and_ the children are selected.

kswenson:
> > I really like the quick and dirty experiment. But are we OK with the conceptual implications? Is selecting all the children of a parent case really the same as also selecting the parent? In V2, of course, it isn't. It also means that selecting child cases in a graph can change parent case selection, also not true in V2. Might be worth discussion with Tim?
> 
> 🤦 I'd forgotten that. This is already true in v3 at the DataSet level, where the implementation of `isCaseSelected` is that a case is selected if all of its items are selected. So if we're sticking with the v2 behavior, then there's a non-trivial bug to fix in v3 and _then_ potentially the rule here would change to showing the relationship lines as selected if the parent _and_ the children are selected.

Actually, looking at it more closely, I don't think the issue you raise affects the evaluation of this PR. If the parent is selected, then all of its children must be selected seems to be true in v2 and v3. It's the reverse proposition -- that the parent must be selected if all of its children are selected -- that is the point at issue here. The rule that is implemented in this PR is that the relationship area is rendered selected if the parent is selected, which seems like it's correct regardless of the answer about whether the parent must be selected if the children are selected.

bfinzer:
At noon tomorrow we have a CODAP Outreach meeting which Dan, Tim and Zac attend. I'll run this by them to see what they think about this level of change.

bfinzer:
Ah, interesting. OK so it's OK to merge the PR after Teale's review. But I'll still run the issue by others. Thanks!

kswenson:
> At noon tomorrow we have a CODAP Outreach meeting which Dan, Tim and Zac attend. I'll run this by them to see what they think about this level of change.

I think our comments crossed in the ether, but just to make sure, I don't think this PR is proposing any radical changes. The selection change you're concerned about is already baked into the `DataSet`. So, having that conversation makes sense, but it's in the context of whether the existing `DataSet` implementation of selection should be changed to match v2 behavior, not in the context of whether this PR should be accepted.

PR #1450, opened 2024-08-31
feat: implement set-aside for cases

kswenson:
[[PT-188003381]](https://www.pivotaltracker.com/story/show/188003381)

- Case table inspector menu items "Set Aside Selected Cases", "Set Aside Unselected Cases", and "Restore %d Set Aside Cases" work as expected.
- `DataSet` has API for hiding/showing cases.
- Unfiltered `itemIds` property renamed to `_itemIds`.
- `itemIds` is now a view method which takes hidden items into account.
- Tables and graphs respond to hiding/showing cases.


v3/src/models/data/data-set.ts
```diff
@@ -132,12 +132,13 @@ export const DataSet = V2Model.named("DataSet").props({
...
   attributesMap: types.map(Attribute),
-  itemIds: types.array(types.string),
+  _itemIds: types.array(types.string),
   sourceName: types.maybe(types.string),
   description: types.maybe(types.string),
   importDate: types.maybe(types.string),
   // for serialization only, not for dynamic selection tracking
-  snapSelection: types.array(types.string)
+  snapSelection: types.array(types.string),
+  hiddenItems: types.array(types.string)
```
bfinzer, line 141: Given that we have "hidden" in the data display world, it would be more consistent to use "setAside" here.

bfinzer:
This is a pretty complex PR. I'd appreciate a synchronous review. Also, if you agree, let's change references to "hidden" to "setAside" and carry that terminology throughout.

bfinzer (changes requested):
… Pending synchronous review

v3/src/models/data/data-set.ts
kswenson: I did consider `setAside` and decided that I prefer `hidden` in this case. The `DataSet` is used in other applications besides CODAP (e.g. CLUE) and is intended to have a data-set-centric API, not a CODAP-centric one. From the standpoint of the `DataSet`, those items/cases are hidden, and I don't think there's any inconsistency with CODAP calling them `setAside` for CODAP reasons, while other applications may call it something else.

tealefristoe (approved):
Looks good 👍 

I left a few minor comments you can take or leave. I also noticed a bunch of code coverage warnings in the code, but I'm seeing the patch coverage at 100%, so I'm not sure what to make of that situation.

kswenson:

PR #1451, opened 2024-09-02
Fix split date axis bug

bfinzer:
[#188109429] Bug fix: Horizontal date axis partly empty when split on top

* Mostly fixed by making Axis an observer.
* But we also have to rerender sub-axes when the number of repetitions changes; e.g. by introducing a new category in an attribute appearing on the top or right.

### DIFF
--- a/v3/src/components/axis/hooks/use-sub-axis.ts
+++ b/v3/src/components/axis/hooks/use-sub-axis.ts
@@ -1,5 +1,5 @@
 import { BaseType, drag, select, Selection } from "d3"
-import { reaction } from "mobx"
+import { comparer, reaction } from "mobx"
 import { mstAutorun } from "../../../utilities/mst-autorun"
 import { mstReaction } from "../../../utilities/mst-reaction"
 import { useCallback, useEffect, useMemo, useRef } from "react"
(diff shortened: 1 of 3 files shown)


v3/src/components/axis/hooks/use-sub-axis.ts
```diff
@@ -331,11 +331,12 @@ export const useSubAxis = ({
...
   }, [dataConfig, updateDomainAndRenderSubAxis])
 
-  // update d3 scale and axis when layout/range changes
+  // Render when axis length or number of sub-axes changes
   useEffect(() => {
     const disposer = reaction(
       () => {
-        return layout.getAxisLength(axisPlace)
+        return [layout.getAxisLength(axisPlace),
+          layout.getAxisMultiScale(axisPlace)?.repetitions]
```
kswenson, line 339: When returning a non-primitive like an array or object, I would add `equals: comparer.structural` to the options below. There's a subtle argument about whether or not it's strictly necessary, but I think the code is clearer that way.

kswenson (approved):
👍 LGTM - one suggestion to add `equals: comparer.structural` when returning in array in a `reaction`.

PR #1452, opened 2024-09-04
188214433 Fix Update Attribute Name Requests

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/188214433

There was a bug in the Collaborative plugin where changing an attribute's name in a shared table was resulting in other users losing data. It turns out this is because `update attribute name` requests were not calling `dataset.setAttributeName`, which is necessary to keep the dataset's `attrNameMap` up to date.


kswenson (approved):
👍 LGTM

PR #1453, opened 2024-09-04
feat: Case card data editing (PT-188001430)

emcelroy:
[#188001430](https://www.pivotaltracker.com/story/show/188001430)


kswenson (changes requested):
👍 Looks good -- bunch of comments/suggestions, but they should be pretty straightforward to address. Also, I didn't double-check this myself, but make sure that the migration of things like `CollectionTitle` to common code didn't adversely affect the case table.

v3/src/components/case-card/case-card-model.ts
```diff
@@ -45,6 +46,46 @@ export const CaseCardModel = TileContentModel
...
+      // Not setting them to "" results in "undefined" being displayed in the case card UI's input fields.
+      collection?.attributes.forEach(attr => {
+        if (attr?.id) {
+          newCase[attr.id] = ""
+        }
+      })
+
+      const newCases = [newCase]
+      const lastExistingCaseId = cases[cases.length - 1].__id__
+      const newCaseId = self.data?.addCases(newCases, {after: lastExistingCaseId})[0]
```
kswenson, line 84: Forgot to mention this earlier, but this can be simplified to:
```suggestion
      const [newCaseId] = self.data?.addCases(newCases) ?? []
```
because adding at the end is the default.

v3/src/components/case-card/case-card.tsx
```diff
@@ -9,6 +9,7 @@ import { CardView } from "./card-view"
 // import { DGDataContext } from "../../models/v2/dg-data-context"
 
 import "./case-card.v2"
```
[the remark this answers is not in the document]
emcelroy: I don't think so. I'm not 100% sure we can drop it entirely at this point, but I re-commented it out for now.

v3/src/components/case-card/case-card.scss
```diff
@@ -1,334 +1,366 @@
...
-  position: relative;
+@import "../vars.scss";
+@import "./case-card-vars.scss";
+
+.codap-case-card {
+  overflow: auto hidden;
+}
+
+input.column-name-input,
+input.collection-title-input,
```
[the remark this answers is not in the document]
emcelroy: These are actually case-card-specific styles. Sorry, I should've placed them all within the `.codap-case-card` rule set, which I've done in the latest commit. I placed them at this level because the input elements rendered by various Case Card child components share these styles.

emcelroy:
> 👍 Looks good -- bunch of comments/suggestions, but they should be pretty straightforward to address. Also, I didn't double-check this myself, but make sure that the migration of things like `CollectionTitle` to common code didn't adversely affect the case table.

@kswenson As far as I can tell, the table elements that are now based on shared code look and behave as they do on the main branch.

v3/src/components/case-card/case-attrs-view.scss
```diff
@@ -1,9 +1,12 @@
...
 .case-card-attrs {
   border-collapse: collapse;
-  border-spacing: 8px;
   font-weight: 600;
-  margin: 5px 5px 5px 29px;
+  margin: 5px $margin-right 5px $margin-left;
   table-layout: auto;
   text-align: right;
-  width: calc(100% - 39px);
+  width: calc(100% - $xMarginOffset);
```
kswenson, line 11: Named constants are helpful. This was `39px` previously and now it appears to be `34px` -- is that intentional?

kswenson (approved):
👍 Changes look good

v3/src/components/case-card/case-attrs-view.scss
emcelroy: Yes, that was intentional. I didn't get the number of pixels quite right originally. Thanks for checking.

PR #1455, opened 2024-09-04
Bar for each point not applicable to dates

bfinzer:
[#188219800] Bug fix: Bar for each point problem with dates

* The observed bug of black bars for points occurs because we are inappropriately allowed to request a bar for each point when the primary attribute is of type "date". The bar would go from a zero date to the date value, but we don't work with dates of value zero. So the solution is to prevent the option from showing.


kswenson (approved):
👍 LGTM

PR #1456, opened 2024-09-04
Count axis should only show only integers

bfinzer:
[#188221613] Bug fix: A Count axis should only show integer values

* We introduce the `integersOnly` property for `NumericAxisModel` and refer to it when rendering a numeric axis

D3 makes it difficult to control the tick marks so we end up with fractional tick marks and gridlines for histograms. But we regard this as an acceptable difference from V2.


kswenson (approved):
👍 LGTM

PR #1458, opened 2024-09-05
fix: caseIds in itemInfo when remapping case ids

kswenson:
While working on #1453, @emcelroy and @lublagg encountered a bug in the persistent id code that failed to properly map the `childIds` associated with an item in the `itemInfoMap`. This PR fixes the bug and extends the jest tests so that they fail without the fix in place.


tealefristoe (approved):
Looks good 👍

PR #1459, opened 2024-09-05
chore: update collection documentation/comments and tweak implementation

kswenson:
- update documentation in `hierarchical-data.md`
- update some comments in the code
- add comment about tradeoffs in caching case id maps
- eliminate `prevCaseIdToIndexMap` is it isn't needed

### DIFF
--- a/v3/src/models/data/collection.ts
+++ b/v3/src/models/data/collection.ts
@@ -179,7 +179,6 @@ export const CollectionModel = V2Model
     if (!self.prevCaseIds) self.prevCaseIds = self.caseIds
     self.caseIds = []
 
-    // no need to maintain a prev copy
     self.caseIdToIndexMap.clear()
 
     if (!self.prevCaseIdToGroupKeyMap) self.prevCaseIdToGroupKeyMap = self.caseIdToGroupKeyMap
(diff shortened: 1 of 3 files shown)


v3/src/models/data/collection.ts
```diff
@@ -181,8 +179,8 @@ export const CollectionModel = V2Model
     if (!self.prevCaseIds) self.prevCaseIds = self.caseIds
     self.caseIds = []
 
-    if (!self.prevCaseIdToIndexMap) self.prevCaseIdToIndexMap = self.caseIdToIndexMap
-    self.caseIdToIndexMap = new Map<string, number>()
+    // no need to maintain a prev copy
```
tealefristoe, line 182: Seems like this comment is unnecessary?

tealefristoe (approved):
Looks good 👍

v3/src/models/data/collection.ts
kswenson: At the time I wrote it I thought people might wonder why all of the other properties needed to have prev copies maintained but not this one, but if you don't think it's necessary I'm happy to remove it.

PR #1460, opened 2024-09-05
Feat: drag attributes from case card to graph (PT-188001758)

lublagg:
[#188001758](https://www.pivotaltracker.com/story/show/188001758)


v3/src/components/case-table-card-common/attribute-header.tsx
```diff
@@ -70,6 +70,11 @@ export const AttributeHeader = observer(function ColumnHeader({
 
   useEffect(() => {
     onCloseMenuRef.current?.()
+    if (dragging) {
+      document.body.style.cursor = "grabbing"
+    } else {
+      document.body.style.cursor = "default"
+    }
```
kswenson, line 77: I would rather handle this with CSS as we've done elsewhere. For instance, if you assign the `grabbing` cursor style in the `drag-overlay.scss` it should kick in for the drag image.

kswenson (changes requested):
👍 Looks good -- just one suggestion about cursor management.

v3/src/components/case-table-card-common/attribute-header.tsx
emcelroy: Yes, that works -- thanks 👍

PR #1461, opened 2024-09-05
Fix excessive categorical axis rerending

bfinzer:
[#188227145] Bug fix: Excessive categorical axis repainting

* In `use-sub-axis.ts` `updateDomainAndRenderSubAxis` we detect that the existing displayed categories are the same as the data configuration's categorical values and, if so, bail without rendering


kswenson (approved):
👍 LGTM

PR #1462, opened 2024-09-05
Fix vertical axis too narrow on open bug

bfinzer:
[#188228770] Bug fix: On opening document categorical axis on y-axis is too narrow

* We were relying incorrectly on the ordinalScale's domain to have the correct list of categories when, in fact, we need to go to the dataConfiguration's categoryArrayForAttrRole.


kswenson (approved):
👍 LGTM

PR #1463, opened 2024-09-05
Fix bar chart all hidden bug

bfinzer:
[#188200067] Bug fix: **Bar Chart** Compresses to Top When Hiding Unselected Cases in Mammals Dataset

* The problem was that when all cases were hidden, the Count axis min and max got set to NaN. We fix by not allowing this to happen. On encountering NaN, we leave the min or max alone.
* With an intact Count axis, the bar chart is able to correctly compute the placement of the bars when the hidden cases are shown.


kswenson (approved):
👍 LGTM

PR #1464, opened 2024-09-06
chore: refactor applyCaseValueChanges for use with case card

kswenson:
With the recent case card work to support editing, there was some redundant code between the case table and the case card which incorporated an outdated assumption about how (not) persistent case ids were in certain situations. This PR refactors the code to correct the assumption and to move the shareable part of the code into the shared code folder.


kswenson:
> Looks good to me aside from the merge conflict 👍 And we'll just update `CaseAttrView` to use `applyCaseValueChanges` after this gets merged?

👍

PR #1465, opened 2024-09-06
Case tile (table/card) refactor; preparation for case card inspector

kswenson:
Terminology: The case table and case card are considered to be instances of case tiles, and so the awkwardly named `case-table-card-common` folder is renamed to `case-tile-common`, and similarly for other components/files with `table-card` in their name. Also, rather than starting with the assumption that all files/components live in the case table folder until they are proven to be useful to the case card, we proactively move all non-case-table-specific UI elements like menus and dialogs to the `case-tile-common` folder since they don't have any case-table-specific code and are potentially shareable with the case card whether or not there are immediate plans to do so. Specifically:
- rename `case-table-card-common` folder to `case-tile-common`
- rename `CaseTableCardTitleBar` to `CaseTileTitleBar`
- refactor `RulerMenuList` to use `StdMenuList`
- move `inspector-panel` folder from `case-table` to `case-tile-common` for sharing with case card
- move `IndexMenuList` and `ColorPicker` components to `case-tile-common`


emcelroy (approved):
Looks good 👍

PR #1466, opened 2024-09-06
Bar chart hide all/show all bug

bfinzer:
[#188200067] Bug fix: **Bar Chart** Compresses to Top When Hiding Unselected Cases in Mammals Dataset

* This is a second try at this bug. It turned out that in the presence of a legend attribute the previous fix was not sufficient. Now, in `updateDomainAndRenderSubAxis` if we determine that there is no attribute associated with `axisPlace`; i.e. we have a Count axis, we bail because there won't be any values to use for the domain.


kswenson (approved):
👍 LGTM

PR #1467, opened 2024-09-07
188235437 add google analytics logging

eireland:
Adds logging to Google Analytics.
Note that V2 only logged a fraction of the events that is logged to the Concord log server.


v3/src/lib/analytics.ts
```diff
@@ -0,0 +1,78 @@
...
+  "Create collection": AnalyticsCategory.TABLE,
+  "Toggle component": AnalyticsCategory.TABLE,
+
+  "sliderThumbDragged": AnalyticsCategory.SLIDER,
+
+  "Edited text component": AnalyticsCategory.TEXT,
+
+  "Logout": AnalyticsCategory.SESSION,
+  "Login": AnalyticsCategory.SESSION,
+}
```
tealefristoe, line 68: This strikes me as being quite frail. I can see a couple possible improvements:
1. Include the `AnalyticsCategory` with the call to `applyModelChange`. I think this is the cleanest option that I would definitely recommend.
2. Define constants for these strings, and use the constants both here and in the calls to `applyModelChange`. At least then the connection between the two places would be explicit, and you'd only have to change one string if the messages ever needed to change.

tealefristoe (changes requested):
Generally looks good, but I think the way the `AnalyticsCategory` is specified should change.

v3/src/lib/analytics.ts
kswenson: I'm inclined to agree with @tealefristoe that option 1 is best. Let's pass the analytics category into `applyModelChange` and skip the map altogether. Aside from the other advantages, that would make it simple to search for all of the places in the code that are associated with a particular category.

v3/src/lib/analytics.ts
```diff
@@ -0,0 +1,78 @@
...
+
+export const mockGA = {
+  gtag: (event: string, eventName: string, data: Record<string, any>) => {
+    console.group("Mock GA4 event payload:")
+    console.debug("Event:", event)
+    console.debug("Event Name:", eventName)
+    console.debug("Data:", JSON.stringify(data, null, 2))
+    console.groupEnd()
```
kswenson, line 76: Add eslint directives to prevent lint warnings if these are going to stay in the code.

kswenson (changes requested):
👍 Generally looks good, but I agree with @tealefristoe on the `eventCategoryMap` and have a few other minor questions/suggestions.

v3/src/lib/analytics.ts
```diff
@@ -0,0 +1,78 @@
...
+  DOCUMENT = "document", // Document interactions
+  PLUGIN = "plugin",     // Game interactions
+  MODEL = "model",    // Model interactions
+  PLOT = "plot",     // Plot interactions
+  SESSION = "session",  // Session events (log in/out)
+  COMPONENT = "component", // General component interactions
+  TABLE = "table",    // Table interactions
+  SLIDER = "slider",    // Slider interactions
+  TEXT = "text",     // Text interactions
+}
```
kswenson, line 13: One other note on this -- you don't need named constants here because TypeScript types can guard against typos. The way we've handled this elsewhere is:
```typescript
export const AnalyticsCategories = [
  "general",      // catch-all for events that we haven't categorized yet
  "data",         // Interactions with generated data
  ...
  "text"          // Text interactions
] as const
export type AnalyticsCategory = typeof AnalyticsCategories[number]
```

v3/src/lib/logger.ts
```diff
@@ -111,9 +122,12 @@ export class Logger {
   private formatAndSend(time: number, event: string, documentTitle: string, args?: Record<string, unknown>) {
     const event_value = JSON.stringify(args)
     const logMessage = this.createLogMessage(time, event, documentTitle, event_value, args)
+    console.log("in formatAndSend args:", args)
```
tealefristoe, line 125: Can this be removed now?

v3/src/lib/log-message.ts
```diff
@@ -1,7 +1,10 @@
 import { t } from "../utilities/translation/translate"
+import { AnalyticsCategory } from "./analytics"
 
 export type LoggableValue = string | number | boolean | undefined
-export type LoggableObject = Record<string, LoggableValue>
+export type LoggableObject = Record<string, LoggableValue> & {
+  category?: AnalyticsCategory;  // If the "category" key exists, it must be an AnalyticsCategory
+}
```
tealefristoe, line 7: I would have added `category` to `ILogMessage` rather than `LoggableObject`.

tealefristoe (approved):
As I mentioned in a comment, I think it would have been better to add `category` to `ILogMessage`, but I'm not sure if it's worth changing now, so I'm giving 👍

kswenson (approved):
👍 Looks good -- I added a commit which carried the `category` through to the logger. Just a couple of minor remaining comments.

PR #1470, opened 2024-09-09
fix: collection serialization; DataSet must call Collection.prepareSnapshot when appropriate

kswenson:
[[PT-188245707]](https://www.pivotaltracker.com/story/show/188245707)


bfinzer (approved):
👍🏻 LGTM
So, in the end a simple, oversight bug. Good to know.

PR #1471, opened 2024-09-10
Bug fix: make case card overflow scrollable. (PT-188200175)

lublagg:
[PT#188200175](https://www.pivotaltracker.com/story/show/188200175)

Allows the user to scroll within a smaller window instead of having to resize the container via the draggable handle to see all of the content.


emcelroy (changes requested):
This looks good 👍 There's a little bug though in that the horizontal scroll bar, when present, appears outside of the tile.

image

I played around with the style rules and one way to fix that is to swap the height definitions between `.case-card.react-data-card` in case-card.scss and `.case-card-content` in card-view.scss. That seems to fix the problem, though I didn't test it super extensively.

PR #1472, opened 2024-09-10
fix: drag attribute before/after itself

kswenson:
[[PT-188252932]](https://www.pivotaltracker.com/story/show/188252932)


tealefristoe (approved):
Looks good 👍

PR #1474, opened 2024-09-10
Bug fix for restore bar chart with hidden cases

bfinzer:
[#188254368] Bug fix: Dot chart not restoring properly with hidden cases

* The bug was caused in graph-controller.ts `syncAxisScalesWithModel` where a misguided attempt to set the categorical axis' domain was done using a category set obtained from the data configuration with `categorySetForAttrRole`. But this category set is obtained from the metadata with no regard for hidden cases. So the fix is to request the categoryArrayForAttrRole from the data configuration since this array is computed taking into hidden cases into account.


kswenson (approved):
👍 LGTM

PR #1476, opened 2024-09-11
fix: restore set-asides after parent value change [PT-188254408]

kswenson:
[[PT-188254408]](https://www.pivotaltracker.com/story/show/188254408)

fix: remove stale entries from group key to case id map

Note: `useRows` change turned out to be unrelated, but seemed worth fixing.


tealefristoe (approved):
Looks good, thanks for getting to this so quickly 👍

PR #1477, opened 2024-09-11
Bug fix: case card overflow is scrollable. (PT-188200175)

lublagg:
Created a new branch off of latest from main instead of fixing rebase issue in other branch.


kswenson (approved):
👍 LGTM

PR #1478, opened 2024-09-12
fix: adding items/cases with set-asides [PT-188263826]

kswenson:
[[PT-188263826]](https://www.pivotaltracker.com/story/show/188263826)

There were a few places where `self.items.length` was being used synonymously with `self._itemIds.length`, which was a reasonable assumption to make before set-asides were implemented, but now one has to be careful about what you mean. `self.items.length` is the number of user-visible items, while `self._itemIds.length` is the total number of items before visibility is taken into account. When set-asides were implemented I searched for instances of `self.itemIds` and converted them to `self._itemIds` where appropriate, but I didn't specifically search for `self.items`. Now I have.

Also, did a little bit of `case` => `item` renaming in code I encountered while debugging.

### DIFF
--- a/v3/src/models/data/data-set.ts
+++ b/v3/src/models/data/data-set.ts
@@ -702,10 +702,10 @@ export const DataSet = V2Model.named("DataSet").props({
       },
       getValueAtItemIndex(index: number, attributeID: string) {
         if (self.isCaching()) {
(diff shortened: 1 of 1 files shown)


tealefristoe (approved):
Looks great 👍

PR #1479, opened 2024-09-12
feat: focus style for attribute buttons (PT-181869693)

emcelroy:
[#181869693](https://www.pivotaltracker.com/story/show/181869693)

Ensures that the attribute header/name buttons in the case card get a focus style similar to the case table's headers.


kswenson (approved):
👍 LGTM

PR #1480, opened 2024-09-12
fix: add case button default/hover styling

emcelroy:
[#181869597](https://www.pivotaltracker.com/story/show/181869597)

Previously, the Add Case and Add Attribute buttons had a default opacity of `.33` which changed to `1` on hover. Since the Add Case button was positioned over a white background, and the Add Attribute buttons were placed on shaded backgrounds, the Add Case button would appear lighter than the Add Attribute buttons and so could've potentially been mistaken for a disabled button.

These changes would drop the use of opacity and instead use solid colors for both types of buttons in their default and hover states. The visual effect is essentially the same except now they appear more similar in their default state.


kswenson (approved):
👍 Looks good -- nice use of `@mixin`

PR #1481, opened 2024-09-12
Fix flaky webpage display text

nstclair-cc:
Hey @tealefristoe,
This test went flaky on us, so I’ve increased the wait time slightly in hopes of stabilizing it. I explored using a visit statement here, but with the iFrame component involved, that doesn’t seem doable. Fingers crossed this adjustment does the trick!

Let me know if you have any thoughts or suggestions.

Cheers,
Talya


tealefristoe (approved):
Looks good, thanks for fixing that 👍

PR #1482, opened 2024-09-12
feat: click on case table background deselects cases [PT-187381671]

kswenson:
[[PT-187381671]](https://www.pivotaltracker.com/story/show/187381671)


tealefristoe (approved):
Looks good 👍

PR #1484, opened 2024-09-13
Fixes bug where user can drag a component under the toolshelf

eireland:
Previously, you can drag a component beyond the `codap-container` and can end up with the component header under the toolshelf. Once there, user can't drag the component back down.
This restricts dragging of component to below the toolshelf. User can still drag left, right and below viewable area.


kswenson (approved):
👍 LGTM

PR #1485, opened 2024-09-13
Adds ability to rerandomize all from table inspector

eireland:


kswenson (changes requested):
👍 Looks good, but should only be enabled when the attribute has a randomizable formula, as with v2.

kswenson (approved):
👍 LGTM

PR #1487, opened 2024-09-13
Captures pointer when dragging resize handle

eireland:
From this [story](https://www.pivotaltracker.com/story/show/187774818):
When resizing webviews with the resize handle, sometimes the pointer loses the resize handle focus causing it to stop resizing.
This forces the resizers to capture the pointer focus.


kswenson (approved):
👍 LGTM

kswenson (commented):
Just noticed that there's a failing cypress test that appears related, so that should be fixed before merging.

PR #1488, opened 2024-09-13
address flaky Cypress test for text tiles

nstclair-cc:
Hey @kswenson - I think the solution for the flaky test for now is to comment out the offending code. I manual tested and things look okay for now, but let me know if more steps are needed or if you see something more obvious. Feel free to merge if you agree.


kswenson (approved):
👍 We test this behavior in other tests so I don't think there's much value in leaving it commented out, so I'd recommend removing it entirely rather than leaving it commented out.

PR #1489, opened 2024-09-13
Quiets MockGA logs if debug_logger is off

eireland:


kswenson (approved):
👍 LGTM -- there are some stray semicolons to remove

PR #1490, opened 2024-09-13
Adds functionality to Help toolshelf button

eireland:
Adds Help toolshelf menu list.
Opens Help page, forum page, project page, and privacy page.

### DIFF
--- a/v3/src/components/tool-shelf/help-button.tsx
+++ b/v3/src/components/tool-shelf/help-button.tsx
@@ -8,14 +8,14 @@ import { showWebView } from "./tool-shelf-utilities"
 
 import "./tool-shelf.scss"
 
-const showHelpURL = 'https://codap.concord.org/help'
-const showHelpURL_ja = 'https://codap.concord.org/resources/latest/help-documents/CODAP解説書.pdf'
-const showHelpForumURL = 'https://codap.concord.org/forums/forum/test/'
-const showWebSiteURL = 'https://codap.concord.org'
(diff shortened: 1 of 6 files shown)


v3/src/components/tool-shelf/help-button.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+    if (url !== undefined) {
+      documentContent?.applyModelChange(() => {
+        const tile = documentContent?.createOrShowTile?.(kWebViewTileType)
+        isWebViewModel(tile?.content) && tile?.content.setUrl(`${url}`)
+      }, {
+        undoStringKey: "V3.Undo.webView.show",
+        redoStringKey: "V3.Redo.webView.show",
+        log: logMessageWithReplacement("Show web view: %@", {url}, "document")
+      })
```
tealefristoe, line 38: This code is duplicated from lines 27-34 of `options-button.tsx`. Instead of duplicating code, you should make a utility function to do this, and then call that utility function here and in `options-button.tsx`.

tealefristoe (approved):
Not sure if you wanted a review for this or not, since you removed me as a reviewer and the PR is still in draft, but here it is either way 😁 

This looks good to me. I left one suggestion about creating a utility function to reduce duplicate code. I also think it would be worth expanding the test a bit to click on the various options, which would probably get you pretty close to 100% patch coverage.

eireland:
Thanks @tealefristoe!

PR #1491, opened 2024-09-14
fix: redo string (copy paste error); skip failing case card cypress test

kswenson:


kswenson:
Merging without review in the interest of expediency as the changes are minor.

PR #1492, opened 2024-09-14
Add blank title space to graph component when there is no data context present

eireland:
Adds an underscore to component title when user hovers over component title bar and there is no title. This is true for graphs when there is no data set in the document.
Note that if user deletes the title, in some instances the component will show the previous title or the data set name.

### DIFF
--- a/v3/src/components/component-title-bar.tsx
+++ b/v3/src/components/component-title-bar.tsx
@@ -1,7 +1,7 @@
 import { Button, CloseButton, Flex, Input } from "@chakra-ui/react"
 import { clsx } from "clsx"
 import { observer } from "mobx-react-lite"
-import React, { useState } from "react"
+import React, { useRef, useState } from "react"
 import { IUseDraggableTile, useDraggableTile } from "../hooks/use-drag-drop"
 import { getTitle } from "../models/tiles/tile-content-info"
 import { uiState } from "../models/ui-state"
@@ -16,9 +16,9 @@ export const ComponentTitleBar = observer(function ComponentTitleBar({
   tile, children, onHandleTitleChange, onMinimizeTile, onCloseTile, preventTitleChange
 }: ITileTitleBarProps) {
   // perform all title-related model access here so only title is re-rendered when properties change
-  const title = (tile && getTitle?.(tile)) || tile?.title || ""
+  const initialTitle = (tile && getTitle?.(tile)) || tile?.title || ""
   const [isEditing, setIsEditing] = useState(false)
-  const [editingTitle, setEditingTitle] = useState(title)
+  const [editingTitle, setEditingTitle] = useState(initialTitle)
   const tileId = tile?.id || ""
   const tileType = tile?.content.type
   const draggableOptions: IUseDraggableTile = { prefix: tileType || "tile", tileId, disabled: isEditing }
@@ -26,9 +26,10 @@ export const ComponentTitleBar = observer(function ComponentTitleBar({
   const classes = clsx("component-title-bar", `${tileType}-title-bar`, {focusTile: uiState.isFocusedTile(tile?.id)})
   const [isHovering, setIsHovering] = useState(false)
   const blankTitle = "_____"
+  const prevTitleRef = useRef(initialTitle)
 
   const handleChangeTitle = (nextValue?: string) => {
-    if (tile != null && nextValue) {
+    if (tile != null && nextValue !== undefined) {
       tile.applyModelChange(() => {
         tile.setTitle(nextValue)
       }, {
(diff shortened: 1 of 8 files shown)


v3/src/components/component-title-bar.tsx
```diff
@@ -86,7 +93,8 @@ export const ComponentTitleBar = observer(function ComponentTitleBar({
               onChange={(e) => setEditingTitle(e.target.value)} onBlur={() => handleSubmit(editingTitle)}
               onFocus={(e) => e.target.select()} onKeyDown={handleInputKeyDown}
             />
-          : <div className="title-text" data-testid="title-text" onClick={handleTitleClick}>{title}</div>
+          : title ? <div className="title-text" data-testid="title-text" onClick={handleTitleClick}>{title}</div>
+                  : <div className="title-text" data-testid="title-text" onClick={handleBlankTitleClick}>_____</div>
```
tealefristoe, line 97:
```suggestion
          : <div className="title-text" data-testid="title-text" onClick={handleTitleClick}>{title || "_____"}</div>
```

tealefristoe (changes requested):
As implemented, this will set the title to "_____" if you change the title to "". I have to imagine that the goal of this story is to display "_____", but never actually set the title to that. I think it should be possible to accomplish this by changing the one line I suggested in a specific comment and removing all other changes you made in `component-title-bar.tsx`.

One other thing: it would be better to create a constant for "_____" instead of hard coding this string in multiple locations. It's a very easy string to type incorrectly (is that 5 or 6 "_"?), so you'll be saving your future self/another engineer some headaches if you standardize it.

tealefristoe (changes requested):
Looking more closely at the PT story and behavior in v2, I now see that the intention is for the "\_\_\_\_\_" to only show up when the user hovers over the title bar, just like the close and minimize buttons. So that adds a wrinkle to this that will need to be addressed.

But this behavior can be implemented without modifying any code that sets the title. The "\_\_\_\_\_" should show up in the `div.title-text`, but it should never show up in the `Input`. So all you need is the change to line 89, and you should remove all other current changes to `component-title-bar.tsx`. You will have to figure out how to determine if the user is hovering over the title bar or not, which will complicate things a bit, but you shouldn't need to modify `handleChangeTitle` or `handleSubmit` at all.

eireland:
Good catch! I forgot it only come up on hover.

tealefristoe (approved):
I ended up submitting some changes. Here's what I did:
- Removed a number of unnecessary lines of code and simplified others.
- Moved the `text-decoration` styling into css.
- Fixed the cypress test, which was failing after the change to make the underline only appear when hovering.
  - I was hoping to use a constant both here and in the component, but I couldn't figure out how to import a `const` from the codebase into a cypress test. Do you know of a way to do this?
  
I tested these changes and think they work, but you should give them a try yourself before merging.

I'll also mention that there are still some discrepancies between v3 and v2 behavior here. For example, in v3 you can't change a tile's title to be empty, but you can in v2. Also, making a tile's title a single space (" ") in v3 seems to make it impossible to change the title. I'm not sure if these issues are within the scope of this story so I didn't try to fix them.

kswenson (approved):
👍 LGTM

eireland:
@tealefristoe I fixed the bugs you found. 
You can now change the title to an empty string "". The falsy of `nextValue` was making it not go through with the `applyModelChange`
I now trim the entered string so typing a space(s) results in a blank title.
I also found a bug where if you have a data context and you delete the title, the title reverts to the data context name. V2 allowed a user to delete a title even with a data context. This is now also fixed here.

tealefristoe (changes requested):
Nice catch on the falsy value of `nextValue` 👍 

I'll note that v2 allowed the user to set the title of a component to " ", which the new trimming strategy doesn't allow. I don't know if the v2 behavior is desirable or not.

It looks like a lot of tests are failing with these changes. I'm guessing a lot of this has to do with using `prevTitleRef`, which is causing the component to rerender less often than it was before, though I haven't looked into it carefully. It also seems like the use of `initialTitle`, `editingTitle`, and `prevTitleRef` is a bit confused in this PR, and I wonder if it would be better to revert a lot of those changes. If you're having trouble getting the tests passed, I'd recommend just punting on some of these extra fixes, since they weren't actually a part of the original PT story.

eireland:
@tealefristoe I had to revert back some of the changes I made to make it work for deleting a component title. Now if there is a data set present and you delete the title, the data set name appears as the title.

tealefristoe (approved):
Looks good 👍 

I made another commit with two changes. I tested things and they seemed to work, so hopefully I didn't break anything.
- I simplified the logic in `case-table-registration.ts`. This is the equivalent of what you had. I'll note that the order here makes the dataset's title trump the tile's title... I'm not sure if that's actually the priority you want.
- I removed the `ref` from `component-title-bar.tsx`. I don't think this was doing anything, and it was complicating the code quite a bit. If I'm mistaken and this was doing something, please let me know what it was doing.

I think it was the right move to not worry about all of the title issues in this one PR. It's already been a long one and covers more than the original PT story requested.

PR #1493, opened 2024-09-17
Adds language menu UI to CFM bar

eireland:
This adds the UI for language menu to the CFM bar. The UI is currently non-functional.

Note that in testing this, if I run the `makeNewLocaleUrl` function manually, CODAP does reload with the new language correctly. But when I select a language from the dropdown, nothing happens.

### DIFF
--- a/v3/cypress/e2e/cfm.spec.ts
+++ b/v3/cypress/e2e/cfm.spec.ts
@@ -92,7 +92,7 @@ context("CloudFileManager", () => {
     // Verify document was closed (Map data table doesn't exist)
     c.checkComponentDoesNotExist("table")
   })
-  it("verify langugae menu is present", () => {
+  it("verify language menu is present", () => {
     cfm.getLanguageMenuButton().should("exist")
     cfm.getLanguageMenu().should("not.exist")
     cfm.getLanguageMenuButton().click()
--- a/v3/cypress/e2e/cfm.spec.ts
+++ b/v3/cypress/e2e/cfm.spec.ts
@@ -92,4 +92,12 @@ context("CloudFileManager", () => {
     // Verify document was closed (Map data table doesn't exist)
     c.checkComponentDoesNotExist("table")
   })
+  it("verify langugae menu is present", () => {
+    cfm.getLanguageMenuButton().should("exist")
+    cfm.getLanguageMenu().should("not.exist")
+    cfm.getLanguageMenuButton().click()
+    cfm.getLanguageMenu().should("exist")
+    cfm.getLanguageMenuButton().click()
+    cfm.getLanguageMenu().should("not.exist")
+  })
 })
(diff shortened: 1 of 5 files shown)


v3/src/lib/use-cloud-file-manager.ts
```diff
@@ -9,6 +9,128 @@ import { appState } from "../models/app-state"
...
+    langName: '简体中文',
+    langDigraph: 'zh-Hans',
+    countryDigraph: 'Hans',
+    icon: 'flag flag-cn'
+  }
+]
+function makeNewLocaleUrl(digraph: string, locationHref: string) {
+  const location = new URL(locationHref)
+  const locMatch = location.pathname.match(/(^.*\/static\/dg\/)[^/]+(\/cert\/.*$)/)
+  let baseURL = "https://codap3.concord.org/branch/main/"
```
tealefristoe, line 115: I assume this `baseURL` basically never comes up, but do we want it to be a branch link? Should it just be `https://codap3.concord.org`?

tealefristoe (approved):
Looks good 👍 

Would `window.location.replace()` redirect to the new url properly? https://developer.mozilla.org/en-US/docs/Web/API/Location/replace

Also, it would be good to increase code coverage before merging. It should be possible to test `makeNewLocaleUrl` in a jest test to make sure it correctly constructs a new url, even if you can't test the actual redirection.

v3/cypress/e2e/cfm.spec.ts
```diff
@@ -92,4 +92,12 @@ context("CloudFileManager", () => {
     // Verify document was closed (Map data table doesn't exist)
     c.checkComponentDoesNotExist("table")
   })
+  it("verify langugae menu is present", () => {
```
kswenson, line 95:
```suggestion
  it("verify language menu is present", () => {
```

v3/src/lib/use-cloud-file-manager.ts
```diff
@@ -19,22 +141,23 @@ export function useCloudFileManager(optionsArg: CFMAppOptions) {
...
+            options: locales.map(function (locale) {
+              return {
+                label: locale.langName,
+                langCode: locale.langDigraph,
+              }
+            }),
+          },
+          onLangChanged: (langCode: string) => {
+            Logger.log(`Changed language: ${langCode}`)
+            window.location.href = makeNewLocaleUrl(langCode, window.location.href)
```
kswenson, line 157: The need to reload the page when switching languages in v2 was an artifact of SproutCore's approach to localization. In v3 we're just going to change some state and force everything to re-render. You can just remove the `makeNewLocaleUrl` function and make this PR UI only for now.

kswenson (changes requested):
👍 Menu code looks fine. Don't worry about the url-switching code -- none of that should be necessary in v3. See comment elsewhere for details. I would just remove it and make this PR UI-only.

kswenson (approved):
👍 Looks good -- one other thing I just noticed which you might want to address is that in v2 the current language isn't listed in the menu, whereas in this PR it is listed. I could see an argument for listing it, but if it is listed it should probably have a checkmark indicating that it's already the currently selected option, so removing it might be simpler. This has the advantage of not tempting users to "switch" to the current language.

PR #1494, opened 2024-09-17
Fix parent visibility toggles

bfinzer:
[#188234714] Bug fix: Parent visibility toggles not working

* There are several things going on here, the first of which has to do with the fact that what was being stored in each toggle button was the *item* ids of the cases associated with that button rather than the *case* id.
* The second of which is that when we render a numeric axis we're unnecessarily removing the current instance of it and rerendering from scratch. We fix that by not removing the axis.
* The third of which is that we rerendering even if the numeric axis domain is not changing. So we detect that condition and don't rerender.
* The fourth of which is that in one place we were incorrectly calling `setNumericDomain` with the complete array of numeric values instead of an array with just the min and max.

### DIFF
--- a/v3/src/components/graph/components/parent-toggles.tsx
+++ b/v3/src/components/graph/components/parent-toggles.tsx
@@ -82,7 +82,7 @@ export const ParentToggles = observer(function ParentToggles() {
   const collectionIndexForPrimaryAttribute = dataset?.getCollectionIndexForAttribute(primaryAttribute) ?? 0
   const isCollectionSet = !!(dataset && dataset.collections?.length > 0)
   const hiddenCases = dataConfig?.hiddenCases ?? []
-  const itemIDs = dataset?.items.map((c) => c.__id__) ?? []
+  const itemIDs = dataset?.itemIds ?? []
   const caseButtons = createCaseButtons(
     { itemIDs, dataset, collectionIndexForPrimaryAttribute, isCollectionSet, hiddenCases })
   const caseButtonsListWidth = caseButtons.reduce((acc, button) => acc + button.width + TEXT_OFFSET, 0)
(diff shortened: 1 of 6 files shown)


v3/src/components/graph/components/parent-toggles.tsx
```diff
@@ -74,11 +77,14 @@ export const ParentToggles = observer(function ParentToggles() {
...
   const graphModel = useGraphContentModelContext()
   const dataConfig = useGraphDataConfigurationContext()
+  const primaryAttribute = dataConfig?.primaryAttributeID ?? ""
   const dataset = dataConfig?.dataset
+  const collectionIndexForPrimaryAttribute = dataset?.getCollectionIndexForAttribute(primaryAttribute) ?? 0
   const isCollectionSet = !!(dataset && dataset.collections?.length > 0)
   const hiddenCases = dataConfig?.hiddenCases ?? []
-  const caseIDs = dataset?.items.map((c) => c.__id__) ?? []
-  const caseButtons = createCaseButtons({ caseIDs, dataset, isCollectionSet, hiddenCases })
+  const itemIDs = dataset?.items.map((c) => c.__id__) ?? []
```
kswenson, line 85: The DataSet has an `itemIds` method, so you should just be able to use:
```suggestion
  const itemIDs = dataset?.itemIds ?? []
```

v3/src/models/data/data-set.ts
```diff
@@ -382,6 +382,11 @@ export const DataSet = V2Model.named("DataSet").props({
     return self.collections.find(coll => coll.getAttribute(attributeId))
   }
 
+  function getCollectionIndexForAttribute(attributeId: string): number | undefined {
+    const id = (self.collections.find(coll => coll.getAttribute(attributeId)))?.id
```
kswenson, line 386:
```suggestion
    const id = getCollectionForAttribute(attributeId)?.id
```

kswenson (approved):
👍 LGTM -- just a couple of suggested code simplifications

PR #1495, opened 2024-09-18
188276295 deselect on whitespace click

eireland:
When the table is in focus, and there are selected cases, a click in the whitespace deselects all the cases.
When the table is not in focus, and there are selected cases, a click in the whitespace focuses the table, but does not deselect the cases.


kswenson (approved):
👍 Looks good -- clever approach to determining when click is outside the cells. I added a commit which moves much of the functionality to a `useWhiteSpaceClick` hook and shares it with the `CollectionTableSpacer` so that now both of them have the same behavior.

PR #1496, opened 2024-09-18
Fix rendering of numeric axis

bfinzer:
[#188300061] Bug fix: Undo of add numeric attribute to axis is broken

This bug occurred as a result of PR that reduced the rerendering of numeric axes, but a little too much so that it broke some things that showed up in cypress tests

* When rendering an empty axis, we first remove whatever is already present
* In `use-sub-axis` the `installDomainSync` was doing an mstAutorun and assuming that if the axis domain and multi-sync domain were the same there was no need to rerender. But this is false since undo/redo can bring about this condition but we still need to rerender.

### DIFF
--- a/v3/src/components/axis/helper-models/axis-helper.ts
+++ b/v3/src/components/axis/helper-models/axis-helper.ts
@@ -94,6 +94,7 @@ export class AxisHelper {
 export class EmptyAxisHelper extends AxisHelper {
 
   render() {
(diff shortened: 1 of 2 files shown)


tealefristoe (commented):
Looks good! Glad cypress tests did their job 👍

tealefristoe (approved):
Meant to approve there.

PR #1497, opened 2024-09-19
188104291 Handle update attribute deleteable API requests.

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/188104291

Previously, the `update attribute` handler would ignore any changes to `deleteable`. This PR fixes that, which is necessary to return unshared tables to full user control in Collaborative.


bfinzer (approved):
👍🏻LGTM

PR #1499, opened 2024-09-19
Fixes bug that disallows attributes to be drag above their containers

eireland:


eireland:
Created a new story for preventing the drag into the toolshelf here https://www.pivotaltracker.com/story/show/188312264

tealefristoe (approved):
Looks good 👍

PR #1500, opened 2024-09-19
No rescale on hide points in graph

bfinzer:
[#188293477] Bug fix: When hiding cases, numeric axes should not rescale

Actually, in **most** situations a user action should not zoom in; i.e. shrink the scale. But there are some where we do. So it's a bit finicky.

* Add `allowRangeToShrink` to `BaseNumericAxisModel` as a flag that clients can set. Then use the flag in `setDomain` to adjust the requested domain.
* Took the opportunity to add undo/redo strings for various parent visibility toggle actions

### DIFF
--- a/v3/src/utilities/translation/lang/en-US.json5
+++ b/v3/src/utilities/translation/lang/en-US.json5
@@ -8,6 +8,17 @@
     "V3.general.yes": "Yes",
     "V3.general.no": "No",
 
+    "V3.Undo.graph.toggleParentVisibility": "Undo toggling visibility",
+    "V3.Redo.graph.toggleParentVisibility": "Redo toggling visibility",
+    "V3.Undo.graph.hideAllCases": "Undo hiding all cases",
+    "V3.Redo.graph.hideAllCases": "Redo hiding all cases",
+    "V3.Undo.graph.showAllCases": "Undo showing all cases",
+    "V3.Redo.graph.showAllCases": "Redo showing all cases",
+    "V3.Undo.graph.showLastParentOnly": "Undo showing only last parent case",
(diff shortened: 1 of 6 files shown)


v3/src/utilities/translation/lang/en-US.json5
```diff
@@ -8,6 +8,17 @@
...
+    "V3.Undo.graph.toggleParentVisibility": "Undo toggling visibility",
+    "V3.Redo.graph.toggleParentVisibility": "Redo toggling visibility",
+    "V3.Undo.graph.hideAllCases": "Undo hiding all cases",
+    "V3.Redo.graph.hideAllCases": "Redo hiding all cases",
+    "V3.Undo.graph.showAllCases": "Undo showing all cases",
+    "V3.Redo.graph.showAllCases": "Redo showing all cases",
+    "V3.Undo.graph.showLastParentOnly": "Undo showing only last parent case",
+    "V3.Redo.graph.showLastParentOnly": "Redo showing only last parent case",
+    "V3.Undo.graph.uncheckLastParentOnly": "Undo unchecking \"Last\" checkbox",
+    "V3.Redo.graph.uncheckLastParentOnly": "Redo unchecking \"Last\" checkbox",
```
kswenson, line 20: We don't have v2 strings for any of these? 🤔

kswenson (approved):
👍 LGTM -- just one question about the new strings.

v3/src/utilities/translation/lang/en-US.json5
bfinzer: Nope. In V2 these actions are not even undoable.

PR #1501, opened 2024-09-20
187573192 units in graph map attrs

eireland:
Adds units to graph axis label if attribute has units.

### DIFF
--- a/v3/src/components/graph/components/graph-attribute-label.tsx
+++ b/v3/src/components/graph/components/graph-attribute-label.tsx
@@ -65,8 +65,10 @@ export const GraphAttributeLabel =
       if (graphModel.pointsFusedIntoBars && place === secondaryPlace) {
         return t("DG.CountAxisView.countLabel")
       }
-      return attrIDs.map(anID => dataset?.attrFromID(anID)?.name)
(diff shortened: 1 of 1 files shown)


bfinzer (approved):
👍🏻 LGTM

v3/src/components/graph/components/graph-attribute-label.tsx
```diff
@@ -65,8 +65,10 @@ export const GraphAttributeLabel =
       if (graphModel.pointsFusedIntoBars && place === secondaryPlace) {
         return t("DG.CountAxisView.countLabel")
       }
-      return attrIDs.map(anID => dataset?.attrFromID(anID)?.name)
-        .filter(aName => aName !== '').join(', ')
+      return attrIDs.map(anID => dataset?.attrFromID(anID))
+                    .filter(attr => attr?.name !== '')
+                    .map(attr => `${attr?.name} ${attr?.units ? `(${attr?.units})` : ""}`.trim())
```
kswenson, line 70: In the case of multiple attributes without units, I think this will result in extraneous spaces. I think it should be:
```typescript
                    .map(attr => `${attr?.name}${attr?.units ? ` (${attr?.units})` : ""}`.trim())
```

PR #1502, opened 2024-09-20
Fixes missing webview inspector panel in Safari.

eireland:
The viewbox and scale were causing a problem with displaying the icon in Safari.


bfinzer (approved):
👍🏻 LGTM

PR #1504, opened 2024-09-20
Fixes deselection when clicking in whitespace in new table

eireland:
There was a problem with the whitespace deselection click with new tables.
The whitespace click only deselects when table is already in focus. When we created new tables, it didn't know that the table tile was already in focus on open. So you would have to click twice to deselect.
This should fix that.


tealefristoe (approved):
Looks great 👍

PR #1505, opened 2024-09-20
Support unicode operators

pjanik:
https://www.pivotaltracker.com/story/show/188274494

MathJS doesn't support Unicode operators by default, and it doesn't appear possible to modify it to do so without changing the parser. Since we are dealing with single Unicode characters, this approach seems good enough.


tealefristoe (approved):
Looks good 👍

PR #1506, opened 2024-09-20
Fix numeric axis crash

bfinzer:
[#188305480] Bug fix: Crash on undo of drag attribute to graph

* Prior PR limited the amount of redrawing we do for numeric axes by not erasing what we find for the axis. But we went too far. If we don't already have a numeric axis, we *do* need to remove what is there or, otherwise, we crash. So this PR makes it possible in `NumericAxisHelper:render` to tell whether we rerendering as opposed to starting fresh. If the latter, remove what is already present.


kswenson (approved):
👍 LGTM

PR #1507, opened 2024-09-20
Date Axis Model needs proper initialization

bfinzer:
[#188312576] Bug fix: Date axis not working properly

* Prior PR for not "shrinking" numeric axes unnecessarily cause date axis model to get spurious min and max. In `graph-model-utils.ts` `setAllowRangeToShrink(true)` was being called for a new numeric axis model, but not for a new date axis model. So, we fix that.


kswenson (approved):
👍 LGTM

PR #1508, opened 2024-09-23
feat: case card summary view (PT-188001359)

emcelroy:
[#188001359](https://www.pivotaltracker.com/story/show/188001359)

Adds a summary view for the case card. Note that the summary view design led to a change in the overall card design. Now all units are displayed after the case values instead of immediately after the attribute names.

**Summary View Toggle Button**

* Card now has a button at the bottom that allows users to either Browse Individual Cases or to Summarize Dataset
* Button is always in view even when scrolling
* Button text initially reads "Browse Individual Cases". When pressed, the individual case in the lowest child collection in the first row of the dataset is selected.
* When in the individual case view, the button reads "Summarize Dataset". When pressed, all cases are deselected and all collections are shown in summary view.

**Summary View Values**

When in the summary view, fields display a summary of their values.

* For categorical data
    * If there is a single unique value/category display this value
    * If there are two unique values/categories, display each separated by a comma
    * If there are 3 or more values/categories, display # followed by "values".
* For numerical data, display the range followed by the units.

**Paging Through Cases**

Text between the navigation arrows displays the total number of selected cases at each hierarchical level. If nothing is selected (i.e. in summary view), it shows the total number of cases at each level. If one case is selected, then it shows the "x of y" feedback for browsing cases.

Paging through selects the case represented by that card and all child cases linked below that parent case/card.

When paging through parent cases from summary view, child collections should show in summary view because they will be part of a sub-selection that occurs when paging through parent cases in the card view.


v3/src/components/case-card/card-view.tsx
```diff
@@ -16,16 +17,31 @@ export const CardView = observer(function CardView({onNewCollectionDrop}: CardVi
...
+  }
+
+  // The first time the card is rendered, summarize all collections unless there is a selection.
+  useEffect(function startWithAllCollectionsSummarized() {
+    if (data?.selection.size === 0) {
+      const allCollections = data?.collections.map(c => c.id) ?? []
+      cardModel?.setSummarizedCollections(allCollections)
+    }
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [])
```
emcelroy, line 43: I'm not sure this is the best way to do this. I had hoped to set the summarized collections list when the card instance is first created, but couldn't figure out how. I explored using `afterCreate` or `afterAttach` in the model, but `self.data` would always return `undefined`, so the collection information wasn't available at that point. I also looked at `createOrShowTableOrCardForDataset` in case-tile-utils.ts but couldn't get it to work there. Am I missing something regarding those options, or perhaps overlooking another alternative?

emcelroy:
> 👍 Generally looks good -- a couple of inline suggestions and one bug. The bug is that when clicking the `Browse Individual Cases` and `Summarize Dataset` button there is a vertical shift that suggests a change in cell height between the two settings.

@kswenson Thanks. I fixed the vertical shift bug. Regarding your other comment about using a MobX autorun or reaction, I think I knew that but forgot. Sorry. I've changed it to use `mstAutorun`. Is that still the preferred method? Or should I actually use MobX's `autorun` instead?

v3/src/components/case-card/card-view.tsx
```diff
@@ -16,16 +18,34 @@ export const CardView = observer(function CardView({onNewCollectionDrop}: CardVi
...
+  useEffect(function startWithAllCollectionsSummarized() {
+    return mstAutorun(
+      () => {
+        if (data?.selection.size === 0) {
+          const allCollections = data?.collections.map(c => c.id) ?? []
+          cardModel?.setSummarizedCollections(allCollections)
+        }
+      }, { name: "CardView.startWithAllCollectionsSummarized" }, cardModel)
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [])
```
kswenson, line 47:
```suggestion
  }, [cardModel, data])
```

kswenson (changes requested):
👍 Changes generally look good -- just one remaining suggestion about `useEffect` dependencies.

>I've changed it to use `mstAutorun`. Is that still the preferred method? Or should I actually use MobX's `autorun` instead?

When I suggested using `autorun` or `reaction` I meant it generically. The rule is to use `mstAutorun`/`mstReaction` when there is a reference to an MST model and the MobX versions when there isn't. As such, your use of `mstAutorun` is correct given the `cardModel` reference it contains.

However, using an autorun here will mean that the function will be triggered when any referenced observable property changes, for instance whenever the collection hierarchy is changed. If you want it to only trigger on initialization as the comment suggests, then an `mstReaction` specifically on the data set may be more appropriate. 

As I write all of that, it occurs to me that maybe we don't need an autorun or reaction at all. If the goal is really just to perform one-time initialization, then it should be sufficient to add the suggested `useEffect` dependencies so that when the data set changes the effect runs and then you don't have to worry about the reaction triggering subsequently.

emcelroy:
> As I write all of that, it occurs to me that maybe we don't need an autorun or reaction at all. If the goal is really just to perform one-time initialization, then it should be sufficient to add the suggested `useEffect` dependencies so that when the data set changes the effect runs and then you don't have to worry about the reaction triggering subsequently.

@kswenson Thanks 👍 That appears to work well and I've committed the suggestion.

I made another small change that fixes a bug I just noticed was introduced by this PR. The dividers between attributes on the card were being placed too far down. Simply removing `kCardCellHeight` from `getDividerBounds` in `CaseAttrsView` fixed that in both summary and single case views in cards with a hierarchy and without.

kswenson (approved):
👍 Changes look good

PR #1509, opened 2024-09-24
Adds units to legend attribute label

eireland:
Also fixes a PR suggestion re where to add the spacing between label and units for graph axis label.


kswenson (approved):
👍 LGTM

kswenson (changes requested):
👍 Code changes look good, but there are failing cypress tests to fix.

PR #1510, opened 2024-09-24
fix: Slate error on closing text component

kswenson:
[[PT-188319487]](https://www.pivotaltracker.com/story/show/188319487)

We were properly disposing of the MobX reaction when the text tile was closed, but not canceling any in-flight `requestAnimationFrame` callback.


bfinzer (approved):
👍🏻 LGTM

PR #1511, opened 2024-09-24
188326911 Fix Title of Imported Datasets

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/188326911

This PR does two things:
- It removes the file extension from the name of datasets imported into Codap (making v3 align with v2).
- It includes the displayed title in response to `get dataContextList` requests, rather than the actual title. This change was needed to make StoryQ work properly in v3.


kswenson (approved):
👍 LGTM

PR #1513, opened 2024-09-24
187972789 undo redo case card table

nstclair-cc:
[PT-187972789](https://www.pivotaltracker.com/story/show/187972789)
- adds undo/redo checks for case card->case table switch


emcelroy (approved):
Looks good 👍

PR #1514, opened 2024-09-25
187951404 two line attribute header

eireland:
Makes table attribute headers appear over two lines if width of column is narrower than attribute and unit text width.


tealefristoe (changes requested):
Looks like a lot of tests are failing with these changes.

v3/src/hooks/use-adjust-header-overflow.ts
```diff
@@ -0,0 +1,107 @@
+import { useEffect, useRef, useState } from 'react'
+import { measureText } from './use-measure-text'
+
+const kPaddingBuffer = 5 // Button width is 5px smaller because of parent padding
+
+// Hook to split headers into 2 rows and ellide the 2nd line if it doesn't fit
```
kswenson, line 6:
```suggestion
// Hook to split headers into 2 rows and elide the 2nd line if it doesn't fit
```
"Ellipsis" has two L's but "elide" only has one. 🤷

v3/src/components/case-tile-common/attribute-header.tsx
```diff
@@ -200,13 +218,15 @@ export const AttributeHeader = observer(function AttributeHeader({
...
+                          className={clsx("codap-attribute-button", {"table": instanceId.includes("table")})}
+                          ref={menuButtonRef}
                           disabled={attributeId === kIndexColumnKey}
                           sx={customButtonStyle}
                           fontWeight="bold" onKeyDown={handleButtonKeyDown}
                           data-testid={`codap-attribute-button ${attrName}`}
                           aria-describedby={`sr-column-header-drag-instructions-${instanceId}`}>
-                        {`${attrName ?? ""}${showUnits ? units : ""}`}
+                        {instanceId.includes("table") ? renderAttributeLabel
+                                                      : `${attrName ?? ""}${showUnits ? attrUnits : ""}`.trim()}
```
kswenson, line 229: Rather than determining whether to support two-line rendering by parsing the `instanceId`, I'd recommend passing in a boolean prop named something like `allowTwoLines` which defaults to `false` but the case table can pass `true`.

v3/src/components/case-tile-common/attribute-header.tsx
```diff
@@ -200,13 +218,15 @@ export const AttributeHeader = observer(function AttributeHeader({
                     />
                   : <>
                       <MenuButton
-                          className="codap-attribute-button" ref={menuButtonRef}
+                          className={clsx("codap-attribute-button", {"table": instanceId.includes("table")})}
```
kswenson, line 221: This class could be something like `allow-two-lines`.

kswenson (changes requested):
👍 Looks good -- the one change I recommend is passing a prop to determine when to allow two lines rather than parsing the `instanceId`.

v3/src/components/case-table/case-table.scss
```diff
@@ -214,8 +214,34 @@ $table-body-font-size: 8pt;
...
+              overflow: hidden;
+            }
+
+            .two-line-header-line-2.truncated {
+              display: inline-block;
+              white-space: nowrap;
+              overflow: hidden;
+              text-overflow: ellipsis;
+              direction: rtl;
+              unicode-bidi: bidi-override;
```
kswenson, line 242: A comment about these would be useful.

kswenson (approved):
👍 Looks good -- one suggested comment to add.

PR #1515, opened 2024-09-25
Update Pixi to v8

pjanik:
https://www.pivotaltracker.com/story/show/187950454

This PR upgrades Pixi from v7 to v8.

The API hasn't changed significantly, in theory. Compile-time errors were relatively easy to identify. The biggest change is that Pixi is now initialized asynchronously (both the Pixi Application, which we don't use, and the renderer).

This required updates to many of our components that previously relied on PixiPoints being available immediately. I modified `usePixiPointsArray` to populate the array only after the async initialization completes, and it now uses the React state (instead of refs) to notify and re-render dependent components.

On a side note, I'm not sure if this hook is worth using for non-map components, but I didn’t look too deeply into that or tried to fully understand its purpose. My goal was just to make the minimal changes necessary to get v8 running without errors.

I also updated the renderer to use the auto-detect feature. Currently, it returns the WebGL Renderer for me, but the docs suggest it might switch to WebGPU shortly. This could offer better performance, but more importantly, we'll always be using the "preferred" version (hopefully more stable).

### DIFF
--- a/v3/package-lock.json
+++ b/v3/package-lock.json
@@ -32,7 +32,7 @@
         "mobx-react-lite": "^4.0.7",
         "nanoid": "^5.0.7",
         "papaparse": "^5.4.1",
-        "pixi.js": "~7.4.2",
+        "pixi.js": "~8.4.1",
         "pluralize": "^8.0.0",
         "prop-types": "^15.8.1",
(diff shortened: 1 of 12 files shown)


kswenson (approved):
👍 Looks good -- nice work!

PR #1516, opened 2024-09-25
feat: inspector panel for case card (PT-188327283)

emcelroy:
[#188327283](https://www.pivotaltracker.com/story/show/188327283)


kswenson (approved):
👍 Looks good -- I added a commit which eliminated the need for common code to import from component-specific code. As usual, double-check to make sure I didn't break anything, but otherwise it's good to go.

emcelroy:
> 👍 Looks good -- I added a commit which eliminated the need for common code to import from component-specific code. As usual, double-check to make sure I didn't break anything, but otherwise it's good to go.

Thanks -- changes look good to me and don't appear to have broken anything.

PR #1517, opened 2024-09-25
fix: graph collision detection for drag/drop

kswenson:
[[PT-188275873]](https://www.pivotaltracker.com/story/show/188275873)

The issue was that we were using rectangle intersection to find the appropriate drop target, but the area of intersection with the plot is usually larger than the area of intersection with the y axis when both intersect. We now prioritize the drop target that the pointer is actually over (if any) and then use the closest center to choose among multiple intersecting drop targets. We also take this opportunity to move the graph collision detection code into the graph folder.


emcelroy (approved):
Looks good 👍 I double checked it in the browser and this definitely fixes the bug.

PR #1519, opened 2024-09-25
Date Attribute Legends

bfinzer:
[#188261676] Bug fix: A date attribute can be added to a map to create a date legend

* In data-configuration-model.ts we provide `getLegendColorForDateValue`
* In legend.tsx we allow for the possibility that the attribute has type 'date'
* Added titles for the choropleth rectangles that show up on hover
* Detecting that attribute is of type 'date' and modifying the formatting of the labels that appear below the choropleth rectangles accordingly
* If label text collides, only show the min and max labels
* In data-configuration-model.ts modify `getCasesForLegendQuantile` to use `dataDisplayGetNumericValue` instead of `Dataset:getNumeric` so that dates will be properly retrieved
* Add a style for choroplet-rect so that pointer cursor will show on hover


kswenson (approved):
👍 LGTM

PR #1520, opened 2024-09-26
Add subplot mask to graph / pixi points [PT-185628921]

pjanik:
https://www.pivotaltracker.com/story/show/185628921

Note that it wasn't really bug but unimplemented feature that was also unimplemented for the SVG graphs in V3. And it's relatively easy to reproduce this problem, see my comment: https://www.pivotaltracker.com/story/show/185628921/comments/242620584

I was struggling a bit to get all the info I needed from various data sources (layout/data configuration). I'm not 100% sure this is done in the optimal way, so if I missed something in the data source, feel free to suggest updates or change it yourself.

There's an assumption that subplots (cells?) are always ordered first by column (left to right), then by rows (bottom to top). This seemed to hold true for both simpler and more complex cases.

Also, I wasn't sure if `CaseDataWithSubPlot` is fully necessary, and if CaseData's `plotNum` wasn’t intended to be `subPlotNum` in fact (but it's always equal to 0 in my test cases). But I added a new type and prop just in case - it can be easily removed if needed.

I tested the following scenarios:
- starting with an empty graph and adding subplots
- loading a document with subplots
- removing subplots
- adding more subplots
- moving a case between subplots by editing the case table value

### DIFF
--- a/v3/src/components/data-display/d3-types.ts
+++ b/v3/src/components/data-display/d3-types.ts
@@ -1,2 +1,5 @@
  // The data stored with each plot element (e.g. 'circle')
 export type CaseData = { plotNum: number, caseID: string }
+
+// The data stored with each plot element (e.g. 'circle') when subplots are used
+export type CaseDataWithSubPlot = { plotNum: number; caseID: string; subPlotNum?: number }
--- a/v3/src/components/data-display/data-display-utils.ts
+++ b/v3/src/components/data-display/data-display-utils.ts
@@ -7,6 +7,7 @@ import {
 } from "../../utilities/color-utils"
 import {between} from "../../utilities/math-utils"
 import { IBarCover } from "../graph/graphing-types"
+import {isGraphDataConfigurationModel} from "../graph/models/graph-data-configuration-model"
 import {ISetPointSelection} from "../graph/utilities/graph-utils"
 import {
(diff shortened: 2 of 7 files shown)


kswenson:
> Also, I wasn't sure if `CaseDataWithSubPlot` is fully necessary, and if CaseData's `plotNum` wasn’t intended to be `subPlotNum` in fact (but it's always equal to 0 in my test cases)

FYI, in the code `plotNum` refers to the index of the attribute on the y-axis when there are multiple attributes, so it would be expected to be zero in these use cases. Yes, I agree that there is probably a better name for it. 😉

kswenson (commented):
👍 Code seems fine to me but I also assigned @bfinzer to review it as well because he may have an opinion on the way the `subPlotNum` is maintained here. I do notice that there are failing cypress tests all of which appear to be graph related and so at least potentially due to the changes in the PR. I tried re-running the tests and the same tests failed. I looked into them briefly and every one of the failing tests that I tried to run locally succeeded when run locally, so I'm not sure what the nature of the issue is but we probably need to understand it.

pjanik:
Tests fixed in a separate PR: https://github.com/concord-consortium/codap/pull/1522 to explain better what might be happening.

bfinzer (approved):
👍🏻LGTM

kswenson (approved):
👍 Looks good!

PR #1521, opened 2024-09-27
Dispose unused Pixi textures

pjanik:
https://www.pivotaltracker.com/story/show/186784013

I think it's unlikely to cause serious problems, but I suppose if there are multiple graphs with large legend sets that change frequently, some cleanup wouldn’t hurt.


kswenson (approved):
👍 Looks great -- thx for doing this.

PR #1522, opened 2024-09-27
Try to handle failed Pixi renderer init, and fix broken tests

pjanik:
@kswenson, so I pushed multiple branch to GH Actions, trying to find what was causing these Cypress tests to fail. Apparently it was `setPointsMask()` call, as this test commit fixed the tests: https://github.com/concord-consortium/codap/commit/06037110209c32cbdc098e581cae8c91f183833a

I don't really know why that happens, but Pixi v8 dropped support of the Canvas renderer. I suspect that GH Actions WebGL support might not be available or reliable. Maybe the mask functionality triggers an error that then breaks the rest of the app flow.

This PR adds error handling for the renderer fail - I've tested that locally with Chrome --disable-webgl flag. But what fixes the Cypress tests indeed is: `if ((window as any).Cypress) { return }` 😉


kswenson (approved):
👍 Looks great -- not ideal that we have to change the rendering path for cypress, but seems like the best solution for now. Nice work sleuthing it out.

PR #1523, opened 2024-09-30
Graph filter formula

pjanik:
https://www.pivotaltracker.com/story/show/187799311

This PR (partially) implements the graph filter formula. However, there's a missing piece - hopefully small, but very important - the graph view is not updating correctly after the filter formula results are applied. That's why I'm opening this as a draft PR.

The new adapter has some elements similar to the more basic `FilterFormulaAdapter` and some elements similar to the `PlottedValue` or `PlottedFunction` adapters. It sits somewhere in between.

I did struggle a bit with the concept of `CaseID` vs `ItemID` and working with the `DataConfiguration`. My (limited) understanding is that `ItemID` represents a leaf, while `CaseID` can encompass multiple `ItemIDs`. However, when we set aside cases/items in the Dataset/Case table, we work with `ItemIDs`, while graphs work with `CaseIDs`. 

I tried calling `self._invalidateCases()` (similar to what I did in the DataSet), but it doesn't seem to update the graph view. However, if you check the `DataConfigurationModel#allCaseIDs` console logs, they appear to be updated correctly. So, I assume it's just a matter of finding a better way to notify the graph, maybe another method? I've added more TODOs in spots that may require additional work or verification. E.g.  `unhiddenCaseIDs` can likely be implemented better - or perhaps should be. I hope this will be an easier task for you.

The second commit is semi-related, as it fixes a bug in the case table filter formula. For example, when I set the filter formula to "Sleep < 5" and then manually modify a single case in the case table to "Sleep = 10", it wasn't disappearing (due to CaseID vs ItemID conflict). Now, it does.

I'm pretty sad that I couldn't wrap up my last CC PR 🤧, but I hope this is still useful.

### DIFF
--- a/v3/cypress/e2e/adornments.spec.ts
+++ b/v3/cypress/e2e/adornments.spec.ts
@@ -24,12 +24,10 @@ context("Graph adornments", () => {
     // add undo/redo to this test
     toolbar.getUndoTool().click()
     ae.getAxisLabel("left").should("have.length", 0)
-    cy.wait(250)
     toolbar.getRedoTool().click()
     ae.getAxisLabel("left").should("have.length", 1)
-    cy.wait(250)
 
-
+    // show the count adornment
     graph.getDisplayValuesButton().click()
     graph.getInspectorPalette().should("be.visible")
     graph.getInspectorPalette().find("[data-testid=adornment-checkbox-count-count]").should("be.visible").click()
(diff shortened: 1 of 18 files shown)


kswenson:
> I'm pretty sad that I couldn't wrap up my last CC PR 🤧, but I hope this is still useful.

No need to feel bad! This will be very helpful and I'm sure we can take it from here. Thx for all your help!

kswenson:
Last couple of commits cause the graphs to update when the filter formula is changed. It makes use of a new set of utilities to hash a set of ids to simplify change detection. We also update one of the cypress tests to exercise the filter formula code.

kswenson:

PR #1526, opened 2024-10-01
Fix summary case card issues. (PT-188001359)

lublagg:
Addresses the following issues:

- Bottom button font size
- If you drag over a selection of cases in a graph, then the data card view should go into summary mode, but just be a summary of those selected cases.
- If you navigate at one of the parent levels, the automatically selects all of the cases present in the child cards, so, since a selection of more than one case occurs, it should go into summary mode.

I also noticed some other bugs while working on the above that were all tied to how we were keeping track of the `summarizedCollections` and whether or not to show the summary view. Instead of having the components pass back information to the model to update the `summarizedCollections` prop on the Case Card model, `summarizedCollections` is now a view on the model that updates whenever there is a change in the selected cases.


v3/cypress/e2e/text.spec.ts
```diff
@@ -11,20 +11,25 @@ context("Text tile", () => {
...
     // Undo title change
+    cy.wait(100)
     toolbar.getUndoTool().click()
+    cy.wait(100)
     c.getComponentTitle(kTextTileTestId).should("have.text", textDefaultTitle)
 
     // Redo title change
     toolbar.getRedoTool().should("be.enabled")
     toolbar.getRedoTool().click()
+    cy.wait(100)
```
emcelroy, line 32: Would it make sense to take out these added waits?

v3/cypress/support/e2e.ts
```diff
@@ -22,6 +22,9 @@ import "./commands"
 // add code coverage support
 import "@cypress/code-coverage/support"
 
+// add support for dispatching native events (https://github.com/dmtrKovalenko/cypress-real-events)
+import "cypress-real-events"
+
```
emcelroy, line 27: Oh, is this what got attribute button clicks working in the tests?

v3/src/components/case-card/card-view.scss
```diff
@@ -30,7 +30,7 @@
     color: #0592af;
     cursor: pointer;
     display: block;
-    font-size: 10px;
+    font-size: 12px;
```
emcelroy, line 33: Is this the right size? The spec says 10px but then Dan said it was too small and that it "should match the font size of the rest of the text on the data card view" which is 11px. Did you and/or Dan decide it should be slightly larger? To me it looks a little too big now which is why I double checked, but maybe that's just me.

v3/src/components/case-card/case-attrs-view.tsx
```diff
@@ -68,7 +68,7 @@ export const CaseAttrsView = observer(function CaseAttrsView({caseItem, collecti
             if (!attr || metadata?.isHidden(attr.id)) return null
             return (
               <CaseAttrView
-                key={attr.id}
+                key={isCollectionSummarized ? `${attr.id}-summary` : attr.id}
```
emcelroy, line 71: Ah :)

emcelroy (commented):
This is looking good to me 👍 Changes to the card code make sense and everything seems to work well in the browser. 

I left a few comments that I don't think should hold up merging this in, but I do wonder about all the changes here. I think something got a little mixed up with the commit history. Did you maybe merge `main` into your branch? I'm not sure but think Kirk may want to clean that up before this gets merged in.

v3/cypress/support/e2e.ts
kswenson: I added that as part of getting cypress tests of the new formula editor to work, so I don't think it's related to the commented out tests.

kswenson:
> Did you maybe merge `main` into your branch? I'm not sure but think Kirk may want to clean that up before this gets merged in.
> 
> There are also conflicts reported in a few files that will need to be fixed.

I merged from `main` and dealt with the conflicts.

kswenson (approved):
👍 Looks good -- I added a commit which merged `main` into the PR branch, simplified the `summarizedCollections()` computation, and tweaked a couple of other things. One thing to note is that if there is only one case in a collection, it needn't be summarized even if nothing is selected.

kswenson (approved):
👍 Changes LGTM

v3/cypress/e2e/text.spec.ts
kswenson: I agree that the `wait`s should be removed unless they've been determined to be necessary (in which case there's probably a better solution anyway).

kswenson:
I gave my approval, but @emcelroy's comments should still be considered/addressed.

v3/cypress/e2e/text.spec.ts
lublagg: I didn't add these in -- these changes came in when I merged in main.

v3/src/components/case-card/card-view.scss
lublagg: Good catch - will change to 11px.

PR #1527, opened 2024-10-01
182849961 table advance selection

eireland:
Adds ability to select multiple cases in the table via:
1. Keyboard navigation - use modifier keys (shift, command, alt) + arrow up or down keys to add additional cases to be selected after user clicks on a case.
2. Marquee select - click and drag over table grid to select cases. Cases are deselected if user changes direction from initial direction of drag. If user moves close to the top or bottom, grid scrolls to continue selecting.

Note that marquee selecting while dragging outside the table component does not work. But shift  + click does extend the selection.

### DIFF
--- a/v3/src/components/case-table/collection-table.tsx
+++ b/v3/src/components/case-table/collection-table.tsx
@@ -1,6 +1,6 @@
 import { comparer } from "mobx"
 import { observer } from "mobx-react-lite"
-import React, { useCallback, useEffect, useMemo, useRef } from "react"
+import React, { useCallback, useEffect, useMemo, useRef, useState } from "react"
 import DataGrid, { CellKeyboardEvent, DataGridHandle } from "react-data-grid"
 import { kCollectionTableBodyDropZoneBaseId } from "./case-table-drag-drop"
 import {
@@ -62,9 +62,11 @@ export const CollectionTable = observer(function CollectionTable(props: IProps) 
                                 .getPropertyValue("--rdg-row-selected-background-color") || undefined
   const visibleAttributes = useVisibleAttributes(collectionId)
   const { selectedRows, setSelectedRows, handleCellClick } = useSelectedRows({ gridRef, onScrollClosestRowIntoView })
-  const { handleWhiteSpaceClick } = useWhiteSpaceClick({ gridRef })
+  const { clearCurrentSelection } = useWhiteSpaceClick({ gridRef })
   const forceUpdate = useForceUpdate()
   const { isTileSelected } = useTileModelContext()
+  const [isSelectDragging, setIsSelectDragging] = useState(false)
+  const [isDragging, setIsDragging] = useState(false)
 
   useEffect(function setGridElement() {
     const element = gridRef.current?.element
@@ -220,10 +222,55 @@ export const CollectionTable = observer(function CollectionTable(props: IProps) 
   }
 
   const handleClick = (event: React.MouseEvent<HTMLDivElement>) => {
+    if (isDragging) {
+      setIsDragging(false)
+      return
+    }
     // the grid element is the target when clicking outside the cells (otherwise, the cell is the target)
     if (isTileSelected() && event.target === gridRef.current?.element) {
-      handleWhiteSpaceClick()
+      clearCurrentSelection()
     }
+  }
+
+  const handleMouseDown = (event: React.MouseEvent<HTMLDivElement>) => {
+    const isExtending = event.shiftKey || event.altKey || event.metaKey
+    setIsSelectDragging(true)
+    if (!isExtending) clearCurrentSelection() // clear current selection
+  }
+
+  const handleMouseMove = (event: React.MouseEvent<HTMLDivElement>) => {
+    if (isSelectDragging) {
+      setIsDragging(true)
+      // Check if the mouse is within the bounds of the DataGrid
+      const gridElement = gridRef.current?.element
+      if (gridElement) {
+        const gridBounds = gridElement.getBoundingClientRect()
+        const { clientX, clientY } = event
+        // Check if the mouse is inside the grid's boundaries
+        if (
+          clientX >= gridBounds.left &&
+          clientX <= gridBounds.right &&
+          clientY >= gridBounds.top &&
+          clientY <= gridBounds.bottom
+        ) {
+          const target = event.target as HTMLDivElement
+          const closestDataCell = target.closest('.codap-data-cell')
+          const className = closestDataCell ? closestDataCell.className : ""
+          const caseId = className.split(" ").find(c => c.startsWith("rowId-"))?.split("-")[1]
+          if (caseId) {
+            if (selectedRows.size > 0) {
+              selectCases([caseId], data)
+            } else {
+              setSelectedCases([caseId], data)
+            }
+          }
+        }
+      }
+    }
+  }
+
+  const handleMouseUp = (event: React.MouseEvent<HTMLDivElement>) => {
+    setIsSelectDragging(false)
   }
 
   if (!data || !rows || !visibleAttributes.length) return null
@@ -231,8 +278,9 @@ export const CollectionTable = observer(function CollectionTable(props: IProps) 
   return (
     <div className={`collection-table collection-${collectionId}`}>
       <CollectionTableSpacer selectedFillColor={selectedFillColor}
-        onWhiteSpaceClick={handleWhiteSpaceClick} onDrop={handleNewCollectionDrop} />
-      <div className="collection-table-and-title" ref={setNodeRef} onClick={handleClick}>
+        onWhiteSpaceClick={clearCurrentSelection} onDrop={handleNewCollectionDrop} />
+      <div className="collection-table-and-title" ref={setNodeRef} onClick={handleClick} onMouseDown={handleMouseDown}
+           onMouseMove={handleMouseMove} onMouseUp={handleMouseUp}>
         <CollectionTitle onAddNewAttribute={handleAddNewAttribute} showCount={true} />
         <DataGrid ref={gridRef} className="rdg-light" data-testid="collection-table-grid" renderers={renderers}
           columns={columns} rows={rows} headerRowHeight={+styles.headerRowHeight} rowKeyGetter={rowKey}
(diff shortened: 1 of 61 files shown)


tealefristoe (changes requested):
Overall this is looking good! I looked through much of the code and thought it was fine, then decided to test it. I think it generally works well, but I did have some problems with dragging to select. These all seem to be inconsistencies with v2. I'm not sure how many of these issues need to be addressed, but I think at least the final one should be, which is reproduceable.

- Selection stops working when you drag outside of the collection table. In v2, selection continues to change as you move the mouse up and down anywhere in the document. If you then move the mouse back into the collection table, you can select discontinuous cases. I can imagine this one would be difficult to fix without a major overhaul of how this is set up.
- Dragging up and down, I encountered issues with cases sometimes deselecting incorrectly. Maybe the most obvious example of this was starting in the middle of the table, dragging down, then dragging up above the starting case. Sometimes doing this, the starting case would deselect, while other times it wouldn't. In v2, the starting case never deselects when you scroll past it.
- When you start dragging on a case that's already selected, it becomes deselected and never gets selected. In v2, if you start dragging on an already selected case, that case remains selected.
- The final issue is that if you start dragging on the top case, then drag downwards, then back up to the top, the second case never becomes deselected. So in that case both the top and second case always remain selected.

v3/src/components/case-table/use-white-space-click.ts
```diff
@@ -36,7 +36,7 @@ export function useWhiteSpaceClick({ gridRef }: IProps) {
     })
   }, [componentRef, data])
 
-  const handleWhiteSpaceClick = useCallback(() => {
+  const clearCurrentSelection = useCallback(() => {
```
kswenson, line 39: The reason I named this function `handleWhiteSpaceClick` originally is because a name like `clearCurrentSelection` implies rather strongly that calling it will clear the selection, but it doesn't actually change the selection under some very important circumstances. If you would like to have access to a function which actually clears the selection, then the `clearCaseSelection` function can be exported from this hook in addition to `handleWhiteSpaceClick`.

v3/src/components/case-table/collection-table.tsx
```diff
@@ -192,22 +200,139 @@ export const CollectionTable = observer(function CollectionTable(props: IProps)
       event.preventGridDefault()
       navigateToNextRow(event.shiftKey)
     }
+    if ((event.key === "ArrowDown" || event.key === "ArrowUp")) {
+      const caseId = args.row.__id__
+      const isCaseSelected = data?.isCaseSelected(caseId)
+      const isExtending = event.shiftKey || event.altKey || event.metaKey
+      const currentSelectionIdx = collectionCaseIndexFromId(caseId, data, collectionId)
+
+      if (currentSelectionIdx !== undefined && currentSelectionIdx !== null) {
```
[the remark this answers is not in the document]
kswenson: I concur that @tealefristoe's simplification is preferable.

kswenson (changes requested):
👍 Keyboard handling looks good. I didn't look too closely at the marquee selection code given @tealefristoe's comments. The traditional way of handling marquee selection is to identify the mouse-down location as the anchor, the current mouse position defines the other corner of a rectangle, and then you select the rows that intersect the rectangle. This avoids any need to track direction of motion and how many times the user may or may not have doubled back on themselves.

tealefristoe (changes requested):
This fixes some of the issues in the last round, but not all of them.
- If you drag outside of the collection table (even still within the table tile, just outside of the collection), the selection stops changing. In v2, you can drag anywhere and the selection will continue to change as the mouse moves up and down.
- If you mouse up outside of the collection table, then move your mouse back to the collection table, the selection continues to be modified, even though you're no longer dragging. This is because the mouseup event listener that triggers the end of the selection is only within the collection table.
- These changes seem to break the whitespace click code. I left a note about `isDragging`, which I suspect is part of the problem. But in many cases with these changes, it's no longer possible to deselect all by clicking on the table background.

For the first two points, I feel like you'll need to add document level mouse event listeners to capture mousemove and mouseup events to recreate the behavior of v2.

kswenson (changes requested):
Given the issues already identified by @tealefristoe, I'm not going to do a detailed review of the drag handling code. Based on his comments, it seems like the down-handler can remain a React handler on the `CollectionTable` component, but that at that point capturing handlers for move/up should be used so that we're not relying on the vagaries of React event dispatch. Also, using pointer events rather than mouse events would make it more likely that touch would work, but if you've tried pointer events and mouse events work better for some reason we can punt that for another day.

v3/src/components/case-table/collection-table.tsx
```diff
@@ -192,22 +199,166 @@ export const CollectionTable = observer(function CollectionTable(props: IProps)
...
+        const selectedCaseIds = newSelectedRows
+                                  .map(idx => collectionCaseIdFromIndex(idx, data, collectionId))
+                                  .filter((id): id is string => id !== undefined)
+        setOrExtendSelection(selectedCaseIds, data)
+      }
+      mouseY.current = event.clientY
+      const grid = gridRef.current?.element
+      if (grid) {
+        const { top, bottom } = grid.getBoundingClientRect()
+        if (mouseY.current < top + 50 || mouseY.current > bottom - 20) {
```
tealefristoe, line 331: More hard-coded numbers... and the top one is different than the one above! You should definitely use named constants for these, and they should be the same constants as the ones from lines 259 and 263.

v3/src/components/case-table/collection-table.tsx
```diff
@@ -192,22 +199,166 @@ export const CollectionTable = observer(function CollectionTable(props: IProps)
...
+    setIsSelecting(false)
+    setSelectionStartRowIdx(null)
+    stopAutoScroll()
+  }, [stopAutoScroll])
+
+  useEffect(() => {
+    return () => {
+      stopAutoScroll()
+    }
+  }, [stopAutoScroll])
```
tealefristoe, line 352: What does this do?

v3/src/components/case-table/collection-table.tsx
```diff
@@ -192,22 +199,166 @@ export const CollectionTable = observer(function CollectionTable(props: IProps)
...
+
   if (!data || !rows || !visibleAttributes.length) return null
 
   return (
     <div className={`collection-table collection-${collectionId}`}>
       <CollectionTableSpacer selectedFillColor={selectedFillColor}
         onWhiteSpaceClick={handleWhiteSpaceClick} onDrop={handleNewCollectionDrop} />
-      <div className="collection-table-and-title" ref={setNodeRef} onClick={handleClick}>
+      <div className="collection-table-and-title" ref={setNodeRef} onClick={handleClick} onPointerDown={handlePointerDown}
```
tealefristoe, line 360: Line length

tealefristoe (approved):
I left several comments that you should address, but after that I think it would be good to merge at this point. I know this PR has been dragging on for a long time.

That said, there are still problems with this that need to be fixed. You should make bug stories for the following:
- If you click and hold near the top or bottom of the table, it will start scrolling but won't select anything.
- Not having mouse handlers outside of the table is particularly bad for this bug: If you click, then drag outside of the table, then release the mouse button, the table will stay in selection mode. So when you move your mouse back to the table, you'll change the selection automatically as you drag, which could be a really annoying surprise for the user.
- If you click, then drag past the top or bottom of the table, then release, the table will continue to scroll by itself, modifying the selection as it goes. If you then interact with another tile, like selecting a point in a graph, that selection will immediately be overwritten by the continually scrolling table. Again, when you return the mouse to the table, you will still be in selection mode, so dragging the mouse will modify the selection until you click.

tealefristoe (changes requested):
On second thought, I think it would be pretty easy to at least put a bandaid on the issues I mentioned in my last review by setting up a `mouseleave` handler that ends the selection just like your `mouseup` handler. This would be easy to do and would prevent the bottom two issues in my last review.

v3/src/components/case-table/collection-table.tsx
```diff
@@ -192,22 +199,166 @@ export const CollectionTable = observer(function CollectionTable(props: IProps)
...
+      const newSelectedRows = []
+      const startIdx = Math.min(selectionStartRowIdx, scrolledToRowIdx)
+      const endIdx = Math.max(selectionStartRowIdx, scrolledToRowIdx)
+      for (let i = startIdx; i <= endIdx; i++) {
+        const rowId = collectionCaseIdFromIndex(i, data, collectionId)
+        if (rowId) {
+          newSelectedRows.push(rowId)
+        }
+      }
+      setOrExtendSelection(newSelectedRows, data)
```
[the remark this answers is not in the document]
eireland: When I use `selectCases`, it doesn't deselect rows when I backtrack on the selection.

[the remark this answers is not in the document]
tealefristoe: Sorry, you should use `setSelectedCases` here. `setOrExtendSelection` is intended to be used when you might want to set the selection, or you might want to extend it. In this situation, you never want to extend the selection.

v3/src/components/case-table/collection-table.tsx
```diff
@@ -243,51 +245,55 @@ export const CollectionTable = observer(function CollectionTable(props: IProps)
...
+        const newSelectedRows = []
+        const start = Math.min(startIdx, endIdx)
+        const end = Math.max(startIdx, endIdx)
+        for (let i = start; i <= end; i++) {
+          newSelectedRows.push(i)
+        }
+        const selectedCaseIds = newSelectedRows
+                                  .map(idx => collectionCaseIdFromIndex(idx, data, collectionId))
+                                  .filter((id): id is string => id !== undefined)
+        setOrExtendSelection(selectedCaseIds, data)
```
tealefristoe, line 261: I was wrong before about using `selectCases`. But you should use `setSelectedCases` here. `setOrExtendSelection` is intended to be used when you sometimes want to set the selection, and sometimes want to modify it. Here, you're always setting the selection.

tealefristoe (approved):
Looks good 👍

kswenson (approved):
👍 LGTM -- I pushed a commit with a few code review tweaks.

kswenson:
One more note -- this doesn't handle extending the selection via marquee selection. I think that should be written up as a separate story.

eireland:
@kswenson V2 also didn't handle extending marquee selection. I will add it as a feature story but probably make it a P3.

PR #1529, opened 2024-10-02
186182704 formula column background

eireland:
Adds background color to columns with formula.

Adds background color to value cell in case card if attribute has a formula

Disables cell editing in case card when attribute has a formula.

Adds cypress test to verify background color for cells with formula values.

Splits case-card cypress test so that two different URLs are used to load CODAP. The `mouseSensor` query param was interfering with test that required the attribute menu.

### DIFF
--- a/v3/src/components/case-table/use-columns.tsx
+++ b/v3/src/components/case-table/use-columns.tsx
@@ -12,6 +12,7 @@ import { kDefaultColumnWidth, TColumn } from "./case-table-types"
 import CellTextEditor from "./cell-text-editor"
 import ColorCellTextEditor from "./color-cell-text-editor"
 import { ColumnHeader } from "./column-header"
+import clsx from "clsx"
 
 interface IUseColumnsProps {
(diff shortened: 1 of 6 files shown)


v3/src/components/case-table/use-columns.tsx
```diff
@@ -47,7 +48,7 @@ export const useColumns = ({ data, indexColumn }: IUseColumnsProps) => {
                 resizable: true,
                 headerCellClass: `codap-column-header`,
                 renderHeaderCell: ColumnHeader,
-                cellClass: "codap-data-cell",
+                cellClass: `codap-data-cell ${hasFormula ? "formula-column" : ""}`,
```
kswenson, line 51:
```suggestion
                cellClass: clsx("codap-data-cell", { "formula-column" : hasFormula }),
```

kswenson (approved):
👍 Looks good -- one suggestion to use `clsx` instead of concatenating class strings ourselves.

PR #1530, opened 2024-10-02
187323603 drag over webview bug

eireland:
Fixes bug where dragging components over a webview makes the component lose the pointer focus. This captures the pointer while dragging a component

### DIFF
--- a/v3/src/components/container/free-tile-component.tsx
+++ b/v3/src/components/container/free-tile-component.tsx
@@ -44,6 +44,21 @@ export const FreeTileComponent = observer(function FreeTileComponent({ row, tile
     activeDrag => {
     const dragTileId = getDragTileId(activeDrag)
     if (dragTileId) {
+      const draggedElement = document.getElementById(dragTileId)
+      if (draggedElement) {
+        // Capture pointer events for the dragged tile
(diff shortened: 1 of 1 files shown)


v3/src/components/container/free-tile-component.tsx
```diff
@@ -44,6 +44,18 @@ export const FreeTileComponent = observer(function FreeTileComponent({ row, tile
     activeDrag => {
     const dragTileId = getDragTileId(activeDrag)
     if (dragTileId) {
+      const draggedElement = document.getElementById(dragTileId)
+      if (draggedElement) {
+        // Capture pointer events for the dragged tile
+        draggedElement.addEventListener('pointermove', (event) => {
+          draggedElement.setPointerCapture(event.pointerId)
```
kswenson, line 51:
```suggestion
          if (!draggedElement.hasPointerCapture(event.pointerId)) {
            draggedElement.setPointerCapture(event.pointerId)
          }
```

kswenson (approved):
👍 Looks good -- I'm impressed that it works as well as it does. One suggestion to add a check for whether the pointer is captured rather than blindly capturing it multiple times.

PR #1531, opened 2024-10-02
fix: crash on delete DataSet

kswenson:
There were several things going on here:
- The primary cause of the hang/crash was an infinite loop in the `GraphDataConfigurationModel`'s attempt to synchronize its `filteredCases` array after its `DataSet` was destroyed.

Once that was fixed, there were still several clients that were attempting to access the defunct `DataSet` triggering warnings. These have also been fixed:
- The `FilteredCases` class now responds to disposal of its `DataSet`, thus preventing further access.
- The `CaseTile`'s `HideShowMenuList` checks whether its `DataSet` is alive before accessing it. (We conventionally avoid `isAlive` in models, but in components there often isn't a better way.)
- The `CountAdornmentComponent` was properly using `mstAutorun`, but the model it was attached to was the `CountAdornmentModel`, not the axis models that were _also_ accessed within the `autorun`.
- Therefore, `mstAutorun` and `mstReaction` have been extended to support an array of model dependencies _or_ a single model and so now the `CountAdornmentComponent` can specify all of its model dependencies. (I have wondered for some time whether we would ever encounter a situation in which a single model dependency was insufficient for these utilities and now we have.)

### DIFF
--- a/v3/src/components/case-tile-common/hide-show-menu-list.tsx
+++ b/v3/src/components/case-tile-common/hide-show-menu-list.tsx
@@ -1,5 +1,6 @@
 import { useDisclosure } from "@chakra-ui/react"
 import { observer } from "mobx-react-lite"
+import { isAlive } from "mobx-state-tree"
 import React from "react"
 import { useCaseMetadata } from "../../hooks/use-case-metadata"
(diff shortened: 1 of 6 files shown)


bfinzer (approved):
👍🏻LGTM

PR #1532, opened 2024-10-02
188286105 Drag from Plugins to Codap

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/188286105

This work can be tested at: https://codap3.concord.org/branch/v3-di-drag-drop/#file=examples:Mammals
Using this version of Multidata: https://models-resources.concord.org/multidata-plugin/branch/drag-v3/index.html

This PR enables dragging attributes from plugins to Codap v3. In the past, this was accomplished using the HTML5 drag and drop system. However, Codap v3 uses `dnd-kit` for drag and drop, so the old system no longer works (and it never worked very well anyway, given security restrictions between iframes and their containers). Now, plugins enable drags within Codap by sending a series of API requests which trigger synthetic pointer events that `dnd-kit` reacts to.
- Plugins cause drags in Codap v3 via `notify attribute` requests. Three requests must be made:
  - A `dragStart` request will begin a drag by creating synthetic `mousedown` and `mousemove` events (because a drag won't actually begin until the mouse is moved at least three pixels).
    - The size of the drag overlay should be specified in this request with `overlayHeight` and `overlayWidth`.
  - `dragOver` requests update the position of the drag overlay, and they should be continually requested as the user moves the mouse. `mouseX` and `mouseY` should be included in the request, and should be in the plugin's coordinate system.
  - A `dragEnd` request will end the drag. It should also include `mouseX` and `mouseY`.
- `PluginAttributeDrag` is a dummy element that is used as the basis of plugin initiated drags. It's a transparent div in the `Container`.
- An `AttributeDragOverlay` is also included in the `Container`, which follows the mouse during plugin initiated drags. Its size and position is controlled by API requests.
- Information is passed from the API handler to the `PluginAttributeDrag` (`dataSetId` and `attributeId`) and `AttributeDragOverlay` (overlay size and position) via new `uiState` properties.
  - This feels inelegant to me, and I am open to alternate methods of achieving this. Using the synthetic events won't work.

Other changes in this PR:
- A `getOverlayDragId` function is used to unify how `AttributeDragOverlay` `activeDragIds` are generated.
- The new `fieldRequiredResult` function generates common error messages to API handlers.
- Fix a bug in `data-set-utils` that failed to broadcast a notification in certain situations when attributes were moved.


v3/src/models/ui-state.ts
```diff
@@ -43,6 +43,13 @@ export class UIState {
   @observable
   private _interruptionCount = 0
 
+  @observable private _draggingDatasetId = ""
+  @observable private _draggingAttributeId = ""
+  @observable private _draggingXOffset = 0
+  @observable private _draggingYOffset = 0
+  @observable private _draggingOverlayHeight = 100
+  @observable private _draggingOverlayWidth = 100
```
kswenson, line 51: >- Information is passed from the API handler to the `PluginAttributeDrag` (`dataSetId` and `attributeId`) and `AttributeDragOverlay` (overlay size and position) via new `uiState` properties.
>   - This feels inelegant to me, and I am open to alternate methods of achieving this. Using the synthetic events won't work.


I thought about a couple of other options that might take more effort and/or might not work as well. In the end, I'm drawn to the simpler suggestion of what about making a new class, perhaps `DataInteractiveState`, and then putting these there?

kswenson (changes requested):
👍 Looks good -- just a few suggestions.

kswenson:
One other comment I forgot to mention in my earlier review -- it would make sense to have constants for the ids/classes referenced in the DI code so they're less likely to get out of sync: e.g. `plugin-attribute-drag`, `codap-container`, `codap-web-view-body`.

v3/src/hooks/use-drag-drop.ts
```diff
@@ -29,6 +30,11 @@ export function getDragAttributeInfo(active: Active | null): Omit<IDragAttribute
   const { dataSet, attributeId } = active?.data.current as IDragAttributeData || {}
   return dataSet && attributeId ? { dataSet, attributeId } : undefined
 }
+export function getOverlayDragId(active: Active | null, instanceId: string, excludeIndexColumn = false) {
```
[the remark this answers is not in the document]
tealefristoe: While I was working on converting this, I realized that `CaseCard` probably doesn't need to exclude the index id, and that its original inclusion was likely a copy/paste artifact. Does that sound right to you? If so I'll just remove the exception from the component.

v3/src/hooks/use-drag-drop.ts
```diff
@@ -2,10 +2,11 @@ import {
   Active, DataRef, DragEndEvent, Modifier, useDndMonitor,
   useDraggable, UseDraggableArguments, useDroppable, UseDroppableArguments
 } from "@dnd-kit/core"
+import { kIndexColumnKey } from "../components/case-tile-common/case-tile-types"
```
kswenson, line 5:
```suggestion
```
Shouldn't need this any more.

kswenson (approved):
👍 Changes look good! Just a couple minor suggestions.

PR #1533, opened 2024-10-03
fix: selection on undo of create case [PT-188376356]

kswenson:
[[PT-188376356]](https://www.pivotaltracker.com/story/show/188376356)

Also, beef up the DataSet jest tests.


tealefristoe (approved):
Looks good 👍

PR #1534, opened 2024-10-04
fix: case card edit formula cypress test

kswenson:


kswenson:
Merging without review as it's a small change that fixes the `main` build.

PR #1535, opened 2024-10-04
fix: import of hierarchical v2 documents

kswenson:
[[PT-188382390]](https://www.pivotaltracker.com/story/show/188382390)

This was broken recently in #1533. The code to prune stale items was over-aggressive and was inadvertently triggered during the v2 import process. This replaces the stale items test with a more conservative/reliable one.


kswenson:

PR #1536, opened 2024-10-05
188363964 v3 DI Drag Notifications

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/188363964

This PR makes Codap broadcast notifications related to dragging attributes. This allows plugins, such as Scrambler, to receive attributes dragged from Codap.
- `dragstart` and `dragend` notifications are broadcast via `handleDragStart` and `handleDragEnd` in `DndContext`. These notifications are broadcast to all plugins.
- `dragenter`, `drag`, `dragleave`, and `drop` notifications are broadcast via the new `WebViewDropOverlay`, a `div` covering a `webView` when the user is dragging an attribute.
  - These notifications are only broadcast to the single plugin the user is dragging over. This required changes to a number of core files like `apply-model-change.ts` and `document-content.ts` so a single `webViewId` could be specified when broadcasting notifications.
  - I had to make the `AttributeDragOverlay` have `pointerEvents: "none"` to get this to work.
  - I ended up having to do a bit of math to determine the mouse's position within the plugin, which is included in `drag` and `drop` notifications. Plugins that I've looked at haven't used this information, so I'm not sure how important it is.
    - I suspect there's a better way to figure out the correct coordinates. And even if there isn't, I suspect there are constants that have the information I use in the calculation somewhere else in the codebase.
    - I created a `getTileInfo` utility function that returns a tile's position and size, since it requires jumping through quite a few hoops to access.
    
I did try to add some cypress tests for this work, but in the short amount of time I spent on it I wasn't able to trigger most of the mouse events the `WebViewDropOverlay` relies on, so coverage is lower than would be ideal.


v3/src/components/web-view/web-view.tsx
```diff
@@ -26,6 +31,68 @@ export const WebViewComponent = observer(function WebViewComponent({ tile }: ITi
...
+  if (!dataSet || !attributeId) return null
+
+  const handleMouseMove: MouseEventHandler<HTMLDivElement> = event => {
+    // Determine position within interactiveFrame based on mouse and tile locations
+    const { position } = getTileInfo(tileId ?? "")
+    const tileX = position?.left ?? 0
+    // TODO Hardcoded header heights
+    const kDocumentHeaderHeight = 94
+    const kTileHeaderHeight = 25
+    const tileY = (position?.top ?? 0) + kDocumentHeaderHeight + kTileHeaderHeight
```
tealefristoe, line 70: @kswenson @eireland Is there a better way to determine the mouse position relative within a tile? If not, are the constants I use here available elsewhere in the codebase?

tealefristoe:
@kswenson This was building and passing all tests before the last commit, which just made minor changes to a cypress command. I have no idea why it's complaining about `bundler` and `Ruby`, which seems totally unrelated to the change. Should I just update bundler like it recommends?

tealefristoe:
Looks like it was because I was accidentally changing the v2 `commands.js`, not the v3 `commands.ts`. 😞

tealefristoe:
That did not fix the problem. I really don't know what's causing the issue.

v3/cypress/support/commands.ts
```diff
@@ -115,6 +116,9 @@ Cypress.Commands.add("dragAttributeToTarget", (source, attribute, target, target
     case ("headerDivider"):
       target_el = el.headerDivider
       break
+    case('webView'):
```
kswenson, line 119:
```suggestion
    case ("webView"):
```
For consistency and to eliminate the lint warning. If you were feeling motivated, for extra credit you could remove the gratuitous parentheses throughout the switch statements in this function, but that's not required.

v3/src/components/drag-drop/attribute-drag-overlay.scss
```diff
@@ -12,3 +12,7 @@
   justify-content: center;
   align-items: center;
 }
+
+.dnd-kit-drag-overlay {
+  pointer-events: none;
```
kswenson, line 17: A comment about this would be useful. Seems a bit counter-intuitive, but as long as regular dragging still works I have no objection.

v3/src/components/web-view/web-view.tsx
```diff
@@ -26,6 +30,68 @@ export const WebViewComponent = observer(function WebViewComponent({ tile }: ITi
       <div className="codap-web-view-iframe-wrapper">
         <iframe className="codap-web-view-iframe" ref={iframeRef} src={webViewModel.url} />
       </div>
+      <WebViewDropOverlay />
```
kswenson, line 33: Any issues with rendering this over the plugin potentially affecting the behavior of the plugin? A priori, I might have thought that the overlay would only be rendered during a drag, which would mitigate any potential impact on the plugin at other times.

v3/src/components/web-view/web-view.tsx
```diff
@@ -26,6 +30,68 @@ export const WebViewComponent = observer(function WebViewComponent({ tile }: ITi
...
+  const setRef = (ref: HTMLDivElement) => {
+    overlayRef.current = ref
+    setNodeRef(ref)
+  }
+
+  return <div
+    className="codap-web-view-drop-overlay"
+    onMouseEnter={() => handleMouseEnterLeave("dragenter")}
+    onMouseLeave={() => handleMouseEnterLeave("dragleave")}
+    onMouseMove={handleMouseMove}
```
kswenson, line 94: Did you try with pointer events instead of mouse events? With pointer events we should get touch support automatically. With mouse events, touch support might require some additional work.

v3/src/models/history/apply-model-change.ts
```diff
@@ -13,6 +13,7 @@ export interface INotification {
 export interface IApplyModelChangeOptions {
   log?: string | ILogMessage | (() => Maybe<string | ILogMessage>)
   notify?: INotification | INotification[] | (() => Maybe<INotification | INotification[]>)
+  webViewId?: string
```
kswenson, line 16:
```suggestion
  notifyTileId?: string
```
From the standpoint of this function, it doesn't matter whether it's a web view. It's the tile that should be notified when we're broadcasting to a single tile rather than all tiles.

v3/src/models/tiles/tile-environment.ts
```diff
@@ -12,7 +12,7 @@ export interface ITileEnvironment {
   sharedModelManager?: ISharedModelManager
   formulaManager?: FormulaManager
   log?: (message: ILogMessage) => void
-  notify?: (message: DIMessage, callback: iframePhone.ListenerCallback) => void
+  notify?: (message: DIMessage, callback: iframePhone.ListenerCallback, webViewId?: string) => void
```
kswenson, line 15:
```suggestion
  notify?: (message: DIMessage, callback: iframePhone.ListenerCallback, targetTileId?: string) => void
```

v3/src/components/web-view/web-view.tsx
```diff
@@ -26,6 +30,68 @@ export const WebViewComponent = observer(function WebViewComponent({ tile }: ITi
       <div className="codap-web-view-iframe-wrapper">
         <iframe className="codap-web-view-iframe" ref={iframeRef} src={webViewModel.url} />
       </div>
+      <WebViewDropOverlay />
     </div>
   )
 })
+
+// The WebViewDropOverlay broadcasts notifications to plugins as the user drags and drops attributes.
+function WebViewDropOverlay() {
```
kswenson, line 39: Even though it's not that large, I think it would be worth moving this into its own file so it's easier to find.

v3/src/lib/dnd-kit/codap-dnd-context.tsx
```diff
@@ -26,6 +31,28 @@ export const CodapDndContext = ({ children }: IProps) => {
...
+    const info = getDragAttributeInfo(e.active)
+    if (info?.dataSet) {
+      const { dataSet, attributeId } = info
+      draggingDataSet.current = dataSet
+      draggingAttributeId.current = attributeId
+      appState.document.applyModelChange(() => {}, {
+        notify: dragStartNotification(dataSet, attributeId)
+      })
+    }
+  }
```
kswenson, line 54: I would move all of this into the `WebViewDropOverlay` using `useDndMonitor` so that it's all unified in one place.

v3/src/models/document/create-document-model.ts
```diff
@@ -62,8 +62,8 @@ export const createDocumentModel = (snapshot?: IDocumentModelSnapshot) => {
     }
 
     // configure notifications
-    fullEnvironment.notify = function(message: DIMessage, callback: iframePhone.ListenerCallback) {
-      document.content?.broadcastMessage(message, callback)
+    fullEnvironment.notify = function(message: DIMessage, callback: iframePhone.ListenerCallback, webViewId?: string) {
+      document.content?.broadcastMessage(message, callback, webViewId)
```
kswenson, line 66:
```suggestion
    fullEnvironment.notify = function(message: DIMessage, callback: iframePhone.ListenerCallback, targetTileId?: string) {
      document.content?.broadcastMessage(message, callback, targetTileId)
```

v3/src/models/document/document-content.ts
```diff
@@ -78,11 +78,11 @@ export const DocumentContentModel = BaseDocumentContentModel
       // complete serialization for each row
       self.rowMap.forEach(row => row.completeSnapshot())
     },
-    broadcastMessage(message: DIMessage, callback: iframePhone.ListenerCallback) {
+    broadcastMessage(message: DIMessage, callback: iframePhone.ListenerCallback, webViewId?: string) {
```
kswenson, line 81:
```suggestion
    broadcastMessage(message: DIMessage, callback: iframePhone.ListenerCallback, targetTileId?: string) {
```

kswenson (approved):
👍 Looks good -- nice work with the drop overlay. I made a bunch of suggestions, but they all should be fairly straightforward.

kswenson (changes requested):
Oops -- clicked the wrong radio button.

v3/src/components/drag-drop/attribute-drag-overlay.scss
tealefristoe: Normal dragging still works, but I'm honestly not totally sure this change didn't have other consequences. I made this change because the `WebViewDropOverlay` needs access to `mouseenter`, `mousemove`, and `mouseleave` events to send notifications, and the dnd overlay was apparently intercepting those events. Is there a less heavy-handed way to give `WebViewDropOverlay` access to the events it needs?

v3/src/components/web-view/web-view.tsx
tealefristoe: Unless an attribute is being dragged, `WebViewDropOverlay` returns `null` 👍

kswenson: 🤦 Right, you are. 😉

v3/src/lib/dnd-kit/codap-dnd-context.tsx
tealefristoe: I put these in a centralized location since `dragstart` and `dragend` notifications go out to all plugins, not just a single plugin like the other notifications handled in this PR. I guess moving them into `WebViewDropOverlay` would mean that each overlay would be in charge of sending the notification to its own plugin, and each would have to keep track of the currently dragging attribute and dataset. Does that sound right?

kswenson: I was thinking they were analogous. In that case, go ahead and leave it where it is.

tealefristoe: Looking into this more closely, I was able to move the code into `WebViewDropOverlay` pretty easily and improve it in a number of ways, like no longer tracking the dragging attribute at all. It does feel a bit strange to have each overlay broadcast the same notification to its own plugin, but overall I think this change is much cleaner 👍

kswenson (approved):
👍 Changes look good

PR #1538, opened 2024-10-07
fix: autocomplete in formula editor should add backticks when appropriate

kswenson:


kswenson:
Merged after sync review with @bfinzer.

PR #1539, opened 2024-10-08
setup github actions for run regression label

scytacki:
This adds two new github workflows:
- `run-regresssion-label.yml` (Regression Label) is triggered when a PR is labeled. If the label is `run regression`, then it looks for the last run of the "CI v3 Regression" workflow and re-runs it. If the "CI v3 Regression" is currently running we cancel it and then re-run it. 
- `v3-regression.yml` (CI v3 Regression) - this checks if the PR associated with the branch that was pushed has the `run regression` label and if it does then the workflow runs all of the Cypress tests in the project. These Cypress tests are organized in a "group" called "Regression Tests". This group shows up in the Cypress Cloud when looking at the test results.

It updates the original `v3.yml` so it only runs a single spec. Because it is running a single spec there is no need for the parallel running of the tests. Currently it is running the `import-v2.spec.ts` spec. We probably want to create a special smoke test spec and run that instead. This spec is organized in a Cypress group called "Smoke Tests".

Finally it updates `ci.yml` (the CODAP v2 file) to ignore the v3 files.

### DIFF
--- a/.github/workflows/run-regression-label.yml
+++ b/.github/workflows/run-regression-label.yml
@@ -1,0 +1,44 @@
+name: Regression Label
+
+on:
+  pull_request:
+    types: [ labeled ]
+jobs:
+  re_run:
+    # only conintue if a push or PR labeled with 'run regression'
+    if: github.event.label.name == 'run regression'
+    runs-on: ubuntu-latest
+    permissions:
+      actions: write
+    steps:
+      - name: Re Run last push
+        # get the last regression run triggered by a pull, and get this run's id
+        # then rerun the run
+        # if the run id can't be found the rerun command will fail which should
+        # fail the job
+        # When the run hasn't finished yet, we try to cancel the run
+        # and wait 30s for it to fnish canceling before trying to re-run it.
+        # If the run isn't complete the rerun command prints this message:
+        # returned: run 6410886572 cannot be rerun; its workflow file may be broken
+        run: |
+          run_id=$(gh run list -e push -b ${{github.head_ref}} -w 'CI v3 Regression' -L 1 --json databaseId -q '.[0].databaseId')
+          run_status=$(gh run view $run_id --json status -q '.status')
+          if [[ "$run_status" != "completed" ]]
+          then
+            echo "run $run_id is $run_status, trying to cancel"
+            gh run cancel $run_id
+            count=0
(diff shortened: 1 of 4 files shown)


.github/workflows/run-regression-label.yml
```diff
@@ -0,0 +1,44 @@
+name: Regression Label
+
+on:
+  pull_request:
+    types: [ labeled ]
+jobs:
+  re_run:
+    # only conintue if a push or PR labeled with 'run regression'
```
kswenson, line 8:
```suggestion
    # only continue if a push or PR labeled with 'run regression'
```

.github/workflows/run-regression-label.yml
```diff
@@ -0,0 +1,44 @@
...
+    permissions:
+      actions: write
+    steps:
+      - name: Re Run last push
+        # get the last regression run triggered by a pull, and get this run's id
+        # then rerun the run
+        # if the run id can't be found the rerun command will fail which should
+        # fail the job
+        # When the run hasn't finished yet, we try to cancel the run
+        # and wait 30s for it to fnish canceling before trying to re-run it.
```
kswenson, line 20:
```suggestion
        # and wait 30s for it to finish canceling before trying to re-run it.
```

.github/workflows/v3.yml
```diff
@@ -66,9 +63,11 @@ jobs:
           wait-on-timeout: 300
           # only record the results to dashboard.cypress.io if CYPRESS_RECORD_KEY is set
           record: ${{ !!secrets.CYPRESS_RECORD_KEY }}
-          # only do parallel if we have a record key
-          parallel: ${{ !!secrets.CYPRESS_RECORD_KEY }}
           browser: chrome
+          # only run a single test
+          # TODO: setup a special test for this purpose
+          spec: cypress/e2e/import-v2.spec.ts
```
kswenson, line 69:
```suggestion
          spec: cypress/e2e/tool-shelf.spec.ts
```
Not that it matters much at this stage, but the tool-shelf spec might be a better choice for a starting smoke test. At least it creates one of every kind of component.

kswenson (approved):
👍 Looks good -- just a couple of small suggestions

nstclair-cc (commented):
lgtm!

PR #1541, opened 2024-10-09
Speed up build

scytacki:
Bring over GitHub Actions improvements from CLUE

## Speed ups
- split the build and jest tasks into 2 jobs
- upload the build result from the build job and download it in the deploy job. This eliminates the need for the deploy job to checkout code, install dependencies and build yet again.
- setup the jest cache which might speed up the jest tests further
- setup npm cache on setup-node step
- get the cores of the machine dynamically, we might be able to bump up the cores on the Jest job to get faster build times.

## GitHub Deployments
This also adds GitHub "deployment" support. This adds a section to the bottom of the PR that includes a link to where the code was deployed. In this case that location is: https://codap3.concord.org/branch/speed-up-build/

### DIFF
--- a/.github/workflows/v3-regression.yml
+++ b/.github/workflows/v3-regression.yml
@@ -59,12 +59,6 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
-      - uses: actions/cache@v4
-        with:
-          path: v3/node_modules/.cache/webpack/
-          key: ${{ github.head_ref || github.ref_name }}-webpack-dev-build
-          restore-keys: |
-            main-webpack-dev-build
       - uses: cypress-io/github-action@v6
         with:
           working-directory: v3
--- a/.github/workflows/v3-regression.yml
+++ b/.github/workflows/v3-regression.yml
@@ -6,6 +6,8 @@ on:
     paths:      # only run this workflow if it contains v3 files
       - 'v3/**' # https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#example-including-paths
       - '.github/workflows/v3-regression.yml'
+      - '.github/workflows/v3.yml'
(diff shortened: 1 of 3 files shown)


scytacki:
Here is the duration of the first build of this PR:
image

And here is the second build after a small change:
image

Note that the build and jest jobs are faster the second time likely due to the caching. Cypress is actually a bit slower the second time. It looks like most of this time is taken up by the webpack build. It looks like 1m50s for this webpack build. Perhaps the files webpack caches from webpack-dev-server is are independent from the files webpack caches when it does a production build. Because the Cypress job and Build job share the same cache key that probably means only one of them "wins" the initial cache upload. Then any secondary runs will get this original upload. So one of the 2 jobs would always be starting from scratch. 

I think we could get more speed up by refining the webpack caching strategy. But I'll put that in a different PR.

scytacki:
So I think the problem is that the Cypress has its own cache of `node_modules` and it is overwriting the restored webpack cache.  I think the Cypress node_modules cache is basically based on the hash of the package.lock file. If this also saved the `.cache/webpack` folder, it might be better than nothing. However it looks like the `node_modules/.cache/` folder is not included in the Cypress cache, because nothing is being restored.  This might be because Cypress does it cache saving before it runs the build process, or maybe it excludes `.` files when saving the cache.

A few ways to fix this:
- disable the Cypress installing and caching. We are already caching the npm dependencies from the previous build step. So we can just add our own caching of the Cypress binary and should get similar performance.
- split the Cypress installing from the test running. We can then restore the webpack cache after Cypress has done its installing. And then run the tests.
- store the webpack cache outside of node_modules

I try to include one of these changes in a separate PR that is all about caching.

kswenson (approved):
👍 Looks good!

PR #1542, opened 2024-10-09
Improve GitHub caching

scytacki:
## General GitHub Cache
- for both eslint and webpack caches we now always save when the actions run on the `main` branch. Saving them is cheap. Builds on main shouldn't happen often. A new branch will use these caches when it is first built, so having them updated will keep these first builds running fast. The first build will then save its own cache. Previously the `main` branch webpack cache had only been saved once and that was year ago, so it wasn't doing doing much good.
- the caches for eslint and webpack have been moved from `v3/node_modules/.cache` to `v3/.cache` folder. The Cypress GitHub action seems to be blowing away node_modules when it restores its own cache. So moving the caches outside prevents this, and it seemed best to just use this folder all of the time not just when running the Cypress GitHub action.
- the Jest cache doesn't follow this pattern, just because it follows the caching recommendation I found before. We might want to change it to be consistent.
- the `npm run clean:cache` script was updated to clean the new location. Note if we find consistent reasons when we have to run `clean:cache`, we should try to update the github cache keys to include hashes of files that trigger this. For example if a change of dependencies breaks the webpack cache, then we should include a hash of `package-lock.json`  in its cache key.

## ESLint
- separate out the lint task into its own job so it can run in parallel
- add caching to the lint job to speed it up.
- move eslint cache to `v3/.cache` for consistency
- use eslint cacheStrategy `content` so it works on GitHub where timestamps change
- skip eslint when running the dev server for Cypress. This cuts a minute off the server startup time.

## Webpack
- remove the branch name from the webpack cache key. GitHub saves the caches in the scope of the branch already so adding it to the key isn't useful. It will automatically look in the "default" branch for the repo if it can't find the cache in the current branch. You can see the caches and their branches here: https://github.com/concord-consortium/codap/actions/caches
- include hash of webpack.config.js in webpack cache key. Webpack will ignore the cache if the config file has changed.
- use "hash" comparison method for webpack cache in order for it to work in the cypress job. Webpack has several categories of files it compares during caching. Two already included the hash check: https://webpack.js.org/configuration/other-options/#snapshot Also note the documentation seems wrong, if both `hash` and `timestamp` are specified webpack checks the timestamp first then checks the hash.

## Jest
- add a prefix to the cache key for Jest, this makes it easier to identify in https://github.com/concord-consortium/codap/actions/caches


kswenson (approved):
👍 Looks good!

>Note if we find consistent reasons when we have to run clean:cache, we should try to update the github cache keys to include hashes of files that trigger this. For example if a change of dependencies breaks the webpack cache, then we should include a hash of package-lock.json in its cache key.

I sometimes encounter cache issues when switching branches locally, particularly if there were local conflicts on the branch switch. I've gotten in the habit of running `start:clean` as a matter of course after switching branches.

PR #1543, opened 2024-10-10
feat: sort case table items by attribute [PT-181890013]

kswenson:
[[PT-181890013]](https://www.pivotaltracker.com/story/show/181890013)

- Required some interesting details including a custom undo/redo implementation.
- Adds `caseIdsHash` and `caseIdsOrderedHash` to the `CollectionModel`.
- Adds `hashOrderedStringSet` function to `js-utils`.
- Code for comparing values largely ported from v2.

### DIFF
--- a/v3/src/utilities/data-utils.ts
+++ b/v3/src/utilities/data-utils.ts
@@ -1,4 +1,5 @@
-import { isDate } from "mathjs"
+import { isDateString, parseDate } from "./date-parser"
+import { isDate } from "./date-utils"
 
 export const valueToString = function (iValue: any): string {
   let valType: string = typeof iValue,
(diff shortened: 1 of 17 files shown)


v3/src/utilities/data-utils.ts
```diff
@@ -42,3 +43,67 @@ export const numericSortComparator = function ({a, b, order}: ICompareProps): nu
...
+    /* istanbul ignore next */
+    default: return kTypeUnknown
+  }
+}
+
+export function sortableValue(value: any) {
+  const type = typeCode(value)
+  let num = type === kTypeNumber ? value : NaN
+  // strings convertible to numbers are treated numerically
+  if (type === kTypeString && value.length) {
```
scytacki, line 76: Shouldn't this check if the attribute has a manually set type? For example if the attribute type is string then I'd expect sorting `["001","001a"]` to treat the `"001"` as a string. If I'm reading the logic right here that array would end up reversed without taking into account the attribute type.

v3/src/utilities/js-utils.ts
```diff
@@ -173,3 +173,15 @@ export function hashStringSets(stringSets: Array<string[]>) {
...
     .reduce((acc, hash) => acc ^ hash, 0) // XOR all individual hashes
 }
+
+/*
+ * hashOrderedStringSet()
+ *
+ * returns an order-dependent hash value for a set of strings (e.g. ids).
+ * developed with the help of ChatGPT.
+ */
+export function hashOrderedStringSet(strings: string[]) {
```
scytacki, line 183: If we find this isn't working well (speed or accuracy) we could try doing `hashString(JSON.stringify(strings))`

scytacki (approved):
This looks good. Perhaps it should be expanded to take into account an attribute type when doing the sort. But I'd guess that is not something v2 does, so it should probably be a different story/PR.

kswenson:
> This looks good. Perhaps it should be expanded to take into account an attribute type when doing the sort. But I'd guess that is not something v2 does, so it should probably be a different story/PR.

Interesting thought, but this is the way v2 handled it and changing it would require a larger conversation.

v3/src/utilities/js-utils.ts
kswenson: I would expect the `JSON.stringify` version to be slower, but it'd be interesting to check.

PR #1546, opened 2024-10-11
fix: graph filter function in restored documents

kswenson:
Fix order-of-initialization issue which prevented filter functions from working in graphs restored from documents.

### DIFF
--- a/v3/src/components/graph/models/graph-content-model.ts
+++ b/v3/src/components/graph/models/graph-content-model.ts
@@ -6,16 +6,17 @@ import {cloneDeep} from "lodash"
 import {when} from "mobx"
 import {addDisposer, IAnyStateTreeNode, Instance, SnapshotIn, types} from "mobx-state-tree"
 import { format } from "d3"
-import {mstAutorun} from "../../../utilities/mst-autorun"
-import {applyModelChange} from "../../../models/history/apply-model-change"
-import {ISharedModel} from "../../../models/shared/shared-model"
-import {SharedModelChangeType} from "../../../models/shared/shared-model-manager"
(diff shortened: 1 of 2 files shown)


v3/src/components/graph/models/graph-content-model.ts
```diff
@@ -6,16 +6,17 @@ import {cloneDeep} from "lodash"
...
+import {applyModelChange} from "../../../models/history/apply-model-change"
 import {
   getDataSetFromId, getSharedCaseMetadataFromDataset, getTileCaseMetadata, getTileDataSet
 } from "../../../models/shared/shared-data-utils"
+import {ISharedModel} from "../../../models/shared/shared-model"
+import {SharedModelChangeType} from "../../../models/shared/shared-model-manager"
+import {ITileContentModel} from "../../../models/tiles/tile-content"
+import { getFormulaManager } from "../../../models/tiles/tile-environment"
+import {typedId} from "../../../utilities/js-utils"
+import {mstAutorun} from "../../../utilities/mst-autorun"
```
kswenson, line 19: Rearranged imports into groups by path.

scytacki (approved):
Makes sense to me. I'll probably want to learn more about this adapter pattern at some point, but not today.

kswenson:
> Makes sense to me. I'll probably want to learn more about this adapter pattern at some point, but not today.

The adapter pattern was implemented by Piotr. It allows different formulas (attribute formulas, filter formulas, plotted value formulas, plotted functions, etc.) to be evaluated in different evaluation contexts with access to different values (attributes, globals, etc.).

PR #1549, opened 2024-10-15
Fix: don't add current level case attribute info unless necessary.

lublagg:
This PR addresses an issue caught by @nstclair-cc in [this](https://www.pivotaltracker.com/story/show/188391026/comments/242757237) story.


kswenson (approved):
👍 LGTM

PR #1551, opened 2024-10-17
182079220 function descriptions

eireland:
- Adds `Insert Values` and `Insert Functions` buttons
- Opens scrollable list of values when `Insert Values` button is clicked. Note that boundary values are currently hard-coded. I don't think they were in V2. When a value is clicked, it is added to the formula editor at the current cursor location, or at the current selected text
- Opens a navigatable list when `Insert Functions` is clicked. Goes from categories, to list of functions available for each category, to information about the function. When function name is clicked, it is added to the formula editor at the current cursor location, or at the current selected text.
- Both list are dynamically positioned based on location of modal and screen size.

### DIFF
--- a/v3/src/components/common/formula-insert-values-menu.tsx
+++ b/v3/src/components/common/formula-insert-values-menu.tsx
@@ -1,6 +1,6 @@
 import { TriangleDownIcon, TriangleUpIcon } from "@chakra-ui/icons"
 import {Divider, Flex, List, ListItem,} from "@chakra-ui/react"
-import React, { useRef } from "react"
+import React, { useEffect, useRef, useState } from "react"
 import { useDataSetContext } from "../../hooks/use-data-set-context"
 import { getGlobalValueManager, getSharedModelManager } from "../../models/tiles/tile-environment"
 import { useFormulaEditorContext } from "./formula-editor-context"
@@ -8,7 +8,7 @@ import { useFormulaEditorContext } from "./formula-editor-context"
 import "./formula-insert-menus.scss"
 
 const kMenuGap = 3
-const kMaxHeight = 570
+const kMaxHeight = 470
 
 interface IProps {
   setShowValuesMenu: (show: boolean) => void
@@ -34,6 +34,9 @@ export const InsertValuesMenu = ({setShowValuesMenu}: IProps) => {
      "US_puma_boundaries", "US_state_boundaries", "country_boundaries" ]
   const constants = ["e", "false", "true", "π"]
   const scrollableContainerRef = useRef<HTMLUListElement>(null)
+  const [listContainerStyle, setListContainerStyle] = useState({})
+  const [, setScrollPosition] = useState(0)
+
   let maxItemLength = 0
 
   const insertValueToFormula = (value: string) => {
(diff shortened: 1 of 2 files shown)


v3/src/components/case-tile-common/attribute-menu/attribute-menu.scss
```diff
@@ -20,8 +22,198 @@
...
   }
 }
+
+.formula-editor-button {
+  padding: 0 10px;
+  &:hover {
+    background: $codap-teal-lighter !important;
+  }
+  &:active, &.menu-open {
+    background: $codap-teal-light !important;
```
tealefristoe, line 44: I didn't check to make sure it's possible, but we should really avoid `!important` whenever we can. It can cause a lot of confusion and frustration down the line.

v3/src/components/case-tile-common/attribute-menu/edit-formula-modal.tsx
```diff
@@ -9,6 +8,9 @@ import { updateAttributesNotification, updateCasesNotification } from "../../../
 import { t } from "../../../utilities/translation/translate"
 import { FormulaEditor } from "../../common/formula-editor"
 import { CodapModal } from "../../codap-modal"
+import { InsertFunctionMenu } from "./formula-insert-function-menu"
+import { InsertValuesMenu } from "./formula-insert-values-menu"
+import { clsx } from "clsx"
```
tealefristoe, line 13: Move up ⬆️

tealefristoe (changes requested):
I added the `run regression` tag to this PR, and it looks like a bunch of cypress tests are failing. I was hoping to get a better feel for patch coverage on this PR, which was at about 15% before the tag was added. Hopefully that number will be much higher once the tests are all passing, but if not it would be good to add a few tests to ensure that this code doesn't crash at the very least.

v3/src/components/case-tile-common/attribute-menu/attribute-menu.scss
eireland: These were to ChakraUI's styling

kswenson (approved):
👍 Looks good -- I added a couple of commits which:
- adds a `FormulaEditorContext` which
  - provides access to the underlying editor
  - provides methods for inserting strings directly
- moves common files into a common folder
- eliminates the need for `!important` in the .scss
- fixes a couple of bugs

v3/src/components/common/formula-insert-values-menu.tsx
```diff
@@ -0,0 +1,190 @@
...
+                {boundary}
+              </ListItem>
+            )
+          })}
+          { globals.map((global) => {
+            return (
+              <ListItem key={global.label} className="formula-operand-list-item"
+                    onClick={() => insertValueToFormula(global.label)}>
+                {global.label}
+              </ListItem>
```
tealefristoe, line 169: This pattern is repeated five times in this component. It would be better to make a separate `InsertValuesMenuListItem` component rather than duplicating the same code this much.

tealefristoe (commented):
I see you've already merged this PR. But since I spent a couple of hours this morning looking over the code and writing suggestions, I figured I'd at least share them.

kswenson:
@eireland In my review I was focused most closely on the editor interactions. I would consider @tealefristoe's suggestions even though the PR has already been merged.

PR #1552, opened 2024-10-17
Forces codap-data-cell color to selection color if row is selected

eireland:


v3/src/components/case-table/case-table.scss
```diff
@@ -266,6 +266,11 @@ $table-body-font-size: 8pt;
           background-color: #D3E9C8;
         }
       }
+      &[aria-selected=true] {
+        .codap-data-cell {
+          background-color: var(--rdg-row-selected-background-color) !important;
```
kswenson, line 271: We should try to avoid use of `!important`. I ran it locally to see why it's necessary and I think it can just be removed in this case.

kswenson (changes requested):
👍 Looks good, but I think you can remove `!important`.

PR #1553, opened 2024-10-17
Map grid display bug fix

bfinzer:
[#188104342] Bug Fix: Grid does not display in Four Seals

* The bug occurs only for restored codap3 documents and is brought about by the way a MapGridModel initializes its latlngGrid property at a moment when the map's data configuration has not yet computed the cases and thus the grid has zero rectangles. We fix this by adding the data configuration's caseDataHash to the things it observes so that it can initialize things properly when the has changes. We also only attempt the initialization if the grid is visible.
* I discovered that the display of grid selection was not synchronized with selection of cases because of an order of operations issue. This is fixed by having the map grid model increment a change count that can be observed by the MapPointGrid component which was previously checking for dataset selection actions.
* Hover over map points was not displaying a caption and this was because 'lat' and 'long' were not included among the attributes to check in the data configuration model's `defaultCaptionAttributeID` method. So we add those.

### DIFF
--- a/v3/src/components/data-display/models/data-configuration-model.ts
+++ b/v3/src/components/data-display/models/data-configuration-model.ts
@@ -119,7 +119,7 @@ export const DataConfigurationModel = types
       const defaultCaptionAttributeID = () => {
         // We find the childmost collection and return the first attribute in that collection. If there is no
         // childmost collection, we return the first attribute in the dataset.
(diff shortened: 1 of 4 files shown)


kswenson (approved):
👍 LGTM -- just one suggested simplification

PR #1555, opened 2024-10-21
feat: dynamic locale switching

kswenson:
In combination with [CFM #359](https://github.com/concord-consortium/cloud-file-manager/pull/359), this PR support dynamic language switching using the CFM language menu. Notable features include:
- new `Locale` class for maintaining current locale and providing locale-related API
- single global instance of `Locale` class (`gLocale`)
- clients that were calling `getDefaultLanguage()` now use `gLocale`
- adds missing Korean strings file
- synchronizes language listings between CFM and CODAP


scytacki (approved):
Looks good to me.

PR #1556, opened 2024-10-21
Add Targeted Cypress Smoke Test for Developer Pushes Without Run Regression Label

nstclair-cc:
[PT-#188412560](https://www.pivotaltracker.com/story/show/188412560)

### Description:
This PR introduces a new targeted Cypress smoke test that checks for essential functionality when developers push code without including the `run regression` label. Below are the key details of the changes included:

### Changes Made:
- **New Folder Structure:**
	•	Added a /smoke/ folder under Cypress to keep the smoke tests organized. This mirrors the structure used in the CLUE project and will help with future scalability if more smoke tests are needed.
- **New Smoke Tests:**
	1.	**Mammals Dataset Test:**
	•	Verifies the Mammals dataset opens and contains data.
	•	Checks the ability to create a hierarchy within the table.
	•	Includes a test to create a graph with both x and y axes.
	2.	**Four Seals Dataset Test:**
	•	Verifies the dataset opens and checks a table and graph with date attributes.
	•	Confirms that the Map component opens successfully.
	3.	**Empty Document Test (Skipped):**
	•	This test checks for an empty document but is skipped for now since document opening is already covered in a Jest test.
	•	Logged a related issue while working on this: [#188439366](https://www.pivotaltracker.com/story/show/188439366).

### Questions/Requests:

**Cypress Setup for Pushes:**
I’m unsure how to configure Cypress to automatically run these smoke tests on every push without the `run regression` label. This might need to be addressed in a separate story—perhaps Scott or Kirk could assist with that setup?
	•	I’d also love to learn how this process works, so please mark me as a reviewer on the PR when it gets set up!

### Testing Notes:
- I added comments in the code to document any minor issues I encountered during testing.
- The smoke tests are passing locally and took my machine 17 seconds to run through, and I believe the implementation is ready for review.

### Next Steps:

- If additional configuration is needed to run these smoke tests automatically, we can create a follow-up story.
- There's a lint error I see, `/Users/nstclair/ConcordConsortium/codap/v3/src/components/case-table/use-columns.tsx
  15:8  warning  Using exported name 'clsx' as identifier for default export  import/no-named-as-default`, which I believe may be tied to the datatest-id I was hoping to add in the code. Feedback on this issue is welcome!
- Feedback on the structure or test coverage is welcome!

Let me know if there are any questions or if further changes are needed. Thanks very much!

### DIFF
--- a/v3/cypress/e2e/smoke/single-codap-smoke.ts
+++ b/v3/cypress/e2e/smoke/single-codap-smoke.ts
@@ -8,6 +8,7 @@ import { CalculatorTileElements as calculator } from "../support/elements/calcul
 import { WebViewTileElements as webView } from "../support/elements/web-view-tile"
 
 const helpURL = 'https://codap.concord.org/help'
+const helpURL_ja = 'https://codap.concord.org/resources/latest/help-documents/CODAP解説書.pdf'
 const helpForumURL = 'https://codap.concord.org/forums/forum/test/'
 const projectWebSiteURL = 'https://codap.concord.org'
 const privacyURL = 'https://codap.concord.org/privacy'
--- a/v3/cypress/e2e/smoke/single-codap-smoke.ts
+++ b/v3/cypress/e2e/smoke/single-codap-smoke.ts
@@ -62,9 +62,9 @@ context("codap single smoke test", () => {
         cy.log("verify selecting cases in the table")
           table.getGridCell(2, 2).should("contain", "African Elephant").click()
           table.getGridCell(2, 2).should('have.attr', 'aria-selected', 'true')
-        // TODO: Add more thorough checks to make sure graph points and table rows actually change color when selected,
-        // once Cypress is configured to interact with the PixiJS canvas. For now, we just check that the buttons in table
-        // are enabled upon selection.
+        // TODO: Add more thorough checks to make sure graph points and table rows actually
+        // change color when selected, once Cypress is configured to interact with the PixiJS canvas.
+        // For now, we just check that the buttons in table are enabled upon selection.
 
         cy.log("test graph axis functionalities in Mammals sample doc")
         // it is possible to create a graph
--- a/v3/cypress/e2e/smoke/single-codap-smoke.ts
+++ b/v3/cypress/e2e/smoke/single-codap-smoke.ts
@@ -17,177 +17,175 @@ context("codap single smoke test", () => {
 
   it("verify Mammals opens from Hamburger menu and tests graph and table functionality", () => {
     cy.log("Open Mammals from Hamburger menu")
-      // hamburger menu is hidden initially
-      cfm.getHamburgerMenuButton().should("exist")
-      cfm.getHamburgerMenu().should("not.exist")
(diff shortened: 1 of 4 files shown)


scytacki (changes requested):
I added the change to run this test instead of the `import-v2.spec.ts` that was being run before.

This looks good to me. 

My only change request is the indentation of the test. I see that you are using indentation to group the various parts of of a large test. However if the file is automatically formatted this would be removed and some lint settings might complain about it.

One way to address this would be to move the different parts into separate functions. So the single `it(..)` block can call a few functions which are used to group the cypress commands. 

Alternatively you could create a utility function like:
```
function cyGroup(msg: string, groupedCommands: () => void) {
  cy.log(msg)
  groupedCommands()
}
```

And then you'd use it like this:
```
cyGroup("Open Mammals from Hamburger menu", () => {
  // hamburger menu is hidden initially
  cfm.getHamburgerMenuButton().should("exist")
  cfm.getHamburgerMenu().should("not.exist")
  // hamburger menu is shows when button is clicked
  cfm.getHamburgerMenuButton().click()
  ...
})

File: v3/cypress/e2e/smoke/single-codap-smoke.ts [hunk 1]
```diff
@@ -0,0 +1,193 @@
...
+
+    cy.log("checks map component")
+      c.getComponentTitle("map").should("have.text", "Measurements")
+      // Some day add a check for map with points, see PT #187534790
+      // Also some day add a legend to the map.
+
+  })
+  // NOTE: This test was drafted on 10/16/2024 but will be skipped since we decided we don't need to
+  // smoke test empty CODAP documents
+  it.skip("verify an empty CODAP document appears", () => {
```
kswenson, line 146: I think it's fine to include (i.e. un-skip) this as it's really a test that each individual component can be created from the tool shelf rather than a test that an empty document can be loaded.

kswenson (approved):
👍 I agree with @scytacki's comment about the indenting and have one other suggestion about including the tool shelf test. I'll approve in the sense that only @scytacki's approval remains a blocker.

nstclair-cc:
Thanks so much for the review!

Scott, I really liked your idea of using a grouping function like cyGroup and experimented with it a bit. I found it helpful for organizing commands, but it started creeping out of scope as I worked through the tests. In the end, I took the easier path and removed the extra indents to align with the formatting standards you mentioned. I hope that makes sense!

Also, I think it would be a good idea to add a chore in the future to clean up the `it` statements across other Cypress tests, reducing the test count. We could also explore incorporating cyGroup into the architecture as part of that effort.

@scytacki - With apologies, I believe I accidentally pushed the code over `import-v2.spec.ts`. I’m really sorry about that! Would it be easy for you to push your changes again, or would you prefer I re-add it myself? Just let me know what’s easiest.

I think we’re good to go with this last change, but if there are any other issues, let me know!

scytacki:
No problem, I just added back in the line to switch the smoke test. I'll approve and merge if it builds.

scytacki (approved):
Looks good

PR #1557, opened 2024-10-22
Hide CFM menu inside of the Activity Player

scytacki:
The CFM menu is hidden when the `interactiveApi` URL param is present.

The interactiveApi url param is added to the URL when a "Interactive API" share link is copied from the CFM share dialog.

The CODAP v2 code also checked for older url params `lara` and `launchFromLara`. https://github.com/concord-consortium/codap/blob/9b54cf1415a6c342b743adc8922ec12c8813e3fa/apps/dg/core.js#L354-L359

I checked and there are no URLs in the LARA authoring system which use those older params, so I didn't include support for them in the CODAPv3 code.


kswenson (approved):
👍 Looks good! In practice the `hideMenuBar` option in the CFM config seems to hide the hamburger menu rather than the menu bar, but that's a preexisting condition.

PR #1560, opened 2024-10-24
188363563 Formula Editor API Requests

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/188363563

This PR adds support for the `notify attribute` with `request: formulaEditor` API request, which displays the formula editor from a plugin.

This required more work than expected. The main problem is that, previously, each table and case card had a separate `EditFormulaModal` for each attribute. This was awkward for a couple of reasons: the API handler had to use another tile's modal, and if there was no table or case card associated with the given attribute, there was no modal to show at all. To fix this issue, a single modal was added to the `Container`, which any tile can use for any attribute via `uiState.editFormulaAttributeId`.
- The table and case card tiles use this new method, which simplifies both and reduces the total number of components.
- The `EditFormulaModal` is no longer wrapped in a `DataSetContext`, so it needs to find the dataset that contains the specified attribute, and then provides its own `DataSetContext` to its children.
- `EditFormulaModal` has been moved from `case-tile-common` to `common`, since it is no longer tied to tables and case cards.
- The only other minor change to `EditFormulaModal` is that the `useEffect` that refreshes the display formula now triggers when the `attribute` changes.

This work can be tested with this Codap link: https://codap3.concord.org/branch/v3-di-formula-editor-request/#file=examples:Mammals
And this Multidata link: https://models-resources.concord.org/multidata-plugin/branch/display-formula-editor/index.html


kswenson (approved):
👍 Looks good -- nice refactor!

PR #1563, opened 2024-10-25
Prevent first map/graph click from clearing selection

bfinzer:
[#187953007] Bug Fix: First click in graph/map empty space does not change selection

The trick for both graph and map is to distinguish between a window event and a PointerDown event sent by PixiPoints. We create use-pointer-down-capture.ts to encapsulate the process and call usePointerDownCapture from background.tsx and map-component.tsx


kswenson (approved):
👍 Looks good -- I added a commit which limits the behavior to the canvases of the current component rather than any canvas element and did a bit of refactoring.

PR #1564, opened 2024-10-26
CFM dirty document support

scytacki:
## revisionId
CODAP now tracks a document revisionId. This serves the same purpose as the changeCount in CODAPv2 documents. The revisionId is the nanoid of the most recent undoable history event.  As a nanoid it should be more resilient if there are multiple people editing a document by at the same time. That isn't supported but could happen. A hash of the document content would be the best, but it is expensive to generate a hash every time we need. Particularly because we serialize the dataset into the document on demand.

The revisionId is stored outside of the document's MST model, so we don't have to add special handling to prevent changes to it from causing an infinite loop. It is stored in the TreeManager.

The revisionId is serialized inside of the document content that is sent to the CFM. Ideally the CFM would be modified to be aware of the revisionId in order to simplify what clients have to do. However CODAP configures the CFM to use unwrapped content so CODAP would still have to implement the storage of it. So it didn't seem worth the effort to update the CFM.

## Cypress Updates
Now that the document is being marked as dirty, a dialog is shown when closing a dirty document either directly or indirectly when opening a new one. This change required updating some Cypress tests to dismiss this dialog. These updated Cypress tests are basically testing the dirty document feature.

The table Cypress spec was updated to check that selecting a cell doesn't dirty the document, and editing a cell does dirty the document. 

Some features listed in the story are not tested:
- checking that tile selection and scrolling do not dirty the document
- Saving should clear the dirty flag
- Auto-save regularly saves, clearing the dirty flag for supported providers. 

## DEBUG_DOCUMENT
Support for the `document` debug flag was added to match the behavior in CLUE. When a document is opened and this flag is set, then `window.currentDocument` is set to the document. This makes it easier to inspect the document and history in the browser console.

## Dirty document monitoring
The `revisionId` is updated by the treeManager whenever a new undoable history entry is recorded. If the action has been flagged with `withoutUndo` the `revisionId` will not be change.

The AppState monitors the `revisionId` for changes with a MobX reaction and then tells the CFM the document is dirty when it changes. The initial setting of the `revisionId` is done before the MobX reaction is added.

## prepareSnapshot and completeSnapshot
These actions exist on many CODAP models and they are problematic for dirty document tracking and document history  (if CODAP ends up supporting history). The solution in this PR is to update the treeMonitor to just ignore actions with these names. That is fragile, but works for now.

A better solution would be to replace these actions with snapshot post processing. Using `types.snapshotProcessor` the post processor has access to the model instance being snapshotted so it can access the volatile state. 

## Removing the changeCount
The MST Document model had a `changeCount` property that wasn't being used in CODAPv3. Since this is redundant with the revisionId I got rid of it. 

## Enable autosave in the CFM config
It is set to 5 seconds just like it was in CODAPv2

## Send content unwrapped
It was being sent wrapped: `({ content })`. This was working in most cases but only because we were fooling the CFM. So now it is sent unwrapped: `(content)`. This fixed an issue in the dirty document handling inside of the Activity Player. 

This was mostly working before because the CFM was thinking **it** had wrapped this content instead of us. So then in the case of local file saving and shared documents the CFM would unwrap the content before saving because of the `wrapFileContent: false` setting. However the interactiveAPI provider doesn't work the same way and was not unwrapping when saving. So the documents were being saved wrapped in the Activity Player. These wrapped documents still worked because the CFM would unwrap the documents when loading them back in. However because the CFM was expecting the document to be unwrapped to start with, the CFM would mark the updated document as dirty since it just "modified" the document by unwrapping it. This marking as dirty would happen every time the document was opened because it was always saved wrapped again. A way to find this code is to search for `requiresConversion` in the CFM. The fix to this was to just send the content unwrapped from CODAP to the CFM. This is how the CFM expects to get the content whether it is configured to be wrapped or unwrapped.

## Slow saving
If the CFM provider takes a while to save the file, the user can change the document during the save. CODAP will tell the CFM the document is dirty. However when the CFM finishes saving the original file it will automatically mark its internal dirty state to be false (thinking it just saved the file so it is no longer dirty). Because of this the change that happened during the save can be lost. This issue was worked around by CODAPv2 by handling the "savedFile" event using its changeCount and savedChangeCount. Here in CODAPv3 we do a similar thing by handling the "savedFile" event. We compare the revisionId of the content that was saved by the CFM with the current documents revisionId. If they are different then tell the CFM the document is dirty again.

A better solution to this would be to send the revisionId to the CFM as an argument to the `dirty()` function. The the CFM could know itself when it was done saving that the revision it saved is different than the current revision. So then clients wouldn't have to handle "savedFile" event and the dirty flag wouldn't flip-flop.

## CFM Local Storage support
A new debug flag `cfmLocalStorage` was added. This way developers and testers can enable the local storage provider. It makes it possible to test the autosave functionality better. With modifications to the CFM it is also possible to add an artificial delay to the local storage provider in order to test the slow saving behavior.

We might want to be able to configure multiple aspects of the CFM. Perhaps disabling autosave, or changing the autosave  delay. It seems unsafe to allow full configuration of the CFM based on browser configuration, but maybe I'm just being overly paranoid.

TODO:
- [x] add documentation about the debug flags and how to access the history from the console
- [x] figure out why mammals document is becoming dirty when selecting a cell of the table

### DIFF
--- a/v3/README.md
+++ b/v3/README.md
@@ -130,7 +130,7 @@ Inside of your `package.json` file:
 
 Various developer features can be enabled by adding a `debug` local storage key with one or more of the following flags separated by spaces. Local storage is specific to the domain that CODAP is running on. In Chrome local storage can be edited by opening the developer tools and going to the "Application" tab. Then find "Storage/Local storage" and the domain that CODAP is running on.
 
-- `cfmLocalStorage` enable the CFM local storage provider so documents can be saved and loaded from the browsers local storage
+- `cfmLocalStorage` enable the CFM local storage provider so documents can be saved and loaded from the browser's local storage
 - `document` this will add the active document as `window.currentDocument`, you can use `currentDocument.toJSON()` to views the current documents content. You can also use `currentDocument.treeManagerAPI.document.toJSON()` to inspect the history of the document.
 - `formulas` print info about recalculating formulas
 - `history` this will: print some info to the console as the history system records changes, print the full history as JSON each time it is loaded from Firestore, and provide a `window.historyDocument` so you can inspect the document while navigating the history.
--- a/v3/README.md
+++ b/v3/README.md
@@ -128,6 +128,17 @@ Inside of your `package.json` file:
 
 ### Debugging
 
+Various developer features can be enabled by adding a `debug` local storage key with one or more of the following flags separated by spaces. Local storage is specific to the domain that CODAP is running on. In Chrome local storage can be edited by opening the developer tools and going to the "Application" tab. Then find "Storage/Local storage" and the domain that CODAP is running on.
+
+- `cfmLocalStorage` enable the CFM local storage provider so documents can be saved and loaded from the browsers local storage
+- `document` this will add the active document as `window.currentDocument`, you can use `currentDocument.toJSON()` to views the current documents content. You can also use `currentDocument.treeManagerAPI.document.toJSON()` to inspect the history of the document.
+- `formulas` print info about recalculating formulas
+- `history` this will: print some info to the console as the history system records changes, print the full history as JSON each time it is loaded from Firestore, and provide a `window.historyDocument` so you can inspect the document while navigating the history.
+- `logger` console log all messages sent to the logging service
+- `map` print info about interactions with the map component
+- `plugins` enable some extra plugins in the plugin menu and print information about interactions with plugins.
+- `undo` this will print information about each action that is added to the undo stack.
+
 ##### Why Did You Render?
 
 To debug React rendering problems, the [why-did-you-render](https://www.npmjs.com/package/@welldone-software/why-did-you-render) utility is available. To use it:
--- a/v3/cypress/e2e/table.spec.ts
+++ b/v3/cypress/e2e/table.spec.ts
@@ -1071,7 +1071,7 @@ context("case table ui", () => {
 
       cy.log("selecting the cell doesn't dirty the document")
       table.getGridCell(2, 2).click()
-      // check for the "UNSAVE" badge
+      // check for the "UNSAVED" badge
       // The badge will not be there before the selection, so we need to wait
       // to give some time for buggy code to add the badge.
       cy.wait(100)
--- a/v3/cypress/e2e/table.spec.ts
+++ b/v3/cypress/e2e/table.spec.ts
@@ -28,10 +28,11 @@ context("case table ui", () => {
   })
 
   describe("table view", () => {
-    it("populates title bar from sample data", () => {
+    it("has a valid title bar and column headers", () => {
+      cy.log("populates title bar from sample data")
       c.getComponentTitle("table").should("contain", collectionName)
-    })
-    it("verify columns and tooltips", () => {
+
+      cy.log("verify columns and tooltips")
       // css width specification caused grid virtualization to only have 9 attributes in the DOM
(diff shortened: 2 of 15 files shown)


v3/cypress/e2e/table.spec.ts
```diff
@@ -1064,17 +1065,33 @@ context("case table ui", () => {
...
   describe("table cell editing", () => {
-    it("edits cells with color swatch", () => {
+    it("edits cells with text and color swatches", () => {
       cy.log("checking cell contents")
       table.getGridCell(2, 2).should("contain", "African Elephant")
 
-      cy.log("double-clicking the cell")
+      cy.log("selecting the cell doesn't dirty the document")
+      table.getGridCell(2, 2).click()
+      // check for the "UNSAVE" badge
```
kswenson, line 1074:
```suggestion
      // check for the "UNSAVED" badge
```

v3/src/lib/handle-cfm-event.ts
```diff
@@ -37,7 +37,7 @@ export function handleCFMEvent(cfmClient: CloudFileManagerClient, event: CloudFi
     //   break
     case "getContent": {
       appState.getDocumentSnapshot().then(content => {
-        event.callback({ content })
+        event.callback(content)
```
kswenson, line 40: 🤦

v3/src/models/app-state.ts
```diff
@@ -44,20 +52,53 @@ class AppState {
...
   async getDocumentSnapshot() {
     // use cloneDeep because MST snapshots are immutable
-    return await serializeDocument(this.currentDocument, doc => cloneDeep(getSnapshot(doc)))
+    const snapshot = await serializeDocument(this.currentDocument, doc => cloneDeep(getSnapshot(doc)))
+    const revisionId = this.treeManager?.revisionId
+    if (revisionId) {
+      return {
+        revisionId: this.treeManager?.revisionId,
+        ...snapshot
+      }
```
kswenson, line 80:
```suggestion
      return { revisionId, ...snapshot }
```

kswenson (approved):
👍 Looks great! Just a couple of small suggestions.

scytacki:
@kswenson I updated this to only dirty the document with 'undoable' history events. It isn't perfect, but it matches the expectations better I think.  I also added the docs about the debug flag and cleaned up some unused debug flags.

v3/README.md
```diff
@@ -128,6 +128,17 @@ Inside of your `package.json` file:
 
 ### Debugging
 
+Various developer features can be enabled by adding a `debug` local storage key with one or more of the following flags separated by spaces. Local storage is specific to the domain that CODAP is running on. In Chrome local storage can be edited by opening the developer tools and going to the "Application" tab. Then find "Storage/Local storage" and the domain that CODAP is running on.
+
+- `cfmLocalStorage` enable the CFM local storage provider so documents can be saved and loaded from the browsers local storage
```
kswenson, line 133:
```suggestion
- `cfmLocalStorage` enable the CFM local storage provider so documents can be saved and loaded from the browser's local storage
```

kswenson (approved):
👍 Looks good

>I updated this to only dirty the document with 'undoable' history events. It isn't perfect, but it matches the expectations better I think.

For posterity, you may want to update the PR description to include this information. It's probably worth moving this information into the `docs` folder at some point as well, but that's not a requirement for merging.

PR #1565, opened 2024-10-27
187738926 Broadcast Sort Attribute Notifications

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187738926

This PR fixes a bug in the Multidata plugin, where the plugin was not updating properly when an attribute was sorted. This is because notifications were never being broadcast when an attribute was sorted.

Note that in v2 this notification has an empty array for `caseIDs`. I kept this in v3, since doing anything more would require more work. Like with many notifications, plugins seem to just care that a notification was broadcast, not any of the specific information in the notification, relying on subsequent API requests to get details about the state of Codap after the change.


kswenson (approved):
👍 LGTM

PR #1566, opened 2024-10-28
187223531 Restore and Replace setAside API Requests

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187223531

This PR adds support for the `operation` value for `notify dataContext setAside` API requests.
- With no `operation`, `caseIDs` are required and will be added to the setAside cases.
- With `operation: replace`, `caseIDs` are required and the existing setAside cases will be cleared before making the specified cases setAside.
- With `operation: restore`, `caseIDs` are not required, and cases are removed from the setAside.
  - If `caseIDs` is specified, those cases will be removed.
  - If no `caseIDs` are specified, all cases will be removed.
    - Note that this is the same as the existing `notify dataContext restoreSetaside` API request, but was added to this new system for consistency. The old API request still works as it did before, though it is now case insensitive, so it supports `notify dataContext restoreSetAside`.
    
While working on this story, I realized there were some problems with the existing setAside implementations.
- Cases should be selected after being restored. This was happening via the Codap menu, but it wasn't happening via API requests.
- Various notifications should be broadcast after setAsides change, and this wasn't happening via the Codap menu or API requests previously.
  - A `deleteCases` notification should be broadcast when cases are set aside.
  - A `createCases` notification should be broadcast when set aside cases are restored.
  - A `selectCases` notification should be broadcast when the selection is modified (either selected cases are set aside or restored cases are selected).
  
To keep the Codap menu options consistent with the behavior of the API handler, utility functions were created that are used in both of these places.


kswenson (approved):
👍 Looks good

PR #1567, opened 2024-10-28
Fix bug whereby wrong number of cases was being displayed in graph

bfinzer:
[#188460088] Bug Fix: Dot chart not respecting number of cases when attribute is not at childmost level

* In data-configuration-model.ts we compute the childmostCollectionIDForPlottedAttributes and this enables us to set up a reaction that clears the filteredCases in response.
* We clean up data-configuration-model.ts and graph-data-configuration-model.ts a bit, getting rid of unused functions and tidying up parameters


kswenson (approved):
👍 Looks good!

PR #1568, opened 2024-10-28
183987088 graph color picker UI

eireland:
This adds the UI for the graph and map color picker for data points.
Included:
- 16 palette color swatches + 1 swatch if user selects from the color input picker.
- More/Less button that shows/hides a color input (using the same color input picker that the case table uses)
- Checkmark on the color swatch when user selects a color.

Note that some of the color selection functionality is present, but the focus of this PR is just to add the UI.


bfinzer:
I get four warnings when I run this. Please fix.

bfinzer:
I notice that in dragging the point size slider the points respond with animation instead of instantly as they do in main. Instant is better. Please fix or log as separate bug.

bfinzer (approved):
Looks good. Just fix warnings and figure out whether you can fix the  point size animation bug of log as a PT story.

kswenson (approved):
👍 LGTM

kswenson:
> 👍 LGTM

Forgot to mention in the comment that the `colorHash` implementation should be changed as described above.

PR #1570, opened 2024-10-29
Fix Hide Unselected Cases for maps and graphs

bfinzer:
[#188351110] Bug Fix: Hide cases is broken in map

Actually, the bug is that **Hide Unselected Cases** is broken for both maps and graphs in that it hides *all* the cases, not just in the map. 

* The bug was in data-configuration-model.ts `get unselectedCases` which was wrongly assuming that `dataset.selection` is an array of case ids when, in fact it is an array of item ids. So, instead of using it, we call the much more transparent `dataset.isCaseSelected` to filter out the selected cases.
* We also clean up both `get selection` and `get unselectedCases` a bit.


kswenson (approved):
👍 LGTM

PR #1573, opened 2024-10-30
183987088 improve graph color selection

eireland:
Uses the 16+1 color palette with the color picker to change color points in:
1. A graph or map with categorical legend
2. A graph or map with no legend

The following bugs still exist:
- With categorical legend, when the point stroke color is changed, the points all change to their non-categorized color until the graph is rerendered. PT story here https://www.pivotaltracker.com/story/show/188498750


bfinzer (approved):
LGTM

kswenson (changes requested):
👍 Code looks good, but some dependencies should be adjusted:
- `data-display` code should not be importing from `case-tile-common`.
- we know that the next story is implementing the same palette for the case table/card, so the palette shouldn't be in `data-display`.

Therefore,
- `case-tile-common/color-picker` should move to `components/common`. (I probably should have put it there initially.)
- `PointColorSetting` should be renamed something like `ColorPickerPalette` and moved to `components/common` as well, so that it can be used by the case table/card and the graph/map.

The general rule is that components import from common/shared folders, not from each other, and common/shared modules shouldn't import from components.

eireland:
The refactor of the ColorPickerPalette will be done when I do the PR for this [story](https://www.pivotaltracker.com/story/show/187502179) when the case table color picker is changed to use the 16+1 color palette.

kswenson (approved):
> The refactor of the ColorPickerPalette will be done when I do the PR for this [story](https://www.pivotaltracker.com/story/show/187502179) when the case table color picker is changed to use the 16+1 color palette.

👍 In that case, LGTM

PR #1574, opened 2024-10-30
188497515 v3 Plotted Function/Value Formula Editor

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/188497515

This PR makes the Plotted Function and Plotted Value adornments use the `EditFormulaModal` to modify their formulas. In doing so they gain access to autocomplete for variables, functions, and other keywords.
- The `EditFormulaModal` was abstracted from the `EditAttributeFormulaModal`, which is used to modify attribute formulas.
- Plotted Function and Plotted Value adornments now use the general `EditFormulaModal`. Their own special modals of the same name were removed.
- I made a few minor changes to the display of the `EditFormulaModal` to try to make it better match v2.


v3/src/components/common/edit-attribute-formula-modal.tsx
```diff
@@ -0,0 +1,51 @@
...
+  return (
+    <DataSetContext.Provider value={dataSet}>
+      <EditFormulaModal
+        applyFormula={applyFormula}
+        formulaPrompt={t("DG.AttrFormView.formulaPrompt")}
+        isOpen={!!uiState.editFormulaAttributeId}
+        onClose={() => uiState.setEditFormulaAttributeId()}
+        titleInput={attribute?.name}
+        titleLabel={t("DG.AttrFormView.attrNamePrompt")}
+        titlePlaceholder="attribute"
```
kswenson, line 46: Assuming this is user-visible (as placeholders generally are), it should be a localizable string. It's a pre-existing condition, so not necessarily part of this PR, but should be fixed or at least add a TODO comment.

kswenson (approved):
👍 Looks great -- nice refactor!

PR #1575, opened 2024-10-30
fix: restore of attribute formula referencing slider

kswenson:
[[PT-187364079]](https://www.pivotaltracker.com/story/show/187364079)

Fixes a bug which caused the `FormulaManager`'s `globalValueManager` property to only be set on newly created documents.


tealefristoe (approved):
Looks good 👍

PR #1576, opened 2024-10-30
188445579 update dependencies; node compatibility

kswenson:
[[PT-188445579]](https://www.pivotaltracker.com/story/show/188445579)

- node 22 (latest LTS) compatibility seems to have been fixed by uninstalling/reinstalling `jest-environment-jsdom`
- most code changes related to Sass deprecating `@import` in favor of `@use`
- fixes npm-reported vulnerabilities
- fixes lingering lint warnings


v3/src/components/common/formula-insert-values-menu.tsx
```diff
@@ -25,46 +26,46 @@ export const InsertValuesMenu = ({setShowValuesMenu}: IProps) => {
   )
   const attributeNames = dataSet?.attributes.map(attr => attr.name)
   const globalManager = dataSet && getGlobalValueManager(getSharedModelManager(dataSet))
-  const globals = globalManager
-                    ? Array.from(globalManager.globals.values()).map(global => ({ label: global.name }))
-                    : []
+  const globals = useMemo(() => globalManager
+    ? Array.from(globalManager.globals.values()).map(global => ({ label: global.name }))
```
scytacki, line 30: If a new global is added to the globalManager I think the `globals` memo will not be updated. That seems like a bug?

scytacki (changes requested):
Mostly looks good, but I think the computation of the globals will be stale in the InsertValuesMenu.
Perhaps the menu component is re-created every time it is shown so the globals will always get recomputed?

v3/src/components/common/formula-insert-values-menu.tsx
kswenson: Good question -- I'll look into it.

scytacki (changes requested):
I think this has the same problem as before. The dataSet object probably does not change when a new global is added. So the getListContainerStyle callback will not get re-created and will be computing its position based on the old list of globals.

kswenson:
>I think this has the same problem as before. The dataSet object probably does not change when a new global is added. So the getListContainerStyle callback will not get re-created and will be computing its position based on the old list of globals.

"Once more unto the breach..." -- I eliminated the caching of `listContainerStyle` altogether.

scytacki (approved):
Looks good. 👍

PR #1577, opened 2024-10-30
Graph hierarchy with legend bug

bfinzer:
[#188502341] Feature: Dot chart not responding correctly to change in hierarchy if it has a legend.

* We were including the legend attribute, if any, in the set of attributes used to determine the childmost collection. But, in fact, it should not participate in that. With this change, a legend attribute that comes from a child collection when the points being plotted are from a parent collection, should not change the number of cases, and the points should be colored gray, the missing color.


kswenson (approved):
👍 LGTM

PR #1578, opened 2024-10-30
chore: track hidden cases in grouping code

kswenson:
In #1566, @tealefristoe pointed out that working with set-aside cases is problematic because hidden (set-aside or filtered) cases are not fully instantiated. This is because when set-aside was implemented, a simplifying assumption was made that hidden items/cases could be ignored internally in addition to being hidden from the standpoint of external clients. With this PR, hidden items/cases are processed and instantiated like visible items/cases, but are tracked as hidden internally where appropriate. It remains the case that `DataSet.itemIds`, `DataSet.items`, `Collection.cases`, and `Collection.caseGroups` only return visible items/cases to clients. Several internal data structures, e.g. `DataSet.caseInfoMap` and `Collection.caseGroupMap` now contain information about hidden and visible cases, however. Jest tests have been updated to validate this behavior.

### DIFF
--- a/v3/src/models/data/collection.test.ts
+++ b/v3/src/models/data/collection.test.ts
@@ -35,7 +35,7 @@ describe("CollectionModel", () => {
     expect(withNameAndTitle.title).toBe("newTitle")
     expect(isCollectionModel(withNameAndTitle)).toBe(true)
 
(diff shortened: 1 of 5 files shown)


tealefristoe (approved):
Looks great 👍

PR #1580, opened 2024-10-31
188497360 v3 Update Categorical Point Color

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/188497360

Evangeline discovered a bug where changing the stroke color for a graph with dots colored according to a categorical attribute resulted in the dots losing their categorical coloring. It turns out this is because a reaction responding to changes in point properties (such as `pointStrokeColor`) was calling `matchCirclesToData`, which is intended to update the number of circles but does not properly style them. The fix is to instead call `callRefreshPointPositions`, which updates the dots' appearance without changing their number. This function is plot specific, so it's not possible to call it in `useGraphModel`. Instead, the reaction that calls `callRefreshPointPositions` has been moved into `usePlotResponders` in `use-plot.ts`.

### DIFF
--- a/v3/src/components/graph/components/graph.tsx
+++ b/v3/src/components/graph/components/graph.tsx
@@ -339,7 +339,7 @@ export const Graph = observer(function Graph({graphController, graphRef, pixiPoi
     return droppables
   }
 
-  useGraphModel({pixiPoints, graphModel, instanceId})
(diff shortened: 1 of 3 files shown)


bfinzer (approved):
👍 LGTM

PR #1581, opened 2024-10-31
fix: auto-complete of function doesn't add redundant parentheses

kswenson:
[[PT-188484892]](https://www.pivotaltracker.com/story/show/188484892)

When auto-completing a function, only insert trailing parentheses if there don't appear to be parentheses already there. This occurs most commonly when replacing one function name with another via auto-complete.

Note: There is one difference from the v2 behavior -- In the case in which only the function name is being replaced, the caret is placed immediately after the function name, which is consistent with what occurs when replacing an attribute name. This seems more intuitive to me than putting the caret between the preexisting parentheses (the v2 behavior).


bfinzer (approved):
LGTM

PR #1582, opened 2024-10-31
fix: selection of first argument after inserting function from list

kswenson:
Fixes two separate issues in the code for inserting a function from the function list menu:
- Short argument names (e.g. `n`) could confuse the code for selecting the first argument on any browser. This is fixed by searching for `(${firstArg}` rather than just `${firstArg}`.
- Setting the selection immediately after changing the string could fail on Safari. This was fixed by adding a `setTimeout()` to delay setting the selection.


bfinzer (approved):
LGTM

PR #1584, opened 2024-11-01
Fixes bug where user can drag components under toolshelf

eireland:


kswenson (changes requested):
The limit should be the top of the document, not the bottom of the tool shelf. If the document is scrolled, the user can place a tile behind the tool shelf (eventually the document should auto-scroll in this case).

tealefristoe (commented):
Not sure if you're waiting on a review from me on this one. But I think you need to incorporate the y scroll of `.codap-container`, otherwise it's still not possible to drag to the top of the document when the document is scrolled down.

eireland:
Current behavior is similar to V2.

kswenson (approved):
👍 Looks good -- I added a commit which takes into account the document scroll and changes the logic for finding the drag container.

PR #1587, opened 2024-11-01
fix: sorting of hierarchical data sets

kswenson:
[[PT-188509712]](https://www.pivotaltracker.com/story/show/188509712)

The previous sort implementation worked for sorting attributes in the top-most collection, but the desired behavior when sorting child attributes is that the order of the parents not change, so sorting must be restricted to sorting within parent cases.

Also eliminated `getLocale()` from the `TileEnvironment`. Adding it seemed like a good idea when I did it but it never quite panned out.


tealefristoe (approved):
Looks great👍

PR #1588, opened 2024-11-02
Categorical legend and axis respond to order changes

bfinzer:
[#187614572] Bug fix: Dot chart and bar chart respond dynamically to change in category order

* One fix is, in `GraphController:syncAxisScalesWithModel` to set the axisMultiScale category set. With this fix, dragging the categories in restored bar chart succeeds in actually changing the category set order.
* An improvement in usePlotResponders:respondToCategorySetChanges we now respond to dataConfiguration.allCategoriesForRoles and do a structural comparison
* A major change is that in data-configuration-model.ts, `getCaseDataArray` is now a `cachedFnWithArgsFactory` and `get caseDataArray` is now gone because it was being cached and not updated when we needed it to be. (Note that this required replacing `.caseDataArray` with .getCaseDataArray(0) in a lot of files.)
* Another bug fixed here is that a categorical legend was not responding to changes in the order of its categories brought about by dragging in a different graph. This was fixed by reacing to `dataConfiguration?.casesChangeCount`

### DIFF
--- a/v3/src/components/graph/hooks/use-plot.ts
+++ b/v3/src/components/graph/hooks/use-plot.ts
@@ -92,12 +92,12 @@ export const usePlotResponders = (props: IPlotResponderProps) => {
 
   useEffect(function respondToCategorySetChanges() {
     return reaction(() => {
-      return dataConfiguration.casesChangeCount
+      return dataConfiguration.allCategoriesForRoles
     }, () => {
       startAnimation()
       callRefreshPointPositions(false)
     }, {name: "usePlot.respondToCategorySetChanges"})
-  }, [callRefreshPointPositions, dataConfiguration.casesChangeCount, startAnimation])
+  }, [callRefreshPointPositions, dataConfiguration, startAnimation])
 
   // respond to attribute assignment changes
   useEffect(() => {
--- a/v3/src/components/graph/hooks/use-plot.ts
+++ b/v3/src/components/graph/hooks/use-plot.ts
@@ -1,5 +1,5 @@
 import {useEffect} from "react"
-import {comparer, reaction} from "mobx"
+import { comparer,reaction } from "mobx"
 import {isAlive} from "mobx-state-tree"
 import {onAnyAction} from "../../../utilities/mst-utils"
 import {mstAutorun} from "../../../utilities/mst-autorun"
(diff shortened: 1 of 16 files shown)


v3/src/components/graph/hooks/use-plot.ts
```diff
@@ -92,14 +92,12 @@ export const usePlotResponders = (props: IPlotResponderProps) => {
...
-        startAnimation()
-        callRefreshPointPositions(false)
-      }
+      return dataConfiguration.casesChangeCount
+    }, () => {
+      startAnimation()
+      callRefreshPointPositions(false)
     }, {name: "usePlot.respondToCategorySetChanges"})
-  }, [callRefreshPointPositions, layout.categorySetArrays, startAnimation])
+  }, [callRefreshPointPositions, dataConfiguration.casesChangeCount, startAnimation])
```
kswenson, line 100: This is going to dispose/reinstall the reaction when `casesChangeCount` changes, which could prevent the reaction from firing at all (since `fireImmediately` isn't specified). I think what you may want is:
```suggestion
  }, [callRefreshPointPositions, dataConfiguration, startAnimation])
```
which says that the reaction will fire when `casesChangeCount` changes, but won't be disposed/reinstalled.

That said, I'm also curious about the change from `categorySetArrays` to `casesChangeCount`. I take it this is what you meant by:

>Another bug fixed here is that a categorical legend was not responding to changes in the order of its categories brought about by dragging in a different graph. This was fixed by reacting to `dataConfiguration?.casesChangeCount`.

Isn't there a concern that this will fire too often? One could listen to the category set itself if only category changes are of interest.

Finally, if we are sticking with `casesChangeCount` then the name of the function (line 93) and the label (line 99) should be changed from `respondToCategorySetChanges` to something more appropriate.

v3/src/components/graph/models/graph-data-configuration-model.test.ts
```diff
@@ -189,20 +189,20 @@ describe("DataConfigurationModel", () => {
     // triggers observers when values change
     const trigger = jest.fn()
     reaction(
-      () => config.caseDataArray,
+      () => config.casesChangeCount,
       () => trigger(),
-      { name: "GraphDataConfigurationTest.caseDataArray reaction" })
+      { name: "GraphDataConfigurationTest.getCaseDataArray(0) reaction" })
```
kswenson, line 194:
```suggestion
      { name: "GraphDataConfigurationTest.casesChangeCount reaction" })
```

kswenson (changes requested):
👍 Looks good -- just some questions about one of the modified reactions.

v3/src/components/graph/hooks/use-plot.ts
bfinzer: Yes, good call. I find I can use `dataConfiguration.allCategoriesForRoles`. Not sure why I didn't see this when I was struggling with it before. `layout.categorySetArrays` was definitely not working, probably because of lack of observability?

kswenson (approved):
👍 Changes LGTM

PR #1589, opened 2024-11-03
Fix restored layout of graph of categorical axis

bfinzer:
[#188403674] Bug fix: Layout of graph axis not correct on restore

* The categorical axis computes its desired extent based on whether there is a collision of categories if the category text is laid out parallel to the axis. We had been attempting to use the layout's multiscale's scale, which does (eventually) compute a bandwidth, but not in time for the initial render. So we do it a bit more brute force from the axis length and number of categories.


kswenson (approved):
👍 LGTM

PR #1590, opened 2024-11-04
feat: Tab key accepts auto-complete suggestion

kswenson:
[[PT-188470760]](https://www.pivotaltracker.com/story/show/188470760)


eireland (approved):
👍 LGTM

PR #1592, opened 2024-11-05
Dragging dot plot point bug

bfinzer:
[#188524694] Bug fix: Dragging dot plot point fails to cause other points to adjust their positions

* When dragging dot plot points, points other than those being dragged were not adjusting their positions. This was because refreshPointPositions was being called limited only to selected points. We change this so that for dot plots the refresh happens for all points.


tealefristoe (approved):
Looks good 👍 

My only suggestion would be to update the comment on line 208 to include the new exception.

PR #1595, opened 2024-11-06
Implements resize all columns to fit longest string

eireland:


v3/src/components/case-tile-common/inspector-panel/case-tile-inspector.tsx
```diff
@@ -36,15 +39,55 @@ export const CaseTileInspector = ({ tile, show, showResizeColumnsButton }: IProp
...
+  }
+
+  const findLongestContentWidth = (attr: IAttribute) => {
+    let longestWidth = 0
+    const widthAdjustment = 25 // width calculation is too wide without this adjustment
+    attr.strValues.forEach((val) => {
+      const contentWidth = measureTextWidth(val?.toString() || "")
+      if (contentWidth > longestWidth) {
+        longestWidth = contentWidth
+      }
```
tealefristoe, line 70:
```suggestion
      longestWidth = Math.max(longestWidth, contentWiddth)
```

v3/src/components/case-tile-common/inspector-panel/case-tile-inspector.tsx
```diff
@@ -36,15 +39,55 @@ export const CaseTileInspector = ({ tile, show, showResizeColumnsButton }: IProp
...
+    })
+
+    return longestWidth - widthAdjustment > kMinColumnWidth
+              ? longestWidth - widthAdjustment :  kMinColumnWidth
```
tealefristoe, line 74:
```suggestion
    return Math.max(longestWidth - widthAdjustment, kMinColumnWidth)
```

v3/src/components/case-tile-common/inspector-panel/case-tile-inspector.tsx
```diff
@@ -36,15 +39,55 @@ export const CaseTileInspector = ({ tile, show, showResizeColumnsButton }: IProp
...
+            redoStringKey: "DG.Redo.caseTable.resizeColumns"
+          })
+        }
+      })
+    })
+  }
+
+  const findLongestContentWidth = (attr: IAttribute) => {
+    let longestWidth = 0
+    const widthAdjustment = 25 // width calculation is too wide without this adjustment
```
tealefristoe, line 65: Do you know why the calculation is off? Maybe it could be fixed by matching the font or other styling? If doing something as complex as creating and then destroying a hidden span is giving you an inaccurate width, maybe it would be better to approximate it in a simpler way, like `string.length * 8`.

v3/src/components/case-tile-common/inspector-panel/case-tile-inspector.tsx
```diff
@@ -36,15 +39,55 @@ export const CaseTileInspector = ({ tile, show, showResizeColumnsButton }: IProp
...
+        if (attr) {
+          const attrId = attr?.id
+          const longestContentWidth = findLongestContentWidth(attr)
+          tableModel?.applyModelChange(() => {
+            tableModel?.setColumnWidth(attrId, longestContentWidth)
+          }, {
+            log: {message: "Resize all columns", args:{}, category: "table"},
+            undoStringKey: "DG.Undo.caseTable.resizeColumns",
+            redoStringKey: "DG.Redo.caseTable.resizeColumns"
+          })
```
tealefristoe, line 57: You should wrap the entire loop in `applyModelChange`, not each column. As is, you'll send a log message and create an undoable action for each attribute in the dataset.

tealefristoe (changes requested):
I left several suggestions. The biggest two are using one `applyModelChange` for resizing all the columns, rather than one for each column, and removing the `columnWidths` ref if it's no longer needed.

v3/src/components/case-table/collection-table.tsx
```diff
@@ -356,7 +358,7 @@ export const CollectionTable = observer(function CollectionTable(props: IProps)
         <DataGrid ref={gridRef} className="rdg-light" data-testid="collection-table-grid" renderers={renderers}
           columns={columns} rows={rows} headerRowHeight={+styles.headerRowHeight} rowKeyGetter={rowKey}
           rowHeight={+styles.bodyRowHeight} selectedRows={selectedRows} onSelectedRowsChange={setSelectedRows}
-          columnWidths={columnWidths.current} onColumnResize={handleColumnResize} onCellClick={handleCellClick}
```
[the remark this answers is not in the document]
eireland: This is used by react-data-grid.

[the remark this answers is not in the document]
tealefristoe: But you removed it, so it's no longer being used...

v3/src/components/case-table/collection-table.tsx
```diff
@@ -356,7 +358,7 @@ export const CollectionTable = observer(function CollectionTable(props: IProps)
...
+          columnWidths={dataGridColumnWidths} onColumnResize={handleColumnResize} onCellClick={handleCellClick}
```
tealefristoe, line 361: Please double check this. `columnWidths.current` was passed to `DataGrid` previously, but you removed it. So now it's not being used for anything. Leaving it in the code slows the application down and makes the code more confusing for future developers.

tealefristoe (changes requested):
Changes look good. But you should remove the `columnWidths` ref from `collection-table.tsx` altogether. If you think it's still being used somewhere, please let me know where that is. To me it looks like it was only being used as the `columnWidths` prop to `DataGrid`, and you removed that as part of this PR.

v3/src/components/case-table/collection-table.tsx
kswenson: I hope to get to reviewing this fully later today, but for now -- I added the API between RDG and CODAP around column widths so that serialization of column widths would work. If you remove that API I expect serialization to break.

kswenson: We certainly don't need two maps of column widths. I'll be adding a commit to address this.

kswenson (approved):
👍 Looks good after adding a commit to tweak a few things. As usual, take a look to make sure I didn't break anything. I also fixed an undo issue (undo of double-click of column divider) that wasn't directly related to the code at hand but that I encountered along the way.

v3/src/components/case-tile-common/inspector-panel/case-tile-inspector.tsx
```diff
@@ -28,23 +34,51 @@ export const CaseTileInspector = ({ tile, show, showResizeColumnsButton }: IProp
...
+
+  const findLongestContentWidth = (attr: IAttribute) => {
+    // include attribute name in content width calculation
+    let longestWidth = Math.max(kMinAutoColumnWidth, Math.min(kMaxAutoColumnWidth, measureHeaderText(attr.name)))
+    for (let i = 0; i < attr.length; ++i) {
+      // use the formatted attribute value in content width calculation
+      const { value } = renderAttributeValue(attr.strValues[i], attr.numValues[i], attr)
+      longestWidth = Math.max(longestWidth, Math.min(kMaxAutoColumnWidth, measureBodyText(value)))
+    }
+    return longestWidth
```
tealefristoe, line 79:
```suggestion
    let longestWidth = Math.max(kMinAutoColumnWidth, measureHeaderText(attr.name))
    for (let i = 0; i < attr.length; ++i) {
      // use the formatted attribute value in content width calculation
      const { value } = renderAttributeValue(attr.strValues[i], attr.numValues[i], attr)
      longestWidth = Math.max(longestWidth, measureBodyText(value))
    }
    return Math.min(kMaxAutoColumnWidth, longestWidth)
```

tealefristoe (approved):
Looks good 👍 

I made one small simplifying suggestion.

PR #1596, opened 2024-11-06
188476241 API Get Collection Hidden Attributes

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/188476241

This PR fixes the `hidden` value for `attributes` included in `get collection` requests. A small oversight (not including the `dataset` when converting the collection to API output format) was preventing this from working previously.

The PR also adds a test for this change, which is the vast majority of line changes.

*Note*: I'm not applying the `run regression` label since this change doesn't impact any cypress tests.


kswenson (approved):
👍 LGTM

PR #1598, opened 2024-11-06
188323297 Allow Domains to Shrink When Rescaling

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/188323297

This PR allows domain axes to shrink when calling `GraphContentModel.rescale()`. This function is called when the user clicks the rescale button in the graph menu, or Show All Cases and Display Only Selected Cases from the graph eye menu.


bfinzer (approved):
👍🏻LGTM

PR #1600, opened 2024-11-07
187502179 fix color palette table placement

eireland:
Improves the color palette placement in the case table to be close to the trigger cell.


kswenson (approved):
👍 LGTM -- I didn't review it too closely, but nothing jumped out.

PR #1601, opened 2024-11-07
fix: case table update on exit from performance mode

kswenson:
[[PT-188524711]](https://www.pivotaltracker.com/story/show/188524711)

The root cause of the bug was an itemId/caseId confusion. In performance mode we keep track of the cells that have been directly altered so that we can trigger a React refresh when exiting performance mode. The code was caching item ids and then querying the cache with case ids. Now we use case ids consistently.


tealefristoe (approved):
Looks good 👍

PR #1602, opened 2024-11-07
188411147 rename click away

eireland:
Currently, the input status of the attribute header is left in limbo status when user clicks outside the input area. This accepts input changes to attribute header when user clicks outside the input.

Currently, attribute menu only closes on user mouse up outside the attribute menu. This closes attribute menu on user mousedown.

### DIFF
--- a/v3/cypress/support/commands.ts
+++ b/v3/cypress/support/commands.ts
@@ -21,7 +21,7 @@ Cypress.Commands.add("clickWhenClickable", (selector: string, shouldCondition = 
         ensuring the action only happens once the assertion is fully satisfied.
   */
   cy.get(selector).should(shouldCondition).then($el => {
-    cy.wrap($el).click()
+    cy.wrap($el).click({force:true})
   })
 })
 
--- a/v3/src/components/case-tile-common/attribute-header.tsx
+++ b/v3/src/components/case-tile-common/attribute-header.tsx
@@ -61,6 +61,7 @@ export const AttributeHeader = observer(function AttributeHeader({
     prefix: instanceId, dataSet: data, attributeId
   }
   const { attributes, listeners, setNodeRef: setDragNodeRef } = useDraggableAttribute(draggableOptions)
+  useOutsidePointerDown({ ref: inputRef, handler: () => onCloseMenuRef.current?.() })
 
   const setHeaderContentRef = (elt: HTMLDivElement | null) => {
     contentRef.current = elt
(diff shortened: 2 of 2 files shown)


v3/src/components/case-tile-common/attribute-header.tsx
```diff
@@ -60,6 +61,7 @@ export const AttributeHeader = observer(function AttributeHeader({
     prefix: instanceId, dataSet: data, attributeId
   }
   const { attributes, listeners, setNodeRef: setDragNodeRef } = useDraggableAttribute(draggableOptions)
+  useOutsidePointerDown({ ref: inputRef, handler: () => onCloseMenuRef.current?.() })
```
tealefristoe, line 64: I'm surprised to see `inputRef` used as the ref here. I would expect it to be `menuRef` or something similar.

tealefristoe (approved):
Looks good 👍 

I did notice one odd thing I pointed out, but I'm not sure if it's actually a problem or not.

v3/src/components/case-tile-common/attribute-header.tsx
eireland: Good catch! Added a `menuListRef`

v3/src/components/case-tile-common/attribute-header.tsx
```diff
@@ -60,6 +62,7 @@ export const AttributeHeader = observer(function AttributeHeader({
     prefix: instanceId, dataSet: data, attributeId
   }
   const { attributes, listeners, setNodeRef: setDragNodeRef } = useDraggableAttribute(draggableOptions)
+  useOutsidePointerDown({ ref: menuListRef, handler: () => onCloseMenuRef.current?.() })
```
kswenson, line 65: Looks fine (with `menuListRef` rather than `inputRef`, but I note that according to the Chakra docs there's an `onPointerDownOutside` prop to `Menu` that would be worth trying rather than replicating a built-in capability.

kswenson (changes requested):
👍 Looks good, but I suggested a couple of things to look into before deciding whether changes are required.

v3/cypress/support/commands.ts
```diff
@@ -21,7 +21,7 @@ Cypress.Commands.add("clickWhenClickable", (selector: string, shouldCondition =
         ensuring the action only happens once the assertion is fully satisfied.
   */
   cy.get(selector).should(shouldCondition).then($el => {
-    cy.wrap($el).click()
+    cy.wrap($el).click({force:true})
```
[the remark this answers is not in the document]
eireland: table.spec.ts is failing at the "creates tables with new collection name" test at `table.createNewTableFromToolShelf()`

v3/src/components/case-tile-common/attribute-header.tsx
eireland: The `closeOnBlur` prop closes the menu on pointerup. So menu stays open while user is resizing.

kswenson: Apparently I was looking at the v3 Chakra docs, which adds the `onPointerDownOutside` prop to `Menu`, so I guess we're stuck with our own implementation for now.

kswenson (approved):
👍 Looks like the best we can do at this point.

PR #1605, opened 2024-11-08
181889643 fit width to content

eireland:
Implements Fit width to content in attribute menu.


v3/src/hooks/use-measure-text.ts
```diff
@@ -22,3 +26,15 @@ export const useMeasureText = (font = defaultFont) => {
...
+export const findLongestContentWidth = (attr: IAttribute) => {
+  // include attribute name in content width calculation
+  let longestWidth = Math.max(kMinAutoColumnWidth,
+                              Math.min(kMaxAutoColumnWidth, measureText(attr.name, kCaseTableHeaderFont)))
+  for (let i = 0; i < attr.length; ++i) {
+    // use the formatted attribute value in content width calculation
+    const { value } = renderAttributeValue(attr.strValues[i], attr.numValues[i], attr)
+    longestWidth = Math.max(longestWidth, Math.min(kMaxAutoColumnWidth, measureText(value, kCaseTableBodyFont)))
+  }
+  return longestWidth
```
tealefristoe, line 39:
```suggestion
  let longestWidth = Math.max(kMinAutoColumnWidth, measureText(attr.name, kCaseTableHeaderFont))
  for (let i = 0; i < attr.length; ++i) {
    // use the formatted attribute value in content width calculation
    const { value } = renderAttributeValue(attr.strValues[i], attr.numValues[i], attr)
    longestWidth = Math.max(longestWidth, measureText(value, kCaseTableBodyFont))
  }
  return Math.min(longestWidth, kMaxAutoColumnWidth)
```

tealefristoe (approved):
I made one suggestion, but looks great 👍

v3/src/hooks/use-measure-text.ts
```diff
@@ -22,3 +26,15 @@ export const useMeasureText = (font = defaultFont) => {
     return measureText(text, font)
   }, [font])
 }
+
+export const findLongestContentWidth = (attr: IAttribute) => {
```
kswenson, line 30: With this function here, components in `case-tile-common` import `findLongestContentWidth()` from the `hooks` directory, which then imports `renderAttributeValue()` from the `case-tile-common` directory, so there's a circular reference at the directory level even if there isn't one at the file level. Since the `findLongestContentWidth()` function is specific to the case tiles, I would recommend putting it into the same file with `renderAttributeValue()` and then renaming the file to something like `attribute-format-utils.ts`.

kswenson (approved):
👍 Looks good -- just one suggestion to move the `findLongestContentWidth()` function into the `case-tile-common` directory.

PR #1606, opened 2024-11-08
unify handling of v2 documents

scytacki:
- the codap3 and codap file extensions are now treated the same by the drop code. It is the content of the file that matters
- remove unnecessary createCodapDocument call when a new document was created by the close option which was doing a lot of duplicate work
- importV2Document now just returns the v3 document instance, which allows it to run synchronously
- handleCFMEvent now returns a promise when it does asynchronous things to help with testing


kswenson (approved):
👍 Looks good -- nice cleanup!

scytacki:
The one failing cypress test was passing before, and it is listed as one of the flakey tests by the cypress cloud. So I'm going to merge without fixing that.

PR #1608, opened 2024-11-09
fix: hierarchical case table updates while dragging points in a graph

kswenson:
[[PT-188539637]](https://www.pivotaltracker.com/story/show/188539637)


tealefristoe (approved):
Looks good 👍

PR #1610, opened 2024-11-11
Bug where split dot plot messes up when hierarchy changes

bfinzer:
[#188539688] Bug fix: Dot plot messed up response to adding parent collection

* This is a subtle bug. Creating the parent collection causes the data configuration's caseDataHash to momentarily go to zero because the cases data array is emptied. This triggers in use-sub-axis.ts the `respondToHiddenCasesChange` mstReaction which calls `updateDomainAndRenderSubAxis` that renders the multiScale's numeric scale to get an invalid domain. It is this invalid domain that sends the points to the graph's topLeft corner. Then the data configuration recomputes is case data array and we come back through the same mstReaction. Unfortunately we weren't setting the numeric scale's domain.
* The immediate fix is to reset the numeric scale's domain.
* But we should also figure out why the case data array is being wiped out and attempt to prevent it in the situation where nothing about it is actually supposed to change.
* We take the opportunity to do some cleanup with respect to setting domains of numeric axis models and multi scales

### DIFF
--- a/v3/src/components/axis/hooks/use-sub-axis.ts
+++ b/v3/src/components/axis/hooks/use-sub-axis.ts
@@ -322,6 +322,7 @@ export const useSubAxis = ({
       setupCategories()
     } else if (isBaseNumericAxisModel(axisModel)) {
       const currentAxisDomain = axisModel.domain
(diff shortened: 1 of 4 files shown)


kswenson (approved):
👍 LGTM

PR #1611, opened 2024-11-11
Fix for map failure to zoom on open bug

bfinzer:
[#188514843] Bug fix: Maps do not open at correct zoom level

* This was essentially a timing issue whereby the data configuration had not been given time to compute its filtered cases before being asked to cough up the lat/long bounds. We fix by wrapping the `MapContentModel`'s call to `rescale` in a setTimeout.
* Apparently as a side effect of this change, the inspector's rescale button now works, whereas it didn't work before this change.

### DIFF
--- a/v3/src/components/map/models/map-content-model.ts
+++ b/v3/src/components/map/models/map-content-model.ts
@@ -199,13 +199,9 @@ export const MapContentModel = DataDisplayContentModel
 
       // Rescale when the layers are changed
       addDisposer(self, reaction(
-        () => {
-          const {layers} = self
-          return {layers}
-        },
-        () => {
-          self.rescale()
-        }, {name: "MapContentModel.reaction rescale when layers change", equals: comparer.structural}
+        () => self.layers.map(layer => layer.id),
+        () => self.rescale(),
+        {name: "MapContentModel.reaction rescale when layers change", equals: comparer.structural}
       ))
     },
     afterAttachToDocument() {
--- a/v3/src/components/map/models/map-content-model.ts
+++ b/v3/src/components/map/models/map-content-model.ts
@@ -196,6 +196,17 @@ export const MapContentModel = DataDisplayContentModel
           }
         }, {name: "MapContentModel.reaction [sync mapModel => leaflet map]", equals: comparer.structural}
       ))
+
+      // Rescale when the layers are changed
+      addDisposer(self, reaction(
+        () => {
+          const {layers} = self
+          return {layers}
+        },
+        () => {
+          self.rescale()
+        }, {name: "MapContentModel.reaction rescale when layers change", equals: comparer.structural}
+      ))
     },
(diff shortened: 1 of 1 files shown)


tealefristoe (commented):
Looks good 👍 

Adding `setTimeout` isn't ideal, but sometimes you gotta do what you gotta do.

kswenson:
> Looks good 👍
> 
> Adding `setTimeout` isn't ideal, but sometimes you gotta do what you gotta do.

The better approach would be to identify the dependencies involved and make sure that the response happens predictably, but the question is whether it's worth the time to figure that out right now when the `setTimeout` seems to work.

bfinzer:
Well, I spent a couple hours prior to creating this PR attempting to figure out how to fix the problem without the setTimeout. But Kirk's comment triggered the experiment to react to a structural change in the `MapContentModel`'s `layers`. I really didn't think it would work because layers have to be added *before* we properly configure the data configuration. **But**, it **did** work, even when I tested with multiple layers. Nice. Thanks for the gentle push. I'll request a re-review.

tealefristoe (approved):
Big improvement 👍

kswenson:
>>But Kirk's comment triggered the experiment to react to a structural change in the `MapContentModel`'s `layers`. I really didn't think it would work because layers have to be added *before* we properly configure the data configuration. **But**, it **did** work, even when I tested with multiple layers. Nice. Thanks for the gentle push. I'll request a re-review.
>
>Big improvement 👍

The change to calling `rescale()` in a reaction to layer changes rather than in `afterAttachToDocument()` clearly is an improvement architecturally, but I was a bit suspicious because it seemed to me that the reaction in question shouldn't actually fire in this case because the layers don't change when restoring a document. 🤔 A little investigation revealed that, in fact, the reaction never fires and that what actually fixes the bug is simply eliminating the call to `rescale()` in `afterAttachToDocument()`. It turns out that `rescale()` is already being called later in the initialization cycle. The original point of the code, however, is to trigger a rescale when there are legitimate layer changes, so it makes sense to have a reaction of some sort to handle that case. I don't think the reaction as written originally will work, however, because the only property accessed during the initial/accessor function is the `layers` property itself, which never changes. One might think that `equals: comparer.structural` would mean that a deep comparison would occur, but if only the top-level `layers` property is cached, any comparison would be of the current state of the property with itself. One could rectify this by returning `JSON.stringify(layers)`, but that's going to trigger when any property of any layer changes, even ones that don't require rescale. My solution was to add a commit which returns `layers.map(layer => layer.id)`, i.e. an array of layer ids so that adding/removing/replacing layers triggers a rescale, but other smaller changes do not.

kswenson (approved):
👍 Looks good -- I added a commit to tweak the reaction for reasons described elsewhere.

tealefristoe (approved):
Looks good 👍 

This was a very interesting PR to follow.

PR #1613, opened 2024-11-12
187849414 Autodetect Boundary Type

tealefristoe:
PT Stories:
https://www.pivotaltracker.com/story/show/187849414
https://www.pivotaltracker.com/story/show/188491735
https://www.pivotaltracker.com/story/show/188491712

This PR implements related behavior requested in three different stories.
- Boundary attributes are now automatically detected.
  - Just like in v2, attributes with one of several special names will be treated as a boundary attribute.
  - Unlike in v2, an attribute will be considered of type boundary if its values all "look" like boundaries.
    - Bill started this one. I cleaned up the functions that he wrote a bit, but I suspect the way boundary data is recognized could be improved. Without having that many examples to work with or rules to follow, though, I just kept what he had.
- Boundary values are rendered as thumbnail images in the table.
- The `lookupBoundary` function has been implemented.
  - This can most easily be seen in the Roller Coasters example document.
  - Currently, we don't have a general solution for how to host boundary documents. Until we have that figured out, I added the state boundary document to the Codap. This means two things:
    - The document is included whenever Codap is loaded, not just when it is needed.
    - Only the state data is currently available, not any other boundary data.

Other changes:
- `basicCanonicalNameToDependency` was moved from `formula-dependency-utils.ts` to `name-mapping-utils.ts` to avoid an import cycle. This was a very straightforward change to make. I'm not sure if there was some organizational reason for this function to be in one file rather than the other.


v3/src/utilities/geojson-utils.test.ts
kswenson, line 1: Rename to `boundary-utils.test.ts`.

v3/src/utilities/geojson-utils.ts
kswenson, line 1: Rename to `boundary-utils.ts`.

kswenson (changes requested):
👍 Looks good -- suggestions are mainly to rearrange some of the dependencies.

kswenson (approved):
👍 Changes look good

PR #1614, opened 2024-11-12
chore: move eslint-webpack-plugin cache from node_modules/.cache to .cache

kswenson:
When the webpack and eslint caches were moved from `node_modules/.cache` to `.cache`, the `eslint-webpack-plugin` cache was left behind. With this PR the caches are reunited (cue happily-ever-after soundtrack and fade-to-black).

I discovered this after `npm run clean:cache` failed to fix a webpack build error caused by stray merge conflict markers in the eslint-plugin-cache.


scytacki (approved):
Looks good to me.

PR #1617, opened 2024-11-14
188497528 case tile filter formula editor

eireland:
Changes case table/card edit filter formula modal to use the regular formula editor modal.

Refactored Cypress formula-helper.ts to avoid circular reference between table and the helper.


v3/cypress/support/helpers/formula-helper.ts
```diff
@@ -16,46 +13,46 @@ export const FormulaHelper = {
...
+  // },
+  // changeSliderValue(sliderValue: string) {
+  //   slider.changeVariableValue(sliderValue)
+  // },
+  // addSlider() {
+  //   c.getIconFromToolShelf("slider").click()
+  // },
+  // deleteSlider() {
+  //   c.closeComponent("slider")
+  // },
```
kswenson, line 55:
```suggestion
```

PR #1618, opened 2024-11-14
fix: table row selection and child table scrolling

kswenson:
[[PT-188036951]](https://www.pivotaltracker.com/story/show/188036951)

Two issues were requested/addressed:
1. Clicking on a single selected case among a set of selected cases should deselect all other cases.
    - This differs from the way most applications behave when clicking on a selected object, but it's what v2 did so we match the behavior here. This was easy to fix.
2. Clicking on a parent case should not only select the given parent, but also scroll the children into view.
    - The code to scroll the children into view is also straightforward to write, but dealing with the resulting scroll synchronization responses took much longer than expected. I ended up adding a mechanism for disabling sync scroll responses that relies on a one-second timer and triggering it from the call to select the child cases. It's not particularly elegant, but it works better than the myriad other things I tried along the way. ¯\\_(ツ)_/¯

### DIFF
--- a/v3/src/components/case-table/case-table-types.ts
+++ b/v3/src/components/case-table/case-table-types.ts
@@ -28,6 +28,13 @@ export interface TCellSelectArgs extends CellSelectArgs<TRow> {}
 export type TCellKeyDownArgs = CellKeyDownArgs<TRow>
 export type TColSpanArgs = ColSpanArgs<TRow, unknown>
 
+export interface IScrollOptions {
+  // "auto" moves immediately; "smooth" animates
+  scrollBehavior?: "auto" | "smooth"
+  // disable parent/child scroll sync for the target table
+  disableScrollSync?: boolean
+}
+
 export interface IBaseTableScrollInfo {
   element: HTMLDivElement
   scrollRowIntoView: (rowIndex: number) => void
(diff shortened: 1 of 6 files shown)


tealefristoe (approved):
Changes look good and seem to work well 👍 

I did notice a couple of differences with scrolling in v2:
- In v2, the autoscrolled collections tend to put the newly selected cases into the middle of the tile (the rules seem more complex than this, but they often end up in the middle). But here it's the top or bottom of the tile.
- In v2, all child collections autoscroll when a parent case is selected. Here, only the immediate child collection autoscrolls.

kswenson:
> Changes look good and seem to work well 👍
> 
> I did notice a couple of differences with scrolling in v2:
> 
> * In v2, the autoscrolled collections tend to put the newly selected cases into the middle of the tile (the rules seem more complex than this, but they often end up in the middle). But here it's the top or bottom of the tile.

This meets the acceptance criteria and I'm inclined to think that auto-scrolling less is potentially less disruptive than auto-scrolling more. In any case, I'm inclined to leave it this way for now and see how it lands.

> * In v2, all child collections autoscroll when a parent case is selected. Here, only the immediate child collection autoscrolls.

Good catch -- I'll look into fixing it as part of this PR but if it looks like a substantial amount of additional work I may put it off to a subsequent story/PR in the interest of moving this one forward.

kswenson:
> > * In v2, all child collections autoscroll when a parent case is selected. Here, only the immediate child collection autoscrolls.
> 
> Good catch -- I'll look into fixing it as part of this PR but if it looks like a substantial amount of additional work I may put it off to a subsequent story/PR in the interest of moving this one forward.

I added a commit to loop through all child collections.

tealefristoe (approved):
Looks good 👍

PR #1619, opened 2024-11-14
use new cfm openLocalFileWithConfirmation

scytacki:
simplify drop handle, it no longer needs to load the file content itself

This addresses: 
PT-188415967
PT-188545557


kswenson (approved):
👍 Looks good!

PR #1620, opened 2024-11-14
182991523 date type precision menu

eireland:
Adds a different precision menu when attribute type is date.

Implements showing the correct precision when user changes precision without having to change attribute type.


kswenson (approved):
👍 Looks good -- I added a commit which cleaned up a few things. Make sure I didn't break anything before you merge. 😉

PR #1621, opened 2024-11-15
Right justifies numeric types and changes font to a monspaced font

eireland:
Precision is implemented with this PR https://github.com/concord-consortium/codap/pull/1620


v3/src/components/case-table/use-columns.tsx
```diff
@@ -49,7 +49,8 @@ export const useColumns = ({ data, indexColumn }: IUseColumnsProps) => {
                 resizable: true,
                 headerCellClass: `codap-column-header`,
                 renderHeaderCell: ColumnHeader,
-                cellClass: row => clsx("codap-data-cell", `rowId-${row.__id__}`, {"formula-column": hasFormula}),
+                cellClass: row => clsx("codap-data-cell", `rowId-${row.__id__}`, {"formula-column": hasFormula,
+                                        "right-justify": userType === "numeric" || type === "numeric"}),
```
kswenson: Number formatting should be applied to values, not columns. I'll add a commit.

v3/src/components/case-table/case-table.scss
```diff
@@ -323,6 +323,13 @@ $table-body-font-size: 8pt;
         background-color: var(--rdg-row-selected-background-color);
       }
 
+      &.right-justify {
+        justify-content: flex-end;
+        // font-family: 'Roboto  Mono', monospace;
+        font-family: 'Inconsolata', monospace;
```
kswenson: I told myself I wasn't going to do this, but I couldn't resist looking into other fonts. Take a look at it after my commit and see what you think.

kswenson (approved):
👍 Looks good after I added a commit:
- Number formatting applies to values, not columns.
- What do you think of `Chivo Mono` as an alternative?

Give it a whirl and see what you think.

PR #1622, opened 2024-11-15
fix: import of v2 documents with minimized components

kswenson:
[[PT-188570070]](https://www.pivotaltracker.com/story/show/188570070)


tealefristoe (approved):
Looks good 👍

PR #1624, opened 2024-11-15
update window title when the document name changes

scytacki:
This replaces the document title in the document state with a volatile title property. It also removes the filename from the properties. Both changes are intended to force the CFM to be the source of truth for the title.
This change of state should not require a migration of existing v3 documents because the filename/name was already managed by the CFM and just not updated in v3 correctly.

To keep the title in sync the filename needs to be processed when the CFM sends the openedFile, savedFile, and renamedFile events.


v3/src/models/app-state.ts
```diff
@@ -120,13 +123,20 @@ class AppState {
         if (metadata) {
           const metadataEntries = Object.entries(metadata)
           metadataEntries.forEach(([key, value]) => {
-            if (value != null) {
+            if (value == null) return
+
+            if (key === "filename") {
+              // We don't save the filename because it is redundant with the filename in the actual
+              // filesystem.
+              // However we need to the extension-less name for the window title
```
kswenson, line 131:
```suggestion
              // However we need the extension-less name for the window title.
```
or something like that.

v3/src/models/app-state.ts
```diff
@@ -42,6 +44,7 @@ class AppState {
   private version = ""
   private cfm: CloudFileManager | undefined
   private dirtyMonitorDisposer: (() => void) | undefined
+  titleMonitorDisposer: any
```
kswenson, line 47:
```suggestion
  private titleMonitorDisposer: (() => void) | undefined
```

v3/src/models/codap/create-codap-document.ts
```diff
@@ -29,6 +29,7 @@ interface IOptions {
 }
 export function createCodapDocument(snapshot?: ICodapDocumentModelSnapshot, options?: IOptions): IDocumentModel {
   const { layout = "free", noGlobals = false } = options || {}
+  // Note: The version and build will not be updated after the document is first created
```
kswenson, line 32: Can/should we do something about that? Seems like the expectation would be that it would be the version that most recently saved it.

kswenson (approved):
👍 LGTM -- just a few suggestions

v3/src/models/codap/create-codap-document.ts
scytacki: @kswenson yeah that'd be my expectation too. 

I think we can fix it, but it seemed possibly tricky and not worth the effort right now.  If you want to prioritize it I can work on it.

PR #1625, opened 2024-11-15
run cypress in docker

scytacki:
This change tells GitHub to run the Cypress job in a docker container. This should prevent issues where the version of Chrome is updated by GitHub actions.

Unfortunately it adds about 25s to each worker while it is downloading the docker image. However I did bump up the number of workers from 5 to 7 so the overall time should go down slightly.

I'm not entirely sure this is worth it.


scytacki:
Using the `setup-chrome` action was faster than this approach of using a docker container.

PR #1626, opened 2024-11-15
fix: bivariate semi-aggregates (linRegrPredicted, linRegrResidual) respect filter argument

kswenson:
[[PT-188459268]](https://www.pivotaltracker.com/story/show/188459268)


tealefristoe (approved):
Looks good 👍

PR #1627, opened 2024-11-15
try using setup-chrome

scytacki:
This is faster than using the cypress built docker container. It only adds 8s instead of 25s to the build.

### DIFF
--- a/.github/workflows/v3-regression.yml
+++ b/.github/workflows/v3-regression.yml
@@ -55,7 +55,7 @@ jobs:
       fail-fast: false
       matrix:
         # run multiple copies of the current job in parallel [1, ..., n]
-        containers: [1, 2, 3, 4, 5]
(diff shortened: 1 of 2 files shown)


scytacki:
The code coverage is down is because the only changes here are to the github actions files. This prevents the main build from running.

.github/workflows/v3-regression.yml
nstclair-cc, line 1: Nice. Now I know where the number of workers lives.

.github/workflows/v3.yml
nstclair-cc, line 1: Using @v1 ensures we get updates within the same major version of Chrome while avoiding breaking changes.

nstclair-cc (approved):

nstclair-cc:
LGTM. The flaky 16+1 color palette test will be addressed by @eireland, yes? @eireland let me know if I can help with fixing that particular test.

PR #1629, opened 2024-11-16
fix: don't show date precision options if user has chosen another type

kswenson:
- fix logic for when to show date options vs. numeric options in Edit Attribute Properties dialog
- disable the precision options if neither are appropriate


eireland (approved):
👍 LGTM

PR #1630, opened 2024-11-16
fix: handling of hidden/minimized tiles

kswenson:
[[PT-188570070]](https://www.pivotaltracker.com/story/show/188570070)

#1622 addressed the case of minimized tiles, but did not properly capture the relationship between hidden and minimized tiles. The document in the PT story has a case table that has been minimized and hidden. In v2, when such a tile is shown/unhidden it is also unminimized. In other words, when a hidden tile is shown, its prior minimized/unminimized state is ignored, which is equivalent to saying that a tile cannot be both hidden and minimized. Therefore, we implement this rule at the model level, so that we don't have to find all clients that show hidden tiles (Calculator button in tool shelf, close button of certain tiles, table menu items, etc.).


tealefristoe (approved):
Looks good to me 👍

PR #1633, opened 2024-11-17
187014862 Color Attribute Points Followup

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/187014862

This PR addresses some issues that were discovered while testing the previous work that colors points based on attribute value when a color attribute is used for the legend of a graph or map.


kswenson (approved):
👍 Looks good -- I added a commit with a few tweaks. @bfinzer As part of your review, take a look at the calls to `callMatchCirclesToData()` and `refreshPoints()` and see if anything looks amiss to you.

bfinzer (approved):
I pushed a commit that moved some code from map-point-layer to data-configuration-model, and also made it so a request for legend color returns `missingColor` when there is no value for the legend attribute.

PR #1635, opened 2024-11-18
188544725 v3 Remote Boundaries

tealefristoe:
PT Stories:
- https://www.pivotaltracker.com/story/show/188544725
- https://www.pivotaltracker.com/story/show/188589165

This PR fetches remote boundary data, removing the baked in boundary data used for previous boundary related stories. Boundary data is not downloaded until it is needed, like when the `lookupBoundary` function is called.

~~Unfortunately, the `lookupBoundary` function is not currently equipped to respond to the data actually being downloaded, so this PR doesn't completely fix the issue. But it is possible to confirm that this works by forcing the function to run again after initially calling it, such as by changing the Boundary attribute formula in the Roller Coasters example doc, then changing it back.~~

Other changes worth noting:
- ~~A cypress test broke with this change, so I skipped it for now. The `skip` should be removed when the `lookupBoundary` function responds to the data being downloaded, probably in a future PR.~~
- Downloaded boundary keys are now used by `formula-insert-values-menu.ts` instead of hard-coded values.
- A call to `Attribute.getBoundaryCount.invalidate()` is now made, ensuring boundary data is correctly analyzed when an attribute's values change.
- A couple of boundary related functions are now mocked in `lookup-functions.test.ts` because fetching remote boundary data does not work in jest tests.

**Update (@kswenson)**
I was originally intending to add a "small" commit integrating @tealefristoe's work with the formula dependency system, but once I got in there it seemed that a larger refactor made more sense. Key features of the refactored code are that:
- `codap-resources` urls are canonicalized
- boundary code moved to `models/boundaries`
- boundary data are now managed by a a singleton observable `BoundaryManager` class rather than as a set of interdependent global variables and utility functions
- async code rewritten using async/await (with the help of ChatGPT)
- a `boundaryManager` instance is added to the formula context, formula api, etc.
- formula manager observes boundaries and triggers reevaluation of formulas when boundaries change (i.e. are downloaded and processed)
- boundary info includes a promise so that boundary data is requested for download only once
- boundary info includes error information in case of error so that pending boundary data can be distinguished from failed boundary data
- boundaries are supported in the formula editor for auto-complete and highlighting purposes
- re-enabled boundary thumbnails cypress test


kswenson (approved):
👍 LGTM 😉 -- I added a commit which fixed the remaining issues mentioned by @tealefristoe, integrated boundaries with the formula manager, formula dependency system, formula editor, etc. Given that so much of this is now my work, I've assigned it to @scytacki for final review.

v3/src/lib/use-cloud-file-manager.ts
```diff
@@ -2,14 +2,15 @@ import { CFMAppOptions, CloudFileManager, CloudFileManagerClientEvent } from "@c
...
-import { handleCFMEvent } from "./handle-cfm-event"
+import { codapResourcesUrl } from "../constants"
 import { appState } from "../models/app-state"
 import { isCodapDocument } from "../models/codap/create-codap-document"
 import { gLocale } from "../utilities/translation/locale"
 import { t } from "../utilities/translation/translate"
 import { removeDevUrlParams, urlParams } from "../utilities/url-params"
+import { clientConnect, createCloudFileManager, renderRoot } from "./cfm-utils"
 import { DEBUG_CFM_LOCAL_STORAGE } from "./debug"
+import { handleCFMEvent } from "./handle-cfm-event"
```
scytacki, line 13: Hmm. I guess we aren't currently seeing lint errors in our PRs. You caught the console log issue and that wasn't showing up either in the PR.

scytacki (approved):
Looks good to me. The only thing I could find that might be unexpected is how the code handles a network error boundary_specs file. In that case the formulas are considered invalid because the symbols for the boundary keys are not known. 

That seems like an edge case that we could punt until it becomes a real problem.

PR #1638, opened 2024-11-19
188392104 case card formula

eireland:
Makes number types follow attribute precision in case cards.
Disables the `Editable` attribute property if the attribute has a formula.


kswenson (approved):
👍 LGTM -- I added a commit with a couple of small tweaks.

PR #1639, opened 2024-11-19
Changes min width for auto rescale

eireland:
Removes attribute neame from longest content width calculation because attribute  names can be split over 2 lines and ellided.

### DIFF
--- a/v3/src/components/case-table/case-table-types.ts
+++ b/v3/src/components/case-table/case-table-types.ts
@@ -55,7 +55,7 @@ export type OnScrollRowRangeIntoViewFn = (collectionId: string, rowIndices: numb
 export const kInputRowKey = "__input__"
 
 export const kDefaultColumnWidth = 80
-export const kMinAutoColumnWidth = 40
+export const kMinAutoColumnWidth = 52
 export const kMaxAutoColumnWidth = 600
 
 export const kCaseTableFontFamily = "Montserrat, sans-serif"
--- a/v3/src/components/case-table/case-table-types.ts
+++ b/v3/src/components/case-table/case-table-types.ts
@@ -55,7 +55,7 @@ export type OnScrollRowRangeIntoViewFn = (collectionId: string, rowIndices: numb
 export const kInputRowKey = "__input__"
 
 export const kDefaultColumnWidth = 80
-export const kMinAutoColumnWidth = 60
+export const kMinAutoColumnWidth = 50
 export const kMaxAutoColumnWidth = 600
 export const kCellPadding = 11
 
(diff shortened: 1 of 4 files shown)


bfinzer (approved):
👍🏻LGTM

v3/src/components/case-tile-common/attribute-format-utils.tsx
```diff
@@ -97,8 +97,10 @@ export function renderAttributeValue(str = "", num = NaN, attr?: IAttribute, key
 }
 
 export const findLongestContentWidth = (attr: IAttribute) => {
-  // include attribute name in content width calculation
-  let longestWidth = Math.max(kMinAutoColumnWidth, measureText(attr.name, kCaseTableHeaderFont))
+  // We do not include attribute name in content width calculation because header text can be split over
+  // two lines and ellided
```
kswenson, line 101: Comment should match what the code is actually doing. Also, `ellipsis` has two `L`s, but `elided` only has one. 😉

v3/src/components/case-tile-common/attribute-format-utils.tsx
```diff
@@ -97,8 +97,10 @@ export function renderAttributeValue(str = "", num = NaN, attr?: IAttribute, key
...
+  const headerWidth = measureText(attr.name, kCaseTableHeaderFont)
+  let longestWidth = Math.max(kMinAutoColumnWidth, Math.ceil(3 + (headerWidth/2)))
```
kswenson, line 103: I'm a bit conflicted here. I think it does make sense to include the column header in the calculation, but I agree with the critique that my previous approach was too simplistic. With the previous minimum width of 40, using the `sample=mammals` data, auto-fit resulted in `LifeSpa⏎n`, which seems wrong. Your change to include `headerWidth/2` in the computation (since the header can be split across two lines) doesn't help for that case. What I think I want as a user is for the first word of the attribute name not to be broken across multiple lines _within reason_. So maybe that means considering half of the entire width of the label (including units -- which are not currently factored in) _and_ considering the width of the first word of the attribute name, but limiting their effect -- like maybe the header width has a maximum contribution of 150, while the maximum data width is 600, or something like that.

On the other hand, I understand the desire to not spend any more time on this, in which case a compromise might be to raise the minimum width to 50 (which fixes the `LifeSpa⏎n` issue) and leaving the rest of the computation you have in place. 🤷

kswenson (changes requested):
👍 Looks good -- I made a couple of small suggestions and a larger consideration that perhaps should be punted for now. 🤷

v3/src/components/case-table/case-table-types.ts
```diff
@@ -55,8 +55,9 @@ export type OnScrollRowRangeIntoViewFn = (collectionId: string, rowIndices: numb
 export const kInputRowKey = "__input__"
 
 export const kDefaultColumnWidth = 80
-export const kMinAutoColumnWidth = 40
+export const kMinAutoColumnWidth = 60
```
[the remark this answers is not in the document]
eireland: To match V2 better

kswenson (approved):
👍 LGTM

PR #1640, opened 2024-11-19
Make number values overflow with ellipsis

eireland:
There was a regression in the number overflow styling

### DIFF
--- a/v3/src/components/case-table/case-table.scss
+++ b/v3/src/components/case-table/case-table.scss
@@ -339,6 +339,9 @@ $table-body-font-size: 8pt;
           font-weight: 500;
           margin-top: 2px;
           text-align: right;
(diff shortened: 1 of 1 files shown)


kswenson (commented):
This does not seem to be the way v2 behaves. Are we sure this is the desired behavior?

eireland:
@kswenson Not sure which behavior we want. This PR was to address this bug report https://www.pivotaltracker.com/story/show/188544375 which mentions that prior to changing to the monospaced font, V3 was showing the ellipsis.

kswenson:
Eliding numeric values is not the v2 behavior and my inclination is that it's not the correct behavior, so I'm closing this pending an affirmative decision to change the behavior for v3. We can always reopen it if such a decision is made.

PR #1641, opened 2024-11-19
Fix: case card has notifications when collections move (PT-188571448)

lublagg:
Addresses the issue caught in [this story](https://www.pivotaltracker.com/story/show/188571448)


kswenson (approved):
👍 LGTM

PR #1642, opened 2024-11-20
Detect lat/lng attribute names independent of case

bfinzer:
[#188571538] Bug fix: Lat/long points will not display in Map component when migrating v2 documents to v3

* We weren't making detection of lat/long attribute names case independent. Now we do.


kswenson (approved):
👍 LGTM

PR #1645, opened 2024-11-22
handle CFM sharing data

scytacki:
Because CODAP is running in "unwrapped" mode it needs to include the sharing data in its own content. However it has to be in a specific place so the CFM can remove it when necessary.

- on save (getContent) we include the sharing data in the content
- on load (openedFile) we need to send the sharing data back to the CFM.

PT-188523684

### DIFF
--- a/v3/src/lib/handle-cfm-event.ts
+++ b/v3/src/lib/handle-cfm-event.ts
@@ -1,3 +1,4 @@
+import { cloneDeep } from "lodash"
 import { CloudFileManagerClient, CloudFileManagerClientEvent } from "@concord-consortium/cloud-file-manager"
 import { appState } from "../models/app-state"
 import { removeDevUrlParams, urlParams } from "../utilities/url-params"
@@ -6,7 +7,7 @@ import { wrapCfmCallback } from "./cfm-utils"
 import build from "../../build_number.json"
 import pkg from "../../package.json"
 
-export function handleCFMEvent(cfmClient: CloudFileManagerClient, event: CloudFileManagerClientEvent) {
+export async function handleCFMEvent(cfmClient: CloudFileManagerClient, event: CloudFileManagerClientEvent) {
   // const { data, state, ...restEvent } = event
   // console.log("cfmEventCallback", JSON.stringify({ ...restEvent }))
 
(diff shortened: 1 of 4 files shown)


kswenson (approved):
👍 Code looks good -- should we also be handling the `sharedFile` and `unsharedFile` events?

scytacki:
Doh. Good point. Without handling them, the user will have to explicitly save again to preserve the sharing info. I'll update this PR.

v3/src/lib/handle-cfm-event.ts
```diff
@@ -1,14 +1,18 @@
...
 import build from "../../build_number.json"
 import pkg from "../../package.json"
 
-export function handleCFMEvent(cfmClient: CloudFileManagerClient, event: CloudFileManagerClientEvent) {
-  // const { data, state, ...restEvent } = event
-  // console.log("cfmEventCallback", JSON.stringify({ ...restEvent }))
+export async function handleCFMEvent(cfmClient: CloudFileManagerClient, event: CloudFileManagerClientEvent) {
+  if (DEBUG_CFM_EVENTS) {
+    // We clone the event because the CFM reuses the same objects within the event
+    console.log("cfmEvent", event.type, cloneDeep(event))
```
kswenson, line 14: Nice addition for debugging/understanding how the CFM works. I expect you'll need to add an eslint-disable comment to avoid warnings in `lint:build`, though.

v3/src/lib/handle-cfm-event.ts
```diff
@@ -66,10 +94,22 @@ export function handleCFMEvent(cfmClient: CloudFileManagerClient, event: CloudFi
...
+    case "sharedFile":
+      // Trigger a save so the sharing info can be updated in the file
+      // If the file is already shared, and the user updates the shared document this event
+      // will happen again. Currently it isn't necessary to update the sharing info, but
+      // perhaps in the future the sharing info will include properties that change each
+      // time, such as a timestamp for when the document was shared.
+      // Due to the design of the CFM event system we need to do this in the next time slice
+      await new Promise(resolve => setTimeout(resolve, 0))
+      cfmClient.dirty(true)
+      break
```
kswenson, line 106:
```suggestion
```
Since the response to `sharedFile` and `unsharedFile` is the same, seems like they could be handled together.

kswenson (approved):
👍 Nice work! Just a couple of suggestions to consider.

PR #1646, opened 2024-11-22
Added back graph rescale Cypress test

nstclair-cc:
[PT-#188370962](https://www.pivotaltracker.com/story/show/188370962)

### Description:

This PR addresses the issue of rescaling the graph after updating values in the table. The fix was to ensure proper interaction with the rescale button and table cells using double-clicks. The issue was identified during testing and is documented in the Pivotal Tracker story below.
- Related Issue: [Autoscale isn't working for 2-numerical attribute graphs in ?sample=mammals document](https://www.pivotaltracker.com/story/show/188601882)

### Fix Details:
- Ensured the rescale button receives a dblclick event to trigger the rescale action.
- Updated table cell interactions to properly double-click for editing values.
- Added comments for clarity and documented the current approach for verifying axis rescaling (checking tick marks).

### Changes:

- Updated the test change points in table and check for autoscale to include:
- dblclick for the rescale button.
- dblclick for editing the second table point.

### DIFF
--- a/v3/cypress/e2e/graph.spec.ts
+++ b/v3/cypress/e2e/graph.spec.ts
@@ -172,9 +172,7 @@ context("Graph UI", () => {
     })
   })
   describe("graph inspector panel", () => {
-    // this test doesn't work. Skipping for now (PT-#188370962)
-    // Logged a bug #188601882
-    it.skip("change points in table and check for autoscale", () => {
+    it("change points in table and check for autoscale", () => {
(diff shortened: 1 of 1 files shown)


kswenson (approved):
👍 I can think of no logical reason why the cypress test should require these additional clicks when they aren't necessary for a human user, but I also don't think we should expend any more time on it at this point. 🤷

PR #1648, opened 2024-11-22
handle invalid documents better

scytacki:
- add cypress tests of several invalid documents
- reject basic strings from the CFM, I believe we will only get strings if the CFM fails to run JSON.parse
- add a try/catch around appState.setDocument, so we can tell the CFM when there is an error
- log the error in a way that makes it easier to inspect MST error messages
- remove try/catch in AppState#setDocument this way errors are sent up to handleCfmEvents
- remove try/catch in createDocumentModel. This was used before in CLUE to so we could return a document that displayed the error. This "pink screen of death" seems less useful in CODAP. There is only one document open at a time.


v3/cypress/fixtures/invalid-docs/index.md
```diff
@@ -0,0 +1,32 @@
+## invalid-content.codap
+This has a invalid content field.
```
kswenson, line 2:
```suggestion
This has an invalid content field.
```

v3/cypress/fixtures/invalid-docs/index.md
```diff
@@ -0,0 +1,32 @@
...
+The invalid content field means that the `isClientContent` function provided to the CFM by CODAP returns false.
+However since there is a content field (even though it is invalid) the CFM thinks this document is "wrapped" so it only sends the value of the content field to CODAP instead of the whole document.
+
+So CODAP receives a document of `{"invalid": "value"}`.
+
+**this does not cause an visible error.**
```
kswenson, line 8:
```suggestion
**this does not cause a visible error.**
```

v3/cypress/fixtures/invalid-docs/index.md
```diff
@@ -0,0 +1,32 @@
...
+It isn't obvious how to handle it better. We don't want to reject documents with unknown fields to allow new fields with crashing old versions of CODAP, so unknown properties are just ignored. We could reject the document because it doesn't have the `type: "CODAP"` or because it doesn't have `content.rowMap` and `content.tileMap` fields, but maybe that is too restrictive?
```
kswenson, line 9:
```suggestion
It isn't obvious how to handle it better. We don't want to reject documents with unknown fields to allow new fields without crashing old versions of CODAP, so unknown properties are just ignored. We could reject the document because it doesn't have the `type: "CODAP"` or because it doesn't have `content.rowMap` and `content.tileMap` fields, but maybe that is too restrictive?
```

v3/cypress/fixtures/invalid-docs/index.md
```diff
@@ -0,0 +1,32 @@
...
+
+## invalid-json.codap
+Not a valid JSON document.
+The CFM catches this and sends CODAP a string of the document.
+CODAP throws an error when it gets a string document.
+This is caught and the CFM callback is called telling the CFM there was an error opening the document.
+
+## invalid-rowmap.docap
```
kswenson, line 17:
```suggestion
## invalid-rowmap.codap
```

v3/cypress/fixtures/invalid-docs/index.md
```diff
@@ -0,0 +1,32 @@
...
+This has a content field but its rowMap is invalid. The fact that there is any rowMap and tileMap fields convinces the `isClientContent` function this is a valid CODAP document.
```
kswenson, line 18:
```suggestion
This has a content field but its rowMap is invalid. The fact that there are rowMap and tileMap fields convinces the `isClientContent` function this is a valid CODAP document.
```

kswenson (approved):
👍 Looks good! Just a smattering of typographical suggestions

PR #1649, opened 2024-11-23
188514789 case card color picker

eireland:
Adds ability to show and edit color attribute types in case card.

Refactored ColorPalettePicker to its own component.


v3/cypress/e2e/table.spec.ts
```diff
@@ -1145,6 +1145,7 @@ context("case table ui", () => {
 
       cy.log("click Set Color button to dismiss color palette and change color")
       cpp.getSetColorButton().click()
+      cy.wait(1000) // Wait for the color change to be reflected
```
kswenson, line 1148: Could we get by with `100` here like with other waits in this test? Surely, it can't take a full second for the color change to be reflected.

kswenson (approved):
👍 LGTM -- just one question about the `wait`.

bfinzer (approved):
👍🏻LGTM

v3/cypress/e2e/table.spec.ts
eireland: I tried 500 and it was still failing 😞

PR #1650, opened 2024-11-24
fix: ruler menu re-renders on undo collection change

kswenson:
[[PT-188548516]](https://www.pivotaltracker.com/story/show/188548516)

The PT story mentions a React warning which is a duplicate of another PT story and an MST warning, which is addressed in this PR. The error occurred in the case table/card inspector's ruler menu, which was accessing a stale collection that no longer existed after the undo. The fix is to make the `RulerMenuList` an `observer`, so that it re-renders when the collections change, flushing any stale collection references.


eireland (approved):
👍 LGTM

PR #1637, opened 2024-11-18
SPIKE: Integrate Cypress with PixiJS

nstclair-cc:
[PT-#188565339](https://www.pivotaltracker.com/story/show/188565339)

### Summary

This PR documents findings from a SPIKE aimed at improving Cypress tests for verifying graph point interaction on a PIXIJS-rendered canvas in CODAP. While partial success was achieved in triggering interactions with the canvas, there remain challenges in validating the visual state of PIXIJS-rendered elements.

### Unskipped Tests

In graph.spec.ts, the following comment explained why certain tests were previously skipped:

```

// Skipping this test because Cypress 13 does not support displaying pixijs canvas elements in CI
// cypress-io/cypress#28289
// This test can be unskipped when the above Cypress bug is resolved
// (Locally, this works fine and the tests can be run successfully)

```

Since the referenced Cypress issue [cypress-io/cypress#28289](https://github.com/cypress-io/cypress/issues/28289) has been resolved, the tests were unskipped and revalidated for performance in both local and CI environments.

---

### Functional Test Code

Below is an updated test example, unskipped and modified during this SPIKE:
```

context("Test graph plot transitions", () => {
  beforeEach(function () {
    const queryParams = "?mouseSensor"
    const url = `${Cypress.config("index")}${queryParams}`
    cy.visit(url)
    cfm.openLocalDoc("cypress/fixtures/3TableGroups.codap")
    cy.wait(2500)
  })

  plots.forEach(test => {
    it(`${test.testName}`, () => {
      c.getIconFromToolShelf("graph").click()
      c.moveComponent("graph", 1000)
      test.axes.forEach(hash => {
        hash.checks.forEach(check => {
          cy.checkDragAttributeHighlights("table", hash.attribute, check.axis, check.active)
        })
        cy.dragAttributeToTarget("table", hash.attribute, hash.target)
        cy.wait(2000)
      })
    })
  })
})
```

---

### Skipped Test Example

The following test remains skipped as it relies on PIXIJS exposing DOM elements or API hooks for verifying interactions, such as graph point highlighting:

```
    it.skip("should highlight a selected graph point", () => {
      // This test is the outcome of a SPIKE to explore testing graph interactions.
      // It partially validates interactions but requires further PIXIJS-level support.
      // https://github.com/concord-consortium/codap/pull/1637

      // Select the target table cell
      table.getGridCell(2, 2).should("contain", "African Elephant").click({ force: true })

      // Verify the graph's component title matches the collection name
[...]

PR #1651, opened 2024-11-25
chore: disable slow graph cypress test

kswenson:
This test was reenabled in #1637, but with it enabled the time to run the cypress tests remotely on GitHub goes from ~7 min to ~23 min, so the benefit is not worth the cost.


kswenson:
Merging without review.

nstclair-cc (commented):
lgtm! sorry about the delay

PR #1653, opened 2024-11-26
188618452 new table clean up

eireland:
On new tables:
- Removes "no cases" div
- Removes empty case above the input line
- Changes default attribute name to "Attribute Name" instead of "AttributeName"
- Changes table created from the `--new--` in the table menu default width to just show the one default column

### DIFF
--- a/v3/cypress/e2e/formula/formula-component.spec.ts
+++ b/v3/cypress/e2e/formula/formula-component.spec.ts
@@ -83,16 +83,18 @@ context("Formula Engine", () => {
         "❌ Undefined symbol v1"
       ])
     })
-    // Skipped for now until insert cases on input row is implemnted
-    it.skip("Formula in a new dataset", () => {
+    // TODO: update test when insert cases on input row is implemnted
+    it("Formula in a new dataset", () => {
       fh.visitURL("")
       table.createNewTableFromToolShelf()
+      table.getGridCell(2, 2).dblclick()
+      table.getGridCell(2, 2).find("input").type("Sloth{enter}")
       table.openIndexMenuForRow(2)
       table.insertCases(4, "after")
       table.addFormula("Attribute Name", "10*10")
       table.verifyFormulaValues("Attribute Name", [100, 100, 100, 100, 100])
       table.addNewAttribute()
-      table.addFormula("newAttr", "Attribute Name+2")
+      table.addFormula("newAttr", "`Attribute Name`+2")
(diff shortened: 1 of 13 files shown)


v3/cypress/e2e/formula/formula-component.spec.ts
```diff
@@ -83,15 +83,16 @@ context("Formula Engine", () => {
         "❌ Undefined symbol v1"
       ])
     })
-    it("Formula in a new dataset", () => {
+    // Skipped for now until insert cases on input row is implemnted
```
tealefristoe, line 86: How difficult would it be to implement this? Isn't this already done  in `table.spec.ts` lines 738-744?

tealefristoe (changes requested):
Overall looks good. I just had a few questions/suggestions that I think are worth addressing.

v3/cypress/e2e/plugin.spec.ts
```diff
@@ -326,19 +332,66 @@ context("codap plugins", () => {
     webView.clearAPITesterResponses()
     // TODO Check for dragleave notification when dragging to plugin then out of plugin
 
+    // For tests involving drag and drop of components with attribute "Attribute Name",
+    // we use a different drag and drop code because the cy.command version includes a
+    // "contains" expectation of text "Attribute Name", which it cannot find because it is over
+    // two spans. This code omits the "contains" expectation because the selector already has enough
+    // information without needing to find a specific text in a span
```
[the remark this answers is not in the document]
eireland: The default column width came from V2. 
I couldn't get the existing function to work with the attribute name over 2 spans. I'll modify it and add it to the table element object.

v3/cypress/e2e/formula/formula-component.spec.ts
eireland: Part of this PR removed the single case created when a new table is first created. 
Input row index cell does not open the index column menu so new cases cannot be inserted without having to enter data.
The `table.spec.ts` lines 738-744 tests that a new case is created at the bottom when user enters data into the input line.

kswenson (approved):
👍 LGTM aside from the issues @tealefristoe pointed out.

v3/cypress/e2e/formula/formula-component.spec.ts
eireland: done

tealefristoe (approved):
Looks good, thanks for addressing those issues 👍

PR #1654, opened 2024-11-26
fix: preserve case and item ids on import of v2 document

kswenson:
[[PT-188383228]](https://www.pivotaltracker.com/story/show/188383228)

- adds a DEBUG_CASE_IDS options which displays case ids in the index column of the case table
- some older documents don't save item ids, so case ids are preserved but not item ids for those documents
- in testing, I tried loading the `24cats.codap` sample document, which led to some unrelated adornments-related typing changes just to make the file importable. There's more work to be done here, for which there is already a separate story: [[PT-188618233]](https://www.pivotaltracker.com/story/show/188618233).

### DIFF
--- a/v3/src/components/graph/adornments/v2-adornment-importer.ts
+++ b/v3/src/components/graph/adornments/v2-adornment-importer.ts
@@ -1,9 +1,8 @@
-
+import { ICodapV2PlotStorage } from "../../../v2/codap-v2-types"
 import { IAttribute } from "../../../models/data/attribute"
 import { typedId } from "../../../utilities/js-utils"
 import { updateCellKey } from "./adornment-utils"
 import { kCountType } from "./count/count-adornment-types"
-import { ILSRLInstance } from "./lsrl/lsrl-adornment-model"
 import { kLSRLType } from "./lsrl/lsrl-adornment-types"
 import { kMovableLineType } from "./movable-line/movable-line-adornment-types"
 import { kMovablePointType } from "./movable-point/movable-point-adornment-types"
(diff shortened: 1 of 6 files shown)


v3/src/components/graph/adornments/v2-adornment-importer.ts
```diff
@@ -322,12 +321,13 @@ export const v2AdornmentImporter = ({data, plotModels, attributeDescriptions, yA
...
 
   // MOVABLE VALUES
-  const movableValuesAdornment = v2Adornments.multipleMovableValues
+  const movableValuesAdornment = v2Adornments?.multipleMovableValues
   if (movableValuesAdornment) {
     const values: Record<string, any> = {}
     instanceKeys?.forEach((key: string) => {
       const plotValues: number[] = []
-      movableValuesAdornment.valueModels.forEach((valueModel: Record<string, any>) => {
+      const valueModels = movableValuesAdornment.valueModels ?? movableValuesAdornment.values
```
scytacki, line 329: I was running into this same problem yesterday and tracked down several documents with `movableValuesAdornment.values`. From all of the documents I looked at, the type of `movableValuesAdornment.values` was `{isVisible: boolean, value: number}[]`. So there would be no `valueModel.values` and `valueModel.values._main` would then crash. 

If you want to just ignore the `movableValuesAdornment.values` in this PR, I can fix it in mine.

v3/src/v2/codap-v2-document.ts
```diff
@@ -161,17 +177,30 @@ export class CodapV2Document {
   }
 
   registerCases(data: IDataSet, cases: ICodapV2Case[], level: number) {
+    const v2CollectionInfo = this.v2CaseIdInfoArray[level]
+    const groupKeyCaseIds = v2CollectionInfo.groupKeyCaseIds
     cases.forEach(_case => {
-      const { guid, values } = _case
+      // some v2 documents don't store item ids, so we generate them if necessary
+      const { guid, itemID = v3Id(kItemIdPrefix), values } = _case
```
scytacki, line 184: I found that the v2 itemIDs were typed as numbers, but in the document I was testing they were strings. I didn't check this branch out locally to test it, but I just wanted to point that out incase it was going to cause problems.

scytacki (changes requested):
Looks good.

v3/src/v2/codap-v2-document.ts
kswenson: I think typing them as`number` was just an error because as far as I know they've always been strings.

v3/src/components/graph/adornments/v2-adornment-importer.ts
kswenson: I added a commit which updates the types, removes the handling of legacy `values`, and adds a comment about the need to handle legacy `values`.

scytacki (approved):
looks good

PR #1655, opened 2024-11-27
Fix for bug in which splitting on top and right not working properly

bfinzer:
[#188589248] Bug fix: Plots not showing when split on top and/or right

* `matchCirclesToData` was being called too early (in GraphController:handleAttributeAssignment); i.e. before we had a chance to call `clearCasesCache` so that `categoryArrayForAttrRole` gets invalidated and can subsequently return the categories belonging to the 'topSplit' or 'rightSplit' attribute. But it turns out we didn't need to call it there since it gets called later.
* Also revised `potentiallyCategoricalRoles` (renamed to `categoricalRoles)` to only return roles that actually have a categorical attribute. Before we were including categories even for numeric attributes if they had a categorySet, which we don't want to do.

### DIFF
--- a/v3/src/components/data-display/models/data-configuration-model.ts
+++ b/v3/src/components/data-display/models/data-configuration-model.ts
@@ -319,8 +319,8 @@ export const DataConfigurationModel = types
         return self.metadata.getCategorySet(attributeID)
       }
     },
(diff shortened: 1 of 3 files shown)


kswenson (approved):
👍 LGTM

PR #1656, opened 2024-11-27
feat: support DEBUG_CFM_NO_AUTO_SAVE debug option to disable auto-save

kswenson:
[[PT-188623447]](https://www.pivotaltracker.com/story/show/188623447)

When testing import/export code, it is useful to be able to open documents without worrying about them possibly being overwritten via auto-save. (I inadvertently overwrote a legacy document I was testing with, which led to this PR.)

Adding `cfmNoAutoSave` to the `debug` settings in the `Application` tab of the Chrome debugger will disable auto-save for any document loaded during that session.


kswenson:
> Can you update the readme?

Done

kswenson:
Merging after addressing @scytacki's code review.

PR #1657, opened 2024-11-27
Turn off animation for map points

bfinzer:
[#188617540] Bug fix: Points in map should not animate in from top left

* For the time being we don't animate the appearance of new points at all. This is better than the animation of points from the top left corner, which was occurring for all points for each refresh. This behavior was preventing meaningful slider animation of map points.


kswenson (approved):
👍 LGTM

PR #1658, opened 2024-11-27
improve adornment v2 importer typing

scytacki:
This turned up three bugs:
- `yAttributeDescriptions.y` doesn't exist
- if the v2 document had a plottedStDev adornment a stDev and stErr adornment would be added to the imported document
- the creation of this erroneous stErr adornment did not set to the number of standard errors correctly.

The importer now handles an older format of the multipleMovableValues adornment.

This also fixes other v2 types that didn't match the one random document from cfm-shared. I'll be checking more documents and will an submit a new PR with those changes.


v3/src/components/graph/adornments/v2-adornment-importer.ts
```diff
@@ -128,21 +153,31 @@ const instanceKeysForAdornments = (props: IInstanceKeysForAdornmentsProps) => {
   return instanceKeys
 }
 
+export type AdornmentAttributeDescriptions = Partial<Record<GraphAttrRole, IAttributeDescriptionSnapshot>>
```
kswenson, line 156: `AdornmentAttributeDescriptions` aren't adornment specific, so I'd be inclined to call it something like `GraphAttributeDescriptions` and move the definition into `data-display-types.ts` or `graphing-types.ts`.

kswenson (approved):
👍 Looks good -- just one naming suggestion

PR #1659, opened 2024-11-27
fix: import of v2 documents with maps

kswenson:
[[PT-188620026]](https://www.pivotaltracker.com/story/show/188620026)
[[PT-188614511]](https://www.pivotaltracker.com/story/show/188614511)

Fixes several issues discovered while importing a number of legacy v2 documents with maps.

- improves types for Map component
- supports legacy Map component format

Note that it doesn't appear that v2 meaningfully supports the legacy v2 Map format, i.e. most of the properties are ignored and the Map reverts to its default behavior. We follow the same approach for v3, whereby the legacy types are acknowledged but not actually migrated.

Also fixes a v3 bug/limitation which prevented the Map component from supporting points and boundaries from the same data set.


scytacki (approved):
Looks good to me.

Would it be worth adding a test of this? I see there is a `v2-graph-importer.test.ts` which uses a special mammals-graph doc to test it.  So perhaps something similar for the map importer would be worthwhile.

kswenson:
> Looks good to me.
> 
> Would it be worth adding a test of this? I see there is a `v2-graph-importer.test.ts` which uses a special mammals-graph doc to test it. So perhaps something similar for the map importer would be worthwhile.

I added jest tests.

bfinzer (commented):
👍🏻LGTM
Nice job, especially of making sense of ICodapV2MapLegacyStorage

PR #1660, opened 2024-11-27
Fix for numeric axis resize & rescale bug

bfinzer:
[#188523090] Bug fix: Graph numeric axis labels do not restore correctly on component resize

* This bug was introduced when we attempted to limit the amount of numeric axis redrawing we had to do. Fix by starting from scratch each time.


kswenson (approved):
👍 LGTM

PR #1661, opened 2024-11-27
Odds and ends

kswenson:
- fix lint warnings
- add `caseIds` debug setting to README


kswenson:
Merging without review as the changes are trivial.

PR #1663, opened 2024-12-03
Restore dynamic update of mean adornment

bfinzer:
[#188640918] Bug fix: When points in dot plot are dragged, mean isn't updating

* This bug first appeared in build 1949, in October, 2024. Some update machinery in data-configuration-model.ts was removed around then so `GraphContentModel:updateAdornments` was no longer being called when case values are changed. The only thing that needed to be fixed was to increment the `DataConfigurationModel`'s `casesChangeCount` in `handleSetCaseValues`. Doing so restored the previous behavior with no apparent added cost.


kswenson (approved):
👍 LGTM

PR #1665, opened 2024-12-03
feat: v2 export infrastructure

kswenson:
[[PT-188613413]](https://www.pivotaltracker.com/story/show/188613413) Initial infrastructure for v2 export
[[PT-188616789]](https://www.pivotaltracker.com/story/show/188616789) Export Calculator in v2 format

- supports `saveAsV2` `debug` property (`DEBUG_SAVE_AS_V2` constant) to enable saving in v2 format by default
- supports registration of tile-specific exporters
- implements exporter for Calculator component

### DIFF
--- a/v3/src/models/app-state.ts
+++ b/v3/src/models/app-state.ts
@@ -101,10 +101,7 @@ class AppState {
   }
 
   @flow
-  *setDocument(
-    snap: ISerializedV3Document | ICodapV2DocumentJson,
-    metadata?: Record<string, any>
-  ) {
+  *setDocument(snap: ISerializedDocument, metadata?: Record<string, any>) {
     // stop monitoring changes for undo/redo on the existing document
     this.disableDocumentMonitoring()
 
@@ -144,11 +141,11 @@ class AppState {
           }
         })
       }
-      if (content.revisionId && this.treeManager) {
+      if (snap.revisionId && this.treeManager) {
         // Restore the revisionId from the stored document
         // This will allow us to consistently compare the local document
         // to the stored document.
-        this.treeManager.setRevisionId(content.revisionId)
+        this.treeManager.setRevisionId(snap.revisionId)
       }
 
       // monitor document changes for undo/redo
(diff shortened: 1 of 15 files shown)


v3/src/models/app-state.ts
```diff
@@ -94,13 +102,13 @@ class AppState {
...
   *setDocument(
-    snap: ISerializedDocumentModel | ICodapV2DocumentJson,
+    snap: ISerializedV3Document | ICodapV2DocumentJson,
     metadata?: Record<string, any>
   ) {
     // stop monitoring changes for undo/redo on the existing document
     this.disableDocumentMonitoring()
 
-    let content: ISerializedDocumentModel
+    let content: ISerializedV3Document
```
scytacki, line 111: Since we might be restoring v2 documents that were saved by v3. The code below that is using `content.revisionId` should probably be changed to `snap.revisionId`. Otherwise I think the conversion from v2 will lose the `revisionId`. That probably also means the type above of `ICodapV2DocumentJson` should be changed to `ISerializedV2Document`

v3/src/v2/codap-v2-tile-exporters.ts
```diff
@@ -0,0 +1,65 @@
...
+    layout: {
+      width: tileWidth,
+      height: layout.isMinimized ? kTitleBarHeight : tileHeight,
+      left: layout.position.x,
+      top: layout.position.y,
+      isVisible: !layout.isHidden,
+      zIndex: layout.zIndex
+    },
+    savedHeight: layout.isMinimized ? tileHeight : null
+  } as Maybe<CodapV2Component>
```
scytacki, line 64:
```suggestion
  } as CodapV2Component
```
Ideally you could get rid of this `as` but then we'd have to figure out how to restrict the `type` property.

scytacki (approved):
Looks good. I just found a typing fix, and a potential problem with revisionId when we restore v2 documents saved by v3.

PR #1674, opened 2024-12-05
fix: handling of set-aside cases in imported/restored documents

kswenson:
[[PT-188652082]](https://www.pivotaltracker.com/story/show/188652082)

This started as part of the effort to improve the TypeScript types for v2 documents. As part of fixing the types for `setAsideItems`, I noticed that `setAsideItems` were not being imported at all. Therefore, this PR:
- adds correct types for `setAsideItems` in v2 documents
- fixes the importer so that `setAsideItems` are imported correctly from v2 documents
- fixes a model initialization bug that prevented saved/restored `setAsideItems` from being handled correctly in newly restored models (restored from either v2 or v3 documents)
- [unrelated] eliminates the `AllowStringIds` and `ICodapV2AttributeV3` types
  - I believe these are left over from early id-handling efforts in which we thought we could get away with passing string ids through the v2 API. It became apparently fairly quickly, however, that some plugins expected ids to be strictly numeric and so now we no longer attempt to pass strings through the v2 API, which means the above types are no longer necessary.


tealefristoe (approved):
Looks good 👍

PR #1676, opened 2024-12-06
chore: use `@stylistic/eslint-plugin` for deprecated style rules

kswenson:
[[PT-187950465]](https://www.pivotaltracker.com/story/show/187950465)

Part 1 of ESLint v9 with some help from ChatGPT as part of AI Day. Not as much help as I would like from ChatGPT, however.

ESLint v9 deprecates a number of stylistic rules that have moved to the `@stylistic/eslint-plugin`. This PR updates the ESLint configuration to use the `@stylistic/eslint-plugin` version of the deprecated rules.

PR #1678, opened 2024-12-08
chore: eslint v9 upgrade [PT-187950465]

kswenson:
Rooted on #1676, which should be merged first.

Upgrade to ESLint v9 and the new flat configuration files with some help from the ESLint migration tool and a bit from ChatGPT.

Note: v9 when uncached appears to be much slower than v8 for reasons I haven't looked into. Once run, the cached times are similarly quick, but the first run after clearing the cache can be quite slow. 🤷 

v8: 18-21s uncached, 1s cached
v9: 2m45 - 3m30s uncached, 1s cached


scytacki (approved):
looks good to me 👍

PR #1679, opened 2024-12-10
chore: enhance attribute value types

kswenson:
The substantive change is adding `object` to `IValueType` and some related types.
Many of the other changes are path changes that resulted from moving some definitions into `attribute-types.ts`.


v3/src/v2/codap-v2-document.ts
```diff
@@ -206,7 +206,7 @@ export class CodapV2Document {
       if (level === 0) {
         // FIXME: values can include objects not just the primitives defined by IValueType
         // FIXME: should the itemID overwrite any __id__ returned by toCanonical?
```
scytacki, line 208: It looks like you addressed both of these issues.

scytacki (approved):
looks good, just the one comment about removing the `FIXME` comments.

The reason for the second FIXME was because TS was complaining when the `__id__` was first. It said `toCanonical` always returned an `__id__` so it was unnecessary to set one. I couldn't figured out why my changes caused this to be a problem so I just moved `__id__` to the end. I guess your changes fixed this if TS isn't complaining anymore.

PR #1680, opened 2024-12-10
feat: v2 export of DataSets and globalValues

kswenson:
Based on #1679, which should be merged first.


scytacki (approved):
Looks good to me.

Originally I was questioning the benefit of having `Attribute#value` return a number if the value can be converted to one. It means information information will be lost by default, for example a string value set as "0100" will just be returned as 100 by default now. 

This trickles down to plugins too. If the v2 plugins have been working fine always receiving string values then perhaps continuing to give them strings would be better so they get more information. 

However after looking though the whole PR it seems like since CODAP is built for working with numbers it seems more reasonable to lose this information by default. At some point it it might be good to introduce an efficient way to save whether the string or numeric value is the "original" value. In most manual entry cases it'd be the string but with plugins and other controls which aren't text fields the number could be the "original" value.

kswenson:
> Looks good to me.
> 
> Originally I was questioning the benefit of having `Attribute#value` return a number if the value can be converted to one. It means information information will be lost by default, for example a string value set as "0100" will just be returned as 100 by default now.
> 
> This trickles down to plugins too. If the v2 plugins have been working fine always receiving string values then perhaps continuing to give them strings would be better so they get more information.
> 
> However after looking though the whole PR it seems like since CODAP is built for working with numbers it seems more reasonable to lose this information by default. At some point it it might be good to introduce an efficient way to save whether the string or numeric value is the "original" value. In most manual entry cases it'd be the string but with plugins and other controls which aren't text fields the number could be the "original" value.

Your point is a good one, but the argument actually cuts the other way, i.e. in favor of the change. In v2 and in early versions of the `DataSet` values were stored in their "typed" form, i.e. numbers as numbers, strings as strings, etc. and so methods like `value()` would return the raw typed value. When updating the `DataSet` for v3, it occurred to me that for serialization purposes and for code simplification, it made more sense for the string version of a given value to be the canonical one, in part for the reason you point out -- that conversion to a number can lose information. So the `DataSet` now always serializes the string values, but maintains cached numeric values as well at runtime. At the time that change was made, functions like `value()` should have been updated so that they continued to return typed values for compatibility with existing code/clients. The result was that a lot of tests were written to test string value returns that would have been simpler and more compatible with v2 precedent to test numeric return values. In theory, this brings us to a better situation in that values are serialized internally as strings, maintaining full fidelity, but clients that expect numbers can get them by default without having to perform additional conversions and maintaining compatibility with v2 expectations.

PR #1682, opened 2024-12-11
handle all properties defined in a sample of 6,000 files from cfm-shared

scytacki:
This also removes a few `any`s from the types. This required updating a few bits of import code. 

It fixes:
- the handling of a null `plotBackgroundImageLockInfo` in the graph
- handling of legacy setAsideItems which didn't have ids and the values were at the top level

It disables `hiddenCases` importing on the map because it would not have worked.

It does not fix:
- the handling of null or undefined values for `*UpperBound` and `*LowerBound` properties in the graph
- the handling of a `*AttributeType=0` in the graph


v3/src/v2/codap-v2-types.ts
```diff
@@ -25,10 +25,15 @@ export interface ICodapV2Attribute {
   deletedFormula?: string
   precision?: number | string | null
   unit?: string | null
+  // TODO_V2_IMPORT decimals is not imported
+  // it occurs 604 times in 36 files in cfm-shared
+  // in 33 cases the value is "2"
+  // the rest of the cases the value is "0"
+  decimals?: string
```
kswenson, line 32: I would guess that `decimals` was a very early precursor to `precision`.

v3/src/v2/codap-v2-types.ts
```diff
@@ -101,11 +111,23 @@ export interface ICodapV2SetAsideItem {
...
   description?: string
+  // TODO_V2_IMPORT data context metadata source is not imported
+  // unknown how many instances in this location
+  source?: string
+  // TODO_V2_IMPORT data context metadata importDate is not imported
+  // at least 20,000 instances in cfm-shared
+  importDate?: string
+  // TODO_V2_IMPORT "import date" is not imported
+  // 2,025 instances in cfm-shared
+  "import date"?: string
```
kswenson, line 130: This just looks like detritus from a bug.

v3/src/v2/codap-v2-types.ts
```diff
@@ -544,10 +723,17 @@ export function isV2MapLegacyStorage(obj: unknown): obj is ICodapV2MapCurrentSto
...
+    // this array was passed right into MST where it is typed as a string array
+    // There are 296 instances where this is a non-empty array in cfm-shared
+    hiddenCases?: IGuidLink<"DG.Case">[],
     legendColl?: IGuidLink<"DG.Collection">,
     // We sometimes see an array of links here
     legendAttr?: IGuidLink<"DG.Attribute"> | IGuidLink<"DG.Attribute">[],
+    // TODO_V2_IMPORT tHiddenCases
+    // this occurs 523 times in cfm-shared
+    // in all cases the value is `[]`
+    tHiddenCases?: unknown[]
```
kswenson, line 736: Looks like detritus from a bug.

v3/src/v2/codap-v2-types.ts
```diff
@@ -642,6 +837,28 @@ export interface ICodapV2BaseComponent {
...
+    // TODO_V2_IMPORT y is not imported
+    // appears 5,258 times in cfm-shared
+    // based on the results for `right`, this must be optional
+    y?: number
+
+    // these *Orig properties only occur in a single file in cfm-shared
+    leftOrig?: number
+    topOrig?: number
+    widthOrig?: number
+    heightOrig?: number
```
kswenson, line 861: These seem ignorable.

kswenson (approved):
👍 Looks good -- just a few comments on what you found.

v3/src/v2/codap-v2-types.ts
scytacki: I'm not sure what to do with these type of properties. 
If I remove them, then if we evaluate the cfm-shared again they will popup and waste time.
If one of these buggy documents is actually used, in some cases these unhandled props might cause problems. 

The best thing I can think of is to just leave them, remove the `TODO_V2_IMPORT`, and update the comment to indicate why they should be ignored.

A further step would be add a second set of types which are used for exporting V2. These would be the canonical v2 types. The importing types would extend them with these extra props which might be seen on import. To do that, I think we'd have to have two copies of the type hierarchy. The import schema would be generated from the import root.

scytacki:
I'm going to merge with the comments as is to see what categories @bfinzer comes up with for the rest of them. 

Perhaps there is a metric based approach: if the property occurs less than X times in cfm-shared, or perhaps in less than Y files, and we've confirmed it isn't supported by existing v2 then we remove the `TODO`. 

I'm not sure how hard it is to confirm if a property is or isn't supported. Hopefully if you know the codebase well you will know what the property should do and can see if it makes a difference in a constructed or discovered file.

If it is worthwhile I can pretty easily collect example files for most of the properties.

PR #1683, opened 2024-12-11
188651851 v3 DI Multiple Y Attributes

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/188651851

This PR adds support for multiple y attributes in graph API requests.
- A graph with multiple y attributes can be created via `yAttributeIDs` or `yAttirbuteNames` in `create component` requests.
- `yAttributeIDs` and `yAttributeNames` are now included in responses to `get component` requests for graphs.
- Multiple y attributes can be set for a graph via `yAttributeIDs` or `yAttributeNames` in `update component` requests.

Additional bug fixes:
- A bug in `get component` responses was fixed in this PR. Previously, attribute ids returned were v3 format, that is, strings starting with "ATTR". Now attribute ids returned are in v2 format, just the number part.
- The dummy model used in `create component type: table` requests is now destroyed after a snapshot is taken of it. Previously, this model was not being garbage collected and reactions associated it were continuing to fire.
- `GraphDataConfigurationModel` now overrides `handleDataSetChange`, a function called in a reaction defined in `DataConfigurationModel`. The base function always sets the number of `filteredCases` to one, which was preventing dots for multiple y attributes from showing up on freshly created graphs. Now, the proper number of `filteredCases` is set for graphs when the dataset changes.

### DIFF
--- a/v3/src/components/graph/component-handler-graph.test.ts
+++ b/v3/src/components/graph/component-handler-graph.test.ts
@@ -84,6 +84,40 @@ describe("DataInteractive ComponentHandler Graph", () => {
     handler.delete!({ component: tileIds })
     expect(documentContent.tileMap.size).toBe(0)
 
+    // Create a graph with multiple y attributes using ids
+    const resultYIDs = handler.create!({}, {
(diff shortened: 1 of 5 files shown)


kswenson (approved):
👍 LGTM

PR #1684, opened 2024-12-12
feat: export case table in v2 format

kswenson:


scytacki (approved):
Looks good. Just has a what is probably a simple merge conflict with some code in my previous PR which has now been merged.

PR #1685, opened 2024-12-13
Fix graph binWidth computation to avoid crash

bfinzer:
[#188590950] Bug fix: Graph: crash on axis switch from categorical parent attribute to date child attribute

* The problem here had to do with the GraphContentModel property `binWidth`. It was being assumed that it could safely default to 1. But this is not the case. It represents the width of a "bin" in world coordinates and is used by dot plots both binned and unbinned. In the unbinned case with an axis with very high numeric values, like a date axis where the values are in seconds, a binWidth of 1 leads to a very large number of bins and therefore very large arrays, which led to the crash in this situation.
* In this PR we no longer choose a default numeric value for `binWidth`. Instead we set it to `undefined` until it can be computed properly from min/max values for the axis. This change necessitated a flurry of places where we have to check to see if `binWidth` is undefined and, essentially, bail if so.

### DIFF
--- a/v3/src/components/graph/adornments/count/count-adornment-component.tsx
+++ b/v3/src/components/graph/adornments/count/count-adornment-component.tsx
@@ -46,13 +46,13 @@ export const CountAdornment = observer(function CountAdornment(props: IAdornment
       // movable values present at the same time.
       if (graphModel.pointDisplayType === "bins") {
         const { binWidth, minBinEdge, maxBinEdge, totalNumberOfBins } = graphModel.binDetails()
-        return [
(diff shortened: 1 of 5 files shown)


tealefristoe (changes requested):
This generally looks good to me, though I'll admit I don't know this part of the code very well. I left one small suggestion that I think you should consider. The main reason I'm requesting changes is that it looks like a couple of the cypress tests are failing.

tealefristoe (approved):
Looks great 👍 

I've encountered similar situations where it's not clear if a view or an action is more appropriate, and I always find it very confusing.

PR #1686, opened 2024-12-13
188651857 v3 DI Top Level Graph Features

tealefristoe:
PT Stories:
https://www.pivotaltracker.com/story/show/188651857
https://www.pivotaltracker.com/story/show/188651865

This PR adds support for many top level graph features to the API. The following features are all supported in `create`, `get`, and `update` requests:
- `backgroundColor`
- `displayOnlySelectedCases`
- `filterFormula`
- `hiddenCases`
- `pointColor`
- `pointConfig` ("points", "bars", "bins", etc)
- `pointsFusedIntoBars`
- `pointSize`
- `showMeasuresForSelection`
- `strokeColor`
- `strokeSameAsFill`
- `transparent` (doesn't seem to do anything)
- `xAttributeType`, `yAttributeType`, `y2AttributeType`

### DIFF
--- a/v3/src/components/graph/component-handler-graph.test.ts
+++ b/v3/src/components/graph/component-handler-graph.test.ts
@@ -15,6 +15,8 @@ import { IGraphContentModel, isGraphContentModel } from "./models/graph-content-
 
 describe("DataInteractive ComponentHandler Graph", () => {
   const handler = diComponentHandler
+  const create = handler.create!
+  const update = handler.update!
   const documentContent = appState.document.content!
   const { dataset: _dataset } = setupTestDataset()
   const dataset = documentContent.createDataSet(getSnapshot(_dataset)).sharedDataSet.dataSet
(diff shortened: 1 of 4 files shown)


tealefristoe:
@bfinzer @kswenson 
This PR is pretty close to being done. I still need to add tests, but before I do I have several questions that I want to ask:
- Are there any other graph features that should be included? (See the PR description.)
- Do we want to include all of the listed features? Should any be in `update` but not `create`?
- Should any feature names be changed?
- There are many different graph types, and not all of these features make sense for every graph type. I'm not doing any compatibility or error checking when assigning the features, I'm just doing whatever the API request asks. Is this ok? I'm guessing it would be pretty difficult and time consuming to thoroughly check everything.
- Some of these features don't work when creating a graph. In particular, bar charts don't work in fresh graphs; it seems like you need to transition a point graph into a bar graph for them to work properly. What should I do about this? I could remove these as options from the create API request. Or I could fix the graph to make bar charts work immediately, though I'm guessing it won't be easy. My inclination is to leave the features in place and write a PT story about fixing newly created bar graphs to deal with down the line.

bfinzer:
* We don't need to support any additional graph features at this time.
* I think it's fine to include all the listed features.
* I think the feature names are fine as you've got them.
* We don't need error checking, at least for now.
* Writing the story for creating a bar chart is a good idea. We don't need this feature to work for now.

kswenson (approved):
👍 LGTM

PR #1689, opened 2024-12-16
188527686 formula editor improvements

eireland:
Makes formula editor modal resizable.
Removes default codap modal header.

Deployed here:
https://codap3.concord.org/branch/formula-editor-improvements/?sample=coasters

### DIFF
--- a/v3/src/components/common/edit-formula-modal.scss
+++ b/v3/src/components/common/edit-formula-modal.scss
@@ -21,7 +21,7 @@ $edit-formula-modal-min-width-px: #{$edit-formula-modal-min-width}px;
     height: inherit;
     margin-bottom: 10px;
     min-height: 72px;
-    min-width: 400px;
+    min-width: $edit-formula-modal-min-width-px;
 
     .formula-form-control {
       display: flex;
@@ -42,7 +42,7 @@ $edit-formula-modal-min-width-px: #{$edit-formula-modal-min-width}px;
         width: 100%;
 
         .formula-editor-input {
-          width: calc(100% - 50px);
+          width: calc(100% - 50px); //50px is to account for modal margins and paddings
 
           .cm-editor {
             height: 100%;
@@ -83,7 +83,7 @@ $edit-formula-modal-min-width-px: #{$edit-formula-modal-min-width}px;
 
   .codap-modal-corner {
     position: absolute;
-    bottom: 0px;
+    bottom: 0;
     right: 0;
     height: 22px;
     width: 22px;
--- a/v3/src/components/common/edit-formula-modal.scss
+++ b/v3/src/components/common/edit-formula-modal.scss
@@ -1,8 +1,18 @@
 @use "../vars";
+$edit-formula-modal-min-height: 180;
+$edit-formula-modal-min-height-px: #{$edit-formula-modal-min-height}px;
+$edit-formula-modal-min-width: 400;
+$edit-formula-modal-min-width-px: #{$edit-formula-modal-min-width}px;
+
+// https://mattferderer.com/use-sass-variables-in-typescript-and-javascript
+:export {
+  editFormulaModalMinHeight: $edit-formula-modal-min-height;
+  editFormulaModalMinWidth: $edit-formula-modal-min-width;
+}
 
 .codap-modal-content {
-  min-height: 180px;
-  min-width: 400px;
+  min-height: $edit-formula-modal-min-height-px;
+  min-width: $edit-formula-modal-min-width-px;
 
   .formula-modal-body {
     cursor: default;
@@ -32,9 +42,8 @@
         width: 100%;
 
         .formula-editor-input {
-          height: 100%;
-          width: 100%;
-          flex-grow: 1;
+          width: calc(100% - 50px);
+
           .cm-editor {
             height: 100%;
             width: 100%;
--- a/v3/src/components/common/edit-formula-modal.scss
+++ b/v3/src/components/common/edit-formula-modal.scss
@@ -3,8 +3,15 @@
 .codap-modal-content {
   .formula-modal-body {
     cursor: default;
+    display: flex;
+    flex-direction: column;
+    height: 100%;
 
     .formula-form-control {
+      display: flex;
+      flex-grow: 1;
+      align-items: flex-start;
+
       .title-label {
         width: 140px;
       }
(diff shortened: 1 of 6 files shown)


bfinzer:
I'm seeing two lint warnings: `ModalCloseButton` and `ModalHeader` are defined but not used in edit-formula-modal.tsx.

bfinzer:
In the V2 formula editor text wraps when it reaches the right edge. This should be the behavior in V3 as well. If this is easy to bring about, it could be part of this PR, otherwise a separate story.

bfinzer (approved):
👍🏻LGTM
Just a couple notes.

v3/src/components/common/edit-formula-modal.tsx
```diff
@@ -28,10 +29,14 @@ interface IProps {
 export const EditFormulaModal = observer(function EditFormulaModal({
   applyFormula, formulaPrompt, isOpen, onClose, titleInput, titleLabel, titlePlaceholder, value
 }: IProps) {
+  const minWidth = 400
+  const minHeight = 180
   const [showValuesMenu, setShowValuesMenu] = useState(false)
   const [showFunctionMenu, setShowFunctionMenu] = useState(false)
   const formulaEditorState = useFormulaEditorState(value ?? "")
   const { formula, setFormula } = formulaEditorState
+  const [dimension, setDimension] = useState({ width: minWidth, height: minHeight })
```
kswenson, line 38:
```suggestion
  const [dimensions, setDimensions] = useState({ width: minWidth, height: minHeight })
```

v3/src/components/common/edit-formula-modal.tsx
```diff
@@ -47,6 +52,7 @@ export const EditFormulaModal = observer(function EditFormulaModal({
     setShowFunctionMenu(false)
     setFormula(value || "")
     onClose?.()
+    setDimension({ width: minWidth, height: minHeight })
```
kswenson, line 55: Are we sure we want to reset the size on close? I can see an argument that retaining its size across invocation would be desirable.

v3/src/components/common/edit-formula-modal.tsx
```diff
@@ -28,10 +29,14 @@ interface IProps {
 export const EditFormulaModal = observer(function EditFormulaModal({
   applyFormula, formulaPrompt, isOpen, onClose, titleInput, titleLabel, titlePlaceholder, value
 }: IProps) {
+  const minWidth = 400
+  const minHeight = 180
```
kswenson, line 33: These values appear here and in the CSS. Either import the values from the .css or use CSS variables, e.g. `--default-codap-modal-width` and `--default-codap-modal-height`, which are accessible from CSS and JS.

v3/src/components/common/edit-formula-modal.tsx
```diff
@@ -86,20 +92,44 @@ export const EditFormulaModal = observer(function EditFormulaModal({
     event.stopPropagation()
   }
 
+  const handleResizeModal = useCallback((e: React.PointerEvent) => {
+    if (e.pointerId !== undefined) {
+      e.currentTarget.setPointerCapture(e.pointerId)
+    }
+
+    const modalRect = document.querySelector(".codap-modal-content")?.getBoundingClientRect()
```
kswenson, line 100: Document-wide queries are fragile -- what if there are two modals showing? Either use a ref or a library like `react-resize-detector`.

v3/src/components/common/edit-formula-modal.tsx
```diff
@@ -86,20 +92,44 @@ export const EditFormulaModal = observer(function EditFormulaModal({
...
+    let resizingWidth = startWidth, resizingHeight = startHeight
+
+    const onPointerMove = (pointerMoveEvent: { pageX: number; pageY: number }) => {
+      const xDelta = pointerMoveEvent.pageX - startPosition.x
+      const yDelta = pointerMoveEvent.pageY - startPosition.y
+      resizingWidth = Math.max(startWidth + xDelta, minWidth)
+      resizingHeight = Math.max(startHeight + yDelta, minHeight)
+      setDimension({
+        width: resizingWidth, height: resizingHeight,
+      })
```
kswenson, line 114: I recommend rounding these values to integer pixels -- they're much easier to deal with when debugging and I don't think fractional pixel widths are necessary for users.

v3/src/components/common/edit-formula-modal.scss
```diff
@@ -1,10 +1,23 @@
 @use "../vars";
 
 .codap-modal-content {
+  min-height: 180px;
+  min-width: 400px;
```
kswenson, line 5: These values appear here and in the JS. They should be shared. See comment elsewhere for more details.

kswenson (changes requested):
👍 Generally looks good -- see inline suggestions.

>In the V2 formula editor text wraps when it reaches the right edge. This should be the behavior in V3 as well. If this is easy to bring about, it could be part of this PR, otherwise a separate story.

According to ChatGPT, CodeMirror can be configured to wrap text by adding `EditorView.lineWrapping` to the list of extensions when initializing the editor.

v3/src/components/common/edit-formula-modal.scss
```diff
@@ -1,10 +1,33 @@
...
+  min-width: $edit-formula-modal-min-width-px;
+
   .formula-modal-body {
     cursor: default;
+    display: flex;
+    flex-direction: column;
+    height: inherit;
+    margin-bottom: 10px;
+    min-height: 72px;
+    min-width: 400px;
```
kswenson, line 24: Could this be:
```suggestion
    min-width: $edit-formula-modal-min-width-px;
```

v3/src/components/common/edit-formula-modal.scss
```diff
@@ -13,24 +36,63 @@
...
-    .formula-editor-button {
-      padding: 0 10px;
-      &:hover {
-        background: vars.$codap-teal-lighter;
-      }
-      &:active, &.menu-open {
-        background: vars.$codap-teal-light;
-        color: white;
+        .formula-editor-input {
+          width: calc(100% - 50px);
```
kswenson, line 45: Comment or named constant for the `50px` value

v3/src/components/common/edit-formula-modal.scss
```diff
@@ -13,24 +36,63 @@
...
     cursor: default;
+    border-radius: 6px;
+    position: relative;
+    bottom: 0;
+    right: 0;
+  }
+
+  .codap-modal-corner {
+    position: absolute;
+    bottom: 0px;
```
kswenson, line 86:
```suggestion
    bottom: 0;
```
For consistency

kswenson (approved):
👍 Looks good -- just a few small suggestions.

PR #1690, opened 2024-12-17
Export game view

scytacki:
This makes `isPlugin` a serialized field in V3. That is necessary so we can keep track of which WebViews are v2 GameViews and which are v2 WebViews.
The `setIsPlugin` is marked as `withoutUndo` so when a plugin is added and it sets up the iframe phone this change of the document state doesn't add an action to the undo stack.

The `contextStorage` field is added to the DataSet conversion. This was necessary to get a document to load with a plugin that had created a data table. However this document is still broken because it data context collections seem to be reversed and disconnected.

Name handling in GameViews adds more complexity to the already complex name and title handling. GameViews store their name in componentStorage.currentGameName instead of componentStorage.name

### DIFF
--- a/v3/src/components/web-view/web-view-model.ts
+++ b/v3/src/components/web-view/web-view-model.ts
@@ -49,7 +49,7 @@ export const WebViewModel = TileContentModel
     setDataInteractiveController(controller?: iframePhone.IframePhoneRpcEndpoint) {
       self.dataInteractiveController = controller
     },
-    setSubType(subType: WebViewSubType) {
+    setSubType(subType?: WebViewSubType) {
       withoutUndo()
       self.subType = subType
     },
--- a/v3/src/components/web-view/web-view-model.ts
+++ b/v3/src/components/web-view/web-view-model.ts
@@ -3,7 +3,7 @@ import { Instance, SnapshotIn, types } from "mobx-state-tree"
 import { DIMessage } from "../../data-interactive/iframe-phone-types"
 import { ITileContentModel, TileContentModel } from "../../models/tiles/tile-content"
 import { withoutUndo } from "../../models/history/without-undo"
-import { kWebViewTileType } from "./web-view-defs"
+import { kWebViewTileType, WebViewSubType, webViewSubTypes } from "./web-view-defs"
 
 export const kDefaultAllowEmptyAttributeDeletion = true
 export const kDefaultBlockAPIRequestsWhileEditing = false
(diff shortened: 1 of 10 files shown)


kswenson (approved):
👍 Looks good -- I added a commit which fixed the game view `name` issue you pointed out.

kswenson (changes requested):
Two thoughts occurred to me this morning:
1. Instead of a Boolean `isPlugin` property, it should be a `subType` property with value `plugin`. That way when `DG.GuideView` is implemented, it can be `subType` `guide`. I would leave it undefined for standard web views.
2. The importer for `DG.GameView` should set the `subType` property to `plugin`.

kswenson:
> Two thoughts occurred to me this morning:
> 
> 1. Instead of a Boolean `isPlugin` property, it should be a `subType` property with value `plugin`. That way when `DG.GuideView` is implemented, it can be `subType` `guide`. I would leave it undefined for standard web views.
> 2. The importer for `DG.GameView` should set the `subType` property to `plugin`.

After consultation with @scytacki, I added a commit which implemented `#1` above. Turns out `#2` was already implemented in what @scytacki had done.

kswenson (approved):
After consultation with @scytacki, I addressed my concerns in a follow-up commit.

v3/src/components/web-view/web-view-model.ts
```diff
@@ -37,15 +37,21 @@ export const WebViewModel = TileContentModel
...
+    get isPlugin() {
+      return self.subType === "plugin"
     }
   }))
   .actions(self => ({
     setDataInteractiveController(controller?: iframePhone.IframePhoneRpcEndpoint) {
       self.dataInteractiveController = controller
     },
-    setIsPlugin(isPlugin: boolean) {
+    setSubType(subType: WebViewSubType) {
```
scytacki, line 52: Maybe you want to allow `undefined` here so the subtype can be reverted back to the default non plugin, non guide.

scytacki (commented):
I can't officially approve but it looks good to me. 

My only comment is about allowing the v3 WebView to be turned back into a v2 WebView via the subType setter.

kswenson:
> I can't officially approve but it looks good to me. 
> 
> 
> 
> My only comment is about allowing the v3 WebView to be turned back into a v2 WebView via the subType setter.

Good catch -- I meant to do that. I'll update before merging.

PR #1691, opened 2024-12-18
feat: import/export case card from/to v2 documents

kswenson:

- fixes a bug in v2 data context export
- implements import of case card tiles from v2 documents
- implements export of case card tiles to v2 documents

### DIFF
--- a/v3/src/components/case-card/case-card-model.ts
+++ b/v3/src/components/case-card/case-card-model.ts
@@ -12,7 +12,7 @@ export const CaseCardModel = TileContentModel
   .named("CaseCardModel")
   .props({
     type: types.optional(types.literal(kCaseCardTileType), kCaseCardTileType),
-    // key is collection id; value is width
+    // key is collection id; value is percentage width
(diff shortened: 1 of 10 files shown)


scytacki (changes requested):
This looks good. You should remove the TODO_V2_IMPORT comments in codap-v2-types.ts that are handled now. There might be more but the ones I noticed are:
- the `isActive` on CaseCardStorage and TableStorage
- the `columnWidthPct` and `columnWidthMap` on CaseCardStorage.
- the general case card `ICodapV2CaseCardComponent`

kswenson:
> This looks good. You should remove the TODO_V2_IMPORT comments in codap-v2-types.ts that are handled now. There might be more but the ones I noticed are:
> 
> * the `isActive` on CaseCardStorage and TableStorage
> * the `columnWidthPct` and `columnWidthMap` on CaseCardStorage.
> * the general case card `ICodapV2CaseCardComponent`

Good idea -- done.

PR #1692, opened 2024-12-18
feat: export graph image snapshot (PT-188650872)

emcelroy:
PT Stories: [#188650872](https://www.pivotaltracker.com/story/show/188650872), [#188700179](https://www.pivotaltracker.com/story/show/188700179)

Adds basic support for exporting an image snapshot of a graph. The image snapshot can be requested via API, as well as exported as a PNG file via the CODAP UI. This work is related to [this DAVAI Plugin PR](https://github.com/concord-consortium/davai-plugin/pull/22), so the plugin can get a graph snapshot for processing by AI.


v3/src/components/data-display/components/legend/categorical-legend.tsx
```diff
@@ -349,7 +349,7 @@ export const CategoricalLegend = observer(
     }, [setDesiredExtent, layerIndex])
 
     return (
-      <svg className='legend-categories' ref={keysElt} data-testid='legend-categories'></svg>
+      <g className='legend-categories' ref={keysElt} data-testid='legend-categories'></g>
```
emcelroy, line 352: It seemed to me these `legend-categories`items don't need to be `svg` elements themselves since they always seem to be contained within a parent `svg`. I changed them to `g` elements because the `svg`s were causing duplications in the image file exports. If I'm overlooking a reason why these should remain `svg`s, the `graphSnaphsot` function in image-utils.ts will need to be adjusted.

kswenson (approved):
👍 Looks fine to me -- just a few minor suggestions/corrections. I also added @tealefristoe (for the API extensions) and @bfinzer (specifically for the graph svg-rendering changes) as reviewers as well.

v3/src/hooks/use-cfm-context.ts
```diff
@@ -0,0 +1,7 @@
+import { createContext, useContext } from "react"
+
+export const CfmContext = createContext<any>(undefined)
```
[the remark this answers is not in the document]
emcelroy: @kswenson Thanks. That change revealed that [this call](https://github.com/concord-consortium/codap/blob/a5c36e1e77b0d46f59535092c4882ef9c27079ff/v3/src/components/graph/components/camera-menu-list.tsx#L56) to `cfm.client.saveSecondaryFileAsDialog` was missing a callback function. It doesn't seem to me that CODAP needs to do anything once the download happens. So for now at least, I just set the callback to `() => null`. Does that seem OK?

v3/src/components/data-display/components/legend/categorical-legend.tsx
kswenson: Seems okay to me, but @bfinzer should take a look as well.

v3/src/data-interactive/data-interactive-types.ts
```diff
@@ -179,8 +182,10 @@ export interface DIResources {
   component?: DIComponent
   dataContext?: IDataSet
   dataContextList?: IDataSet[]
+  dataDisplay?: DIDataDisplay
   error?: string
   global?: IGlobalValue
+  imageSnapshot?: DIComponent
```
tealefristoe, line 188: I don't think this is being used anywhere. Can it be removed?

v3/src/data-interactive/handlers/data-display-handler.ts
```diff
@@ -0,0 +1,43 @@
...
+      const imgSnapshotRes = await handler?.get(component?.content)
+      const { exportDataUri, success } = imgSnapshotRes ?? {}
+      const values = success
+        ? { exportDataUri }
+        : { error: t("V3.DI.Error.dataDisplayNotFound") }
+
+      return { success, values }
+    } catch (e) {
+      console.error("Error in diDataDisplayHandler", e)
+      return { success: false, values: { error: t("V3.DI.Error.dataDisplayNotFound") } }
```
tealefristoe, line 38: Might as well use `dataDisplayNotFoundResult` here.

tealefristoe (approved):
The API extension looks great to me 👍

v3/src/components/data-display/components/legend/categorical-legend.tsx
bfinzer: Seems OK to me as well.

PR #1693, opened 2024-12-19
add dataContextFromURL api

scytacki:
PT Story: https://www.pivotaltracker.com/story/show/187738928

This PR implements a synchronous version of the `create dataContextFromURL` API request. It works similarly to v2, where the result is success if there's enough information to attempt to download the data context, even if the actual download fails. In this case, an alert is displayed so the user knows the download failed, but the plugin will not be informed.

A separate PR implements an asynchronous version of this API request, which will return a fail response if the download does not succeed: https://github.com/concord-consortium/codap/pull/1699.


tealefristoe (approved):
Looks good 👍

PR #1694, opened 2024-12-19
Update some TODO_V2_IMPORT comments

bfinzer:
* In this branch we're marking instances of TODO_V2_IMPORT with indications of what should be done to address the issue. All relate to properties that are present in V2 documents that have not yet been brought into the V3 document model. About 7 stories have been created based on these markers. But the remainder are all in codap-v2-types.ts and can be dealt with in situ. Here is a rough classification.
  * Properties that were present briefly in V2 documents but are no longer used. These can be ignored in the V3 import.
  * Properties that are also present in V3 and which refer to implemented features, but the carrying them over during the input process has not yet been brought about.
  * Properties present in the V3 model which do not yet have an implementation.
  * Properties not present in V3 but where the feature is present but defined differently. An example of this is `transparency` referring to the opacity of a graph point where this is stored as a separate property in V2 but is an intrinsic part of the `fill` of a point in V3.
  * Properties referring to a feature present in V2 but which is not yet implemented in V3.

### DIFF
--- a/v3/src/v2/codap-v2-types.ts
+++ b/v3/src/v2/codap-v2-types.ts
@@ -25,7 +25,7 @@ export interface ICodapV2Attribute {
   deletedFormula?: string
   precision?: number | string | null
   unit?: string | null
-  // TODO_V2_IMPORT decimals is not imported
+  // TODO_V2_IMPORT [Story:#188701222] decimals is not imported
   // it occurs 604 times in 36 files in cfm-shared
   // in 33 cases the value is "2"
   // the rest of the cases the value is "0"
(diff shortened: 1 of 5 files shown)


v3/src/v2/codap-v2-types.ts
```diff
@@ -115,33 +117,37 @@ export interface ICodapV2ExternalContext {
 }
 export interface ICodapV2DataContextMetadata {
   description?: string
-  // TODO_V2_IMPORT data context metadata source is not imported
+  // TODO_V2_IMPORT_CARRY_OVER data context metadata source is not imported
   // unknown how many instances in this location
+  // In V3 this is stored as sourceName in dataset
   source?: string
-  // TODO_V2_IMPORT data context metadata importDate is not imported
+  // TODO_V2_IMPORT_CARRY_OVER G data context metadata importDate is not imported
```
kswenson, line 124:
```suggestion
  // TODO_V2_IMPORT_CARRY_OVER data context metadata importDate is not imported
```

kswenson (approved):
👍 LGTM -- just one potential typo

PR #1695, opened 2024-12-19
feat: v2 slider export

kswenson:
[[PT-188616852]](https://www.pivotaltracker.com/story/show/188616852)

### DIFF
--- a/v3/src/components/slider/slider-model.ts
+++ b/v3/src/components/slider/slider-model.ts
@@ -8,9 +8,11 @@ import { getGlobalValueManager, getSharedModelManager } from "../../models/tiles
 import { ITileContentModel, TileContentModel } from "../../models/tiles/tile-content"
 import { DateUnit, dateUnits, unitsStringToMilliseconds } from "../../utilities/date-utils"
 import { kSliderTileType } from "./slider-defs"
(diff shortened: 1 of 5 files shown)


scytacki (approved):
Looks OK to me.

I'd assume that the round tripping of the `v3` property only works if the document is **not** opened in CODAPv2 and saved again?

kswenson:
> Looks OK to me.
> 
> I'd assume that the round tripping of the `v3` property only works if the document is **not** opened in CODAPv2 and saved again?

Correct -- if we're considering having v3 save to v2 format by default, then it should be full round-trip for documents only saved/opened in v3, but fidelity can/will be lost if the document is opened/saved in v2 and then re-opened in v3.

kswenson:
> > Looks OK to me.
> > I'd assume that the round tripping of the `v3` property only works if the document is **not** opened in CODAPv2 and saved again?
> 
> Correct -- if we're considering having v3 save to v2 format by default, then it should be full round-trip for documents only saved/opened in v3, but fidelity can/will be lost if the document is opened/saved in v2 and then re-opened in v3.

If we really wanted to support round-trip fidelity through v2 save/restore we'd have to update v2 to preserve these `v3` properties.

PR #1696, opened 2024-12-19
188651857 Fix update graph displayOnlySelectedCases.

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/188651857

This PR fixes `update graph displayOnlySelectedCases` API calls. Previously, these requests just toggled `graphContentModel.displayOnlySelectedCases`, but did not actually hide or show cases. It now does the same actions that the "Display Only Selected Cases" and "Show All Cases" eye menu options perform.

It's worth noting that "Show All Cases" isn't exactly the same as just revealing unselected cases (this is what occurs in response to `update graph displayOnlySelectedCases: false` requests), but it's the only option users have to revert "Display Only Selected Cases".

I also renamed the jest test for the graph component handler, which I've been meaning to do for a while now. There are no actual changes to the test.


kswenson (approved):
👍 LGTM

PR #1697, opened 2024-12-20
188651494 pixi js integration

nstclair-cc:
[PT-#188651494](https://www.pivotaltracker.com/story/show/188651494)

### Summary

This PR introduces improvements and proof-of-concept tests using PixiJS for graph and map interactions in CODAP v3. We’ve added multiple checks, tools, and tests to enhance our ability to debug and validate graphical behaviors.

---

### New Features

**1. Point Selection and Coloring in Graphs and Maps**
	•	Added functionality to validate coloring of points (e.g., selection, legend plots).
	•	Example proof-of-concept tests now check:
	  •	Colors of points are as expected.
	  •	Colors are valid hex codes (e.g., #4682B4, #E6805B).

---
**2. Counting Points in Graphs and Maps**
	•	Implemented logic to count the number of points in a graph dynamically.
	•	Validates expected point counts in various scenarios, such as parent-child grouping and attribute selections.

---
**3. Checking Position of Points in Graphs and Maps**
	•	Implemented logic to check the (x, y) position of points in a graph or map.
	•	Validates expected position in various scenarios, such as undo redo.

---
**4. Debugger Statement**
	•	Introduced a new debugger tool to make it easier to inspect PixiJS point metadata.
	•	In the browser Console, you can call `pixiPointsMap` to access the array of Pixi points.

---
### Tests Added

**1. Proof-of-Concept Tests**
Included tests for new functionality, such as:
	   •	Point selection validation (with colors).
	   •	Attribute grouping with dynamic point counting.
	   •	Graph updates after selection, hierarchy, and legend interactions.
	   •	Point (x, y) position.

**2. PixiJS Checks**
Multiple checks added that check:
  	  •	Colors of points are accurate and match expectations.
	  •	Graphs update correctly after selections and attribute changes.
	  •	Graph legends and plots dynamically adapt.
          •	Added logs for pixiPoints, textures, and fill values for better visibility (this is because we don't have visibility with the canvas element; logs help the developer get feedback on the issue when tests fail)

---	   
**How to Use the Debugger**
	1.	Open the graph or map component.
	2.	In the Console, call `pixiPointsMap` after enabling the debugger with Key: `debug` and Value: `pixiPoints`.
	3.	Inspect `pixiPointsMap` to view point metadata (e.g., texture, position).

**Next Steps**
	•	These changes lay the groundwork for:
	•	Extending point-coloring logic for additional use cases.
	•	Enhancing legend interactions for categorical attributes.

---
### Proof of Concept in Action

Here’s an example of what’s now possible:
	•	Validate graph points are colored correctly.
	•	Validate graph points are in correct position.
	•	Verify dynamic updates when selecting attributes or grouping data.
	•	Debug PixiJS behavior directly in the Console with a single statement.


.github/workflows/graph-pixi-tests.yml
```diff
@@ -0,0 +1,55 @@
...
+      actions: write
+    steps:
+      - name: Run Cypress tests for graph-pixi
+        run: npm run cypress:run -- --spec "cypress/e2e/pixi-interaction/*"
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          # set repository so we don't have to check out all of the code
+          GH_REPO: ${{github.repository}}
+
+cypress:
```
scytacki, line 19: This cypress section, a GitHub Actions "job", needs to be indented to the same spot as `re_run:`.
I found that by looking for the log of running this workflow: https://github.com/concord-consortium/codap/actions/runs/12474798288

Since the workflow isn't running yet it is hard to know if there are other things broken.

.github/workflows/graph-pixi-tests.yml
```diff
@@ -0,0 +1,55 @@
+name: Regression Label
```
scytacki, line 1: This should be renamed to something like "Graph Pixi Tests". 
That will make it easier to tell if it is working or not. It currently has the same name as another workflow.

Below are the results right now. This workflow is the one with the `on: push` below it. It doesn't have any running jobs because of the syntax error (see comment below).
image

scytacki (changes requested):
Getting closer. I left some comments of things to change so we can see if the cypress tests will run in GitHub Actions.

kswenson (approved):
👍 LGTM -- just a couple of suggestions. Beyond that, @scytacki should have the final word.

.github/workflows/graph-pixi-tests.yml
```diff
@@ -0,0 +1,127 @@
...
+        working-directory: v3
+        run: npm run lint:build
+      - name: Save ESLint Cache
+        uses: actions/cache/save@v4
+        if: github.ref_name == 'main' || steps.restore-eslint-cache.outputs.cache-hit != 'true'
+        with:
+          path: v3/.cache/eslint/
+          key: ${{ steps.restore-eslint-cache.outputs.cache-primary-key }}
+
+  jest:
```
scytacki, line 41: I don't think it is necessary to run the jest tests here. They should already be run by the main workflow.  But perhaps you added them just make sure they were working here since they are failing in the main workflow.

.github/workflows/graph-pixi-tests.yml
```diff
@@ -0,0 +1,127 @@
...
+    name: Run Cypress Pixi Tests
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+    steps:
+      - uses: snow-actions/sparse-checkout@v1.2.0
+        with:
+          patterns: |
+            v3
+            cypress/e2e/pixi-interaction
```
scytacki, line 83: I'd assume the cypress tests are inside of the v3 folder so it shouldn't be necessary to add that here. Just listing v3 should be enough.

scytacki (changes requested):
I think this is mostly done. 

I left a couple of inline comments of things to change.

The reason the Jest tests look like they are failing is because the changed files are triggering the CODAP v2 workflow. So it is trying to run the v2 jest tests.

That happens because of the filters setup for the various workflows. So to fix it you could update this section of v3.yml: https://github.com/concord-consortium/codap/blob/9b994cc5f7f1835cf7fae2ef5f0381c6ec008e5e/.github/workflows/v3.yml#L5-L7
To include `- '.github/workflows/graph-pixi-tests.yml'

And update this section:
https://github.com/concord-consortium/codap/blob/9b994cc5f7f1835cf7fae2ef5f0381c6ec008e5e/.github/workflows/ci.yml#L5-L12
to exclude the workflow.

At one point I thought Teale was just going to remove the v2 workflows from the v3 branch but it looks like that didn't happen.

.github/workflows/graph-pixi-tests.yml
```diff
@@ -0,0 +1,143 @@
+name: Graph Pixi Tests
+
+on:
+  push:
+    paths:
+      - 'cypress/e2e/pixi-interaction/**'
+      - '.github/workflows/graph-pixi-tests.yml'
+
+jobs:
+  lint:
```
scytacki, line 10: I'm not sure if I missed this before or you added it recently, but I don't think `lint` should be run in this worlkflow. It is redundant with the lint that is run during main `v3.yml` workflow.

.github/workflows/graph-pixi-tests.yml
```diff
@@ -0,0 +1,143 @@
...
+        run: npm run lint:build
+      - name: Save ESLint Cache
+        uses: actions/cache/save@v4
+        # Always save the cache on the main branch, it won't be built often and
+        # this way features branches will have an up-to-date main cache
+        if: github.ref_name == 'main' || steps.restore-eslint-cache.outputs.cache-hit != 'true'
+        with:
+          path: v3/.cache/eslint/
+          key: ${{steps.restore-eslint-cache.outputs.cache-primary-key}}  
+  build:
```
scytacki, line 42: I'm also not sure if I missed this before, but this whole `build` job shouldn't be needed. The cypress job below runs the webpack dev server which will build the app.

scytacki (changes requested):
Closer. :) I'm not sure if I missed them before, or you just added them but the `lint` and `build` jobs should not be needed in the `graph-pixi-tests.yml` workflow.

scytacki (approved):
Looks good. 👍

nstclair-cc:
> Closer. :) I'm not sure if I missed them before, or you just added them but the `lint` and `build` jobs should not be needed in the `graph-pixi-tests.yml` workflow.

Oh, phew—you’re right! I left that in because I didn’t want to assume it wasn’t needed. Thanks for catching that! Let’s try again with the review, @scytacki.

PR #1698, opened 2024-12-20
async api handlers

scytacki:
Refactor the handling of api requests from interactives.
The handlers can now be asynchronous. If an array of api requests is sent by the interactive they will be run one after the other with each waiting for the previous one to finish.
Individual api requests will be run when they come in even if there is another api request or array of request waiting to finish.

The api request loop is changed from an autorun based observer to a reaction observer. This way properties accessed by each handler are not observed. We only want to run the loop when a new request comes in, not when a property changes that a handler accessed.

The cell editing state handling will be broken when an async handler is used. While the async handler is still running `incrementInterruptionCount` will called. This is because `processItems` doesn't wait. Simply making `processItems` wait would not fix the problem because additional requests might come and finish in while one is waiting. 

It seems OK to punt this problem since it should only be a problem when the shared data table is mixed with a plugin which uses async apis, and that plugin calls an async api in the middle of someone editing. When @kswenson comes back, if he thinks it is isn't OK to punt this problem, I can put a little more time into it. I think I could add some better waiting that would prevent this problem. The additional waiting would mean that a slow async handler could hold up a whole chain of handlers, but given our limited number of async handlers this might be a safer approach.

### DIFF
--- a/v3/src/components/web-view/use-data-interactive-controller.ts
+++ b/v3/src/components/web-view/use-data-interactive-controller.ts
@@ -13,6 +13,7 @@ import { uiState } from "../../models/ui-state"
 import { t } from "../../utilities/translation/translate"
 import { RequestQueue } from "./request-queue"
 import { isWebViewModel } from "./web-view-model"
+import { errorResult } from "../../data-interactive/handlers/di-results"
 
 function extractOrigin(url?: string) {
   if (!url) return
@@ -71,7 +72,6 @@ export function useDataInteractiveController(iframeRef: React.RefObject<HTMLIFra
           debugLog(DEBUG_PLUGINS, `Processing data-interactive: ${JSON.stringify(request)}`)
           let result: DIRequestResponse = { success: false }
 
-          const errorResult = (error: string) => ({ success: false, values: { error }} as const)
           const processAction = async (action: DIAction) => {
             if (!action) return errorResult(t("V3.DI.Error.noAction"))
             if (!tile) return errorResult(t("V3.DI.Error.noTile"))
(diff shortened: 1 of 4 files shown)


tealefristoe:
@scytacki I tested the shared table plugin on this branch and things look like they're working correctly 👍  This is the plugin that required us to add the API request queue, which is blocked when a table is being edited. Here's how to test it if you want to see for yourself:
1. Visit https://codap3.concord.org/branch/async-api-handlers/?di=https://codap-shared-table-plugin.concord.org/branch/master/ on two different computers.
2. Create a new shared table on one computer, then join the shared table on the other.
3. On one computer, start editing the table, but don't finish--leave the cursor blinking in the table cell.
4. On the other computer, make several changes to the table.
5. On the first computer, finish editing the table.

On the first computer, you should not see changes made on the second computer while you are actively editing the cell. When you finish editing the cell, all of the changes made on the second computer should appear. This is how the plugin is designed to work.

v3/src/components/web-view/use-data-interactive-controller.ts
```diff
@@ -52,53 +52,65 @@ export function useDataInteractiveController(iframeRef: React.RefObject<HTMLIFra
...
+
+        requestQueue.processItems(async ({ request, callback }) => {
+          debugLog(DEBUG_PLUGINS, `Processing data-interactive: ${JSON.stringify(request)}`)
+          let result: DIRequestResponse = { success: false }
 
-            const errorResult = (error: string) => ({ success: false, values: { error }} as const)
-            const processAction = (action: DIAction) => {
-              if (!action) return errorResult(t("V3.DI.Error.noAction"))
-              if (!tile) return errorResult(t("V3.DI.Error.noTile"))
+          const errorResult = (error: string) => ({ success: false, values: { error }} as const)
```
tealefristoe, line 74: I know this was already in the code, but this function is already defined in `di-results.ts`.

tealefristoe (approved):
Looks good 👍

PR #1699, opened 2024-12-20
switch dataContextFromURL to be async

scytacki:
This is an example of an async handler.

The code in the async `create` is a more complicated than a simple async handler. Because `downloadCsvFile` uses callbacks instead of promises, so we can't use the async/await construct.

This can be tested with the API tester. If you send send it an invalid URL you will now get a error response instead of success like what happens in the [non-async PR](https://github.com/concord-consortium/codap/pull/1693)


tealefristoe (approved):
Looks great 👍

PR #1700, opened 2024-12-20
181846570 adjustable row height

eireland:
Adds a row divider that acts as a handle for adjusting row heights.
Rows snaps to heights that are multiples of 18 (the default min row height)
Text wraps when row height is taller than minimum row height.
Row height is saved and restored.


kswenson (approved):
👍 Looks good -- I pushed a commit which tweaked a few things. Run it through its paces to make sure everything works before merging.

PR #1701, opened 2024-12-21
fix: v2 export of hierarchical Dataset

kswenson:
[[PT-188713991]](https://www.pivotaltracker.com/story/show/188713991)

v2 relies on `parent` links for collections and cases, even though they can be derived at runtime, so v3 must export them. Separately, fixes a bug in which parent attribute values would be redundantly exported in child cases. Also adds a jest test for a hierarchical export.


tealefristoe (approved):
Looks great 👍

PR #1702, opened 2024-12-23
fix: lookup functions match user-set title

kswenson:
[[PT-188719111]](https://www.pivotaltracker.com/story/show/188719111)


tealefristoe (approved):
Looks good 👍 

I'm beginning to think it might be better to scrap having both name and title. But that's probably not a discussion for right before Christmas 😆

PR #1707, opened 2025-01-02
Categorical axis limits number of categories to space available

bfinzer:
[#181912260] Feature: A categorical axis fits as many category labels as it can without overlap and then clumps the remaining categories into *OTHER*.

* We compute that categories that can be displayed in use-sub-axis.ts and pass them to `CategoricalAxisHelper`
* A categorical axis responds dynamically to changes in the amount of space available.
* Added volatile `numberofCategoriesLimitByRole` to DataConfigurationModel to allow useSubAxis to communicate with other parts of a graph
* DataConfigurationModel:mapOfIndicesByCase and GraphDataConfigurationModel:cellMap modified to take into account `numberofCategoriesLimitByRole`
* Took the opportunity to replace all '__main__' with a constant kMain = '__main__'
* Modify graph-data-configuration-model.ts `subPlotCases` to take into account that the given subplot is one for the OTHER category
* Commented out test in single-codap-smoke.ts that counts the number of categories displayed on a categorical axis because that number now depends how many fit on the axis
* Cleanup of movable-value-adornment-component.tsx to prevent errors when height of movable value computes as negative


kswenson (approved):
👍 Looks good -- just a smattering of minor suggestions.

PR #1709, opened 2025-01-03
fix: don't write out empty case table name when exporting to v2 document

kswenson:

Addresses the issue identified in the PT story that a new case table loses its title when exported to v2.


tealefristoe (approved):
Looks good 👍

PR #1710, opened 2025-01-03
fix: default numeric precision is two decimal places

kswenson:
[[PT-188590488]](https://www.pivotaltracker.com/story/show/188590488)


tealefristoe (approved):
Short and sweet 👍

PR #1713, opened 2025-01-07
188565497 map filter formula

eireland:
Adds a DataDisplayFilterFormulaAdapter as a base class for the new MapFilterFormulaAdapter and the refactored GraphFilterFormulaAdapter.
Adds a MapFilterFormulaAdapter to make the filter formula editor in the map functional. 
Note that the current implementation of the map filter formula editor only works for the first layer of the first dataset attached to the map.

Demo at https://codap3.concord.org/branch/map-filter-formula/#file=examples:Four%20Seals


kswenson (approved):
👍 Looks good -- nice work with the refactoring. I added a commit which took it a little further, eliminated some circular imports, and cleaned up some types a bit. Take a look and make sure I didn't break anything before you merge.

PR #1715, opened 2025-01-07
Adds ability to rename attribute in formula editor modal.

eireland:
If formula editor modal is a filter formula, then attribute name input is disabled because we don't have an attribute.

Attribute name input was changed from ChakraUI Input to an html input tag because the ChakraUI Input kept stealing the focus from the formula editor.


kswenson (approved):
👍 Looks good -- I'm not sure I _like_ the new UI, but it is what was requested. Just a couple of suggestions, the most important of which is to use the `DataSet` API for changing the attribute name.

PR #1716, opened 2025-01-07
fix: wordListMatches() updates on dependent changes

kswenson:
[[PT-188749138]](https://www.pivotaltracker.com/story/show/188749138)

The `wordListMatches()` function is a lookup function in all but name. As such, it needs to follow the patterns laid out by the other lookup functions like `lookupByIndex()` and `lookupByKey()`. In particular, it needs to implement the `getDependency()` and `canonicalize()` functions for registering the `lookup` dependency appropriately. Once that's done, the existing lookup dependency tracking kicks in and the formula updates as expected.


tealefristoe (approved):
Looks good, hope this wasn't too difficult to figure out.

PR #1717, opened 2025-01-07
188276475 case table filter formula UI

eireland:
Adds a filter formula bar under the case table/card title bar if a filter formula is present.

When user clicks on the filter formula, the edit formula modal opens to allow user to edit the filter formula.

Demo here: https://codap3.concord.org/branch/case-table-filter-formula-ui/

### DIFF
--- a/v3/src/components/case-tile-common/filter-formula-bar.tsx
+++ b/v3/src/components/case-tile-common/filter-formula-bar.tsx
@@ -35,7 +35,7 @@ export const FilterFormulaBar = () => {
 
   return (
     <>
-      <div className="filter-formula-container">
+      <div className="filter-formula-container" data-testid="filter-formula-bar">
         <span className="filter-formula-label">{t("DG.CaseTable.formulaFilterBar.label")}:</span>
         <span className="filter-formula-value" onClick={handleOpenEditFormulaModal}>
           {filterFormula}
--- a/v3/src/components/case-tile-common/filter-formula-bar.tsx
+++ b/v3/src/components/case-tile-common/filter-formula-bar.tsx
@@ -1,8 +1,8 @@
 import React, { useState } from "react"
-import { useDataSetContext } from "../../hooks/use-data-set-context"
-import { EditFormulaModal } from "../common/edit-formula-modal"
 import { useDisclosure } from "@chakra-ui/react"
 import { t } from "../../utilities/translation/translate"
+import { useDataSetContext } from "../../hooks/use-data-set-context"
+import { EditFormulaModal } from "../common/edit-formula-modal"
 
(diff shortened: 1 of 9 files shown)


v3/src/components/case-tile-common/filter-formula-bar.tsx
```diff
@@ -0,0 +1,55 @@
+import React, { useState } from "react"
+import { useDataSetContext } from "../../hooks/use-data-set-context"
+import { EditFormulaModal } from "../common/edit-formula-modal"
+import { useDisclosure } from "@chakra-ui/react"
```
kswenson, line 4: Move ⬆️ to top.

v3/src/components/case-tile-common/filter-formula-bar.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { t } from "../../utilities/translation/translate"
```
kswenson, line 5: Move ⬆️ to after `node_modules` imports.

v3/src/utilities/translation/lang/en-US.json5
```diff
@@ -823,6 +823,7 @@
     "DG.CaseTable.dividerView.collapseAllTooltip": "collapse all groups",
     "DG.CaseTable.dividerView.collapseGroupTooltip": "collapse group",
     "DG.CaseTable.indexProto.tooltip": "Drag to place data entry row within this table",
+    "DG.CaseTable.formulaFilterBar.label": "Show cases only if",
```
kswenson, line 826: Move this to the "V3." section for now so we can keep it separate.

kswenson (changes requested):
Good start, but several issues are visible with the Roller Coasters example document on the deployed branch (https://codap3.concord.org/branch/case-table-filter-formula-ui/#file=examples:Roller%20Coasters):
1. Black on gray label is nearly illegible. This appears to be the v2 and v3 behavior for plotted functions, but surely we can do better. Needs some design, though, so could be deferred to a separate PR.
2. Adding filter function causes case table to grow (or draw outside its bounds) vertically (see screen shot). Note that horizontal scroll bar overlaps the graph title bars and the position of the corner resize icon.
3. Clicking on filter function brings up modal dialog with label "Plotted Function" and buttons clipped (see screen shot).
4. On the Case Card, the filter formula should not be scrollable, i.e. it should be fixed under the title bar and the content should scroll underneath it.

Original appearance before adding filter:
CoastersInitial

Appearance after adding filter:
CoastersFilter

Formula editor after clicking filter:
CoastersFilterDialog

kswenson (approved):
👍 Changes look good

PR #1719, opened 2025-01-08
breakup some dependency trees

scytacki:
Add a IBaseDataDisplayModel for use by the legend code so it doesn't depend on the full DataDisplayModel which brings in a dependency on the Tile System.

Update the types and type checking of the GraphContentModelContext and MapContentModelContext to be safer. They will now throw a runtime error if the wrong context object is used.

Remove the default data display model provided by the context. It was not fully functional so could cause weird issues if used in ways Typescript allowed. The default model was used by tests and the Axis code when the Slider rendered the Axis. To handle the Slider usage of the Axis, a new useDataDisplayModelContextMaybe was added.  

The Axis system did rely on two of the methods provided by either the default or real DataDisplayModel. Those methods were added to the IAxisProvider interface. And the implementation from the default DataDisplayModel was added to the Slider content model. Now the SubAxis just uses the AxisProvider. The Axis still tries to use the DataDisplayModel if it is exists to get the dataConfiguration.

The AxisProviderContext was also providing a default axis provider which was not fully functional. From what I can tell this is default axis provider was never actually used. So now the useAxisProviderContext throws a runtime error if it is called without an axis provider. 

Split out axis-constants which used to be in graphing-types.

In axis-helper there was an assertion that the axisBounds would not be undefined. Since it technically be it now throws an explict runtime error if it is.

You can see how the the WIP PR of CODAP files are currently being brought into the dst-plugin here:
https://github.com/concord-consortium/dst-codap-plugin/pull/9
With a status of what is left here:
https://github.com/concord-consortium/dst-codap-plugin/blob/try-bringing-in-legend/src/codap/seperation-todo.md

My plan is:
1. make PRs like this within the CODAP repository which break up dependencies and copy these files over without changes into the dst-plugin until the PR compiles and lint without errors. 
2. Try to actually use the legend within the dst-plugin, which will probably require more fixes.
3. Try to take all of the files in the `src/codap` folder of the PR and move then into a single folder in CODAP v3 repo. Perhaps something like `src/data-core`

Doing this while the CODAPv3 code is changing will probably be challenging, so that is why I'm hoping to be able to merge in incremental PRs like this one. 

And step 3 will probably in several merge conflicts. It be a bunch of renames and lots of updates to import statements. If I can, I'll try to make a script to do this refactoring. This way the actual PR can be created from main just before it is merged. This will likely still cause merge conflicts for any in progress PRs, so if this looks like it will be a problem, we could try to find a time in the future when there aren't any in progress PRs.


kswenson:
> My plan is:
> 
> 1. make PRs like this within the CODAP repository which break up dependencies and copy these files over without changes into the dst-plugin until the PR compiles and lint without errors.

FWIW, I think the plan of creating more incremental PRs with smaller changes (like this one) will make any conflicts easier to deal with, so I'd recommend submitting this PR in something close to its current form -- on a quick glance it looks pretty good already -- and then keep making incremental improvements as you encounter them.

kswenson (approved):
👍 LGTM -- just a handful of stylistic suggestions.

PR #1720, opened 2025-01-08
Focuses cursor in new table title bar on create.

eireland:


kswenson (approved):
👍 Looks good -- I added a commit which changed the way the initial edit is triggered so that it doesn't do so on saved documents, for instance.

PR #1721, opened 2025-01-09
feat: export text tile in v2 format

kswenson:
[[PT-188616819]](https://www.pivotaltracker.com/story/show/188616819)

We also take this opportunity to change the way the slate value is stored in the model. Previously, we were storing the `JSON.stringify`ed value directly in the model. Now, we store it in the model as a `types.frozen<SlateExchangeValue>()` and just `JSON.stringify()`/`JSON.parse()` it on export/import. This should provide performance benefits (less `JSON.stringify`ing) as well as additional type-safety.


scytacki (approved):
Looks good to me. 👍

PR #1722, opened 2025-01-09
extract kIndexColumnKey

scytacki:
This constant is used by data-set-utils so shouldn't be in src/components/case-tile-common


kswenson (changes requested):
>This constant is used by data-set-utils so shouldn't be in src/components/case-tile-common

I agree that there's a mismatch here, but `kIndexColumnKey` is really a case table-specific notion, not a data set notion. I think the better fix would be to change the signature of the `moveAttribute()` function so that there's another way to specify that the attribute should be inserted at the beginning, which I believe is the only reason for using the `kIndexColumnKey`.

scytacki:
@kswenson how's this look now?

PR #1723, opened 2025-01-10
simplify plugin drop support

scytacki:
This replaces the use of applyModelChange and just calls broadcastMessage directly on the tile.

The reason for tackling this is that I'm trying to type the `INotification`s better. Without the drag notification this type is pretty simple and consistent. So by having the drag notifications not use applyModelChange, it is now possible for applyModelChange to take a more restrictive and simple set of arguments.

With that change it is easier for a simple version of applyModelChange to be included in the CODAP "library" I'm trying to create. The basic idea is that this library would support a notification system which it uses internally. The system isn't specific to sending messages to data interactives, but that is the way it will be used when the components of the library are used within the main CODAP app.  

A further step could be removing the `notifyTileId` from the applyModelChange arguments. A tileId will not really have a meaning when the CODAP library is used outside of CODAP, but it is just a string so won't cause problems to leave it in.

This is the first in a series of PRs:
- https://github.com/concord-consortium/codap/pull/1724
- https://github.com/concord-consortium/codap/pull/1728

### DIFF
--- a/v3/src/components/web-view/web-view-drop-overlay.tsx
+++ b/v3/src/components/web-view/web-view-drop-overlay.tsx
@@ -23,16 +23,20 @@ export function WebViewDropOverlay() {
   const mouseX = useRef<number|undefined>()
   const mouseY = useRef<number|undefined>()
 
+  const sendNotification = (notification: INotification) => {
+    const { message, callback } = notification
+    tile?.content.broadcastMessage(message, callback ?? (() => null))
+  }
+
   // Broadcast dragstart and dragend notifications
   const handleDragStartEnd = (
     notification: (dataSet: IDataSet, attributeId: string) => INotification, _active: Active
   ) => {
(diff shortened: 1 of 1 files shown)


tealefristoe (approved):
This looks good to me. I don't remember exactly why we used `applyModelChange` for this originally--it might have been to make all notifications consistently go through `applyModelChange`, or we may have just not given it much thought and went with what we'd already been doing. Either way, it's probably best to run this by Kirk before merging.

I'd also suggest removing `notifyTileId` from `applyModelChange` as a part of this work, assuming it's no longer used for anything. Otherwise, it will likely linger in the codebase and become mysterious and confusing sometime in the future.

kswenson (approved):
👍 LGTM

scytacki:
I've merged this as is for now. I'll try to remove the notifyTileId infrastructure in a follow on PR.

PR #1724, opened 2025-01-10
beginning of history service

scytacki:
The idea is that within CODAP the "app" history service will handle the applyModelChange. When the code is used outside of CODAP a no-op history service can be used.

Part of this work is also to type the INotification better. Instead of just using the iframePhone generic types, now the notifications are type specific to how CODAP is using them. The core of CODAP will work with ICoreNotifications. And then things like drag and drop support for iframes extend ICoreNotifications.  The rational is that the notification system isn't necessary specific to iframePhone or iframes at all. It is just a generic notification system which is then exposed to plugins via the iframe API.

When the tests are made part of the CODAP "library", instead of importing `models/history/register-app-history-service` they'd import a no-op history service. Or if they are actually checking the history then a simplified history service could be used.

A variation of this PR could be to register a default no-op history service. This would remove the need for most of the tests to import `models/history/register-app-history-service`.  However this will make it less obvious how to fix a new test that is expecting the history to work. So my preference is to fail fast and require the more explicit import.

### DIFF
--- a/v3/src/models/history/apply-model-change.ts
+++ b/v3/src/models/history/apply-model-change.ts
@@ -1,17 +1,12 @@
-import iframePhone from "iframe-phone"
 import { IAnyStateTreeNode } from "mobx-state-tree"
-import { DIMessage } from "../../data-interactive/iframe-phone-types"
+import { ICoreNotification } from "../../data-interactive/notification-core-types"
 import { ILogMessage } from "../../lib/log-message"
-import { getTileEnvironment } from "../tiles/tile-environment"
-import { withUndoRedoStrings } from "./codap-undo-types"
-import { withoutUndo } from "./without-undo"
+import { getHistoryService } from "./history-service"
 
-export interface INotification {
-  message: DIMessage
-  callback?: iframePhone.ListenerCallback
-}
 export type INotify =
-  INotification | (INotification | (() => Maybe<INotification>))[] | (() => Maybe<INotification | INotification[]>)
+  ICoreNotification |
+  (ICoreNotification | (() => Maybe<ICoreNotification>))[] |
+  (() => Maybe<ICoreNotification | ICoreNotification[]>)
 export interface IApplyModelChangeOptions {
   log?: string | ILogMessage | (() => Maybe<string | ILogMessage>)
   notify?: INotify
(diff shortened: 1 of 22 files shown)


v3/src/models/history/apply-model-change.ts
```diff
@@ -1,69 +1,29 @@
...
-              const __notification = typeof _notification === "function" ? _notification() : _notification
-              if (__notification) {
-                const { message, callback } = __notification
-                tileEnv.notify?.(message, callback ?? (() => null), notifyTileId)
-              }
-            })
-          }
-        }
-      }
+      getHistoryService().handleApplyModelChange(self, options)
```
kswenson, line 26: Couldn't you also attach the history service to the MST environment like we do with logging, which would avoid the need to register the history service explicitly for tests that only require it for `applyModelChange`? You probably considered that and rejected it for some reason, but it seems like it would have some advantages.

v3/src/models/history/apply-model-change.ts
```diff
@@ -1,69 +1,29 @@
...
 
-export interface INotification {
-  message: DIMessage
-  callback?: iframePhone.ListenerCallback
-}
 export type INotify =
-  INotification | (INotification | (() => Maybe<INotification>))[] | (() => Maybe<INotification | INotification[]>)
+  ICoreNotification |
+  (ICoreNotification | (() => Maybe<ICoreNotification>))[] |
+  (() => Maybe<ICoreNotification | ICoreNotification[]>)
```
kswenson, line 9: This type is redundantly defined in `history-service.ts`, so could be imported instead.

v3/src/models/history/history-service.ts
```diff
@@ -0,0 +1,33 @@
...
+  notifyTileId?: string
+  undoStringKey?: string
+  redoStringKey?: string
+}
+
+export interface IHistoryService {
+  handleApplyModelChange: (mstNode: IAnyStateTreeNode, options?: IApplyModelChangeOptions) => void
+}
+
+const gHistoryServiceRef: {ref?: IHistoryService} = {}
```
kswenson, line 22: What's the advantage to this over the simpler:
```suggestion
const gHistoryService: Maybe<IHistoryService>
```
Are you expecting to extend the object to more than just the `ref` property in the future?

kswenson (changes requested):
👍 Generally looks good -- just a few suggestions to consider.

v3/src/models/history/apply-model-change.ts
scytacki: I think I ruled it out because I was trying to avoid depending on the `ITileEnvironment`. However as I've continued on this work I now think using ITileEnvironment makes sense. 

I also was thinking about not changing all of the code that calls `withoutUndo()`. But I've figure out a way to look up the MST environment from a simple `withoutUndo()` so I think its do-able. I'm running out of time on this, but hopefully I can get it wrapped up.

v3/src/models/history/history-service.ts
scytacki: No you're right it is unnecessary. I was thinking some modules might import gHistoryService directly. So then it'd have to have a ref property so the value could be changed after import. But then I switched to using a `getHistoryService()` method which seems cleaner than `gHistoryService.ref`.  Anyhow this will be somewhat mute when this is moved into the ITileEnvironment.

scytacki: With the change to putting the history service in the tile environment, yes I could keep the global access available. However there aren't any clients that need that as far as I know. So it seems better to take that away, and then interface of the history service can be simplified.  In particular it shouldn't be necessary to pass in MST nodes to the methods just so they can find the right tree to work with. Additionally the history service implementation can assume the tile environment is always available.

scytacki:
This PR has been replaced by: #1735

PR #1725, opened 2025-01-10
fix: v2 text export

kswenson:
[[PT-188616819]](https://www.pivotaltracker.com/story/show/188616819)

CODAP v2 is using an older version of the slate editor library that requires additional metadata to be exported.


tealefristoe (approved):
Looks good 👍

PR #1726, opened 2025-01-11
181904766 graph title attr collection

eireland:
Changes the graph title shown when there is a data set present. 
The graph title now finds the child-most collection of any attribute associated with the graph and uses that collection title. If there are no attributes associated with the graph, then use the title of the child-most collection.
If user changes the title, then graph will always use the user defined title.

### DIFF
--- a/v3/cypress/e2e/cfm.spec.ts
+++ b/v3/cypress/e2e/cfm.spec.ts
@@ -86,7 +86,7 @@ context("CloudFileManager", () => {
 
     // Close the document
     // Note: because we aren't changing the document we can close it without the dialog
-    cfm.closeDocument({discardChanges: true})
+    cfm.closeDocument()
 
     // Verify document was closed (table doesn't exist)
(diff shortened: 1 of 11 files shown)


bfinzer (changes requested):
I've encountered a bug. In http://localhost:8080/?sample=coasters I drag `Type` to the left, make a graph, and drag `Type` to its x-axis. The title of the graph is still **Cases** whereas it should be **Types**. If I then drag `Top_Speed` to the x-axis, the title of the graph becomes **Types** whereas it should now be **Cases**.

kswenson (approved):
👍 Approved after adding a commit which consolidated the implementation in the graph's `getTitle()` function and eliminated the changes to the `Graph` and `GraphComponent`. Double-check that everything still works as expected, but I believe this provides the right behavior with fewer moving parts.

bfinzer (approved):
👍🏻LGTM

PR #1729, opened 2025-01-11
Fixes broken case table filter cypress test

eireland:


nstclair-cc (approved):
lgtm! Thanks for getting that fix in. We might need to unskip the test https://github.com/concord-consortium/codap/blob/main/v3/cypress/e2e/table.spec.ts#L341-L386

PR #1730, opened 2025-01-13
Table filter bar refinements

eireland:
Changes pointer cursor for filter formula label in case table to default
Changes Formula name width to fit "Plotted Function"


kswenson (approved):
👍 LGTM

PR #1732, opened 2025-01-13
188753758 Broadcast notifications when selection is changed via API requests.

tealefristoe:
PT Story: https://www.pivotaltracker.com/story/show/188753758

This PR makes Codap broadcast notifications when a plugin changes which cases are selected.


kswenson (approved):
👍 LGTM

PR #1733, opened 2025-01-14
make TileContentModel more independent

scytacki:
This removes the dependency on iframe-phone and shared-model-manager.

The tile-environment still brings in more dependencies, but that should be taken care of in a future PR.


kswenson (approved):
👍 LGTM

PR #1734, opened 2025-01-15
break dependencies around the formula system

scytacki:
I'm not really satisfied with this one, but I think it will be enough to have a clean split. 

The basic split here is between `IFormulaManager` and `FormulaManager` and `IFormulaManagerAdapater` and `FormulaManagerAdapter`. I couldn't find a good way to split off all of the formula types. Because DataSet and DataConfigurationModel have optional filter formulas. It would be possible to use a MST Union to define types with and without Formulas, but it seemed like that would require lots of changes to the CODAP code. So instead I opt'd to bring in the Formula type but have it be a shell whose implementation is moved into the FormulaManager provided by the tile environment.

Doing this avoids bringing in:
- mathjs
- the Boundary system
- global value system

What I'm hoping is that once this is split is used in an example case (DST) it will become more clear what is needed and what is not. 

- move a couple formula adapter registrations to app.tsx
- add a IFormulaManager interface to be used by the TileEnvironment so the full FormulaManager doesn't need to be exposed to tile content models and their sub objects.
- move abstract base class FormulaManagerAdapter to its own file along with the IFormulaAdapterApi. This breaks the dependency on the BoundaryManager and GlobalValueManager
- createCodapDocument makes the assumption that the IFormulaManager returned by getFormulaManager is really a FormulaManager. Given that it basically configured all this in createDocumentModel it is safe to assume.
- update tests so Formula object an get the formulaManager from its MST environment.
- add 3 new methods to FormulaManager and IFormulaManager so that the Formula model does not directly depend on the formula system. Instead the Formula model uses the formulaManager to do its work. Unfortunately, this re-enforces the existing circular dependency between the Formula and FormulaManager. But at least now the circle is based on interfaces not actual class types.
- move ICase out of formula-types.ts since that class imported mathjs.

Note: this does not fix some existing circular dependencies. For example: Formula -> IFormulaManager -> IFormulaManagerAdapter -> IFormula
I spent some time trying to figure out a nice way to break that cycle and didn't come up with anything simple. The best thing I came up with was to add a separate formulaService to the TileEnvironment which is intended just for Formulas. This way Formulas could use this service instead of the formulaManager. However Formulas couldn't use the TileEnvironment interface to get this service, they'd have to get the MST environment and cast it to an different interface. This approach of having multiple environment interfaces for a single MST environment doesn't seem good, but MST only provides a single environment object. Essentially it means the name of the key in the environment is the service identifier. It is probably worth leaning into this, and break apart ITileEnvironment into separate interfaces with separate methods for getting each typed service from the environment.

Note 2: The formula adapter system could probably be simplified by having the adapters look up their content models via a new method on IFormulaAdapterApi. This Api already knows the document so it could just search the document for the specified tile types and return their contentModels. This way content models don't have to add themselves to the adapters when they are added to the document. As long as the adapters are MobX objects and the contentModels is a view (like it is already) it will be cached so should be just as efficient as the current approach.  I didn't make this change because things were getting too large already.


v3/src/models/codap/create-codap-document.ts
```diff
@@ -43,7 +44,8 @@ export function createCodapDocument(snapshot?: ICodapDocumentModelSnapshot, opti
       document.content?.addSharedModel(globalValueManager)
     }
     // Add the global value manager to the formula manager
-    getFormulaManager(document)?.addGlobalValueManager(globalValueManager)
+    const formulaManager = getFormulaManager(document) as Maybe<FormulaManager>
```
kswenson, line 47: Cast shouldn't be required because it matches return value of `getFormulaManager()`.

kswenson (approved):
👍 Looks good -- just one typing consideration.

v3/src/models/codap/create-codap-document.ts
scytacki: It is required because `getFormulaManager` returns a `IFormulaManager` which doesn't include `addGlobalValueManager`. That is setup that way so the global value manager doesn't have to be included in the CODAP "library.

PR #1735, opened 2025-01-15
HistoryService take 2

scytacki:
The idea is that within CODAP the AppHistoryService will handle the applyModelChange. When the code is used outside of CODAP a no-op history service can be used.

Part of this work is also to type the INotification better. Instead of just using the iframePhone generic types, now the notifications are typed specific to how CODAP is using them. The core of CODAP will work with ICoreNotifications. And then things like drag and drop support for iframes extend ICoreNotifications. The rational is that the notification system isn't necessary specific to iframePhone or iframes at all. It is just a generic notification system which is then exposed to plugins via the iframe API.

Additionally this takes a new approach for the ITileEnvironment. It is now created by intersecting several individual environment interfaces. This makes it possible for services like the history service to depend on them in a way where the "key" of the service in the environment is typed in a single place while not making the history service dependent on the full ITileEnvironment.

When the tests are made part of the CODAP "library", instead of importing using AppHistoryService they'd use a no-op history service. Or if they are actually checking the history then a simplified history service could be used.

A variation of this PR could be have applyModelChange handle the case of not finding the history service. This would remove the need for most of the tests to setup MST environments with a historyService. However this will make it less obvious how to fix a new test that is expecting the history to work. So my preference is to fail fast and require the more explicit configuration.

### DIFF
--- a/v3/src/components/axis/models/axis-model.test.ts
+++ b/v3/src/components/axis/models/axis-model.test.ts
@@ -1,8 +1,9 @@
+import {getSnapshot, types} from "mobx-state-tree"
 import {
   AxisModel, AxisModelUnion, CategoricalAxisModel, EmptyAxisModel, IAxisModelUnion,
   isCategoricalAxisModel, isEmptyAxisModel, isNumericAxisModel, NumericAxisModel
 } from "./axis-model"
-import {getSnapshot, types} from "mobx-state-tree"
+import { AppHistoryService } from "../../../models/history/app-history-service"
 
 describe("AxisModel", () => {
   it("should error if AxisModel is instantiated directly", () => {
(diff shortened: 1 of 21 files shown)


v3/src/components/axis/models/axis-model.test.ts
```diff
@@ -1,8 +1,9 @@
+import {getSnapshot, types} from "mobx-state-tree"
 import {
   AxisModel, AxisModelUnion, CategoricalAxisModel, EmptyAxisModel, IAxisModelUnion,
   isCategoricalAxisModel, isEmptyAxisModel, isNumericAxisModel, NumericAxisModel
 } from "./axis-model"
-import {getSnapshot, types} from "mobx-state-tree"
+import { AppHistoryService } from "../../../models/history/app-history-service"
```
kswenson, line 6: Could/should there be a `TestHistoryService` available in `src/test`? It seems weird for "low-level" jest tests to be reaching out for the application's history service.

kswenson (approved):
👍 Looks good -- just one question about whether jest tests really need the full `AppHistoryService` or could get by with a lesser `TestHistoryService`, which could have mock functions in it, for instance.

v3/src/components/axis/models/axis-model.test.ts
scytacki: Yes they could be switched to use a Test or Noop history service. I was going to add that when I started moving tests over that are within to the "library". But it does seem good to add it now so I don't forget to update the tests that are outside of the "library".

scytacki: However on second thought, given the amount of files changed in this PR. I'd rather get it merged sooner than later. So I'm going to punt this change. I've added it to the the document where I'm tracking what is left to finish.

PR #1736, opened 2025-01-15
188616850 map export v2

eireland:
Adds map export from V3 to V2


v3/src/components/map/v2-map-exporter.ts
```diff
@@ -0,0 +1,129 @@
...
+          areaColor: polygonDescription._itemColors[0] ?? "",
+          // TODO_V2_EXPORT: areaTransparency
+          areaTransparency: 1,
+          areaStrokeColor: polygonDescription.itemStrokeColor ?? "",
+          // TODO_V2_EXPORT: areaStrokeTransparency
+          areaStrokeTransparency: 1,
+          strokeSameAsFill: polygonDescription.itemStrokeSameAsFill ?? false
+        }
+        // V2 needs polygon layers appear before the point layers
+        v2LayerModels.unshift(polygonLayerModel)
```
kswenson, line 110: As near as I can tell from looking at the v2 code, point layer models are written out before polygon layer models. I think this needs further investigation before proceeding down this path.

kswenson (changes requested):
👍 Generally looks good. I added a commit which tweaks a few things. The remaining issue is that a glance at the v2 code suggests that point layer models are written out before polygon layer models. I think we need to investigate the issue of order-dependence among the layers before proceeding.

kswenson (approved):
👍 Looks good -- much better solution for the layer ordering. I added a commit which tweaked a few things, including many of the same things I tweaked in my original review commit, which seems to have been lost when you force-pushed after merging from main. Please take care when merging/force-pushing, particularly when other engineers have been working on a branch. As usual, make sure I didn't break anything in the process before merging.

PR #1738, opened 2025-01-15
fix: plugin API matches data context title as well as name; table inspector info shows title

kswenson:
For historical reasons CODAP objects (attributes, data sets, etc.) generally have an internal `name` property and an external `title` property. When created the `name` and `title` are the same, but the user has the ability to change the `title` (but generally not the `name`). This PR fixes two name/title issues:
1. The plugin API matches data sets (nee data contexts) by title as well as name. This is an improvement over v2 behavior.
2. The case table and case card data set information dialog display the title rather than the name (matching v2 behavior).


tealefristoe (approved):
Looks good 👍

PR #1739, opened 2025-01-16
split getGlobalValueManager

scytacki:
this way tile-environment doesn't depend on a specific shared model directly


kswenson (approved):
👍 LGTM

PR #1740, opened 2025-01-16
split data set linking from shared-data-utils

scytacki:
The data set linking methods were only used by the document. They also required the ITileModel, so this split allows shared-data-utils to be used in the library without the ITileModel.


kswenson (approved):
👍 LGTM

PR #1741, opened 2025-01-16
split out v2 dataset conversion

scytacki:
DataSet notifications in CODAP convert cases and attributes to v2 form.
This conversion brings in several files and probably is unnecessary in a library.

This PR adds an adapter which can convert the attributes and cases.
A default adapter is configured which does no conversion.

Hopefully this won't cause too much confusion if someone is tracing through the code and trying to figure out why what they see in the plugin messages doesn't match what they expect.

This also probably resulted in reduced code coverage by the Jest tests. However the actual coverage only dropped a little, probably because the Cypress tests covered it. Perhaps some Jest tests should install the V2Adapter.

### DIFF
--- a/v3/src/components/app.tsx
+++ b/v3/src/components/app.tsx
@@ -12,6 +12,8 @@ import { addDefaultComponents } from "../models/codap/add-default-content"
 import {gDataBroker} from "../models/data/data-broker"
 import {IDataSet} from "../models/data/data-set"
 import { dataContextCountChangedNotification } from "../models/data/data-set-notifications"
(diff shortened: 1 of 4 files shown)


kswenson (approved):
👍 LGTM -- presumably some of the v2 API handler tests could/should set the adapter.

scytacki:
This change only affects the notifications. And it looks like there aren't any tests of the notifications. It seems most notifications happen in response to UI actions. So they will be hard to test in isolation with this current design. 

Ideally the `applyModelChange` blocks could be moved into models somehow. This would make testing notifications easier. Project for another day. 😄

PR #1745, opened 2025-01-17
feat: initial v2 graph export

kswenson:
Implements v2 export of the basic graph configuration of attributes, types, roles, and axes. Includes jest test of full round-trip fidelity of opening a v2 document via the v2 importer, exporting it as a v2 document via the v2 exporter, and then comparing the exported graph to the original for each graph in the document, modulo an explicit list of unimplemented properties. Tests this round-trip fidelity against three documents with many different graph configurations. Moving forward, as new properties are supported in the export (e.g. `adornments`), they can be removed from the list of unsupported properties. When the list is empty, the basic work is complete.


scytacki (approved):
Looks good to me.

bfinzer (approved):
👍🏻LGTM

PR #1746, opened 2025-01-17
handle vertical movable lines

scytacki:
This adds a JsonNumber type which serializes NaN, Infinity, and -Infinity as strings.
This makes the javascript number type safe to serialize with MST.

It fixes a bug where vertical lines with Infinity slope were not round tripping.


kswenson (approved):
👍 Looks good

PR #1747, opened 2025-01-18
fix: case selection is saved/restored

kswenson:
[[PT-188794630]](https://www.pivotaltracker.com/story/show/188794630)


bfinzer (approved):
👍🏻LGTM

PR #1748, opened 2025-01-19
Fix graph's displayOnlySelected and axis rerendering touchups

bfinzer:
[#188719090] Bug fix: displayOnlySelectedCases shows all cases after a click away

* The problem was that on choosing the option, unselected cases were being added to hidden cases instead of allowing filtering in the data configuration model to dynamically respond to selection. So part of the fix is in `_filterCase` to detect that displayOnlySelectedCases is true and return false if the given case is not selected.
* We also take this opportunity to reduce the aggressiveness with which the entire axis repaints. For numeric axes we don't wish the axis to rescale or repaint when selection changes, even if displayOnlySelectedCases is true. We check more carefully for identical numeric axis domains and don't rerender if they aren't changing.
* In debugging we found it useful for svg elements to have a class name, so we added some that didn't already have them.
* Fixing one unnecessary rerendering of numeric axis required detection of `pointDisplayType` in useSubAxis. But this meant changing `IAxisProvider` to be able to cough this up. Adding this caused a ripple through some cypress test files and slider-model.ts.
* A slight modification was required for axis.spec.ts because changing from points as bars back to points doesn't needlessly rescale the axis for the test situation.

### DIFF
--- a/v3/cypress/e2e/axis.spec.ts
+++ b/v3/cypress/e2e/axis.spec.ts
@@ -343,7 +343,7 @@ context("Test graph axes with various attribute types", () => {
     // Switch back to points without undo/redo to clean up state
     cy.get("[data-testid=graph-display-config-button").click()
     cy.get("[data-testid=points-radio-button]").click()
-    ah.verifyAxisTickLabel("bottom", "−0.5", 0)
+    ah.verifyAxisTickLabel("bottom", "0", 0)
   })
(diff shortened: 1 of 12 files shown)


kswenson (approved):
👍 LGTM

eireland (approved):
Mostly sync reviewed. Looks good

PR #1749, opened 2025-01-20
Fix crash on resize categorical

bfinzer:
[#188794755] Bug fix: Crash on categorical/categorical plot graph resize

* The fix is to prevent the number of categories limit from going negative.


eireland (approved):
Sync reviewed. Looks good

PR #1750, opened 2025-01-20
Allow 2 y-attrs when x-attribute is date

bfinzer:
[#188523118] Bug fix: Cannot create multiple y axes in graph with date-time axis

* The fix is in `GraphDataConfigurationModel:placeCanAcceptAttributeIDDrop` to weaken the restriction on the x-attribute type to allow date as well as numeric


eireland (approved):
Sync revied. Looks good.

PR #1752, opened 2025-01-21
Trimming imported and manually entered values

bfinzer:
[#188801759] Bug fix: Values and attribute names in an imported CSV should be trimmed
[#188804884] Bug fix: Values entered in case table/card are not being trimmed.

* Added trimming to importValueToString. This applies to both bug situations.


kswenson (approved):
👍 LGTM

PR #1753, opened 2025-01-22
split out the data set types from data-interactive-types

scytacki:
The full data-interactive-types depend on the slate editor types. The data set portion of the types will be the most commonly used by plugins.

Note more work will be needed to split th `v2/codap-v2-types` since those also depend on the slate-editor types.


v3/src/data-interactive/handlers/collection-handler.ts
```diff
@@ -4,8 +4,8 @@ import { getSharedCaseMetadataFromDataset } from "../../models/shared/shared-dat
 import { toV2Id } from "../../utilities/codap-utils"
 import { registerDIHandler } from "../data-interactive-handler"
 import {
-  DIHandler, DIResources, DIValues, DICreateCollection, DICollection, DIUpdateCollection
-} from "../data-interactive-types"
+  DIHandler, DIResources, DIValues} from "../data-interactive-types"
```
kswenson, line 7: Should fit on one line now.

kswenson (approved):
👍 LGTM

v3/src/data-interactive/handlers/collection-handler.ts
scytacki: I'll add that fix to the next PR: https://github.com/concord-consortium/codap/pull/1755

PR #1756, opened 2025-01-22
Fix for legend response to selection bug

bfinzer:
[#188492167] Bug fix: Selection of numeric legend key is not showing that the range is selected

* There were actually problems with selection in the categorical legend also. Both are fixed by using an `mstReaction` instead of an `onAnyAction` in the relevant useEffects.


kswenson (approved):
👍 LGTM

eireland (approved):
Sync reviewed. Looks good.

PR #1759, opened 2025-01-23
Trim attribute names on csv import

bfinzer:
[#188801759] Bug fix: Attribute name needs to be trimmed

* When attributes are being created during csv import, their names were not being trimmed. We fix by trimming the name when creating the attribute snapshot in convertParsedCsvToDataSet
* For good measure, we trim any string passed to Attribute:setName


kswenson (approved):
👍 LGTM

PR #1760, opened 2025-01-23
extract v2 data set importer

scytacki:
This extracted importer will be used by the DST plugin.
There is other code in the data-interactives folder which also handles converting v2 like datasets sent by plugins. However it wasn't easy to unify that code too.

This also fixes an issue with the `attrFromName` map in dataset. The map is converted to view so it gets automatically recreated when an attribute's name changes or an attribute is added or removed from the list. The new approach does behave a little differently. A new map is created when something changes. So this means users of this map should not hold onto it and assume it will be updated. From what I could tell there was no code that was doing that, so no other code needed to be updated. I assumed the number of attributes is small enough and updates to attributes are infrequent enough, that recreating the map each time would not be a problem.


kswenson (approved):
👍 LGTM

PR #1761, opened 2025-01-24
188575507 color functions

eireland:
Adds color functions to the formula editor.
- rgb(red,green,blue ,[alpha])
- hsl (hue,saturation,lightness,[alpha])
- darken(color, percentage)
- lighten(color, percentage)


v3/src/models/formula/functions/color-functions.ts
```diff
@@ -0,0 +1,89 @@
...
+      let [r, g, b, a] = args
+
+      if (!isValidRgbValue(r) || !isValidRgbValue(g) || !isValidRgbValue(b)) return ""
+      if (a !== undefined) {
+        if (typeof a === 'number' && a > 1) {
+          a = a / 100
+        }
+        if (!isValidPercentage(a)) return ""
+      }
+      const rgbString = a !== undefined ? `rgba(${r}, ${g}, ${b}, ${a})` : `rgb(${r}, ${g}, ${b})`
```
bfinzer, line 32: Is there a reason not to use objects instead of strings, as in:
const color = colord({ r: Number(r), g: Number(g), b: Number(b), a: Number(a) })

v3/src/models/formula/functions/color-functions.ts
```diff
@@ -0,0 +1,89 @@
...
+      const lightness = typeof args[2] === 'number' ? `${args[2]}%` : args[2]
+      let alpha = args.length === 4 ? args[3] : undefined
+      if (alpha !== undefined) {
+        alpha = isValidPercentage(alpha) && typeof alpha === "number" ? alpha > 1 ? alpha / 100 : alpha : ""
+      }
+
+      const hslString = alpha !== undefined
+        ? `hsla(${hue}, ${saturation}, ${lightness}, ${alpha})`
+        : `hsl(${hue}, ${saturation}, ${lightness})`
+
```
bfinzer, line 59: Same question about using objects rather than strings?

v3/src/models/formula/functions/color-functions.ts
```diff
@@ -0,0 +1,89 @@
...
+      return colord(hslString).isValid() ? colord(hslString).toHex() : ""
+    }
+  },
+
+  // darken(color, percentage)
+  darken: {
```
bfinzer, line 65: When I evaluate `darken("#FF0000",0.5)` I get "#000000" whereas I would think I would get "#880000". Is this an error in the colord library or a misunderstanding on my part?

v3/src/models/formula/functions/color-functions.ts
```diff
@@ -0,0 +1,89 @@
...
+      const colorVal = typeof color === "string" ?  parseColorToHex(color, { colorNames: true }) : color
+      if (typeof colorVal !== "string" || !colord(colorVal).isValid() || !isValidPercentage(percentage)) return ""
+      const percentageVal = typeof percentage === "number" ? percentage > 1 ? percentage / 100 : percentage : 0
+      const darkenedColor = colord(colorVal).darken(percentageVal).toHex()
+      return darkenedColor
+    }
+  },
+
+  // darken(color, percentage)
+  lighten: {
```
bfinzer, line 78: Similar problem here in that a percentage of 50% seems to take the color to white.

bfinzer (changes requested):
* I revised the descriptions, hopefully improving them somewhat.
* I ask a question about using objects rather than strings.
* I ask questions about the behavior of darken and lighten not being what I would expect.
I think the questions about `darken` and `lighten` need to be resolved.

v3/src/models/formula/functions/color-functions.ts
kswenson: I agree that passing the values independently makes more sense than packaging them as a string which just requires that `colord` parse the resulting string to extract the individual values again.

v3/src/models/formula/functions/color-functions.ts
kswenson: As above, I agree that passing the values independently makes more sense than packaging them as a string which just requires that `colord` parse the resulting string to extract the individual values again.

v3/src/models/formula/functions/color-functions.ts
kswenson: I looked into this and there's a design decision to be made. What `colord` is doing is consistent with what other "standard" lighten/darken functions do (e.g. SASS), in that the lighten/darken parameter is additive, not proportional. In other words, in your example, it's not moving 50% of the way toward the dark extreme, it's converting to HSL and subtracting 50% from the L value. `#FF0000` converted to HSL is `hsl(0,100%,50%)` minus 50% is `hsl(0,100%,0%)`. Intuitively, I expected it to be proportional (as I suspect you do), i.e. that the request is to move 50% of the distance to 0%, which would be `hsl(0,100%,25%)`. But if we want it to work that way, we have to code it that way because that's not the way the function behaves out of the box, and it's not the way other "standard" lighten/darken functions behave.

v3/src/models/formula/functions/color-functions.ts
kswenson: See comment/explanation above for the `darken` function.

v3/src/models/formula/functions/math.ts
```diff
@@ -88,7 +89,9 @@ export const fnRegistry = {
 
   ...bivariateStatsFunctions,
 
-  ...semiAggregateFunctions
+  ...semiAggregateFunctions,
+
+  ...colorFunctions
```
kswenson, line 94: There's a loose simple to complex ordering going on here, so I'd put this up between the logic functions and the date functions with the other type-specific functions.

v3/src/models/formula/functions/color-functions.ts
```diff
@@ -0,0 +1,89 @@
...
+    numOfRequiredArguments: 2,
+    evaluate: (...args: FValue[]) => {
+      const [color, percentage] = args as [ColorFValue, FValue]
+      const colorVal = typeof color === "string" ?  parseColorToHex(color, { colorNames: true }) : color
```
kswenson, line 82:
```suggestion
      const colorVal = typeof color === "string" ? parseColorToHex(color, { colorNames: true }) : color
```

kswenson (changes requested):
👍 Generally looks good -- I'm generally in agreement with @bfinzer's comments and we need to figure out how we want to handle the lighten/darken functions.

kswenson (approved):
👍 Looks good -- I added a commit which improves the TypeScript types and fixes the darken computation.

bfinzer (approved):
👍🏻 LGTM

PR #1764, opened 2025-01-24
Fixes date precision in case card

eireland:


v3/src/components/case-tile-common/attribute-format-utils.tsx
```diff
@@ -87,8 +87,9 @@ export function renderAttributeValue(str = "", num = NaN, showUnits = false, att
     const date = parseDate(str, true)
     if (date) {
       const formattedDate = formatDate(date, datePrecision)
+      console.log('formattedDate', formattedDate)
```
kswenson, line 90:
```suggestion
```

kswenson (approved):
👍 LGTM, but remove the console.log.

PR #1767, opened 2025-01-27
Adornments respond to changes in selection if only showing measures for selection

bfinzer:
[#187711050] Feature: Adornments respond to Show Measures for Selection option

* The property governing this feature was being stored as part of the `GraphContentModel` which made it inaccessible to the `UnivariateMeasureAdornmentModel` which needs to know the value in its `getCaseValues` method. Accordingly, we move `showMeasuresForSelection` into the `GraphDataConfigurationModel` where it is more accessible. This causes a ripple effect of changes through the files that were assuming it belonged to `GraphContentModel`. Among these were graph-component-handler.test.ts where I simply removed references to `showMeasuresForSelection`.
* graph-content-model.ts (where I reorganized the imports), add an mstReaction that responds to selection changes and updates the adornments if showMeasuresForSelection is true.

### DIFF
--- a/v3/src/components/graph/models/graph-content-model.ts
+++ b/v3/src/components/graph/models/graph-content-model.ts
@@ -3,7 +3,7 @@
  * Its array of DataDisplayLayerModels has just one element, a GraphPointLayerModel.
  */
 import {cloneDeep} from "lodash"
-import {when} from "mobx"
+import { comparer, when } from "mobx"
 import { format } from "d3"
 import {addDisposer, Instance, SnapshotIn, types} from "mobx-state-tree"
 import { t } from "../../../utilities/translation/translate"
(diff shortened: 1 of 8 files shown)


v3/src/components/graph/models/graph-content-model.ts
```diff
@@ -270,6 +270,17 @@ export const GraphContentModel = DataDisplayContentModel
         const updateCategoriesOptions = self.getUpdateCategoriesOptions()
         self.adornmentsStore.updateAdornments(updateCategoriesOptions)
       }, {name: "GraphContentModel.afterAttachToDocument.updateAdornments"}, self.dataConfiguration))
+
+      // When showMeasuresForSelection is true, update adornments when selection changes
+      addDisposer(self, mstReaction(() => {
+        return self.dataConfiguration.selection
+      },
+        () => {
+          if (self.dataConfiguration.showMeasuresForSelection) {
```
kswenson, line 279: You also presumably want to run the reaction when `showMeasuresForSelection` changes, in which case you could do something like:
```suggestion
        return [self.dataConfiguration.showMeasuresForSelection, self.dataConfiguration.selection]
      },
        ([showMeasuresForSelection, selection]) => {
          if (showMeasuresForSelection) {
```
and then you probably want `equals: comparer.structural` in the options below.

kswenson (changes requested):
👍 Looks good -- just a couple suggestions around the reaction.

kswenson (approved):
👍 LGTM

PR #1769, opened 2025-01-27
make cachedFnWithArgsFactory observable

scytacki:
This should not impact performance unless a lot of keys are used. I'd assume this cached function approach shouldn't be used when there are lots and lots of different keys.

I'd guess making this observable could eliminate several places where extra reactions are used around the dataConfigurationModel. I'm only aware of one specific case within the categorical legend which will be part of a different PR.

### DIFF
--- a/v3/src/utilities/mst-utils.ts
+++ b/v3/src/utilities/mst-utils.ts
@@ -1,3 +1,4 @@
+import { observable } from "mobx"
 import {
   IAnyStateTreeNode, IDisposer, isAlive, ISerializedActionCall, getParent, getType,
   IOnActionOptions, onAction, hasParent, types, getSnapshot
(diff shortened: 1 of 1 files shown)


v3/src/utilities/mst-utils.ts
```diff
@@ -122,7 +123,13 @@ export function cachedFnWithArgsFactory<FunDef extends (...args: any[]) => any>(
   // with the same arguments as the calculate() function. It will work even if the client code completely skips
   // explicit type definition between < and >.
   const { key, calculate } = options
-  const cacheMap = new Map<string, ReturnType<FunDef>>()
+
+  // The map is observable so any observers will be triggered when the cache is updated
+  // The values within the map are not automatically made observable since
+  // cachedFnWithArgsFactory is usually used in cases where the values are large objects
+  // and usually a whole new value object is created by on each calculate call
+  const cacheMap = observable.map<string, ReturnType<FunDef>>({}, {deep: false})
```
kswenson, line 131: I like this -- I wish I had thought of it myself. 😉

scytacki:
@kswenson I haven't changed the PR, but it was still a draft and I hadn't run the regression tests when you looked at it last. 
Now you should be able to approve it.

kswenson (approved):
👍 Looks good!

PR #1770, opened 2025-01-28
Propagate null value in arithmetic

bfinzer:
[#188590675] Bug fix: **Formula Editor** prev() and next () function produces invalid argument error in first cell

[#188517456] Feature: You can use the "modulo" operator ('%') in formulas

* The problem is that the formula engine was not propagating null values ("").
* We modify `subtract` and add `multiply` and `divide` in operators.ts to return "" if either of the operators are the empty string.
* As a bonus, we add `mod` so that V3 can evaluate the `%` operator.

### DIFF
--- a/v3/src/models/formula/functions/operators.ts
+++ b/v3/src/models/formula/functions/operators.ts
@@ -140,6 +140,11 @@ export const operators = {
       const [isANumber, aNumber] = checkNumber(a)
       const [isBNumber, bNumber] = checkNumber(b)
 
+      // Empty strings
+      if (a === "" || b === "") {
+        return ""
+      }
+
       // Date objects
       if (isADate || isBDate) {
         // both are dates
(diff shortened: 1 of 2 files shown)


v3/src/models/formula/functions/operators.ts
```diff
@@ -140,6 +140,11 @@ export const operators = {
       const [isANumber, aNumber] = checkNumber(a)
       const [isBNumber, bNumber] = checkNumber(b)
 
+      // Empty strings
+      if (a === "" || b === "") {
+        return ""
+      }
```
kswenson, line 146: Move this up above the `checkDate` and `checkNumber` lines. There's no reason to check those things if we're just going to bail. Also, maybe we should make the check broader just in case:
```suggestion
      // Empty strings
      if (a == null || a === "" || b == null || b === "") {
        return ""
      }
```

v3/src/models/formula/functions/operators.ts
```diff
@@ -161,5 +166,71 @@ export const operators = {
...
+    evaluateOperator: (a: any, b: any) => {
+      const multiplyError = new Error(`Invalid arguments for multiply operator: ${a}, ${b}`)
+
+      const [isANumber, aNumber] = checkNumber(a)
+      const [isBNumber, bNumber] = checkNumber(b)
+
+      // Empty strings
+      if (a === "" || b === "") {
+        return ""
+      }
```
kswenson, line 183: Same comment as above.

v3/src/models/formula/functions/operators.ts
```diff
@@ -161,5 +166,71 @@ export const operators = {
...
+    evaluateOperator: (a: any, b: any) => {
+      const divideError = new Error(`Invalid arguments for multiply operator: ${a}, ${b}`)
+
+      const [isANumber, aNumber] = checkNumber(a)
+      const [isBNumber, bNumber] = checkNumber(b)
+
+      // Empty strings
+      if (a === "" || b === "") {
+        return ""
+      }
```
kswenson, line 205: Same comment as above.

v3/src/models/formula/functions/operators.ts
```diff
@@ -161,5 +166,71 @@ export const operators = {
...
+    evaluateOperator: (a: any, b: any) => {
+      const modError = new Error(`Invalid arguments for multiply operator: ${a}, ${b}`)
+
+      const [isANumber, aNumber] = checkNumber(a)
+      const [isBNumber, bNumber] = checkNumber(b)
+
+      // Empty strings
+      if (a === "" || b === "") {
+        return ""
+      }
```
kswenson, line 227: Same comment as above.

kswenson (changes requested):
👍 Looks good! Just a couple suggestions about the nature and placement of the empty value checks.

kswenson:
> 👍 Looks good! Just a couple suggestions about the nature and placement of the empty value checks.

Also, awesome that you were able to fix the `%` behavior as well!

kswenson (changes requested):
👍 Looks good -- I should have mentioned in my prior review that you should enhance the jest tests to test the new behaviors.

kswenson (approved):
👍 LGTM

PR #1772, opened 2025-01-28
categorical legend refactor

scytacki:
create categorical legend model

- use autorun to do the d3 rendering
- leverage MobX computed properties to do caching
- use a single d3 join instead of two, the double d3 join was the source of the error shown before
- don't modify the categorySet during dragging, only modify at the end. This makes undo/redo work better, but there is still a lingering issue with it. I've found a single legend key move is undone correctly, but when I do multiple legend moves undo doesn't work anymore.

### DIFF
--- a/v3/src/components/data-display/components/legend/categorical-legend-model.ts
+++ b/v3/src/components/data-display/components/legend/categorical-legend-model.ts
@@ -1,0 +1,254 @@
+import { makeAutoObservable } from "mobx"
+import { measureText } from "../../../../hooks/use-measure-text"
+import { logMessageWithReplacement } from "../../../../lib/log-message"
+import { missingColor } from "../../../../utilities/color-utils"
+import { axisGap } from "../../../axis/axis-types"
+import { getStringBounds } from "../../../axis/axis-utils"
+import vars from "../../../vars.scss"
+import { kMain, kDataDisplayFont } from "../../data-display-types"
+import { IDataConfigurationModel } from "../../models/data-configuration-model"
+import { DataDisplayLayout } from "../../models/data-display-layout"
+
+export interface Key {
+  category: string;
+  color: string;
+  index: number;
+  column: number;
+  row: number;
+}
+interface Layout {
+  maxWidth: number;
+  fullWidth: number;
+  numColumns: number;
+  numRows: number;
+  columnWidth: number;
+}
+
+export const keySize = 15
+export const padding = 5
+export const labelHeight = getStringBounds('Wy', vars.labelFont).height + axisGap
+
+export class CategoricalLegendModel {
+  dragInfo = {
+    category: "",
+    initialOffset: { x: 0, y: 0 },
+    currentDragPosition: { x: 0, y: 0 },
+    initialIndex: -1,
+    currentIndex: -1
+  }
+
+  dataDisplayLayout: DataDisplayLayout
+  dataConfiguration: IDataConfigurationModel | undefined
+
+  categoriesAtDragStart: string[] = []
+  colorsAtDragStart: Record<string, string> = {}
+
+  constructor(dataConfiguration: IDataConfigurationModel | undefined, dataDisplayLayout: DataDisplayLayout) {
+    this.dataDisplayLayout = dataDisplayLayout
+    this.dataConfiguration = dataConfiguration
+    makeAutoObservable(this)
+  }
+
+  setDataConfiguration(dataConfiguration: IDataConfigurationModel | undefined) {
+    this.dataConfiguration = dataConfiguration
+  }
+
+  setDataDisplayLayout(dataDisplayLayout: DataDisplayLayout) {
+    this.dataDisplayLayout = dataDisplayLayout
+  }
+
+  get categoriesFromDoc() {
+    const categories = this.dataConfiguration?.categoryArrayForAttrRole('legend') || []
+    return categories.filter(cat => cat !== kMain)
+  }
+
+  getColorForCategoryFromDoc(cat: string, defaultColor = missingColor) {
+    return this.dataConfiguration?.getLegendColorForCategory(cat) ?? defaultColor
+  }
+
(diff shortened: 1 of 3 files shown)


v3/src/components/data-display/components/legend/categorical-legend-model.ts
```diff
@@ -0,0 +1,254 @@
+import { makeAutoObservable } from "mobx"
+import { measureText } from "../../../../hooks/use-measure-text"
+import { logMessageWithReplacement } from "../../../../lib/log-message"
+import { missingColor } from "../../../../utilities/color-utils"
+import { axisGap } from "../../../axis/axis-types"
+import { getStringBounds } from "../../../axis/axis-utils"
+import vars from "../../../vars.scss"
```
kswenson, line 7: This is an interesting case. I would normally put this at the bottom given that it's a CSS import, but since it's being used as a JavaScript import in this case, it could arguably be placed here as well. 🤷

v3/src/components/data-display/components/legend/categorical-legend.tsx
```diff
@@ -1,351 +1,162 @@
...
+    // Also the setDesiredExtent might cause extra re-renders, so it only
+    // runs when the desiredExtent actually changes
+    useEffect(function updateDesiredExtent() {
+      return mstReaction(
+        () => {
+          if (dataConfiguration?.placeCanHaveZeroExtent('legend')) {
+            return 0
+          }
+          const lod = legendModel.layoutData
+          return lod.numRows * (keySize + padding) + labelHeight + axisGap
```
kswenson, line 41: `lod` seems unnecessarily obtuse. Why not simply:
```suggestion
          return legendModel.layoutData.numRows * (keySize + padding) + labelHeight + axisGap
```

v3/src/components/data-display/components/legend/categorical-legend.tsx
```diff
@@ -1,351 +1,162 @@
...
+        })
+        .style('fill', (d) => d.color)
+        .transition().duration(duration.current)
+        .on('end', () => {
+          duration.current = 0
+        })
+        .attr('x', (d) => {
+          return dI.category === d.category
+            ? dI.currentDragPosition.x - dI.initialOffset.x
+            : axisGap + (d.column || 0)* legendModel.layoutData.columnWidth
```
kswenson, line 132:
```suggestion
            : axisGap + (d.column || 0) * legendModel.layoutData.columnWidth
```

v3/src/components/data-display/components/legend/categorical-legend.tsx
```diff
@@ -1,351 +1,162 @@
...
+      keysSelection.select('text')
+        .text((d) => d.category)
+        .transition().duration(duration.current)
+        .on('end', () => {
+          duration.current = 0
+        })
+        .attr('x', (d) => {
+          return keySize + 3 + (dI.category === d.category
+            ? dI.currentDragPosition.x - dI.initialOffset.x
+            : axisGap + (d.column || 0)* legendModel.layoutData.columnWidth)
```
kswenson, line 148:
```suggestion
            : axisGap + (d.column || 0) * legendModel.layoutData.columnWidth)
```

v3/src/components/data-display/components/legend/categorical-legend-model.ts
```diff
@@ -0,0 +1,254 @@
...
+    const row = Math.floor(y / (keySize + padding))
+    const catIndex = row * lod.numColumns + col
+
+    return catIndex >= 0 && catIndex < this.numCategories ? catIndex : -1
+  }
+
+  catLocation(categoryKey: Key) {
+    return {
+      x: axisGap + categoryKey.column * this.layoutData.columnWidth,
+      y: /*labelHeight + */ categoryKey.row * (keySize + padding)
```
kswenson, line 112: Do you want to keep this comment?

v3/src/components/data-display/components/legend/categorical-legend-model.ts
```diff
@@ -0,0 +1,254 @@
...
+
+    return lod
+  }
+
+  coordinatesToCatIndex(localPoint: { x: number; y: number; }) {
+    const lod = this.layoutData
+    const { x, y } = localPoint
+    const col = Math.floor(x / lod.columnWidth)
+    const row = Math.floor(y / (keySize + padding))
+    const catIndex = row * lod.numColumns + col
```
kswenson, line 104: I don't think `lod` enhances readability here:
```suggestion
    const { x, y } = localPoint
    const col = Math.floor(x / this.layoutData.columnWidth)
    const row = Math.floor(y / (keySize + padding))
    const catIndex = row * this.layoutData.numColumns + col
```

kswenson (approved):
👍 LGTM -- all of my comments are minor. I added @bfinzer as a reviewer as well since he's more familiar with the graph D3 code and he will also find these techniques interesting.

v3/src/components/data-display/components/legend/categorical-legend-model.ts
scytacki: I'll move it down because it does look strange and I wasn't paying attention to it being a css file.

v3/src/components/data-display/components/legend/categorical-legend-model.ts
scytacki: It was there before. I guess since it is rendering correctly it makes sense to remove it.

bfinzer (approved):
👍🏻 LGTM
I appreciate both the refactoring and the D3 niceties!

scytacki:
I just realized the code coverage for the changes is low. I'm going to merge it as is, but if I can find time I'll come back and improve that. The biggest missing coverage is around dragging which was not covered before. Dragging is hard to test with Cypress. Now that most of the dragging code is in a model it will be possible to test it with pretty simple Jest tests.

PR #1773, opened 2025-01-28
Remove the defocus when we are already in input mode

eireland:
We were telling the component state that a click inside the Input component should set the focus to false even though it should still be in edit mode.
Removed the setFocus so we keep the focus in the input until we click away.


kswenson (approved):
👍 LGTM

PR #1774, opened 2025-01-29
feat: v2 export of LSRL adornment

kswenson:
Includes round-trip import/export test using document `mammals-lsrl-adornment.codap`.

### DIFF
--- a/v3/src/components/graph/v2-graph-exporter.test.ts
+++ b/v3/src/components/graph/v2-graph-exporter.test.ts
@@ -157,4 +157,22 @@ describe("V2GraphImporter", () => {
       expect(v2GraphTileOutStorage).toEqual(v2GraphTileStorage)
     })
   })
+
+  it("exports graph components with LSLR adornments", () => {
+    const { v2Document } = loadCodapDocument("mammals-lsrl-adornment.codap")
+    const v2GraphTiles = v2Document.components.filter(c => c.type === "DG.GraphView")
+    v2GraphTiles.forEach(v2GraphTile => {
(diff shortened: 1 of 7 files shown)


v3/src/components/graph/v2-graph-exporter.test.ts
```diff
@@ -157,4 +157,22 @@ describe("V2GraphImporter", () => {
       expect(v2GraphTileOutStorage).toEqual(v2GraphTileStorage)
     })
   })
+
+  it("exports graph components with LSLR adornments", () => {
+    const { v2Document } = loadCodapDocument("mammals-lsrl-adornment.codap")
+    const v2GraphTiles = v2Document.components.filter(c => c.type === "DG.GraphView")
+    v2GraphTiles.forEach(v2GraphTile => {
+      // console.log("v2GraphTile", v2GraphTile.componentStorage.title || v2GraphTile.componentStorage.name)
```
eireland, line 165: remove console.log

kswenson: I'm intentionally leaving those in (but commented out) because they're useful for determining which graph is failing when the test fails, so during development uncommenting those console.logs is often the first step.

bfinzer (approved):
👍🏻LGTM

eireland (approved):
👍 LGTM

kswenson:
I'm leaving those `console.log`s in  because they're necessary when developing/debugging the tests to indicate what graph is failing the test. I don't want them to log in the general case, however, because successful tests should be silent when not under active development. Since they are apparently a source of some confusion, I'll convert it to a function call like `logGraphTitle()` that takes a boolean parameter and has a comment explaining its purpose in the last PR in the chain.

PR #1775, opened 2025-01-29
Change attribute type to categorical with a date on opposite axis plot bug fix

eireland:
Adds check for date type and makes that primaryRole in the abscence of a numeric type in either if the axis

### DIFF
--- a/v3/src/components/graph/models/graph-model-utils.ts
+++ b/v3/src/components/graph/models/graph-model-utils.ts
@@ -27,10 +27,11 @@ function setPrimaryRoleAndPlotType(graphModel: IGraphContentModel) {
   // Numeric attributes get priority for primaryRole when present. First one that is already present
   // and then the newly assigned one. If there is an already assigned categorical then its place is
   // the primaryRole, or, lastly, the newly assigned place
-  primaryRole = attributeType === 'numeric' ? oldPrimaryRole
-    : otherAttributeType === 'numeric' ? otherAttrRole
-      : otherAttributeType === 'date' ? otherAttrRole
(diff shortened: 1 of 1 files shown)


tealefristoe (approved):
Looks good 👍

v3/src/components/graph/models/graph-model-utils.ts
```diff
@@ -29,7 +29,8 @@ function setPrimaryRoleAndPlotType(graphModel: IGraphContentModel) {
   // the primaryRole, or, lastly, the newly assigned place
   primaryRole = attributeType === 'numeric' ? oldPrimaryRole
     : otherAttributeType === 'numeric' ? otherAttrRole
-      : attributeType !== 'empty' ? oldPrimaryRole : otherAttrRole
+      : otherAttributeType === 'date' ? otherAttrRole
+        : attributeType !== 'empty' ? oldPrimaryRole : otherAttrRole
```
kswenson, line 33: The logic here is subtle, but I think perhaps what is desired is something more like:
```typescript
  primaryRole = ['numeric', 'date'].includes(attributeType) ? oldPrimaryRole
    : ['numeric', 'date'].includes(otherAttributeType) ? otherAttrRole
      : attributeType !== 'empty' ? oldPrimaryRole : otherAttrRole
```

kswenson (changes requested):
I don't think the logic is quite right. I made another suggestion, but it should be double-checked as well.

eireland:
@bfinzer Need your opinion on @kswenson's suggested change.

bfinzer (approved):
LGTM after acceding to Kirk's recommended change.

kswenson (approved):
👍 LGTM

PR #1776, opened 2025-01-29
feat: v2 export of movable lines and points

kswenson:
Supports v2 export of movable line and movable point adornments by round-tripping a document with numerous examples created by v2, imported to v3 and then re-exported to v2 format.
See enclosed document `mammals-movable-line-point-adornments.codap` to see what graph configurations are successfully round-tripped.

Note that v2 sometimes renders multiple movable lines or movable points in a single graph, but it never serializes more than a single line or point. v3 handles more of these multiple lines/points situations appropriately, and so there will be a loss of information in these cases when round-tripping a v3-created document through the v2 export/import process. We'll need to extend the v2 format (hybrid format?) to handle that case.

This PR is based on #1774, which should be merged first (with merge commit).

Also fixes import of vertical movable lines from v2 documents.

Last major adornment-related feature to address is label/equation coordinates.

### DIFF
--- a/v3/src/components/graph/v2-graph-exporter.test.ts
+++ b/v3/src/components/graph/v2-graph-exporter.test.ts
@@ -5,13 +5,20 @@ import { SharedModelDocumentManager } from "../../models/document/shared-model-d
 import { ITileModelSnapshotIn } from "../../models/tiles/tile-model"
 import { safeJsonParse } from "../../utilities/js-utils"
 import { CodapV2Document } from "../../v2/codap-v2-document"
-import { ICodapV2DocumentJson } from "../../v2/codap-v2-types"
+import { ICodapV2DocumentJson, ICodapV2GraphComponent } from "../../v2/codap-v2-types"
 import { v2GraphExporter } from "./v2-graph-exporter"
 import { v2GraphImporter } from "./v2-graph-importer"
 import "./graph-registration"
 
 const fs = require("fs")
 const path = require("path")
+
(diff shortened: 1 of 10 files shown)


bfinzer (approved):
👍🏻LGTM

v3/src/components/graph/v2-graph-exporter.test.ts
```diff
@@ -175,4 +175,22 @@ describe("V2GraphImporter", () => {
       expect(v2GraphTileOutStorage).toEqual(v2GraphTileStorage)
     })
   })
+
+  it("exports graph components with movable line and point adornments", () => {
+    const { v2Document } = loadCodapDocument("mammals-movable-line-point-adornments.codap")
+    const v2GraphTiles = v2Document.components.filter(c => c.type === "DG.GraphView")
+    v2GraphTiles.forEach(v2GraphTile => {
+      // console.log("v2GraphTile", v2GraphTile.componentStorage.title || v2GraphTile.componentStorage.name)
```
eireland, line 183: remove

eireland (approved):
One console log, otherwise LGTM

kswenson:
I'm leaving those `console.log`s in  because they're necessary when developing/debugging the tests to indicate what graph is failing the test. I don't want them to log in the general case, however, because successful tests should be silent when not under active development. Since they are apparently a source of some confusion, I'll convert it to a function call like `logGraphTitle()` that takes a boolean parameter and has a comment explaining its purpose in the last PR in the chain.

PR #1777, opened 2025-01-29
Fix transition from numeric to categorical axis

bfinzer:
[#188819856] Bug fix: Graph axis is incorrect when switching from numerical to categorical plots

* The problem is that a previous PR assumed that it was unnecessary for the category-axis-helper to remove everything before rendering. But this left behind numeric axis stuff.
* The fix is to add a class "numeric-axis" when we're rendering a numeric axis that we can detect when rendering a categorical axis and, if present, remove the numeric axis elements

### DIFF
--- a/v3/src/components/axis/helper-models/numeric-axis-helper.ts
+++ b/v3/src/components/axis/helper-models/numeric-axis-helper.ts
@@ -70,7 +70,7 @@ export class NumericAxisHelper extends AxisHelper {
         return Number.isInteger(d) ? d.toString() : ""
       })
     }
-    subAxisSelection  //.append("g").attr("class", "numeric")
+    subAxisSelection
       .attr("class", "numeric-axis")
       .attr("transform", this.initialTransform)
       .transition().duration(duration)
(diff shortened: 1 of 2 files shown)


v3/src/components/axis/helper-models/numeric-axis-helper.ts
```diff
@@ -70,7 +70,8 @@ export class NumericAxisHelper extends AxisHelper {
         return Number.isInteger(d) ? d.toString() : ""
       })
     }
-    subAxisSelection
+    subAxisSelection  //.append("g").attr("class", "numeric")
```
kswenson, line 73: Is the comment still necessary/useful or can it be removed?

v3/src/components/axis/hooks/use-sub-axis.ts
```diff
@@ -177,6 +177,10 @@ export const useSubAxis = ({
       if (!subAxisElt) return
       subAxisSelectionRef.current = select(subAxisElt)
       const sAS = subAxisSelectionRef.current
+      if (sAS.classed('numeric-axis')) {
+        sAS.selectAll('g').remove()
+        sAS.classed('numeric-axis', false)
+      }
```
kswenson, line 183: Is there anything comparable that needs to happen when switching from categorical back to numeric?

kswenson (approved):
👍 LGTM -- just a couple of questions.

PR #1778, opened 2025-01-31
188839221 Fix Legend Select

tealefristoe:
PT Stories:
https://www.pivotaltracker.com/story/show/188839221
https://www.pivotaltracker.com/story/show/188820212

This PR fixes a bug where legends with both strings and numbers would not select the cases with numbers ([here's a better description](https://www.pivotaltracker.com/story/show/188820212)). It fixes the problem by always comparing string values rather than raw values.

It also fixes another bug where number entries would not highlight in a categorical legend, even when all of the relevant cases were selected. This was a result of the same problem, comparing a string with `getValue` instead of `getStrValue`.


kswenson (approved):
👍 LGTM -- I added @bfinzer as a reviewer as well to see if he agrees.

bfinzer (approved):
👍🏻 LGTM

PR #1779, opened 2025-01-31
feat: import/export univariate measure adornment labels

kswenson:
V2 stores measure label coordinates as proportional position of the top-left of the label in the plot's coordinate system, but this proportional position is then converted to an absolute position for rendering. V3 was both storing and rendering measure label coordinates as absolute position of the top-left of the label in cell coordinates. This absolute positioning meant that the label didn't move when the graph was resized. With this PR, V3 now stores _and_ renders the label coordinates as the proportional position of the top-left of the label in the cell's coordinate system. (V3 also clips labels to the cell, along with other cell rendering.)

With this PR we also convert measure label coordinates appropriately when importing from/exporting to v2 documents.

Along the way, we also:
- fix a bug in which the units were rendered redundantly in the standard error adornment's label.
- eliminate redundant `showMeasureLabels` properties in individual adornments. (It's a graph-wide property.)
- round label coordinates to two decimal places before comparing them for import/export fidelity.
- eliminate passing of `plotWidth`/`plotHeight` arguments when they're not actually used.


v3/src/components/graph/adornments/univariate-measures/normal-curve/normal-curve-adornment-component.tsx
```diff
@@ -115,14 +115,12 @@
       if (!numericAttrId || !dataConfig) return
       const labelSelection = select(labelRef.current)
       const labelCoords = measure.labelCoords
-      const lineHeight = 20
-      const topOffset = lineHeight * (adornmentsStore?.getLabelLinesAboveAdornment(model) ?? 0)
```
bfinzer, line 119: By not using getLabelLinesAboveAdornment adornment labels start overlapping. E.g. create a dot plot with a normal curve, mean and median. Then turn on "Show Measure Labels." Observe the difference in main compared to this branch.

v3/src/components/graph/adornments/univariate-measures/standard-error/standard-error-adornment-component.tsx
```diff
@@ -83,14 +80,12 @@ export const StandardErrorAdornmentComponent = observer(
       if (!numericAttrId || !dataConfig) return
       const labelSelection = select(labelRef.current)
       const labelCoords = measure.labelCoords
-      const lineHeight = 20
-      const topOffset = lineHeight * (adornmentsStore?.getLabelLinesAboveAdornment(model, isGaussianFit) ?? 0)
```
bfinzer, line 87: As with the normal curve adornment, if the user has not manually moved the measure label, getLabelLinesAboveAdornment is necessary to prevent overlap with other adornments.

v3/src/components/graph/adornments/univariate-measures/univariate-measure-adornment-simple-component.tsx
```diff
@@ -61,17 +59,13 @@ export const UnivariateMeasureAdornmentSimpleComponent = observer(
     ) => {
       const labelSelection = select(labelRef.current)
       const labelCoords = measure.labelCoords
-      const lineHeight = 20
-      // We have to pass isGaussianFit to getLabelLinesAboveAdornment because the Gaussian fit is a special case where
-      // if there is a gaussianFit (not just a normal curve), we need to add an extra line for the adornment title.
-      const topOffset = lineHeight * (adornmentsStore?.getLabelLinesAboveAdornment(model, isGaussianFit) ?? 0)
```
bfinzer, line 67: As with normal curve and standard error, getLabelLinesAboveAdornment is necessary to prevent measure label overlap

bfinzer:
Bug: When the user has dragged a measure label to a non-default position, resizing the graph causes the measure label to snap back to its default position. It should stay proportionately to where it was dragged.

bfinzer (changes requested):

bfinzer (changes requested):
There are two important bugs related to positioning of measure labels when they are showing

v3/src/components/graph/adornments/univariate-measures/normal-curve/normal-curve-adornment-component.tsx
kswenson: Ah, I misunderstood the intent. I've restored the default vertical label offsets in another commit which centralizes the computation in the `useAdornmentAttributes()` hook to avoid redundancy.

kswenson:
> There are two important bugs related to positioning of measure labels when they are showing.

I've added a couple of commits that fix the default vertical positioning of labels and the label drag behavior.

v3/src/components/graph/adornments/univariate-measures/univariate-measure-adornment-helper.ts
```diff
@@ -249,30 +261,40 @@
...
+    const labelBounds = labelElt.getBoundingClientRect()
+    const labelWidth = labelBounds.width
+    const labelHeight = labelBounds.height
+    const parentBounds = parentElt.getBoundingClientRect()
+    const parentWidth = parentBounds.width
+    const parentHeight = parentBounds.height
+    const left = event.x - labelWidth / 2
+    const top = event.y - labelHeight / 2
+    const leftPct = left / parentWidth
+    const topPct = top / parentHeight
```
bfinzer, line 278: These locals would now be better named leftProp and topProp

bfinzer (approved):
👍🏻LGTM
Just one trivial suggestion

PR #1782, opened 2025-02-03
legend type based on the data configuration

scytacki:
The user can set an attribute type in the data configuration of the graph or map. The legend was ignoring this and just showing the type from the table.

Additionally the component wasn't observing the attribute type. So if the user changed the type in the table and they hadn't overridden it with a setting in the graph or map it wouldn't update.


v3/src/components/data-display/components/legend/legend.tsx
```diff
@@ -7,6 +7,7 @@ import {CategoricalLegend} from "./categorical-legend"
 import { ColorLegend } from "./color-legend"
 import { IBaseLegendProps } from "./legend-common"
 import {NumericLegend} from "./numeric-legend"
+import { observer } from "mobx-react-lite"
```
kswenson, line 10: Move import ⬆️

kswenson (approved):
👍 LGTM

PR #1783, opened 2025-02-04
Changes to support a size legend

scytacki:
The legendComponentManager is added so the component can be chosen based on more criteria than just the type of the attribute.


kswenson (approved):
👍 Seems fine as far as it goes, but is it sufficient for your needs? Seems like you could make it easier to work with by moving the legend manager into its own module (for easier mocking) and perhaps adding a `setLegendManager()` function (or something like that) to make it easier for clients to override.

scytacki:
Yes those would be good changes, but since you aren't requiring them, I'll leave it as is for now.  ;)
When we come back to this work I'll try to make them. 

Also when we come back this and try to bring in the size legends into CODAP itself, there might be some further refactoring which would overlap with this. Currently a custom "legend" consists of both the component below the graph as well as methods in the data configuration which are then used by the rendering within the visualization. In both places there are basically switch which determine how different attribute types should be handled. It might be useful to set this up so these methods on the data configuration are encapsulated with the legend. I'm not sure how to do that yet, but I think I could find a way.

PR #1786, opened 2025-02-05
feat: import/export regression line labels from/to v2

kswenson:
This turned into more of an odyssey than expected. The existing LSRL code treated label positions as absolute coordinates rather than proportions, similar to the measure labels handled earlier. Unlike measure labels, however, LSRL labels split on legend categories. The existing LSRL code handled this with arrays, which embedded a problematic dependency on category order. The existing LSRL code also failed to include units in its regression equations. In addition, v2 made things more complicated by 1) storing label positions as relative position of the center of the label (rather than top-left as was true for measures), which embeds a dependency on label size, which means you have to have a rendering engine to determine the placement; and 2) having a bug in which the wrong height is used when performing the aforementioned label placement computations.

The result is that this PR:
- fixes import of LSRL label positions
- tracks LSRL label positions imported from v2 for adjustment
- changes internal representation of LSRL label instances to be a map keyed by category
- changes rendering of LSRL label positions to be proportional
- includes units in LSRL equation strings when available
- exports LSRL label positions correctly
- reverses (approximately) the v2 height computation bug
- factors out some redundancy in the v2 adornment importer's `instanceKeysForAdornments()` function
- restores the behavior of the `DataConfigurationModel.categoryArrayForAttrRole()` method that it returns the client-provided empty array value when there are no categories (lost in #1707)

### DIFF
--- a/v3/src/components/data-display/models/data-configuration-model.ts
+++ b/v3/src/components/data-display/models/data-configuration-model.ts
@@ -347,10 +347,9 @@ export const DataConfigurationModel = types
       calculate: (role: AttrRole, emptyCategoryArray = [kMain]) => {
         const valuesSet = new Set(self.valuesForAttrRole(role)),
           categoryLimitForRole = self.numberOfCategoriesLimitByRole.get(role)
+        if (valuesSet.size === 0) return emptyCategoryArray
+
         let resultArray: string[] = []
-        if (valuesSet.size === 0) {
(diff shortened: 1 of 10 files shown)


bfinzer (approved):
👍🏻LGTM
What a great piece of work with worthwhile improvements! Sorry that V2 made things difficult.

PR #1788, opened 2025-02-05
chore: update chakra-ui

kswenson:
- update Chakra UI to address performance regression
- adjust `import/no-cycle` configuration to reduce lint times
- npm audit fix


kswenson:
This PR with Chakra UI upgrade to 2.10.5 was deemed not as performant as the downgraded Chakra UI in #1789, so we're closing this PR and merging that one instead.

PR #1790, opened 2025-02-06
feat: table does not become selected on attribute drag

kswenson:
[[PT-188514669]](https://www.pivotaltracker.com/story/show/188514669)

My first attempt at this was purely focused on not selecting the table when an attribute drag is initiated in the table. I maintain that this worked as I expected.

@scytacki correctly pointed out that when the table and graph are overlapping, there is an additional problem in that the table handles the _drop_ even when it's behind the graph, and the act of handling the drop results in the table becoming selected. This turns out to be a more fundamental problem with our collision-detection code. I have now added a commit which sorts the available drop targets by the z-index of their corresponding tile and finds the appropriate handler for the most salient collision, rather than finding the first matching collision for the earliest-registered handler.

### DIFF
--- a/v3/src/components/case-table/case-table-component.test.tsx
+++ b/v3/src/components/case-table/case-table-component.test.tsx
@@ -25,29 +25,44 @@ const UseKeyStatesWrapper = () => {
 describe("Case Table", () => {
   let broker: DataBroker
   let tile: ITileModel
+
+  const tileSelection: ITileSelection = {
+    isTileSelected() {
+      return false
+    },
+    selectTile() {
+    },
+    addFocusFilter() {
+      return () => null
+    }
+  }
+
   beforeEach(() => {
     broker = new DataBroker()
     tile = TileModel.create({ content: getSnapshot(CaseTableModel.create()) })
   })
 
   it("renders nothing with no broker", () => {
-    render(<DndContext><CaseTableComponent tile={tile}/></DndContext>)
+    render(
+      <DndContext>
+        <TileSelectionContext.Provider value={tileSelection}>
+          <CaseTableComponent tile={tile}/>
+        </TileSelectionContext.Provider>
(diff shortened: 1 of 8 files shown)


kswenson:
> When I tested with this link it didn't fix the problem as far as I understand it: https://codap3.concord.org/branch/no-table-select-on-drag/#file=examples:Roller%20Coasters

Works as expected for me -- initiating a drag from the column headers when the table is not frontmost doesn't bring the table to the front. Simply clicking on the attribute headers does still select the table.

> Also, I'd recommend changing the logic of filter to match that of Array.filter, so a true response means to keep the item and a false response means to filter it out. I've never been a fan of that, but since it is such a core behavior of JS it seems like the behavior should be consistent. But perhaps there are other functions that use the verb "filter" in the correct way that I'm forgetting.

I specifically want disabling the tile selection to require an affirmative response from the function, i.e. an empty function (or falling off the end of a function) doesn't change the tile selection behavior. I didn't think that the naming would be confusing, but we could call it something like `addSelectionDisableFn` or something like that. 🤷

scytacki:
So I start like this:
image
Then when I mouse down on the Max Height attribute on the table, it doesn't bring the table to foreground (as you said), but then when I drop the attribute on the graph the table does go to the foreground. And also the attribute ends up being dropped on the table instead of graph so the now the table as a different attribute order. So after the drag and drop it looks like this:
image

For a more nuanced case, if I start like this:
image
Then if I drag and drop the attribute to the area of the graph which doesn't intersect the table it works as I would expect. But if I drop the attribute on the area of the graph that intersects the table the attribute just moves in the table and the table goes to the foreground.

"disable" sounds good, or another option would be "ignore".

bfinzer (approved):
👍🏻LGTM
Subtle stuff! Nice work!

PR #1791, opened 2025-02-06
[#188862999] - Back ticks to attributes names with non-characters added via Insert Value button

eireland:
Adds back ticks to attribute names when attributes are added via Insert Value button


kswenson (approved):
👍 LGTM

PR #1792, opened 2025-02-06
188463252 checkbox case value

eireland:
User should be able to create an attribute of checkbox type.
When an attribute is specified as checkbox, any non-empty cells should show a checkbox input in the inderteminate state (dash line). Any cells that already have data should retain their data.
User should be able to click on a checkbox to change its state. When user clicks on an indeterminate checkbox, checkbox should show checked. Thereafter, clicks just toggle the input from checked to unchecked.

Checkbox tooltip should show `true | false`. When checkbox is indeterminate, tooltip should not come up.
Input row does not show checkbox, but entering `true` or `false` into the cell will show the checkbox in the appropriate state once input row is entered.

All pointer events are disabled for cell inputs in a formula column.

When user creates a formula on an attribute that is typed checkbox, checkboxes are at 50% opacity.

Demo at https://codap3.concord.org/branch/checkbox-case-value/


bfinzer:
Bug: Checkbox attribute cannot be plotted. OK with me if you want to postpone fixing this and create a PT story for it.

bfinzer:
In V2 I can double-click on a checkbox cell and edit it. But I don't seem to be able to do that here. I think we should conform to V2 for this.

v3/src/components/case-table/use-columns.tsx
```diff
@@ -56,10 +56,12 @@ export const useColumns = ({ data, indexColumn }: IUseColumnsProps) => {
                                         {"formula-column": hasFormula, "multi-line": rowHeight > kDefaultRowHeight}),
                 renderCell: AttributeValueCell,
                 editable: row => isCaseEditable(data, row.__id__),
-                renderEditCell: isEditable
+                renderEditCell: isEditable && userType !== "checkbox"
```
bfinzer, line 59: This must be what to change to allow editability?

bfinzer (approved):
👍🏻LGTM
Looks good except for graphing problem which could become a story. and editability which looks like a simple fix.

v3/src/components/case-table/checkbox-cell.tsx
```diff
@@ -0,0 +1,49 @@
...
+export default function CheckboxCell ({ rowId, attrId }: ICheckboxCellProps) {
+  const data = useDataSetContext()
+  // We need checkRef to show indeterminate state
+  const checkRef = useRef<HTMLInputElement>(null)
+  const cellValue = data?.getValue(rowId, attrId)
+
+  useEffect(() => {
+    if (checkRef.current) {
+      const isBool = isBoolean(cellValue)
+      const isTrue = typeof cellValue === "string" && cellValue.toLowerCase() === "true"
```
kswenson, line 27:
```suggestion
      const isTrue = cellValue === true || (typeof cellValue === "string" && cellValue.toLowerCase() === "true")
```
`data?.getValue()` can return `boolean` values.

v3/src/components/case-table/checkbox-cell.tsx
kswenson, line 1: Since this file is used by the case table and case card, it should be in `case-tile-common`.

v3/src/components/case-table/checkbox-cell.tsx
```diff
@@ -0,0 +1,49 @@
...
+  const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    const newValue = e.target.checked
+    data?.applyModelChange(() => {
+      data.setCaseValues([{ __id__: rowId, [attrId]: newValue }])
+    }, {
+      log: `update checkbox state: ${attrId} to ${newValue ? "checked" : "unchecked"}`
+    })
+  }
+  // title is used to show the value of the cell when hovering over the checkbox
+  // When checkbox is in the indeterminate state, we want the tooltip to show "undefined"
```
kswenson, line 42:
```suggestion
```
I don't think that's true and it's not how v2 behaves.

v3/src/components/case-table/checkbox-cell.tsx
```diff
@@ -0,0 +1,49 @@
...
+  return (
+    <span className="cell-checkbox">
+      <input type="checkbox" ref={checkRef}  onChange={handleChange}
+              title={String(cellValue) === "" ? "undefined" : String(cellValue)}/>
```
kswenson, line 46:
```suggestion
      <input type="checkbox" ref={checkRef} onChange={handleChange} title={String(cellValue)}/>
```

v3/src/components/case-table/use-columns.tsx
kswenson, line 1: Given @bfinzer's comment, you can simply revert all changes to this file.

v3/src/components/case-table/checkbox-cell.tsx
```diff
@@ -0,0 +1,49 @@
...
+  }
+  return v === undefined || ["", "false", "true", false, true, null].includes(v as string | boolean | null)
+}
+
+interface ICheckboxCellProps {
+  rowId: string
+  attrId: string
+}
+
+export default function CheckboxCell ({ rowId, attrId }: ICheckboxCellProps) {
```
kswenson, line 18:
```suggestion
export function CheckboxCell ({ rowId, attrId }: ICheckboxCellProps) {
```
We've been avoiding default imports lately in favor of named imports.

v3/src/components/case-table/checkbox-cell.tsx
```diff
@@ -0,0 +1,49 @@
...
+  let v = value
+  if (typeof v === "string") {
+    v = v.toLowerCase()
+  }
+  return v === undefined || ["", "false", "true", false, true, null].includes(v as string | boolean | null)
+}
+
+interface ICheckboxCellProps {
+  rowId: string
+  attrId: string
```
kswenson, line 15:
```suggestion
  attrId: string
  caseId: string
```
`rowId` doesn't make sense outside the context of the case table.

kswenson (changes requested):
👍 Generally looks good -- In addition to a number of suggestions, most notably moving `checkbox-cell.ts` to the `case-tile-common` folder, I added a commit which changes the signature of the `renderAttributeValue()` function. Make sure you pull my changes before merging from `main` or otherwise potentially losing them.

v3/src/components/case-table/checkbox-cell.tsx
eireland: V2 tooltip shows "undefined".
image

v3/src/components/case-table/checkbox-cell.tsx
eireland: Without this, tooltip does not show `undefined` when checkbox is in indeterminate state.

kswenson (approved):
👍 LGTM

PR #1793, opened 2025-02-07
feat: v2 import/export of movable line labels

kswenson:
The movable line adornment uses the same label management system as the LSRL line, so functionally this was mostly about sharing the implementation between the two. Toward that end, there's a little bit of refactoring to separate the `LineLabelInstance` model into its own module for easier sharing. That then led me to a broader refactoring of the unwieldy number of files in the `adornments` folder into a few sub-folders, and since I was updating a number of imports to address the path changes I took the opportunity to sort a number of the imports. The result is that the functional part of this PR is limited to the movable line files and a few files used by them, but the number of files touched by the refactor is much larger. These changes can be skimmed quickly on review. Finally, the movable line equation building code ignored units (just as the LSRL equation code had) and so units support was added there as well. There remains some redundant code between the LSRL and movable line equation building and rendering/components, but I refrained from attempting any further refactoring.


bfinzer:
Two problems you may want to defer to new stories.
1. Undo of showing movable line generates 11 mobx-state-tree errors.
2. After moving a movable line, there is no action-specific undo string on hover over Undo button.

bfinzer (approved):
👍🏻LGTM
Just the two problems noted in comment.

kswenson:
> Two problems you may want to defer to new stories.
> 
> 1. Undo of showing movable line generates 11 mobx-state-tree errors.
> 2. After moving a movable line, there is no action-specific undo string on hover over Undo button.

I created [[PT-188872789]](https://www.pivotaltracker.com/story/show/188872789) for these issues.

bfinzer:
Hi Kirk,

I couldn’t recreate the mst errors on main either, so I’m going to edit the story to remove that.

Bill

PR #1794, opened 2025-02-08
fix: create a copy of document with boundaries

kswenson:
[[PT-188781023]](https://www.pivotaltracker.com/story/show/188781023)

The problem was that in the copy, formulas were being compiled before the boundary data was loaded. With this PR, we add a `boundariesLoaded` property to the metadata that is used by the formula manager. This property is observed, so when the boundaries are loaded formulas are recompiled.


bfinzer (approved):
👍🏻LGTM

PR #1796, opened 2025-02-10
add check for point compression interaction with categorical plots

nstclair-cc:
[PT-#188713862](https://www.pivotaltracker.com/story/show/188713862)

This is a simple PR that enables a Cypress check for checking regression with graph point compression. It was built as part of the PixiJS test suite and unblocked by PT-[188601933](https://www.pivotaltracker.com/story/show/188601933)


kswenson (approved):
👍 LGTM

PR #1797, opened 2025-02-10
add option to use quantized choropleth legend

scytacki:
many of the changes are just renaming methods from quantile to numericColor, choropleth, or bin

To support reactions the colorHash method was replaced with a legendColorDomain function. I'm not sure that is the best name. `legendColorDomain` has the same downside as colorHash:  changing a single category color will cause all points to be re-rendered.

Additionally to simplify the monitoring of category colors the hashing of colors in colorHash is replaced with a colorMap property. This colorMap is then used to speed up the the access of each category color. With the colorHash approach MobX was monitoring every category anyway, so it makes sense to expose colorMap. 

Even though "color" type attributes are not correctly monitored by legendColorDomain the plot does re-render the points when a color is changed. So there must be some other reaction monitoring color values somehow.


v3/src/components/data-display/models/data-configuration-model.ts
```diff
@@ -462,15 +464,15 @@ export const DataConfigurationModel = types
     get caseDataHash() {
       return hashStringSets(self.filteredCases.map(cases => cases.caseIds))
     },
-    get quantileScaleColors() {
-      return getQuantileScale(
+    get choroplethColors() {
+      return getCholorplethColors(
```
kswenson, line 468:
```suggestion
      return getChoroplethColors(
```

v3/src/components/data-display/models/data-configuration-model.ts
```diff
@@ -21,12 +21,14 @@ import {
   kDefaultHighAttributeColor, kDefaultLowAttributeColor
 } from "../../../models/shared/shared-case-metadata-constants"
 import {hashStringSets, typedId, uniqueId} from "../../../utilities/js-utils"
-import {getQuantileScale, missingColor, parseColor} from "../../../utilities/color-utils"
+import {getCholorplethColors, missingColor, parseColor} from "../../../utilities/color-utils"
```
kswenson, line 24:
```suggestion
import {getChoroplethColors, missingColor, parseColor} from "../../../utilities/color-utils"
```

v3/src/utilities/color-utils.ts
```diff
@@ -118,7 +118,7 @@ export function interpolateColors(color1: string, color2: string, percentage: nu
 }
 
 // Returns an array of five colors transitioning between color1 and color2
-export function getQuantileScale(color1: string, color2: string) {
+export function getCholorplethColors(color1: string, color2: string) {
```
kswenson, line 121:
```suggestion
export function getChoroplethColors(color1: string, color2: string) {
```

kswenson (approved):
👍 LGTM -- adding @bfinzer as a reviewer since it impacts graph code API.

bfinzer (approved):
👍🏻 LGTM

PR #1801, opened 2025-02-11
Adds "index" to index column header

eireland:


kswenson (approved):
I added a commit which:
- moves the fix to the `AttributeHeader` so that attribute dragging isn't affected
- uses a localizable string for the label

PR #1802, opened 2025-02-11
188616847 stroke point plot background color import/export V2

eireland:
Adds import from V2 and export to V2 for graph:
- point color and transparency 
- stroke color and transparency
- plot background color and transparency


v3/src/components/graph/v2-graph-exporter.ts
```diff
@@ -230,13 +231,34 @@ function getPlotModels(graph: IGraphContentModel): Partial<ICodapV2GraphStorage>
...
+    pointColor: removeTransparencyFromColor(graph.pointDescription.pointColor),
+    transparency: getTransparency(graph.pointDescription.pointColor, "point"),
+    strokeColor: graph.pointDescription.pointStrokeSameAsFill
+                    ? "white" // v2 uses white for stroke when stroke is same as fill
+                    : strokeColorStr(graph.pointDescription.pointStrokeColor),
+    strokeTransparency: graph.pointDescription.pointStrokeSameAsFill
+                          ? 0.4 : getTransparency(graph.pointDescription.pointStrokeColor, "stroke"),
+    pointSizeMultiplier: graph.pointDescription.pointSizeMultiplier,
+    strokeSameAsFill: graph.pointDescription.pointStrokeSameAsFill,
+    plotBackgroundColor: graph.plotBackgroundColor === "#FFFFFF"
```
kswenson, line 264: Do we need to worry about other variations (e.g. `#ffffff` or `white`?) or is this string canonicalized to this format?

v3/src/components/graph/v2-graph-exporter.ts
```diff
@@ -230,13 +232,39 @@ function getPlotModels(graph: IGraphContentModel): Partial<ICodapV2GraphStorage>
   return storage
 }
 
+const getTransparency = (color: string, type: "point" | "stroke") => {
+  const rgbaColor = colord(color).toRgb()
+  return rgbaColor.a ?? (type === "point" ? 0.84 : 1)
```
kswenson, line 237: I moved this function here because it contains graph-specific logic rather than something appropriate to all clients of `color-utils`. That said, in practice it doesn't matter because `toRgb()` is documented to always return an alpha, so the `?? (type === "point" ? 0.84 : 1)` will never have any effect. If there are some circumstances in which `0.84` should be returned, then the logic will need to be clarified.

kswenson (approved):
👍 LGTM -- I added a commit which tweaked some of the types and moved some of the graph-specific logic from `color-utils` to `v2-graph-exporter`. Note the remaining comments/questions.

eireland:
@kswenson I tweaked the code based on your comments.

PR #1804, opened 2025-02-12
feat: implement `notify` method of `formulaEngine` handler (evalExpression)

kswenson:
[[PT-188879551]](https://www.pivotaltracker.com/story/show/188879551)

### DIFF
--- a/v3/src/data-interactive/handlers/formula-engine-handler.ts
+++ b/v3/src/data-interactive/handlers/formula-engine-handler.ts
@@ -1,9 +1,53 @@
+import { EvalFunction } from "mathjs"
+import { math } from "../../models/formula/functions/math"
+import { t } from "../../utilities/translation/translate"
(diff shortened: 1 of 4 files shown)


v3/src/data-interactive/handlers/formula-engine-handler.ts
```diff
@@ -1,9 +1,53 @@
...
+          ...record
+        }
+        try {
+          return compiledFormula.evaluate(scope)
+        }
+        catch (e) {
+          return String(e)
+        }
+      }
+      return t("V3.DI.Error.invalidEvaluationRecord")
```
tealefristoe, line 47: This will add the error message to the list of results rather than invalidating the whole API call. That's probably intended, but just pointing it out in case.

tealefristoe (approved):
Looks great 👍

PR #1805, opened 2025-02-12
fix: movable line undo/redo issues

kswenson:
[[PT-188872789]](https://www.pivotaltracker.com/story/show/188872789)

The MST warnings were fixed by using `mstAutorun` instead of `autorun`. The undo/redo strings were just a matter of using `applyModelChange()` and providing undo/redo strings. Curiously, adjusting the movable line is not undoable in v2, so I had to add a new pair of localizable undo/redo strings.


bfinzer (approved):
👍🏻LGTM

PR #1806, opened 2025-02-12
feat: implement first() and last() semi-aggregate functions

kswenson:
[[PT-188810519]](https://www.pivotaltracker.com/story/show/188810519)

### DIFF
--- a/v3/src/models/formula/functions/semi-aggregate-functions.ts
+++ b/v3/src/models/formula/functions/semi-aggregate-functions.ts
@@ -1,8 +1,77 @@
 import { MathNode } from "mathjs"
-import { CurrentScope, FValue } from "../formula-types"
+import { CODAPMathjsFunctionRegistry, CurrentScope, FValue } from "../formula-types"
 import { UNDEF_RESULT, evaluateNode, getRootScope, isValueTruthy } from "./function-utils"
 
-export const semiAggregateFunctions = {
+export const semiAggregateFunctions: CODAPMathjsFunctionRegistry = {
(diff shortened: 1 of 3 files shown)


v3/src/models/formula/functions/semi-aggregate-functions.ts
```diff
@@ -1,8 +1,77 @@
...
+      const [ expression, filter ] = args
+
+      const scope = getRootScope(currentScope)
+      const caseCount = scope.getNumberOfCases()
+      const caseGroupId = scope.getCaseGroupId()
+      const cacheKey = `last(${args.toString()})-${caseGroupId}`
+      let result = scope.getCached<Maybe<FValue>>(cacheKey)
+
+      if (result == null) {
+        // loop until the end, tracking the last valid result
```
scytacki, line 55: Could you loop backwards to make this more efficient? Or perhaps there are side effects that happen during the looping that need to be preserved?

scytacki (approved):
Looks good to me, just a question about iterating backwards.

v3/src/models/formula/functions/semi-aggregate-functions.ts
kswenson: I believe the way `prev()` and `next()` are implemented requires that the iteration be forward, so that, for instance, `first(next(LifeSpan))` would potentially be affected.

scytacki: Ah makes sense.

PR #1807, opened 2025-02-12
[#188809846] Points plot in {0,0} when Other is a category

eireland:
Fixes bug where points do not have a valid position when Other is a acategory.

The bug would only manifest when secondary attribute has the Other category.


kswenson (approved):
👍 LGTM

PR #1808, opened 2025-02-13
Prevents selection of row when user clicks on checkbox

eireland:
Previously, when user clicked on checkbox, the row would get selected.
This only allows case selection if user clicks on the cell outside the checkbox. If checkbox is clicked, it only changes the state of the checkbox and not the state of case selection.


kswenson (approved):
👍 LGTM

PR #1810, opened 2025-02-13
fix flakey clicks

scytacki:
This fixes the flakey clicks that were first found on the graph inspector panel, but can happen anywhere inside of a component that has been dragged to a new position.

Additionally it:
- reduces the number of pointerdown listeners by disabling the useOutsidePointerDown hook when possible
- adds info the useOutsidePointerDown to make it easier to see which components have active listeners
- updates the useCallbackRef with the latest code from chakra-ui


kswenson (approved):
👍 Looks good -- nice work!

PR #1811, opened 2025-02-13
add eventModification debug flag

scytacki:


v3/src/lib/debug-event-modification.ts
```diff
@@ -0,0 +1,36 @@
+import { DEBUG_EVENT_MODIFICATION } from "./debug"
+
+const eventMethods = [
+  "preventDefault",
+  "stopPropagation",
+  "stopImmediatePropagation"
+] as const
+
+const elementMethods = [
+  "setPointerCapture"
```
kswenson, line 10:
```suggestion
  "setPointerCapture",
  "releasePointerCapture"
```
I think it would be useful to track both to make it clear whether the capture was released or not.

kswenson (approved):
👍 Looks good -- I suggest adding `releasePointerCapture` to the events tracked.

v3/src/lib/debug-event-modification.ts
scytacki: I'll add it, but I'd also note that it isn't necessary to call `releasePointerCapture`. According to https://developer.mozilla.org/en-US/docs/Web/API/Element/setPointerCapture the capture will be released automatically when the `pointerup` event is fired.

kswenson: Good to know.

PR #1815, opened 2025-02-16
Fix date axis scaling bug

bfinzer:
[#188819773] Feature: Graph axes and scaling are incorrect when switching from a date attribute to another date attribute

* The immediate problem is fixed by calling setAllowRangeToShrink for the date axis.
* But another problem is that attempting to rescale the date axis with the new date attribute wasn't working because the numericValuesForAttrRole factory method wasn't being invalidated and therefore the rescale was happening with the previous values.


kswenson (approved):

kswenson:
👍 LGTM

PR #1817, opened 2025-02-19
fix: first() and last() functions

kswenson:
In my initial implementation of `first()` and `last()` (#1806), I assumed that `first()` and `last()` were like their cousins `next()` and `prev()` in requiring a semi-aggregate implementation in which the looping and filtering of appropriate cases is handled within the function implementation. Upon further reflection, `first()` and `last()` are actually full aggregate functions, i.e. they're more like `mean()` and `median()` in returning the same value for every case in a group. 🤦 This means we can make use of aggregate function utilities like `cachedAggregateFnFactory` and `aggregateFnWithFilterFactory` which handle all of the looping and filtering for us and simplify the implementation dramatically. With this realization, I've also renamed the former `semi-aggregate-functions.ts` module to `local-lookup-functions.ts`, reflecting the fact that not all of the functions within it are semi-aggregate (and most people don't know what "semi-aggregate" means anyway 😉).

### DIFF
--- a/v3/src/models/formula/functions/aggregate-functions.ts
+++ b/v3/src/models/formula/functions/aggregate-functions.ts
@@ -92,7 +92,6 @@ export const aggregateFnWithFilterFactory = (fn: ClientFn) => {
       filterValues = [ filterValues ]
     }
     expressionValues = expressionValues.filter((v: FValue, i: number) =>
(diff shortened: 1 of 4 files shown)


bfinzer (approved):
👍🏻LGTM

PR #1798, opened 2025-02-11
fix: case table cell click editing behavior

kswenson:
[[PT-188624240]](https://www.pivotaltracker.com/story/show/188624240)

- single-click in a regular cell selects the row and does not initiate editing
- code to handle extending the case/row selection (e.g. shift-click) is unchanged
- double-click in a regular cell initiates editing of that cell
- single-click in the input row initiates editing immediately

This required a patch to RDG to support clearing the current cell selection programmatically. See [CC/RDG #4](https://github.com/concord-consortium/react-data-grid/pull/4) for details of the patch to RDG. In this PR, those changes appear as a patch to be applied to RDG by `patch-package`. With the patch in place, the CODAP changes are straightforward.

You will need to run `npm install` to update `react-data-grid` and have `patch-package` apply the patch and clear the webpack cache to make sure that webpack sees it. If you encounter any errors:
```shell
# rm -rf node_modules/react-data-grid # remove current `react-data-grid` to start with a fresh install
# npm install                         # install fresh copy of `react-data-grid` and apply the patch
# npm run start:clean                 # clear the webpack cache and start the webpack dev server
```

PR #1818, opened 2025-02-20
fix: case table cell selection/editing (take two)

kswenson:
[[PT-188892243](https://www.pivotaltracker.com/story/show/188892243)]

The crash was introduced in #1798 in an attempt to fix [[PT-188624240](https://www.pivotaltracker.com/story/show/188624240)].

There were two parts to [[PT-188624240](https://www.pivotaltracker.com/story/show/188624240)]:

>1. When in the input row, a single click on a cell should bring up the blinking cursor for the input component.
>2. When not in input row, a single cell selects the case and not the cell so that user cannot type into the cell. User would have to double click on the cell to edit the cell contents.

The first part is successfully fixed in #1798. The patch to RDG that introduces the crash was part of an attempt to fix the second point. Instead of attempting an alternative fix to the second point, I'm going to make the argument that the prior behavior is preferable to the requested behavior, even though that's how v2 behaved.

The default behavior of RDG when single-clicking a cell is to select the cell and to show a focus highlight on the cell to indicate that to the user. This is useful for accessibility -- it indicates, for instance, that if the return key is pressed, the selected cell will become editable. We add the behavior of selecting the row and thus showing it as highlighted as well.

If the user presses the up/down arrows at this point, the adjacent row is selected and the corresponding cell in the neighboring row is selected/focused and the focus highlight is shown on the new cell. Again, pressing the enter/return key initiates editing on the selected cell, and so the focus highlight is useful for accessibility reasons.

So to reiterate this state of affairs, RDG maintains a current/selected cell internally and this is indicated with a focus highlight. The selected/focused cell receives keyboard events, some of which, e.g. the enter/return key, will initiate editing. I believe this is correct/desired behavior and the focus highlight on the current/selected cell is therefore useful for accessibility.

RDG makes the further assumption, that any normal key (e.g. not cell navigation keys, etc.) should also initiate editing. This is what I attempted to prevent in #1798. But there's an argument that this is reasonable/correct behavior. If the user has selected a cell, and then started typing, it seems more reasonable to assume that they mean to type into the contents of that cell than to simply ignore their typing entirely.

So rather than attempting to intercept these keyboard events in order to ignore them, I propose to allow the RDG behavior of highlighting the selected cell and initiating editing when the user types into the selected cell.

### Notes

I've also added a few scripts to facilitate updating to the new patch:

- `npm run clean:patches`
  - Removes the patched `node_modules` libraries `@dnd-kit` and `react-data-grid` and runs `npm install`, which reinstalls the libraries and the patches.
- `npm run clean:all`
  - Runs `npm run clean:cache` and `npm run clean:patches`
- `npm run start:clean:all`
  - Runs `npm run clean:all` followed by `npm start`.


bfinzer:
I'm persuaded by your discussion that the RDG behavior is preferable to V2 behavior. And I've tried it out and I like it!

bfinzer (approved):
👍🏻 LGTM

PR #1819, opened 2025-02-20
Matches table collection colors to case card collection colors

eireland:


kswenson (approved):
👍 LGTM -- personally, I think it would make more sense to apply the colors child-most to parent-most rather than the way it's currently done. This way, when an attribute is dragged from a flat collection to create a parent collection the child collection changes color so the parent can adopt the previous color of the children, and this process is repeated for each collection created. I would prefer that the child-most collection always have a particular color and then parent collections get assigned colors as needed. But that's not what the design specified and this seems to correctly implement that design. I'm assigning @bfinzer for review as well so he can weigh in.

bfinzer:
I'm going to run Kirk's thoughts by Dan for his reaction.

bfinzer:
Under certain drag conditions I get this error:
> Warning: Cannot update a component (`CollectionTableSpacer`) while rendering a different component (`CollectionTableSpacer`). To locate the bad setState() call inside `CollectionTableSpacer`, follow the stack trace as described in https://reactjs.org/link/setstate-in-render 

Please pin down the conditions and prevent the warning.

bfinzer (changes requested):
See comment about warning

kswenson:
> Under certain drag conditions I get this error:
> 
> > Warning: Cannot update a component (`CollectionTableSpacer`) while rendering a different component (`CollectionTableSpacer`). To locate the bad setState() call inside `CollectionTableSpacer`, follow the stack trace as described in https://reactjs.org/link/setstate-in-render
> 
> Please pin down the conditions and prevent the warning.

I think this may be preexisting, i.e. not related to this PR.

bfinzer:
This morning I appear to be unable to reproduce the above error in either main or this branch. I'll review the code later today.

bfinzer (approved):
👍🏻 LGTM

PR #1820, opened 2025-02-20
Plot "0" in dot charts

eireland:
If attribute is categorical, treat numerical values as strings so `0` is plotted


kswenson (approved):
👍 LGTM -- adding the `categorical` case to the switch makes sense. I added a commit which also fixes the default empty test (I told you that's what it would be 😉) and adds a `TODO` to handle dates properly, since it's not obvious what the right answer is there and I don't want to hold up the PR to figure it out.

kswenson:
Merging to get it into the test/build pipeline.

PR #1823, opened 2025-02-23
Fix date-to-categorical axis bug

bfinzer:
[#188919673] Bug fix: Replacing date axis with categorical attribute doesn't work properly

* In the transition from *numeric* axis to categorical we were properly detecting the presence of the numeric axis elements and removing them. Do the same for date axis elements.


kswenson (approved):
👍 LGTM

PR #1824, opened 2025-02-23
Fix to categorical axis update bug

bfinzer:
[#188914180] Bug fix: Graph axis doesn't update when initial checkbox value changes

* The bug does not have to do with checkboxes in particular, just a failure to detect changes in the category set determining the categorical axis. By paying attention to the appropriate category set rather than to the multiScale for the axis, we properly detect and react to changes.
* We also take a chance that this change will obviate the need for the comparison of current values with previously saved values of the categories.

### DIFF
--- a/v3/src/components/axis/hooks/use-sub-axis.ts
+++ b/v3/src/components/axis/hooks/use-sub-axis.ts
@@ -317,14 +317,14 @@ export const useSubAxis = ({
   // Refresh when category set, if any, changes
   useEffect(function installCategorySetSync() {
     if (isCategorical) {
-      const disposer = reaction(() => {
+      const disposer = mstReaction(() => {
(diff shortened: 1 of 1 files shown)


v3/src/components/axis/hooks/use-sub-axis.ts
```diff
@@ -319,22 +318,16 @@ export const useSubAxis = ({
   useEffect(function installCategorySetSync() {
     if (isCategorical) {
       const disposer = reaction(() => {
```
kswenson, line 320:
```suggestion
      const disposer = mstReaction(() => {
```

kswenson (approved):
👍 LGTM -- but you should also make the `reaction` an `mstReaction` so that it will be disposed when the `dataConfig` is destroyed.

PR #1826, opened 2025-02-25
Sorts attribute categories numerically and alphabetically

eireland:


eireland:
Closed this branch because of the failing jest test. The current code sorts all categorical axes (including secondary axes) numerically and then alphabetically. V2 only sort the primary x & y axes.

PR #1828, opened 2025-02-26
CODAP-2 Initial splash screen

scytacki:
This adds a splash screen to the html file. When the page has loaded the splash screen is hidden.
This is different than v2. In v2 the splashscreen was rendered by the javascript code so it would take longer to show up. 

Technically this might be better called a loading screen. I'd think a splash screen would stick around for a little while even if the application loads quickly. In v2 the splash screen would be quickly hidden by the the choice of opening a document or creating a new one.


kswenson (approved):
👍 This seems fine for the Activity Player use case, which is the point of the story. It seems jarringly quick on directly loading CODAP, where I think it should linger for a reasonable splash screen interval (~5 sec?) but be dismissable by any user action before that. This would complicate the implementation, however, so I think it's fine to defer such niceties to a later story.

scytacki:
@kswenson I could delay the showing of it for 1 second, that should prevent the flicker effect, and I don't think it would be too hard to implement.

PR #1829, opened 2025-02-27
Codap 4 hierarchy outlines for tables

eireland:
Adds spline and case outline highlights when useer selects cases.


v3/src/components/case-table/case-table.scss
```diff
@@ -294,7 +294,28 @@ $table-body-font-size: 8pt;
         .codap-data-cell {
           background-color: var(--rdg-row-selected-background-color);
         }
+        &.has-selected-item.first-child {
```
bfinzer, line 297: There's probably a reason this and the following are repeated below starting on 308, but just in case it's an oversight, I note it.

v3/src/components/case-table/collection-table-spacer.tsx
```diff
@@ -9,13 +9,16 @@ import { measureText } from "../../hooks/use-measure-text"
...
 import { IDataSet } from "../../models/data/data-set"
+import { symParent } from "../../models/data/data-set-types"
 // import { getNumericCssVariable } from "../../utilities/css-utils"
 import { preventAttributeMove, preventCollectionReorg } from "../../utilities/plugin-utils"
 import { t } from "../../utilities/translation/translate"
 import { kInputRowKey } from "./case-table-types"
 import { CurvedSpline } from "./curved-spline"
 import { useCollectionTableModel } from "./use-collection-table-model"
 
+const kRelationSelectedStrokeColor = '#66afe9' // blue
```
bfinzer, line 20: Since this color also appears in the '.scss', is there a way to make a global out of it?

bfinzer (approved):
👍🏻LGTM
I made a couple of inconsequential comments. But OK with me as is.

kswenson:
@eireland My review will involve adding a commit. I recommend _not_ trying to address @bfinzer's comments until my commit is in, which may obviate the need for some of them.

v3/src/components/case-table/case-table.scss
kswenson: Resolved in my code review tweaks commit.

v3/src/components/case-table/collection-table-spacer.tsx
kswenson: Resolved in my code review tweaks commit.

kswenson (approved):
👍 Looks good -- I added a commit which fixed the behavior with discontiguous selection, cleaned up the logic in places, refactored the curved spline code, uses RDG's CSS variables rather than defining local constants, moved `isAnyChildSelected()` from the `DataSet` to a utility function and simplified its implementation, etc.

PR #1830, opened 2025-02-27
fix: restore of graph adornments

kswenson:
The bug was that on restore the `UnivariateMeasureAdornmentHelper` was caching `xScale` and `yScale` properties passed to it on construction, but these were preliminary scales that were later configured/replaced. The fix is simply to look them up when needed rather than caching them, guaranteeing that the current/correct scales are always used.


bfinzer (approved):
👍🏻LGTM

PR #1834, opened 2025-03-06
fix: positioning of dots/bars in dot/bar charts

kswenson:
[[CODAP-132](https://concord-consortium.atlassian.net/browse/CODAP-132)]

In PixiPoints, there is a notion of an `anchorPoint` which determines offsets used when rendering points or bars. The correct anchor point to use is plot-dependent, but wasn't always being reset correctly after switching plots.

[CODAP-132]: https://concord-consortium.atlassian.net/browse/CODAP-132?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bfinzer (approved):
👍🏻LGTM

PR #1835, opened 2025-03-06
Adds plugin notification when user changes case focused in case card

eireland:
Case card navigation did not send notifications to plugins. This caused issues for the MultiData plugin, which needed notifications to show the correct selected case.
This makes it that case card navigation triggers the necessary notifications.


kswenson (approved):
👍 LGTM

PR #1836, opened 2025-03-06
Fixes table display bug when collection has one attribute and it is hidden

eireland:
Fix is by showing the hidden attribute, which automatically makes copllection visible.

### DIFF
--- a/v3/src/components/case-table/use-columns.tsx
+++ b/v3/src/components/case-table/use-columns.tsx
@@ -33,10 +33,6 @@ export const useColumns = ({ data, indexColumn }: IUseColumnsProps) => {
       () => {
         const collection = data?.getCollection(collectionId)
         const attrs: IAttribute[] = collection ? getCollectionAttrs(collection, data) : []
-        // if attribute is the only attribute in the collection and it is hidden, make the attribute visible
-        if (attrs.length === 1 && caseMetadata?.isHidden(attrs[0].id)) {
-          caseMetadata?.setIsHidden(attrs[0].id, false)
-        }
         const visible: IAttribute[] = attrs.filter(attr => attr && !caseMetadata?.isHidden(attr.id))
         return visible.map(({ id, name, type, userType, isEditable, hasFormula, precision }) =>
                   ({ id, name, type, userType, isEditable, hasFormula, precision }))
--- a/v3/src/components/case-table/use-columns.tsx
+++ b/v3/src/components/case-table/use-columns.tsx
@@ -33,6 +33,10 @@ export const useColumns = ({ data, indexColumn }: IUseColumnsProps) => {
       () => {
         const collection = data?.getCollection(collectionId)
         const attrs: IAttribute[] = collection ? getCollectionAttrs(collection, data) : []
+        // if attribute is the only attribute in the collection and it is hidden, make the attribute visible
+        if (attrs.length === 1 && caseMetadata?.isHidden(attrs[0].id)) {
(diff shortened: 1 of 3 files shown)


v3/src/components/case-table/use-columns.tsx
```diff
@@ -33,6 +33,10 @@ export const useColumns = ({ data, indexColumn }: IUseColumnsProps) => {
       () => {
         const collection = data?.getCollection(collectionId)
         const attrs: IAttribute[] = collection ? getCollectionAttrs(collection, data) : []
+        // if attribute is the only attribute in the collection and it is hidden, make the attribute visible
+        if (attrs.length === 1 && caseMetadata?.isHidden(attrs[0].id)) {
+          caseMetadata?.setIsHidden(attrs[0].id, false)
+        }
```
kswenson, line 39: Couple of issues:
- The case table is not the right place to address an issue that doesn't require the case table. (Attributes can be hidden from other components, e.g. the case card.)
- The accessor function of a reaction is just for returning values to the action function, not for actually changing models.
- The number of attributes isn't important. The relevant test is whether all remaining attributes in a collection are hidden, in which case they should all be shown.

I suspect that a reaction that is installed in the `afterCreate()` of the `SharedCaseMetadata` would make the most sense.

kswenson (changes requested):
See comments below for details/suggestions.

v3/src/models/shared/shared-case-metadata.ts
```diff
@@ -194,6 +194,31 @@ export const SharedCaseMetadata = SharedModel
...
+        () => {
+          return {
+            hiddenAttrs: Array.from(self.hidden.keys()),
+            collections: self.data?.collections.map(collection => collection.attributes.map(attr => attr?.id)) || []
+          }
+        },
+        ({hiddenAttrs}) => {
+          self.data?.collections.forEach(collection => {
+            const attrs = collection.attributes
+            if (attrs && attrs.every(attr => attr && hiddenAttrs.includes(attr.id))) {
```
kswenson, line 210:
```suggestion
            if (attrs && attrs.every(attr => attr && !!self.hidden.get(attr.id))) {
```
We don't expect the list of hidden attributes to get particularly long, but as a general rule it's good to avoid linear search through an array and the corresponding O(n^2) behavior where possible, and in this case we've got ready access to a map of hidden attributes for efficient lookup. Even though we will no longer be using the `hiddenAttrs` value returned by the accessor function, it's still worth returning it so that changes will trigger the reaction.

v3/src/models/shared/shared-case-metadata.ts
```diff
@@ -194,6 +194,31 @@ export const SharedCaseMetadata = SharedModel
       return categorySet
     }
   }))
+  .actions(self => ({
+    afterCreate() {
+      // respond to change of collections by showing hidden attributes if all attributes are hidden
+      addDisposer(self, reaction(
```
kswenson, line 200: This should be an `mstReaction` on `self.data`, so that if the `DataSet` is destroyed before the `SharedCaseMetadata`, the reaction will be disposed of before firing after the `DataSet` is defunct.

kswenson (approved):
👍 LGTM -- just a couple of final suggested tweaks.

v3/src/models/shared/shared-case-metadata.ts
kswenson: I tried this locally and it's not as straightforward as I'd hoped, so let's skip it for now.

PR #1837, opened 2025-03-06
Bar chart & histogram bar behavior fixes

bfinzer:
[#CODAP-123] Bug fix: Bar chart selection misbehaves in Safari

* Working on this bug flushed out a number of problems in bar charts:
  * As the mouse moves over a bar, the data tip flashes, once for each portion of the bar corresponding to a case
  * In Safari, the hover tip was not showing and click did not select cases
  * The data tip of a bar chart for no legend was incorrect
* There were similar problems for histogram bars
--
## Fixes
* DataDisplayContentModel how has volatile properties `showDataTip` and `hideDateTip` so that these can be accessed when rendering bar cover rectangles for bar charts and histograms. The DataTip component, of which there is only one, sets these properties.
* If the browser is Safari, PisiPoints treats events specially for fused bars so that the different way Safari handles events in the presence of WebGl can be dealt with.

### DIFF
--- a/v3/src/components/data-display/components/data-tip.tsx
+++ b/v3/src/components/data-display/components/data-tip.tsx
@@ -1,11 +1,12 @@
 import React, { Fragment, useCallback, useEffect, useRef, useState } from "react"
 import * as PIXI from "pixi.js"
 import { computePosition, offset, useFloating } from "@floating-ui/react"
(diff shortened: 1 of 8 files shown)


kswenson (approved):
👍 Looks good -- I added a commit to tweak a couple of things and to add a TODO comment about how the Safari-specific fix could be made more robust in the future. There are also a couple of remaining small comments/suggestions.

PR #1840, opened 2025-03-07
Fix spurious circular reference error

bfinzer:
[#CODAP-53] Bug fix: Circular ref error

* The problem is that `getFormulaDependencies` in formula-dependency-utils.ts was not protecting against adding the same attribute more than once to the result when a formula has two occurrences of a given attribute. Because of this `isDependencyCyclePresent` was returning true when there actually was no cycle.


kswenson (approved):
👍 Sync review with @bfinzer.

PR #1841, opened 2025-03-07
Remove duplicate plugin on copy

eireland:
The bug: When creating a copy from a browser tab with a url that includes `di=` query param, the new document created opens a second tile with the plugin specified in the the query param.

This checks whether the plugin already exists in the document when it opens. If it does, then we ignore the `di` query param.

Note: It turns out the save and restore path in create a copy does not make use of the CODAP save and restore. CFM has its own methods for save and restore for create a copy.


kswenson (approved):
👍 LGTM

PR #1838, opened 2025-03-06
fix: better fix for PixiPoints cell masking

kswenson:
[[CODAP-129](https://concord-consortium.atlassian.net/browse/CODAP-129)]

Improves on fix in #1833 -- once more unto the breach.

[CODAP-129]: https://concord-consortium.atlassian.net/browse/CODAP-129?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1842, opened 2025-03-07
fix: cellKey indexing in GraphDataConfigurationModel

kswenson:
[[CODAP-129](https://concord-consortium.atlassian.net/browse/CODAP-129)]

In the presence of multiple categorical attributes on a graph, the `GraphDataConfigurationModel` is responsible for computing cell keys for each combination of values, and the PixiPoints class is responsible for assigning cell masks for each cell key. In #1833 and #1838 the order of assignment of cell masks was fixed to match the expected order of cell keys. With this PR, we fix an underlying bug in the assignment of cell keys by the `GraphDataConfigurationModel`, so that the two should be (finally) in sync.

[CODAP-129]: https://concord-consortium.atlassian.net/browse/CODAP-129?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bfinzer (approved):
👍🏻LGTM

PR #1845, opened 2025-03-07
Attempt to fix flaky cypress tests

eireland:
Attempted to fix flaky slider, table, and bivariate-adornment cypress tests.


nstclair-cc (approved):
lgtm! Thank you so much! 😊

PR #1846, opened 2025-03-08
Data tips respect date and numeric precision

bfinzer:
[#CODAP-20] Bug fix: Hover dates in graph component should match precision in table component

* Respecting date precision as described in the story was straightforward. But it presented the opportunity to do a bit more with numeric values.
  * Respect the numeric precision, if any, specified by the user.
  * Display units, if any, along with numeric values

### DIFF
--- a/v3/src/components/data-display/models/data-display-content-model.ts
+++ b/v3/src/components/data-display/models/data-display-content-model.ts
@@ -5,6 +5,9 @@
 import {comparer, reaction} from "mobx"
 import {addDisposer, Instance, types} from "mobx-state-tree"
 import { format } from "d3"
+import { formatDate } from "../../../utilities/date-utils"
+import { IValueType } from "../../../models/data/attribute-types"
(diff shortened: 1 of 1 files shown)


v3/src/components/data-display/models/data-display-content-model.ts
```diff
@@ -68,14 +71,39 @@ export const DataDisplayContentModel = TileContentModel
       return getTileContentInfo(self.type)?.getFormulaAdapters?.(self) ?? []
     },
     caseTipText(attributeIDs: string[], caseID: string, dataset?: IDataSet) {
+
+      const getValueToDisplay = (numValue:number | undefined, value: IValueType, attribute: IAttribute | undefined) => {
```
kswenson, line 75:
```suggestion
      const getValueToDisplay = (numValue: number | undefined, value: IValueType, attribute?: IAttribute) => {
```

kswenson (approved):
👍 LGTM -- just a couple minor formatting suggestions.

PR #1848, opened 2025-03-09
iPad drag and resize components fix

bfinzer:
[#CODAP-89] Bug fix: Component reposition and resize in iPadOS

* With some help from co-pilot we learn that adding `touch-action: none;` to component css governing the title bar, borders and corners serves to prevent touch actions from propagating to the browser, thus limiting their effect to the desired drag and resize behavior in CODAP.
* In order to be able to test using an iPad connected to one's development computer, we add `allowedHosts: 'all'` the webpack configuration `devserver:`

Note that it is also possible to test using xCode's **Simulator** running locally, and, for that, the `allowedHosts` change is not necessary.

### DIFF
--- a/v3/src/components/codap-component.scss
+++ b/v3/src/components/codap-component.scss
@@ -27,6 +27,7 @@ $corner-drag-size: calc($border-drag-width * 2);
   margin: 0;
   user-select: none;
   z-index: 11;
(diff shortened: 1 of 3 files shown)


kswenson (approved):
👍 LGTM

PR #1855, opened 2025-03-13
Shows month and day when multiple of value is 1 and unit is year

eireland:


eireland:
After playing around with the slider in this branch, @bfinzer decided he didn't like this feature. So closing this PR.

PR #1859, opened 2025-03-14
feat: case table relation lines have symmetric widths (updated design)

kswenson:
Updated design based on design meeting.


eireland (changes requested):
Current implementation only changes the thickness of the curved spline and not the table line. I'm assuming they want to extend the same thickness across the table cells? So you also have to change the logic in collection-table.tsx line 420.

kswenson:
> Current implementation only changes the thickness of the curved spline and not the table line. I'm assuming they want to extend the same thickness across the table cells? So you also have to change the logic in collection-table.tsx line 420.

No, the current decision was just to change the relation lines for now. We can discuss the table part of it in a separate conversation.

eireland (approved):
👍 LGTM

PR #1860, opened 2025-03-14
chore: update typescript

kswenson:
This PR updates all updatable dependencies and includes `npm audit fix` so we have zero reported vulnerabilities. The `dependencies-notes.md` document is also updated. Some notes of interest:
- Earlier attempts to update to TypeScript 5.6+ (5.8.2 is current) failed when performing the cypress build on GitHub Actions, but succeeded locally and when building for non-cypress targets on GitHub Actions. 🤷 With the help of ChatGPT it turned out to be a TypeScript module resolution issue which was fixed by adding explicit module resolution instructions to the `tsconfig.json` for some of the `@codemirror` libraries. Apparently, TS 5.6 
- The latest `@dnd-kit/core` (6.3.1) incorporates one of our bug-fixes that I submitted, so our patch is somewhat smaller than the previous one. Also, recent versions incorporate some changes to reduce the number of re-renders under some circumstances, so there may be some performance benefits as well.

**Note:** If you get build errors after `npm install`, you can use `npm run clean:patches` or `npm run clean:all` (which calls `clean:patches` and then `npm install`) to address any patch conflicts.


v3/src/models/formula/formula-mathjs-scope.ts
```diff
@@ -119,7 +119,7 @@ export class FormulaMathJsScope {
   //     based on: // https://github.com/josdejong/mathjs/blob/develop/examples/advanced/custom_scope_objects.js ---
 
   get(key: string): FValue {
-    return this.extraScope?.has(key) ? this.extraScope.get(key) : this.dataStorage[key]
+    return this.extraScope?.get(key) ?? this.dataStorage[key]
```
scytacki, line 122: Is it possible the extraScope is used to override dataStorage with an undefined or null value for a key? Looking at the types of where it is used in `PlottedFunctionFormulaAdapter` that doesn't seem possible. I'm just checking since that would no longer be possible after this change.

scytacki (approved):
Looks good, just a comment on a change to the possible usage for the formula scope.

v3/src/models/formula/formula-mathjs-scope.ts
kswenson: It's a good question, but `extraScope` is a `Map<string, FValue>`, where `FValue` is defined as `type FValue = string | number | boolean | Date | object` so unless our types are off, it can't be used to store `null` or `undefined`.

PR #1861, opened 2025-03-15
Codap 64 Add category set color map API request handler

eireland:
This adds a handler for colorMap API requests. 
Requests can be `get`, `create`, or `update`.


v3/src/data-interactive/handlers/color-map-handler.ts
```diff
@@ -0,0 +1,87 @@
...
+    const createColorMap = (_values && 'colorMap' in _values) ? _values.colorMap as DICategoryColorMap : undefined
+    if (!createColorMap) return errorResult("New color map not found")
+    const categorySet = createProvisionalCategorySet(dataContext, attributeId)
+    metadata.promoteProvisionalCategorySet(categorySet)
+    if (!categorySet) return errorResult("Category set not found")
+    Object.entries(createColorMap || {}).forEach(([category, color]) => {
+      categorySet.setColorForCategory(category, color ?? categoryColorMap[category] ?? "")
+    })
+    categorySet.storeAllCurrentColors()
+    metadata.promoteProvisionalCategorySet(categorySet)
```
tealefristoe, line 28: Does this need to happen twice?

v3/src/data-interactive/handlers/color-map-handler.ts
```diff
@@ -0,0 +1,87 @@
...
+    const categoryColorMap: Record<string, string | undefined> = categorySet?.colorMap || {}
```
tealefristoe, line 29: ⬆️ This should be moved up to before it's used in the `forEach` statement above. It's just confusing to define it down here after it's used.

v3/src/data-interactive/handlers/color-map-handler.ts
```diff
@@ -0,0 +1,87 @@
...
+    if (!categoryColorMap) return errorResult("Category color map not found")
```
tealefristoe, line 30: Would this ever happen? You define `categoryColorMap` immediately before this and I don't see how it could become falsy.

v3/src/data-interactive/handlers/color-map-handler.ts
```diff
@@ -0,0 +1,87 @@
...
+    if (!metadata) return errorResult("Metadata not found")
+    const attributeId = attribute.id
+    const categorySet = metadata.getCategorySet(attributeId)
+    const categoryColorMap: Record<string, string | undefined> = categorySet?.colorMap || {}
+    if (!categoryColorMap) return errorResult("Category color map not found")
+
+    return {
+      success: true,
+      values: categoryColorMap
+    }
```
tealefristoe, line 55: Except follow the usual pattern for error results like `attributeNotFoundResult`, don't use a hard coded string here.
```suggestion
    if (!categorySet?.colorMap) return errorResult("Category color map not found")

    return {
      success: true,
      values: categorySet.colorMap
    }
```

v3/src/data-interactive/handlers/color-map-handler.ts
```diff
@@ -0,0 +1,87 @@
...
+  },
+
+  update(resources: DIResources, _values?: DIValues & { metadata?: any }) {
+    const { dataContext, collection, attribute } = resources
+    if (!dataContext) return dataContextNotFoundResult
```
tealefristoe, line 60: You should do what you did at the top of `get` here. In other words, check for `dataContext`, `collection`, and `attribute` immediately and bail if any of them are missing. It's a lot easier to follow that way and avoids running unnecessary code when you already know the handler is going to fail.

v3/src/data-interactive/handlers/color-map-handler.ts
```diff
@@ -0,0 +1,87 @@
...
+    const metadata = getSharedCaseMetadataFromDataset(dataContext) ?? _values?.metadata
+    if (!metadata) return errorResult("Metadata not found")
+    if (!collection) return collectionNotFoundResult
+    const attributeId = attribute && typeof attribute === 'object' ? attribute.id : undefined
+    if (!attributeId) return attributeNotFoundResult
+    const newColorMap = (_values && 'colorMap' in _values) ? _values.colorMap as DICategoryColorMap : undefined
+    if (!newColorMap) return errorResult("New color map not found")
+    const categorySet = metadata.getCategorySet(attributeId)
+    if (!categorySet) return errorResult("Category set not found")
+      const categoryColorMap: Record<string, string | undefined> = categorySet?.colorMap || {}
```
tealefristoe, line 71:
```suggestion
    const categoryColorMap: Record<string, string | undefined> = categorySet?.colorMap || {}
```

v3/src/data-interactive/handlers/color-map-handler.ts
```diff
@@ -0,0 +1,87 @@
...
+import { registerDIHandler } from "../data-interactive-handler"
+import { DIHandler, DIResources, DIValues } from "../data-interactive-types"
+import {
+  attributeNotFoundResult, collectionNotFoundResult, dataContextNotFoundResult, errorResult,
+} from "./di-results"
+
+export const diCategoryColorMapHandler: DIHandler = {
+  create(resources: DIResources, _values?: DIValues & { metadata?: any }) {
+    const { dataContext, collection, attribute } = resources
+    if (!dataContext) return dataContextNotFoundResult
```
tealefristoe, line 13: You should check for `collection` and `attribute` here as well, like you do in `get`. There's no need to do anything else if we know the handler will fail. It's also easier to engineers to follow to group these things together.

v3/src/data-interactive/handlers/color-map-handler.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    const metadata = getSharedCaseMetadataFromDataset(dataContext) ?? _values?.metadata
+    if (!metadata) return errorResult("Metadata not found")
+    if (!collection) return collectionNotFoundResult
+    const attributeId = attribute && typeof attribute === 'object' ? attribute.id : undefined
+    if (!attributeId) return attributeNotFoundResult
+    const newColorMap = (_values && 'colorMap' in _values) ? _values.colorMap as DICategoryColorMap : undefined
+    if (!newColorMap) return errorResult("New color map not found")
```
tealefristoe, line 68: This is a very confusing error message. Imagine receiving this when you're trying to send an `update colorMap` request. There is no new color map, right? You're just setting values in an existing color map. Remember that these error messages are intended to help plugin writers understand what's going on when things fail.

v3/src/data-interactive/data-interactive-data-set-types.ts
```diff
@@ -114,3 +114,5 @@ export interface DIUpdateItemResult {
 export interface DIDeleteCollectionResult {
   collections?: number[]
 }
+
+export type DICategoryColorMap = Record<string, string | undefined>
```
tealefristoe, line 118: Does including `undefined` as a possible value make this type more useful? It seems to complicate the actual handler code, and I suspect it's not necessary.

v3/src/data-interactive/handlers/color-map-handler.ts
```diff
@@ -0,0 +1,87 @@
...
+    Object.entries(newColorMap || {}).forEach(([category, color]) => {
+      categorySet.setColorForCategory(category, color ?? categoryColorMap[category] ?? "")
+    })
+    categorySet.storeAllCurrentColors()
+    const updatedColorMap = categorySet.colorMap
+    if (!updatedColorMap) return errorResult("Updated color map not found")
+    return {
+      success: true,
+      values: updatedColorMap
+    }
```
tealefristoe, line 83: There are lots of similarities between this code and the code for `create`, so please apply relevant suggestions for that handler here as well.

tealefristoe (changes requested):
We talked separately about changing the test to remove jumping through so many hoops with `metadata`--hopefully that's working out alright. But I looked over the handler code and left a bunch of suggestions. There are probably more improvements/simplifications that can be made, but let's stick with these for a first round.

kswenson (changes requested):
Before getting to the code of this PR, I don't think we need new API or a new handler for the color map. The `colormap` is a property of the attribute that is already supported by the existing attribute API as indicated by the API documentation:
```json
{
  "action": "create",
  "resource": "dataContext[DataCard2].collection[Measurements].attribute",
  "values": [
    {
      "name": "sampleDate",
      "title": "date of sample",
      "type": "dateTime"
    },
    {
      "name": "Age",
      "title": "Age",
      "type": "numeric",
      "description": "Age of person in years",
      "precision": 0
    },
    {
      "name": "Height",
      "title": "Height",
      "type": "numeric",
      "description": "Height of person in inches",
      "precision": 1,
      "colormap": {
        "high-attribute-color": "#0000ff",
        "attribute-color": "#ccccff"
      }
    },
    {
      "name": "Flavor",
      "title": "Favorite ice cream flavor",
      "type": "categorical",
      "colormap": {
        "vanilla": "#f3e5ab",
        "chocolate": "#d2691e",
        "strawberry": "#fc5a8d"
      }
    }
  ]
}
```

I would think we should just support `colormap` as a property when handling update attribute requests. It doesn't really make sense to create/delete color maps independently, for instance. The one wrinkle with color maps that doesn't exist for simple properties like `precision` is that you may want to add new color assignments to the existing map or you may want to replace the entire color map with a new one. So maybe we need an additional parameter to specify that distinction.

Peeking into the code a little bit, I don't think you should need to deal with creating/promoting provisional category sets at all. That's an implementation detail that clients shouldn't be concerned with. You simply `getCategorySet` when you need it and use it normally and it should be created/promoted as needed. If the existing implementation doesn't work for this use case, then we should understand why and see if it can be rectified.

v3/src/data-interactive/data-interactive-data-set-types.ts
tealefristoe, line 118: Can you explain why you need the `undefined` here? It seems like it's unnecessary to me, and it would simplify your code elsewhere.

v3/src/data-interactive/handlers/di-handler-utils.ts
```diff
@@ -17,6 +17,12 @@ export function createAttribute(value: DIAttribute, dataContext: IDataSet, colle
     const attribute = dataContext.addAttribute(attributeSnapshot, { collection: collection?.id })
     if (value.formula) attribute.formula?.setDisplayExpression(value.formula)
     metadata?.setIsHidden(attribute.id, !!value.hidden)
+    if (value.colormap != null && metadata) {
+        const categorySet = metadata.getCategorySet(attribute.id)
+        Object.entries(value.colormap as DICategoryColorMap || {}).forEach(([category, color]) => {
+          categorySet?.setColorForCategory(category, color ?? (categorySet.colorMap[category] || ""))
```
tealefristoe, line 23: If `DICategoryColorMap` didn't have `undefined` as a possible value type, this could just be:
```suggestion
          categorySet?.setColorForCategory(category, color)
```

v3/src/data-interactive/handlers/di-handler-utils.ts
```diff
@@ -55,4 +61,13 @@ export function updateAttribute(attribute: IAttribute, value: DIAttribute, dataC
       metadata?.setIsHidden(attribute.id, value.hidden)
     }
   }
+  if (value?.colormap != null) {
+    if (dataContext) {
+      const metadata = getSharedCaseMetadataFromDataset(dataContext)
+      const categorySet = metadata?.getCategorySet(attribute.id)
+      Object.entries(value?.colormap as DICategoryColorMap || {}).forEach(([category, color]) => {
+        categorySet?.setColorForCategory(category, color || categorySet.colorMap[category] || "")
```
tealefristoe, line 69:
```suggestion
        categorySet?.setColorForCategory(category, color)
```
Assuming you can remove `undefined` as a possible type from `DICategoryColorMap` values.

v3/src/data-interactive/handlers/di-handler-utils.ts
```diff
@@ -55,4 +61,13 @@ export function updateAttribute(attribute: IAttribute, value: DIAttribute, dataC
...
+      })
+    }
+  }
```
tealefristoe, line 72: You should reformat this block and the one above it (`value.hidden`) so they are both under a single `if (dataContext)` conditional, rather than duplicating a bunch of code.

v3/src/data-interactive/handlers/attribute-handler.ts
```diff
@@ -33,7 +34,9 @@ export const diAttributeHandler: DIHandler = {
       return { success: true }
     }).filter(error => !error.success)
     if (attributeErrors.length > 0) return attributeErrors[0]
-
+    if (!metadata && !Array.isArray(values) && 'colormap' in values && values.colormap) {
```
tealefristoe, line 37: This fail case should occur if _any_ attribute includes `colormap` and we can't find `metadata`.
```suggestion
    if (attributeValues.some(attribute => !!attribute.colormap) && !metadata) {
```

v3/src/data-interactive/handlers/attribute-handler.ts
```diff
@@ -33,7 +34,9 @@ export const diAttributeHandler: DIHandler = {
...
+      return sharedCaseMetadaNotFoundResult
```
tealefristoe, line 38: I'm afraid this error message is going to be confusing and unhelpful. Hopefully it will never occur because the `metadata` is actually always found. But if it is missing, there's no indication that it's the color map portion of the request that requires metadata. This message should say something like "SharedCaseMetadata not found but required for colormaps". Even that is confusing--will plugin developers have any idea what SharedCaseMetadata is?

tealefristoe (changes requested):
Looks close, but there's one error I found and some changes you could make to simplify the code. Please do let me know if I'm mistaken about the `DICategoryColorMap` definition, but I really see no reason why the value should ever be `undefined`.

v3/src/utilities/translation/lang/en-US.json5
```diff
@@ -136,6 +136,7 @@
     "V3.DI.Error.interactiveFrameNotFound": "Interactive Frame not found",
     "V3.DI.Error.itemNotFound": "Item not found",
     "V3.DI.Error.notFound": "Not found",
+    "V3.DI.Error.sharedCaseMetadatanotFound": "SharedCaseMetadata not found",
```
kswenson, line 139: As @tealefristoe mentioned elsewhere, this could be more helpful. How about something like:
```suggestion
    "V3.DI.Error.sharedCaseMetadataNotFound": "Internal error prevented color map access",
```

v3/src/data-interactive/handlers/di-results.ts
```diff
@@ -9,6 +9,7 @@ export const couldNotParseQueryResult = errorResult(t("V3.DI.Error.couldNotParse
 export const dataContextNotFoundResult = errorResult(t("V3.DI.Error.dataContextNotFound"))
 export const dataDisplayNotFoundResult = errorResult(t("V3.DI.Error.dataDisplayNotFound"))
 export const itemNotFoundResult = errorResult(t("V3.DI.Error.itemNotFound"))
+export const sharedCaseMetadaNotFoundResult = errorResult(t("V3.DI.Error.sharedCaseMetadatanotFound"))
```
kswenson, line 12:
```suggestion
export const sharedCaseMetadaNotFoundResult = errorResult(t("V3.DI.Error.sharedCaseMetadataNotFound"))
```

v3/src/data-interactive/data-interactive-data-set-types.ts
kswenson: I think the API question here is what is the mechanism for removing a color assignment from the color map? I think you're right that `undefined` is unnecessary, but `null` may be useful for indicating that a color should be removed from the color map (and thus returned to its default color). Alternatively, we could say that `""` is used for that purpose. As I think about it, using `""` seems reasonable from an API perspective and simpler to implement so I'd second @tealefristoe's suggestion of simply removing `undefined` from the definition.

v3/src/data-interactive/data-interactive-types.ts
```diff
@@ -94,14 +95,14 @@ export interface DIResources {
 // types for values accepted as inputs by the API
 export type DISingleValues = DIAttribute | DINotifyAttribute | DIAttributeLocationValues | DICase | DIDataContext |
   DINotifyDataContext | DIGlobal | DIInteractiveFrame | DIItemValues | DICreateCollection | DINewCase | DIUpdateCase |
-  DINotification | DIItemSearchNotify | DILogMessage | DIUrl | V2SpecificComponent
+  DINotification | DIItemSearchNotify | DILogMessage | DIUrl | V2SpecificComponent | DICategoryColorMap
```
kswenson, line 98:
```suggestion
  DINotification | DIItemSearchNotify | DILogMessage | DIUrl | V2SpecificComponent
```
Shouldn't be necessary now that colormap isn't a separate API call.

v3/src/data-interactive/data-interactive-types.ts
```diff
@@ -94,14 +95,14 @@ export interface DIResources {
...
 export type DIValues = DISingleValues | DISingleValues[] | number | string[]
 
 // types returned as outputs by the API
 export type DIResultSingleValues = DICase | DIComponentInfo |  DIDataDisplay | DIGetCaseResult | DIGlobal
   | DIInteractiveFrame
 
-export type DIResultValues = DIResultSingleValues | DIResultSingleValues[] |
+export type DIResultValues = DIResultSingleValues | DIResultSingleValues[] | DICategoryColorMap |
```
kswenson, line 105:
```suggestion
export type DIResultValues = DIResultSingleValues | DIResultSingleValues[] |
```
Shouldn't be necessary now that colormap isn't a separate API call.

v3/src/data-interactive/handlers/attribute-handler.ts
kswenson: I suggested another option elsewhere.

v3/src/models/data/category-set.ts
```diff
@@ -275,9 +275,7 @@ export const CategorySet = types.model("CategorySet", {
     self.invalidate()
   },
   setColorForCategory(value: string, color: string) {
-    if (self.index(value) != null) {
-      self.colors.set(value, color)
-    }
+    self.colors.set(value, color)
```
kswenson, line 278: I think you should special case the empty string here to remove a color from the map:
```suggestion
    if (color) {
      self.colors.set(value, color)
    } else {
      self.colors.delete(value)
    }
```

kswenson (changes requested):
👍 This looks better than the prior attempt. I second @tealefristoe's suggestions and added a few of my own.

v3/src/utilities/translation/lang/en-US.json5
```diff
@@ -136,6 +136,7 @@
     "V3.DI.Error.interactiveFrameNotFound": "Interactive Frame not found",
     "V3.DI.Error.itemNotFound": "Item not found",
     "V3.DI.Error.notFound": "Not found",
+    "V3.DI.Error.sharedCaseMetadataNotFound": "Internal error prevented color map access",
```
tealefristoe, line 139:
```suggestion
    "V3.DI.Error.colorMapAccess": "Internal error prevented color map access",
```

v3/src/data-interactive/handlers/di-handler-utils.ts
```diff
@@ -7,16 +7,28 @@ import { v2NameTitleToV3Title } from "../../models/data/v2-model"
...
+
 export function createAttribute(value: DIAttribute, dataContext: IDataSet, collection?: ICollectionModel,
                                 metadata?: ISharedCaseMetadata) {
   const attributeSnapshot = convertValuesToAttributeSnapshot(value)
   if (attributeSnapshot) {
     const attribute = dataContext.addAttribute(attributeSnapshot, { collection: collection?.id })
     if (value.formula) attribute.formula?.setDisplayExpression(value.formula)
     metadata?.setIsHidden(attribute.id, !!value.hidden)
+    if (value.colormap) {
+      applyColormap(attribute.id, value.colormap as DICategoryColorMap, metadata)
```
tealefristoe, line 30: It shouldn't be necessary to use an `as` here. The problem is that `colormap` is typed as `any` in `ICodapV2Attribute`. I think that if you type that properly, you won't have to use `as` here or on line 70 below.

tealefristoe (changes requested):
Looks very close! I made two suggestions that I think should be quick to fix.

A little more context for the `as` issue--you generally want to avoid `as` if at all possible. It signals that we're giving up on type protections. Sometimes you have to do that, but it should be a last resort (like `!important` in css). If it seems like you need to use `as`, it's a good idea to look through the type definitions to see if you can find a place where something is mistyped. Here, `ICodapV2Attribute.colormap` is typed as `any`, which was probably a shortcut way back when types were first being defined and we didn't know what the type should actually be. Now we know what the type should be, so we should type it properly, which means we won't have to cast the field in `di-handler-utils.ts`.

v3/src/v2/codap-v2-data-set-types.ts
```diff
@@ -14,8 +14,7 @@ export interface ICodapV2Attribute {
   // TODO_V2_IMPORT_EXPORT
   categoryMap?: any
   _categoryMap?: any
-  // TODO_V2_IMPORT_EXPORT
-  colormap?: any
+  colormap?: Record<string, string>
```
tealefristoe, line 17:
```suggestion
  colormap?: DICategoryColorMap
```
And you will need to move the definition of `DICategoryColorMap` into this file to avoid an import cycle.

v3/src/data-interactive/handlers/di-results.ts
```diff
@@ -12,6 +12,7 @@ export const couldNotParseQueryResult = errorResult(t("V3.DI.Error.couldNotParse
 export const dataContextNotFoundResult = errorResult(t("V3.DI.Error.dataContextNotFound"))
 export const dataDisplayNotFoundResult = errorResult(t("V3.DI.Error.dataDisplayNotFound"))
 export const itemNotFoundResult = errorResult(t("V3.DI.Error.itemNotFound"))
+export const sharedCaseMetadaNotFoundResult = errorResult(t("V3.DI.Error.colorMapAccess"))
```
tealefristoe, line 15:
```suggestion
export const noColorMapAccessResult = errorResult(t("V3.DI.Error.colorMapAccess"))
```

v3/src/data-interactive/data-interactive-type-utils.ts
```diff
@@ -110,12 +110,16 @@ export function convertAttributeToV2(attribute: IAttribute, dataContext?: IDataS
   const metadata = dataContext && getSharedCaseMetadataFromDataset(dataContext)
   const { cid, name, type, title, description, deleteable, editable, id, precision } = attribute
   const v2Id = toV2Id(id)
+  const rawColorMap = metadata?.getCategorySet(attribute.id)?.colorMap ?? {}
+  const entries = Object.entries(rawColorMap).filter((entry): entry is [string, string] => entry[1] !== undefined)
+  const colormap: Record<string, string> = Object.fromEntries(entries)
```
tealefristoe, line 115:
```suggestion
  const colormap: DICategoryColorMap = Object.fromEntries(entries)
```

tealefristoe (approved):
Just a few more small changes and I think you're good to go 👍

PR #1863, opened 2025-03-17
Codap-9 Heatmaps

tealefristoe:
Jira Story: https://concord-consortium.atlassian.net/browse/CODAP-9

This PR adds a heatmap option to the map, which allows data to be displayed as a heatmap instead of as points. I'm anticipating there will be some required changes to where this code is located in the codebase, so I'm requesting a review before adding tests. After an initial review, I'll add tests to make sure coverage is decent.

I used the `simpleheat` library for this work, which generally worked well, but did have one very confusing problem. For some reason, `simpleheat` always saw the canvas it draws on as being 300x150 pixels, despite the fact that the canvas can be any size and is sized correctly when examining elements. I could never figure out why this is the case, or where those numbers came from. The library is indeed simple and can be found [here](https://github.com/mourner/simpleheat) if you want to take a look.


v3/src/components/map/components/map-point-layer.tsx
```diff
@@ -123,6 +128,87 @@ export const MapPointLayer = observer(function MapPointLayer({mapLayerModel, set
...
+    simpleheatRef.current.resize()
+
+    // If we should draw the heatmap, update simpleheat
+    const { isVisible, pointsAreVisible, pointType } = mapLayerModel
+    if (legendAttributeId && pointType === "heatmap" && pointsAreVisible && isVisible) {
+      // For some reason, the simpleheat canvas always changes to 300x150, so we have to scale the positions accordingly
+      const mapContainer = leafletMap.getContainer()
+      const mapRect = mapContainer.getBoundingClientRect()
+      const scaleX = 300 / mapRect.width
+      const scaleY = 150 / mapRect.height
```
kswenson, line 161: This is sufficiently bizarre (among other reasons) that I'd be inclined to consider other libraries, but given the speculative nature of the feature we can table that discussion for another day.

v3/src/components/map/components/map-point-layer.tsx
```diff
@@ -364,6 +461,9 @@ export const MapPointLayer = observer(function MapPointLayer({mapLayerModel, set
       </svg>
       <div ref={pixiContainerRef} className="map-dot-area"/>
       <MapPointGrid mapLayerModel={mapLayerModel} />
+      <div className="heatmap-area">
+        <canvas ref={heatmapCanvasRef} className="heatmap-canvas" />
+      </div>
```
kswenson, line 466: Can we only render the heat map canvas when the heat map is enabled? Otherwise, it just seems like it's cluttering up the DOM.

v3/src/components/map/models/map-point-layer-model.ts
```diff
@@ -15,6 +15,7 @@ export const MapPointLayerModel = MapLayerModel
     gridModel: types.optional(MapGridModel, () => MapGridModel.create()),
     pointsAreVisible: true, // This is different than layer visibility
     connectingLinesAreVisible: false,
+    pointType: types.optional(types.string, "points"), // "points" or "heatmap"
```
kswenson, line 18: I'd be inclined to call this `displayType` and to define it as an enumerated property:
```suggestion
    displayType: types.optional(types.enumeration([...MapPointDisplayTypes]), "points")
```
and then add:
```typescript
export const MapPointDisplayTypes = ['points', 'heatmap'] as const
export type MapPointDisplayType = typeof MapPlaces[number]
```
```

kswenson (approved):
👍 LGTM -- I made a couple of small suggestions, but given the speculative nature of the feature at this point, I suspect any discussion of larger changes like potentially switching heat map libraries can be put off for another day.

PR #1865, opened 2025-03-19
fix: update graph cell masks when categories change

kswenson:
[[CODAP-337](https://concord-consortium.atlassian.net/browse/CODAP-337)]

[CODAP-337]: https://concord-consortium.atlassian.net/browse/CODAP-337?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1867, opened 2025-03-20
fix: invalidate sub-plot assignments when attribute values change

kswenson:
[[CODAP-337](https://concord-consortium.atlassian.net/browse/CODAP-337)]

In #1865, a graph rendering/masking bug which resulted in some points not being displayed was fixed. In testing that fix, another scenario was identified that resulted in a similar symptom of some points not being displayed. Under the hood, however, the two bugs have different root causes. The bug fixed in #1865 was a straight rendering bug -- the model provided the right data to the graph which simply failed to render it correctly. The remaining issue is a model/cache invalidation bug in which stale data was being passed to the graph. It turns out that some of our reactive code was relying on things like the set of categories to be plotted changing to perform some necessary computations. In the scenario described in the bug report, the values of a computed attribute were changed, but the set of categories was the same and so some necessary invalidations were bypassed. With this PR, instead of relying on the set of categories to be plotted, we react to any changes to the categorical attributes being plotted. The new `GraphDataConfigurationModel` property -- `categoricalAttrsWithChangeCounts` is more efficient to compute and to compare than some of the other properties previously being observed (e.g. `caseDataWithSubPlot`, `allCategoriesForRoles`, `getAllCellKeys`) and so we use it preferentially where it makes sense.

[CODAP-337]: https://concord-consortium.atlassian.net/browse/CODAP-337?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/v3/src/components/data-display/data-display-utils.ts
+++ b/v3/src/components/data-display/data-display-utils.ts
@@ -90,6 +90,7 @@ export interface IMatchCirclesProps {
 export function matchCirclesToData(props: IMatchCirclesProps) {
   const { dataConfiguration, pixiPoints, startAnimation, pointRadius, pointColor, pointStrokeColor,
           pointDisplayType = "points" } = props
+  // TODO: eliminate dependence on GraphDataConfigurationModel
   const allCaseData: CaseDataWithSubPlot[] = isGraphDataConfigurationModel(dataConfiguration)
     ? dataConfiguration.caseDataWithSubPlot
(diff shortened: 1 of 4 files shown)


bfinzer (approved):
👍🏻LGTM
And thanks for the clear explanation!

PR #1870, opened 2025-03-21
imports and exports V2 guides

eireland:


kswenson (approved):
👍 Generally looks good -- I added a commit which:
- supports multi-page guides
- consolidates redundant types
- imports legacy `currentItemTitle` and `currentURL` properties
- exports the `isVisible` property

kswenson:
I just approved the PR, but one thing that would be useful before merge would be to add a test that imports the legacy guide storage. I can send you a document separately that can be used to write an import/export test.

PR #1874, opened 2025-03-22
fix: switching plot type results in adornmentList API confusion (CODAP-342)

emcelroy:
[CODAP-342](https://concord-consortium.atlassian.net/browse/CODAP-342)

This fixes an issue where adornments activated in one plot configuration, were still being listed for `adornmentList` requests after the plot had changed to a type that those adornments don't support. It also adds a listing for a "Percent" adornment. Even though Percent is part of Count, plugin developers may not be aware of that since the CODAP UI presents Count and Percent as totally separate options.

[CODAP-342]: https://concord-consortium.atlassian.net/browse/CODAP-342?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/v3/src/data-interactive/resource-parser.ts
+++ b/v3/src/data-interactive/resource-parser.ts
@@ -1,3 +1,7 @@
+import { getAdornmentContentInfo } from "../components/graph/adornments/adornment-content-info"
+import { IAdornmentModel } from "../components/graph/adornments/adornment-models"
+import { isCountAdornment } from "../components/graph/adornments/count/count-adornment-model"
+import { kPercentType } from "../components/graph/adornments/count/count-adornment-types"
 import { isGraphContentModel } from "../components/graph/models/graph-content-model"
 import { appState } from "../models/app-state"
 import { IAttribute } from "../models/data/attribute"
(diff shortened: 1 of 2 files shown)


v3/src/data-interactive/resource-parser.ts
```diff
@@ -158,7 +162,29 @@ export function resolveResources(
...
-    result.adornmentList = result.component.content.adornmentsStore.adornments ?? []
+    const graphPlotType = result.component?.content.plotType
+    const adornmentList = result.component.content.adornmentsStore.adornments.reduce((list, adornment) => {
+      if (isCountAdornment(adornment)) {
+        // If the Count adornment is present, we add a separate Percent adornment item to the list. Even though
+        // Percent is part of the Count adornment, clients may not be aware of that since the UI presents them
+        // as separate entities.
+        const percentAdornment = {
+          ...adornment,
+          isVisible: adornment.showPercent,
```
tealefristoe, line 173: Should this be `adornment.showPercent && adornment.isVisible`? This also makes me wonder if `isVisible` is correct for the count adornment.

tealefristoe (approved):
Looks good 👍

v3/src/data-interactive/resource-parser.ts
emcelroy: Yes, you're right. Thanks. And I do think it makes sense to adjust the Count adornment's `isVisible` value, too. I've done that in the latest commit. Does that look good?

kswenson (commented):
👍 LGTM

PR #1875, opened 2025-03-24
feat: Allow plugins to create, update, and delete specific graph adornments (CODAP-47)

emcelroy:
[CODAP-47](https://concord-consortium.atlassian.net/browse/CODAP-47)

Adds `create`, `update`, and `delete` API endpoints for specific graph adornments. Support for all three endpoints has not been added for every single adornment type. These changes cover the adornments the DAVAI plugin will mainly be concerned with: Count/Percent, Least Squares Line, Mean, Median, Standard Deviation, and Movable Value. Support for the other adornments will have to be added later.

[CODAP-47]: https://concord-consortium.atlassian.net/browse/CODAP-47?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/v3/src/components/graph/adornments/lsrl/lsrl-adornment-handler.ts
+++ b/v3/src/components/graph/adornments/lsrl/lsrl-adornment-handler.ts
@@ -1,21 +1,35 @@
-import { isAdornmentValues } from "../../../../data-interactive/data-interactive-adornment-types"
+import { DIAdornmentValues, DILsrlAdornmentValues, isAdornmentValues }
+  from "../../../../data-interactive/data-interactive-adornment-types"
 import { DIAdornmentHandler } from "../../../../data-interactive/handlers/adornment-handler"
 import { adornmentNotFoundResult, adornmentNotSupportedByPlotTypeResult }
   from "../../../../data-interactive/handlers/di-results"
 import { IGraphContentModel } from "../../models/graph-content-model"
-import { getAdornmentContentInfo } from "../adornment-content-info"
+import { getAdornmentContentInfo, isCompatibleWithPlotType } from "../adornment-content-info"
 import { IAdornmentModel } from "../adornment-models"
 import { IAdornmentsBaseStore } from "../store/adornments-base-store"
-import { AdornmentData, adornmentMismatchResult, cellKeyToCategories, isAdornmentSupportedByPlotType }
-  from "../utilities/adornment-handler-utils"
+import { AdornmentData, adornmentMismatchResult, cellKeyToCategories } from "../utilities/adornment-handler-utils"
 import { ILSRLAdornmentModel, isLSRLAdornment } from "./lsrl-adornment-model"
 import { kLSRLType } from "./lsrl-adornment-types"
+
+const setAdornmentProperties = (adornment: ILSRLAdornmentModel, values: DIAdornmentValues) => {
+  if (isAdornmentValues(values)) {
+    const { isVisible, showConfidenceBands } = values as DILsrlAdornmentValues
+    if (isVisible != null) {
+      adornment.setVisibility(isVisible)
+    }
+    if (showConfidenceBands != null) {
+      adornment.setShowConfidenceBands(showConfidenceBands)
+    }
+  }
+}
 
 export const lsrlAdornmentHandler: DIAdornmentHandler = {
   create(args) {
     const { graphContent, values } = args
-    const isAdornmentSupported = isAdornmentSupportedByPlotType(kLSRLType, graphContent.plotType)
-    if (!isAdornmentSupported) return adornmentNotSupportedByPlotTypeResult
+
+    if (!isCompatibleWithPlotType(kLSRLType, graphContent.plotType)) {
+      return adornmentNotSupportedByPlotTypeResult
+    }
 
     const adornmentsStore = graphContent.adornmentsStore as IAdornmentsBaseStore
     const dataConfig = graphContent.dataConfiguration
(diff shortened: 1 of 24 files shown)


v3/src/components/graph/adornments/lsrl/lsrl-adornment-handler.ts
```diff
@@ -39,6 +96,19 @@ export const lsrlAdornmentHandler: DIAdornmentHandler = {
...
     return { data, showConfidenceBands }
+  },
+
+  update(args) {
+    const { graphContent, values } = args
+    const adornmentsStore = graphContent.adornmentsStore as IAdornmentsBaseStore
+    const existingLsrlAdornment = adornmentsStore.findAdornmentOfType<ILSRLAdornmentModel>(kLSRLType)
+    if (!existingLsrlAdornment) return adornmentNotFoundResult
+
+    if (isAdornmentValues(values) && "showConfidenceBands" in values && values.showConfidenceBands !== undefined) {
```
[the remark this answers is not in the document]
kswenson: to handle `null` and `undefined`. Also, what about `isVisible`?

v3/src/components/graph/adornments/movable-value/movable-value-adornment-handler.test.ts
```diff
@@ -17,13 +45,23 @@ describe("DataInteractive movableValueAdornmentHandler", () => {
...
+      adornmentsStore: {
+        addAdornment: jest.fn((adornment: any, options: any) => null),
+        findAdornmentOfType: jest.fn()
+      },
       dataConfiguration: mockDataConfig
     }
-    
+  
+    mockWrongGraphContent = {
+      plotType: "linePlot"
```
kswenson, line 58: `linePlot` _does_ support movable values.

v3/src/components/graph/adornments/movable-value/movable-value-adornment-handler.ts
```diff
@@ -1,11 +1,65 @@
...
+    const data: AdornmentData[] = []
+    const existingMovableValueAdornment =
+      adornmentsStore.findAdornmentOfType<IMovableValueAdornmentModel>(kMovableValueType)
+    const componentContentInfo = getAdornmentContentInfo(kMovableValueType)
+    const adornment =
+      existingMovableValueAdornment ?? componentContentInfo.modelClass.create() as IMovableValueAdornmentModel
+
+    if (!existingMovableValueAdornment) {
+      const options: IUpdateCategoriesOptions = { ...graphContent.getUpdateCategoriesOptions(), addMovableValue: true }
+      adornmentsStore.addAdornment(adornment, options)
```
kswenson, line 31: Shouldn't `create` create an additional movable value, even if there's already one present? And shouldn't the API allow the client to specify the value rather than using a default value?

kswenson (changes requested):
👍 Generally looks good -- some smaller suggestions and one that's a bit more work:
1. Use `isCompatibleWithPlotType()` in `adornment-content-info.ts` instead of the new  `isAdornmentSupportedByPlotType()` function, which doesn't handle the new plot types correctly.
2. Across the board, use `!= null` rather than `!== undefined`. Personally, I prefer this as a matter of style, but here it's important because the client can easily pass `null` as well as `undefined`.
3. There's a lot of redundancy. Utility methods like `createUnivariateMeasureAdornment()` and `updateUnivariateMeasureAdornment()` could be used to eliminate a lot of it. There may be other opportunities as well but those were the most obvious ones.

v3/src/components/graph/adornments/movable-value/movable-value-adornment-handler.test.ts
```diff
@@ -1,10 +1,76 @@
...
+//   return {
+//     ...jest.requireActual("../adornment-content-info"),
+//     getAdornmentContentInfo: jest.fn().mockReturnValue({
+//       modelClass: mockCountModel,
+//       plots: ["dotPlot"],
+//       prefix: "movable-value",
+//       type: "Movable Value",
+//     }),
+//   }
+// })
```
kswenson, line 67: Do you want to keep this around or can it be deleted now?

v3/src/components/graph/adornments/movable-value/movable-value-adornment-registration.tsx
```diff
@@ -73,7 +73,7 @@ const Controls = () => {
 
 registerAdornmentContentInfo({
   type: kMovableValueType,
-  plots: ['dotPlot'],
+  plots: ['dotPlot', 'linePlot'],
```
kswenson, line 76:
```suggestion
  plots: ['dotPlot'],
```
You shouldn't need this. `isCompatibleWithPlotType()` handles extending `dotPlot` to other compatible plot types like `linePlot` (there are actually four of them).

v3/src/components/graph/adornments/univariate-measures/standard-deviation/standard-deviation-adornment-handler.test.ts
```diff
@@ -1,6 +1,39 @@
+import { types } from "@concord-consortium/mobx-state-tree"
```
kswenson, line 1:
```suggestion
import { types } from "mobx-state-tree"
```

kswenson (approved):
👍 LGTM -- just a few remaining tweaks.

PR #1881, opened 2025-03-26
CODAP-354 fixed the check for percent adornments

nstclair-cc:
[CODAP-354](https://concord-consortium.atlassian.net/browse/CODAP-354)

I decided to fix the percent adornment test today to prevent regressions in future builds.

Changes include:
- Replaced the flaky graph-adornments-grid__cell selector with more stable checks on [data-testid=adornment-wrapper].
- Confirmed percent values using [data-testid^=graph-count] to verify actual content is rendered as expected.
- Added a lightweight check to ensure the percent values aren’t all 100%. The expected values reflect what I saw when doing the test by hand.
- Introduced a small tolerance in the value checks to account for rounding differences, as per [CODAP-359](https://concord-consortium.atlassian.net/browse/CODAP-359), which notes that percent rounding behavior should match v2.

If desired, I can add more detailed checks in another story to validate the percent columns and cell display when radials are toggled on. For now, I just wanted to get a reliable test working in CI.

[CODAP-354]: https://concord-consortium.atlassian.net/browse/CODAP-354?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CODAP-359]: https://concord-consortium.atlassian.net/browse/CODAP-359?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/v3/cypress/e2e/adornments.spec.ts
+++ b/v3/cypress/e2e/adornments.spec.ts
@@ -4,7 +4,7 @@ import { AxisElements as ae } from "../support/elements/axis-elements"
 import { ToolbarElements as toolbar } from "../support/elements/toolbar-elements"
 import { FormulaHelper as fh } from "../support/helpers/formula-helper"
 
-const expectedPercents = ["0", "100", "0", "0", "100", "0", "29.17", "33.33", "37.5"]
+const expectedPercents = [0, 100, 0, 0, 100, 0, 29, 33, 37]
 
 context("Graph adornments", () => {
   beforeEach(function () {
@@ -118,10 +118,13 @@ context("Graph adornments", () => {
 
     // Verify the approximate values of the percentages
     cy.get("[data-testid^=graph-count]").should("have.length", expectedPercents.length)
-      .each(($el, index) => {
-        const expected = expectedPercents[index]
-        cy.wrap($el).invoke("text").should("contain", expected)
+    .each(($el, index) => {
+      const expected = expectedPercents[index]
+      cy.wrap($el).invoke("text").then(text => {
+        const actual = parseFloat(text.replace("%", ""))
+        expect(actual).to.be.closeTo(expected, 1)  // Tolerance of 1 is safe due to CODAP-359
       })
+    })
   
     // Hide the percent values
     graph.getInspectorPalette()
--- a/v3/cypress/e2e/adornments.spec.ts
+++ b/v3/cypress/e2e/adornments.spec.ts
@@ -4,6 +4,8 @@ import { AxisElements as ae } from "../support/elements/axis-elements"
 import { ToolbarElements as toolbar } from "../support/elements/toolbar-elements"
 import { FormulaHelper as fh } from "../support/helpers/formula-helper"
 
+const expectedPercents = ["0", "100", "0", "0", "100", "0", "29.17", "33.33", "37.5"]
+
 context("Graph adornments", () => {
   beforeEach(function () {
     const queryParams = "?sample=mammals&dashboard&mouseSensor"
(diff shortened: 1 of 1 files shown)


v3/cypress/e2e/adornments.spec.ts
```diff
@@ -91,45 +93,56 @@ context("Graph adornments", () => {
...
+    // Verify that at least one percent value is shown
+    cy.get("[data-testid^=graph-count]").should("have.length.at.least", 1)
+    cy.get("[data-testid^=graph-count]").first().should("contain.text", "%")
+
+    // Verify the approximate values of the percentages
+    cy.get("[data-testid^=graph-count]").should("have.length", expectedPercents.length)
+      .each(($el, index) => {
+        const expected = expectedPercents[index]
+        cy.wrap($el).invoke("text").should("contain", expected)
+      })
```
emcelroy, line 124: Is this actually verifying _approximate_ values? If I'm reading it correctly, it's checking that the text contains the associated value in `expectedPercents`. So if the text is "33%" it's not going to match "33.33" which I think was your intention. Is that right? If so, would updating `expectedPercents` to only include whole numbers fix that?

I'm not sure this is good advice given the code hasn't been approved yet, but in a separate Cypress test that I recently updated, instead of checking for specific percent values, [we check that the sum of percent values present equals about ~100](https://github.com/concord-consortium/codap/blob/c04dfc7cf17dd14810d01db38c8273acb5fd50f1/v3/cypress/e2e/plugin.spec.ts#L230-L235). That won't work here exactly since the sum would be more than 100, but maybe you could adjust that general approach to work? I think that way we wouldn't have to worry about maintaining the array of expected values.

emcelroy (approved):
Looks good 👍 I did question the approach of checking the percent values. I think it may not quite be right even though it should work OK as is.

kswenson (commented):
I agree with @emcelroy that testing approximate values would be better than precise equality tests of percentages. Beyond that, I have nothing to add.

v3/cypress/e2e/adornments.spec.ts
nstclair-cc: Hi Ethan, great catch! I pushed a commit that switches to numerical matching with a tolerance instead of relying on string comparisons. The tolerance might be a bit tight, but I added a comment referencing CODAP-359 so we’re aware that changes to percent rounding are in the pipeline.

kswenson (approved):
👍 LGTM

v3/cypress/e2e/adornments.spec.ts
```diff
@@ -91,45 +93,59 @@ context("Graph adornments", () => {
...
+    // Verify that at least one percent value is shown
+    cy.get("[data-testid^=graph-count]").should("have.length.at.least", 1)
+    cy.get("[data-testid^=graph-count]").first().should("contain.text", "%")
+
+    // Verify the approximate values of the percentages
+    cy.get("[data-testid^=graph-count]").should("have.length", expectedPercents.length)
+    .each(($el, index) => {
+      const expected = expectedPercents[index]
+      cy.wrap($el).invoke("text").then(text => {
+        const actual = parseFloat(text.replace("%", ""))
```
emcelroy, line 124: You don't have to change this because it'll work fine as is, but for future reference, I'm pretty sure the `.replace("%", "")` is unnecessary. `parseFloat` will strip out that part of the string.

emcelroy (approved):
Looks good 👍 And thanks for resurrecting that set of tests!

PR #1884, opened 2025-03-28
fix: adornments spec after category order change

kswenson:
#1880 changed the default order of categories, which required a minor change to the adornments percent test.


kswenson:
Merging without review as it's a minor tweak to a test.

PR #1886, opened 2025-03-28
Codap 320 component import export

eireland:
Adds import of the component layout `x` and `y` values.
Adds import of component `cannotClose` value.

### DIFF
--- a/v3/src/v2/import-v2-document.ts
+++ b/v3/src/v2/import-v2-document.ts
@@ -55,7 +55,7 @@ export function importV2Document(v2Document: CodapV2Document) {
           const _zIndex = zIndex != null ? { zIndex } : {}
           if (zIndex != null && zIndex > maxZIndex) maxZIndex = zIndex
           const _layout: IFreeTileInRowOptions = {
-            x: x ?? left, y: y ?? top, ..._width, ..._height, ..._zIndex, isHidden, isMinimized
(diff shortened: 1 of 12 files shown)


v3/src/v2/import-v2-document.ts
```diff
@@ -55,7 +55,7 @@ export function importV2Document(v2Document: CodapV2Document) {
           const _zIndex = zIndex != null ? { zIndex } : {}
           if (zIndex != null && zIndex > maxZIndex) maxZIndex = zIndex
           const _layout: IFreeTileInRowOptions = {
-            x: left, y: top, ..._width, ..._height, ..._zIndex, isHidden, isMinimized
+            x: x ?? left, y: y ?? top, ..._width, ..._height, ..._zIndex, isHidden, isMinimized
```
kswenson, line 58:
```suggestion
            x: left ?? x, y: top ?? y, ..._width, ..._height, ..._zIndex, isHidden, isMinimized
```
We only want to use `x` and `y` if `left` and `top` are not present.

kswenson (changes requested):
👍 Looks good -- two suggestions that should be straightforward.

kswenson (approved):
👍 LGTM -- I added a commit to address the `cannotClose` issue I mentioned previously.

PR #1887, opened 2025-03-28
Percent Bar Charts

bfinzer:
[#CODAP-290] Feature: The perpendicular axis of a bar chart can show percentages and the heights of the bars (assuming no legend) is the percentage for each category.

* We add "formula" as a breakdown type
* In display-config-palette.tsx we detect that we should show breakdown types and, if so, append the list of breakdown types as radio buttons
* Introduce PercentAxisModel
* Put responsibility for producing an axis label for count, percent or formula in BarChartModel
* We fix a bug whereby adding a categorical attribute to the "secondary" (x or y) axis of a bar chart was not changing the plot to a dot chart when it should because we don't support multiple count axes on x or y when there is also a categorical attribute on the axis.
* We fix another bug whereby splitting a bar chart on the top or right was not computing new bounds for the count axis.
* We need to be able to compute the maximum percent when the breakdownType is that. This led to a reordering of the cells in graph-data-configuration-model.ts cellMap so that it's easier to compute the maximum percent for each sub-plot.
* We add PercentAxisModel and install an `onPatch` in `GraphContentModel` that allows the graph model's plot to let the graph model know that a new axis model is needed for the secondary place.
* Similarly, in order to synchronize the axis scale with the axis model, we install an mstReaction in GraphController that carries this out when either the 'left' or 'bottom' axis changes
* When displaying percent bar charts, the thickness of a rectangle for a case depends on the number of cases in the sub-plot. Unfortunately, because we're iterating through cases irrespective of sub-plot, this requires determining the sub-plot of each case and the number of cases in that sub-plot.
* Also when displaying bar charts, the bar covers have to be adjusted to take into account the number of cases in the sub-plot.

### DIFF
--- a/v3/src/components/axis/models/axis-model.ts
+++ b/v3/src/components/axis/models/axis-model.ts
@@ -200,7 +200,7 @@ export function isCountAxisModel(axisModel?: IAxisModel): axisModel is ICountAxi
 }
 
 export function isNumericOrCountOrPercentAxisModel(axisModel?: IAxisModel):
-    axisModel is INumericAxisModel | IDateAxisModel | ICountAxisModel | IPercentAxisModel {
+    axisModel is INumericAxisModel | ICountAxisModel | IPercentAxisModel {
   return isNumericAxisModel(axisModel) || isCountAxisModel(axisModel) || isPercentAxisModel(axisModel)
 }
 
--- a/v3/src/components/axis/models/axis-model.ts
+++ b/v3/src/components/axis/models/axis-model.ts
@@ -162,18 +162,36 @@ export function isNumericAxisModel(axisModel?: IAxisModel): axisModel is INumeri
   return axisModel?.type === "numeric"
 }
 
+export const PercentAxisModel = BaseNumericAxisModel
+  .named("PercentAxisModel")
+  .props({
+    type: types.optional(types.literal("percent"), "percent")
+  })
+  .views(self => ({
+    get lockZero() {
+      return true
+    },
+  }))
+export interface IPercentAxisModel extends Instance<typeof PercentAxisModel> {}
+export interface IPercentAxisModelSnapshot extends SnapshotIn<typeof PercentAxisModel> {}
+export function isPercentAxisModel(axisModel?: IAxisModel): axisModel is IPercentAxisModel {
+  return axisModel?.type === "percent"
+}
+
+// ToDo: It _should_ be possible to have CountAxisModel inherit from PercentAxisModel and get rid of lockZero
+//  but it causes a strange typescript error in graph-content-model.ts
 export const CountAxisModel = BaseNumericAxisModel
   .named("CountAxisModel")
   .props({
     type: types.optional(types.literal("count"), "count")
   })
   .views(self => ({
-    get integersOnly() {
-      return true
-    },
     get lockZero() {
       return true
     },
+    get integersOnly() {
+      return true
+    }
   }))
 export interface ICountAxisModel extends Instance<typeof CountAxisModel> {}
 export interface ICountAxisModelSnapshot extends SnapshotIn<typeof CountAxisModel> {}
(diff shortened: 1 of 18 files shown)


v3/src/components/graph/plots/bar-chart/bar-chart-model.ts
```diff
@@ -16,31 +16,68 @@ export const BarChartModel = DotChartModel
   .named("BarChartModel")
   .props({
     type: typesPlotType("barChart"),
-    breakdownType: types.maybe(types.enumeration([...BreakdownTypes])),
+    breakdownType: types.optional(types.enumeration([...BreakdownTypes]), "count"),
     expression: types.maybe(Formula)
   })
+  .actions(self => ({
+    setBreakdownType(type: BreakdownType) {
+      self.breakdownType = type
```
kswenson, line 24: Should setting a breakdown type of `count` or `percent` clear the expression? If matching v2 behavior were the goal, the answer would be yes.

v3/src/components/graph/plots/bar-chart/bar-chart-model.ts
```diff
@@ -16,31 +16,68 @@ export const BarChartModel = DotChartModel
...
+    },
+    setExpression(expression: IFormula) {
+      self.expression = expression
```
kswenson, line 27: Should setting the expression also set the `breakdownType` to `formula`? If matching v2 behavior were the goal, the answer would be no. Because `breakdownType` was in the base class, switching from `BarChart` to `ComputedBarChart` wouldn't clear the `breakdownType`, but going in reverse _would_ clear the expression.

v3/src/components/axis/models/axis-model.ts
```diff
@@ -162,27 +162,46 @@ export function isNumericAxisModel(axisModel?: IAxisModel): axisModel is INumeri
...
 export interface ICountAxisModel extends Instance<typeof CountAxisModel> {}
 export interface ICountAxisModelSnapshot extends SnapshotIn<typeof CountAxisModel> {}
 export function isCountAxisModel(axisModel?: IAxisModel): axisModel is ICountAxisModel {
   return axisModel?.type === "count"
 }
 
-export function isNumericOrCountAxisModel(axisModel?: IAxisModel): axisModel is INumericAxisModel | IDateAxisModel {
-  return isNumericAxisModel(axisModel) || isCountAxisModel(axisModel)
+export function isNumericOrCountOrPercentAxisModel(axisModel?: IAxisModel):
+    axisModel is INumericAxisModel | IDateAxisModel | ICountAxisModel | IPercentAxisModel {
```
kswenson, line 203: It seems that the goal of this function is to exclude date axes, in which case it should be:
```suggestion
    axisModel is INumericAxisModel | ICountAxisModel | IPercentAxisModel {
```
On the other hand, if we don't care about excluding date axes then it's redundant with `isBaseNumericAxisModel()` below.

v3/src/components/graph/models/graph-content-model.ts
```diff
@@ -186,6 +189,23 @@ export const GraphContentModel = DataDisplayContentModel
...
+    setAxis(place: AxisPlace, axis: IAxisModel) {
+      if (isAxisModelInUnion(axis)) {
+        self.axes.set(place, axis)
+      }
+    },
+    installPlotPatchDisposer() {
+      // At the plot level we can't install a new axis. But a plot can change in such a way that it
+      // requires a new secondary axis. When our plot is patched, we pass the patch to it and allow it to
+      // tell us about any required new secondary axis.
+      // Todo: Kirk will figure out how to do the right thing with disposing stuff
```
kswenson, line 201: Done

kswenson (approved):
👍 Looks good -- I added commit which cleaned up the plot patch disposer as requested. I have some ideas on how to simplify the polymorphic axis model hierarchy, but I'll leave those for another day. Just a few remaining questions to consider.

v3/src/components/graph/plots/bar-chart/bar-chart-model.ts
bfinzer: I think sticking with V2 behavior is fine.

PR #1889, opened 2025-03-31
fix: create, update, delete graph adornments API handlers (CODAP-47)

emcelroy:
[CODAP-47](https://concord-consortium.atlassian.net/browse/CODAP-47)

These changes fix the following issues uncovered in review.

- Any adornment: `update` request for an adornment that doesn't exist receives "success" response
- Percent adornment: `create` request always receives error response
- Movable Value: a few inconsistencies with `create` and `update` due to issues involving cell key values -- most importantly that `create` requests with specified numeric values per subplot for split plots resulted in multiple Movable Line instances per subplot instead of one instance per subplot.

[CODAP-47]: https://concord-consortium.atlassian.net/browse/CODAP-47?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/v3/src/data-interactive/data-interactive-adornment-types.ts
+++ b/v3/src/data-interactive/data-interactive-adornment-types.ts
@@ -1,6 +1,6 @@
 import { SnapshotIn } from "@concord-consortium/mobx-state-tree"
 import { CountAdornmentModel } from "../components/graph/adornments/count/count-adornment-model"
-import { kCountType } from "../components/graph/adornments/count/count-adornment-types"
+import { kCountType, kPercentType } from "../components/graph/adornments/count/count-adornment-types"
 import { LSRLAdornmentModel } from "../components/graph/adornments/lsrl/lsrl-adornment-model"
 import { kLSRLType } from "../components/graph/adornments/lsrl/lsrl-adornment-types"
 import { MeanAdornmentModel } from "../components/graph/adornments/univariate-measures/mean/mean-adornment-model"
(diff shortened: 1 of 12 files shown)


v3/src/data-interactive/data-interactive-adornment-types.ts
```diff
@@ -23,7 +23,7 @@ export type DIStandardDeviationAdornmentValues = Partial<SnapshotIn<typeof Stand
 export type DIAdornmentValues = DICountAdornmentValues | DILsrlAdornmentValues | DIMeanAdornmentValues |
   DIMedianAdornmentValues | DIMovableValueAdornmentValues | DIStandardDeviationAdornmentValues
 
-const kAdornmentTypes = [kCountType, kLSRLType, kMeanType, kMedianType, kMovableValueType, kStandardDeviationType]
+const kAdornmentTypes = [kCountType, kLSRLType, kMeanType, kMedianType, kMovableValueType, kPercentType, kStandardDeviationType]
```
kswenson: The reason I added the type alias resolution system is so that the "Percent" adornment could be mapped to the `Count` adornment without having to pretend it exists elsewhere in the system. I'd rather leave this as a list of real adornments and change the `isAdornmentValues` function below. I've added a commit which addresses this.

kswenson (approved):
👍 LGTM -- there were a few changes I was going to suggest but I ended up trying them out locally and then adding them as a commit to the PR. The remaining suggestions are quite minor. Let me know if any of the changes I made need more explanation and as usual, make sure I didn't break anything. 😉

PR #1891, opened 2025-04-01
[CODAP-355]: Region of Interest Adornment on Numeric Graphs

lublagg:
[CODAP-355](https://concord-consortium.atlassian.net/browse/CODAP-355?atlOrigin=eyJpIjoiYjFkYjgyMzcxOTBkNGZlYjlkODU1M2FkNzg4ZTVhOTIiLCJwIjoiaiJ9)

This PR implements the "Region of Interest" adornment. 
Note that this currently only works for numeric graphs, and does not support complicated scenarios (such as [this one](https://concord-consortium.slack.com/archives/C035J6RDAK0/p1743177801431339?thread_ts=1743171039.736129&cid=C035J6RDAK0)). 
This feature was moved out of its original epic and into the DAVAI sonification epic so that we could get a visual of a line moving across a graph, and we are punting on the work that is not relevant to that need for the moment.

To test out the functionality you can use [this CODAP document](https://codap3.concord.org/branch/roi-graph-adornment/#shared=https%3A%2F%2Fcfm-shared.concord.org%2FX8zWVD5ayRE6UTSltvrr%2Ffile.json).

Example create request:
```
{
  "action": "create",
  "resource": "component[43].adornment",
  "values": {
    "type": "Region of Interest",
    "width": { "unit": "percent", "value": 1 },
    "xPosition": { "unit": "coordinate", "value": 15 },
    "yPosition": { "unit": "coordinate", "value": 0 }
  }
}
```

[CODAP-355]: https://concord-consortium.atlassian.net/browse/CODAP-355?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


v3/src/components/graph/adornments/region-of-interest/region-of-interest-adornment-model.ts
```diff
@@ -0,0 +1,52 @@
...
+export const RegionOfInterestAdornmentModel = AdornmentModel
+  .named("RegionOfInterestAdornmentModel")
+  .props({
+    type: types.optional(types.literal(kRegionOfInterestType), kRegionOfInterestType),
+    height: types.maybe(RoiUnitValueModel),
+    width: types.maybe(RoiUnitValueModel),
+    xAttribute: types.optional(types.string, ""),
+    xPosition: types.maybe(RoiUnitValueModel),
+    yAttribute: types.optional(types.string, ""),
+    yPosition: types.maybe(RoiUnitValueModel)
```
kswenson, line 21: This model representation will result in the ROI adornment being less robust to graph configuration changes than other adornments. See suggestion elsewhere for how to improve it.

kswenson (changes requested):
👍 Looks good so far. Upon review and further consideration, as well as a conversation with @bfinzer, I’m going to suggest a change to the structure of the API request (and the way it’s stored in the model). The proposed API results in adornments that are more brittle than other adornments. Right now if you put movable values (or any other adornment) on a dot plot on the x axis and then replace the numeric attribute on the x axis with a different one, the adornment is unaffected. Similarly, if you drag the same attribute to the y axis, the plot reconfigures to be a dot plot on y and all adornments render themselves appropriately in the new configuration. By making the ROI API (and model representation) overly specific in terms of x and y and specific attributes, the ROI adornments are much less robust. Therefore, my proposed change to the API would be something like this:
```
{
  primary: {
    position: [number/number%],
    extent: [number/number%]
  }
  secondary: {
    position: [number/number%],
    extent: [number/number%],
    axis?: "y" | "y2" // defaults to y, need not be implemented until y2 support is required
  }
}
```
I believe this handles all of the use cases discussed above. In a numeric/categorical plot the numeric axis is always primary. In a numeric/numeric plot (i.e. a scatter plot), the x-axis is primary and the y-axis is secondary. This would allow the ROI adornment to behave like other adornments in response to graph configuration changes.

PR #1892, opened 2025-04-01
Codap-6 Map Pins

tealefristoe:
Jira story: https://concord-consortium.atlassian.net/browse/CODAP-6

This PR adds support for pins/markers to Codap maps.
- Pins are enabled if a dataset has an appropriately named latitude and longitude attribute.
  - The attribute keyword names are the same as point latitude and longitude names with `pin` appended to them.
- When pins are enabled, add pin and remove pin buttons are displayed on the map.
  - Clicking the add pin button enters add pin mode. The next time the user clicks of the map, a new pin will be added where the user clicked (a new case is added to the pin table with the correct latitude and longitude), and then add pin mode is exited.
  - Clicking the remove pin button will immediately remove all selected pins (removing their cases from the pin table).
- Pin selection is tied to selecting the corresponding case in the pin table.
  - Clicking a pin will select it.
  - Shift+clicking a pin will add or remove it from the current selection.
- Pins are colored in order according to a set of ten colors.
  - If a color attribute is in the pin table, the color is stored here. Changing this color will change the color of the pin.
  - If no color attribute is in the pin table, the color is based on the index. This isn't great, because pins will change colors if their case order changes (for example, removing an earlier pin). However, I didn't think it would be appropriate to add a color attribute if one wasn't already present.

There are a few places where I left `FIX-ME` comments in the code, where I welcome suggestions on how to improve things.
- `z-index` for the pin layer and pin controls.
- Hard coded pin svg dimensions.


v3/src/components/map/map-types.ts
```diff
@@ -46,7 +47,15 @@ export const
 
 
   kLatNames = ['latitude', 'lat', 'latitud', 'breitengrad', '緯度', 'عرض جغرافیایی'],
-  kLongNames = ['longitude', 'long', 'lng', 'lon', 'longitud', 'längengrad', '経度', 'طول جغرافیایی']
+  kLongNames = ['longitude', 'long', 'lng', 'lon', 'longitud', 'längengrad', '経度', 'طول جغرافیایی'],
+
+  kPinLatNames = ['pinlatitude', 'pinlat', 'pinlatitud', 'pinbreitengrad', 'pin緯度', 'pinعرض جغرافیایی'],
+  kPinLongNames = [
+    'pinlongitude', 'pinlong', 'pinlng', 'pinlon', 'pinlongitud', 'pinlängengrad', 'pin経度', 'pinطول جغرافیایی'
+  ],
```
kswenson, line 55: That's probably not the right solution for languages like Chinese that use a different alphabet, but it's probably the best we can do right now. Ultimately, this should be integrated into the rest of our localization system, but that's a project for another day. Maybe for now add a TODO comment about figuring out a better solution.

v3/src/components/map/models/map-pin-layer-model.ts
```diff
@@ -0,0 +1,63 @@
...
+import {kMapPinLayerType} from "../map-types"
+import {latLongAttributesFromDataSet} from "../utilities/map-utils"
+import {MapGridModel} from "./map-grid-model"
+import {MapLayerModel} from "./map-layer-model"
+
+export const MapPointDisplayTypes = ["points", "heatmap"] as const
+export type MapPointDisplayType = typeof MapPointDisplayTypes[number]
+export function isMapPointDisplayType(value: any): value is MapPointDisplayType {
+  return MapPointDisplayTypes.includes(value)
+}
```
kswenson, line 15:
```suggestion
```
Don't think you need these here. Seems like copy/paste detritus.

kswenson (approved):
👍 Looks good -- a handful of suggestions that should be straightforward. I also added @bfinzer as a reviewer to at least take a look at the layer architecture and play around with the UI.

bfinzer:
I renamed a `pinLat` attribute to `pinLatX` and got a runtime error. Probably you are looking up by name when you should be using an ID? (I got worried that regular points would misbehave in the same way, but they don't.)

v3/src/components/map/map-types.ts
```diff
@@ -46,7 +47,16 @@ export const
 
 
   kLatNames = ['latitude', 'lat', 'latitud', 'breitengrad', '緯度', 'عرض جغرافیایی'],
-  kLongNames = ['longitude', 'long', 'lng', 'lon', 'longitud', 'längengrad', '経度', 'طول جغرافیایی']
+  kLongNames = ['longitude', 'long', 'lng', 'lon', 'longitud', 'längengrad', '経度', 'طول جغرافیایی'],
+
+  // TODO: Figure out a better names for these in languages that don't use the latin alphabet.
```
bfinzer, line 52: Actually, since these are user-visible strings, the word 'pin' should be translated for all the non-English names. That should at least be noted in the TODO.

bfinzer (changes requested):
Just a few minor suggestions and a crash to fix

tealefristoe:
@bfinzer Good catch on that crash bug. I tried it with points, and you're right that it doesn't break immediately when you change the name, but it does crash if you change a value after you change the name. You'll also see the same crash if you delete either the `lat` or `long` attribute.

I'm not sure the best approach for fixing this problem. I could store the id of the attributes when they're first discovered, which would fix any name changes. But it wouldn't fix the issue if one of the attributes was deleted. Another approach would be to stop rendering the pins if one of the special attributes changes. I think this would be pretty easy to implement for pins, but I'm not sure how easy this would be for points.

tealefristoe:
@bfinzer I went ahead and disabled the pin layer if the dataset it's using stops having the necessary attributes (they're renamed or deleted). This seems to work pretty well. It would probably be good to make bug stories for the equivalent point layer problems, and we should check the boundary layer as well.

PR #1894, opened 2025-04-01
fix: tool shelf cypress test after guide PR

kswenson:


kswenson:
Merging without review as it's just a simple cypress test fix.

PR #1896, opened 2025-04-02
Codap 78 attr hover indicator

eireland:
Adds hover styling to attribute headers in case table and card. Attribute text should be blue and underlined when hovered.


kswenson (approved):
👍 LGTM

PR #1897, opened 2025-04-03
Percent Bar Charts with Legends

bfinzer:
[#CODAP-372] Feature: Percentage bar charts with legend divide 100% bar for each category into sub-categories

* Computing the maximum to display on a percent axis is more complicated than computing the maximum count.
  * If there is a legend, the maximum is 100 because the bars always go to 100%.
  * If no legend, then we have to iterate over the sub-plots to find the maximum percentage over all the categories in that sub-plot, and take the maximum of these.
* Bars are composed of thin rectangular slices. Whereas when displaying counts, the thickness of slices is constant across all bars and sub-plots, when displaying percentages, this thickness varies from bar to bar in order to guarantee that the height of the bar will come out to 100.
* In `GraphDataConfigurationModel` we introduce some additional convenience views. And we turn `cellMap` into a `cachedFnWithArgsFactory` because it now gets called once for each case in a percent bar chart.

### DIFF
--- a/v3/src/components/graph/models/graph-data-configuration-model.ts
+++ b/v3/src/components/graph/models/graph-data-configuration-model.ts
@@ -324,13 +324,14 @@ export const GraphDataConfigurationModel = DataConfigurationModel
     cellMap: cachedFnWithArgsFactory({
       key: (extraPrimaryAttrRole: AttrRole, extraSecondaryAttrRole: AttrRole,
             binWidth = 0, minValue = 0, totalNumberOfBins = 0) => {
-        return JSON.stringify({
-          ep: extraPrimaryAttrRole, es: extraSecondaryAttrRole,
-          bw: binWidth, min: minValue, tnb: totalNumberOfBins
-        })
-      },
+        return totalNumberOfBins === 0 ? kMain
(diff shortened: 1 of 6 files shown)


kswenson (changes requested):
👍 Generally looks good -- I added a commit which tweaked the implementation in a few places. As noted in our sync review, need to look at the caching in the `cellMap` function, where the `key` function can be simplified and there was some question about whether the fact that it is sometimes called with binning arguments and sometimes without raised the question of whether all clients receive the correct cell map.

bfinzer:
I've changed the key to `cellMap` to be `kMain` if the `totalNumberOfBins === 0`. This isn't actually necessary, since in this case, `invalidateAll` gets called when `extraPrimaryAttrRole` or `extraSecondaryAttrRole` changes. In the other situation, I've left the key dependent on the numeric parameters partly because we don't have an easy way to invalidate when they change and partly because caching the results will probably help responsiveness for large datasets when the user makes a binned plot and drags the bin boundaries.

kswenson (approved):
👍 LGTM -- just one stray comment that seems removable.

PR #1901, opened 2025-04-04
CODAP-375 Added a working (locally) Activity Player Cypress test

nstclair-cc:
[CODAP-375](https://concord-consortium.atlassian.net/browse/CODAP-375)
- Full disclosure, I got help from Cursor to get this test working, mostly help with the iFrames.
- The main line of code I wanted to flag was the uncaught exception handling, I'm not sure that's something we can make an exception for in the case of the Activity Player. There may be a better way to handle the iFrame embedding issue.

## Changes
- Added new test to verify CODAP v3 display within the Activity Player
- Implemented iframe handling for Cypress tests

## Details
Added a new test that verifies CODAP v3 can properly display within the Activity Player. This test:
   - Navigates to the Activity Player
   - Verifies proper loading of the activity
   - Clicks to a specific page
   - Handles iframe content verification

## Testing
The test has been verified to run successfully in headless mode locally. 

Note: This test includes specific handling for uncaught exceptions that occur within the Activity Player, but keeps this exception handling isolated to just this test.

### DIFF
--- a/v3/cypress/e2e/cfm.spec.ts
+++ b/v3/cypress/e2e/cfm.spec.ts
@@ -138,4 +138,50 @@ context("CloudFileManager", () => {
     cfm.getLanguageMenuButton().click()
     cfm.getLanguageMenu().should("not.exist")
   })
+  it("should display in the Activity Player", () => {
+    // Ignore uncaught exceptions from the application for this test only
+    Cypress.on('uncaught:exception', (err) => {
+      // returning false here prevents Cypress from failing the test
+      return false
+    })
+
+    const activityPlayerUrl = Cypress.config("v3ActivityPlayerUrl")
+    cy.visit(activityPlayerUrl)
+
+    // Verify activity player loaded
+    cy.get("[data-cy='activity-title']", { timeout: 10000 })
+      .should("exist")
+      .and("contain.text", "Test CODAP v3")
+
+    cy.get("[data-cy='activity-nav-header']").should("exist")
+
+    // Navigate to page 8
+    cy.get("[data-cy='nav-pages-button']")
+      .filter("[aria-label='Page 8']")
+      .first()
+      .should("have.text", "8")
+      .click()
+      .then(() => {
+        // Wait for page content to load including iframe
+        cy.wait(15000)
+      })
+
+    // Verify iframe exists and has loaded
+    cy.get('iframe')
+      .should('be.visible')
+      .then($iframe => {
+        cy.wrap($iframe)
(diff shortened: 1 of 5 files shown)


v3/cypress/e2e/cfm.spec.ts
```diff
@@ -138,4 +138,50 @@ context("CloudFileManager", () => {
...
     cfm.getLanguageMenu().should("not.exist")
   })
+  it("should display in the Activity Player", () => {
+    // Ignore uncaught exceptions from the application for this test only
+    Cypress.on('uncaught:exception', (err) => {
+      // returning false here prevents Cypress from failing the test
+      return false
+    })
+
+    const activityPlayerUrl = Cypress.config("v3ActivityPlayerUrl")
```
scytacki, line 148: Rather than using Cypress configs for this, I think it'd be better to have a set of global constants perhaps in `cypress/support/constants.ts`. Then in the test you can just import the constants you need for example `import { v3ActivityPlayerUrl } from "../support/constants.ts"` This is more type safe than using the configs, and you don't have to repeat it in each config.

I know CLUE uses this `Cypress.config(...)` approach but that is something I've been meaning to fix, so isn't something to follow.

v3/cypress/e2e/cfm.spec.ts
```diff
@@ -138,4 +138,50 @@ context("CloudFileManager", () => {
...
+
+    // Navigate to page 8
+    cy.get("[data-cy='nav-pages-button']")
+      .filter("[aria-label='Page 8']")
+      .first()
+      .should("have.text", "8")
+      .click()
+      .then(() => {
+        // Wait for page content to load including iframe
+        cy.wait(15000)
```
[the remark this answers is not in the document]
scytacki: This would be a nice change, you might need to increase the timeout for the `get` https://docs.cypress.io/app/core-concepts/retry-ability#Increase-time-to-retry

I think the default is only 5 seconds, but perhaps we increase it to 30s or something.

scytacki (approved):
Looks good.  I just left a few comments that you could address or ignore.

kswenson (commented):
I have nothing to add beyond @scytacki's suggestions.

v3/cypress/e2e/cfm.spec.ts
nstclair-cc: Great, I'll start a Jira story for this one here: https://concord-consortium.atlassian.net/browse/CODAP-602

For some reason this was confusing Cursor and I don't have the bandwidth to address it right now. 🙃

PR #1903, opened 2025-04-08
Removes droppable legend highlight when there are no legends present.

eireland:
Currently, when you drag an attribute into the graph, all droppable areas highlight, which shows an extraneous highlight at the bottom of the graph outside the tile component.
image

This removes it by adding logic to not show the droppable highlight if the graph does not have a legend.


bfinzer (approved):
👍🏻LGTM

PR #1904, opened 2025-04-09
CODAP-373 Bar Chart Formula UI 

bfinzer:
[#CODAP-373] Feature: A formula can be applied to a bar chart

* We modify the `BarChart` component so that it can be triggered to display the formula editor when `BarChartModel.formulaEditorIsOpen` is set to `true`. The component can also provide the necessary handlers. The actual initial triggering is done in `DisplayConfigPalette.handleBreakdownTyeChange`.
* Because I renamed `BarChart.expression` to `formula`, small changes to v2-graph-exporter.ts and v2-plot-importer.ts were required
* When a bar chart has a formula, we display that formula as a label on the secondary axis and we allow the user to click the displayed formula to bring up the formula editor. To bring this about we introduce a `ClickableAxisLabel` component and use that in place of `AttributeLabel` when the plot responds to `axisLabelClickHandler(role)` with something other than undefined.
* We fine tune the radio options so that the breakdownType only gets applied when a formula is accepted and we keep the chosen radio button in sync with the actual breakdownType

### DIFF
--- a/v3/src/components/graph/components/clickable-axis-label.tsx
+++ b/v3/src/components/graph/components/clickable-axis-label.tsx
@@ -1,0 +1,38 @@
+import React, { ForwardedRef, forwardRef, useEffect, useState } from "react"
+import { autorun } from "mobx"
+import { useDataDisplayLayout } from "../../data-display/hooks/use-data-display-layout"
+import { GraphPlace } from "../../axis-graph-shared"
+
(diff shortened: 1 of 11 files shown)


kswenson (approved):
👍 Looks good -- I added a commit with a few tweaks. Take a look and make sure that everything makes sense (and still works).

PR #1906, opened 2025-04-10
Codap-381 Map Pin UI

tealefristoe:
Jira story: https://concord-consortium.atlassian.net/browse/CODAP-381

This PR makes a number of improvements to the UI for map pins:
- Pins now show up as an option in the ruler menu of maps. Pins will only display if this checkbox is checked.
- Pin layers now show up in the layers menu. Pins will only display if this checkbox is checked.
- It is no longer possible to pan the map by clicking on a map pin and dragging.
- Clicking on the map background now deselects all pins.
- The cursor now appears as a colored pin when in add pin mode.
- The add pin button is now disabled when a pin is selected.

Other bug fixes:
- A copy and paste error in my last PR made `MapPinLayerModel.setDataset` set the `lat` and `long` attributes, rather than `pinLat` and `pinLong`. This was actually throwing an error in the console 😬 
- Probably due to a copy and paste error, the log messages for all checkbox toggles in `map-measure-palette.tsx` were `${operation}Grid`, even when toggling points or connecting lines. These actions all log appropriate messages now.


v3/src/components/map/components/pin-controls.scss
```diff
@@ -4,7 +4,7 @@
   left: 10px;
   position: absolute;
   top: 90px;
-  z-index: 11;
+  z-index: 13;
```
kswenson, line 7: This won't have the effect you're expecting because it's within the stacking context set by the `.map-pin-layer`, which is behind the point layer.

kswenson (changes requested):
👍 Looks good expect for the z-index issue that puts the point layer above the pin layer.

kswenson (approved):
👍 LGTM

PR #1908, opened 2025-04-14
fixes bug where slider does not update slider value at the end of the run if the scale type is date

eireland:
We were containing the slider value twice and so the last value of a date scale wasn't getting updated in the slider model.

Demo here: https://codap3.concord.org/branch/fix-slider-data-value/

### DIFF
--- a/v3/src/components/slider/slider-model.ts
+++ b/v3/src/components/slider/slider-model.ts
@@ -78,7 +78,7 @@ export const SliderModel = TileContentModel
 
       // The bounds check for date scale type happens in the use-slider-animation
       // so we skip the bounds check if the axis is a date axis here
-      if (self.multipleOf && self.increment && self.scaleType !== "date") {
+      if (self.multipleOf && self.increment) {
         value = Math.round(value / self.increment) * self.increment
         value = value > self.axis.max ? value - self.increment : value
         value = value < self.axis.min ? value + self.increment : value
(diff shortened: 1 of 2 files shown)


bfinzer (approved):
👍🏻LGTM

v3/src/components/slider/slider-model.ts
```diff
@@ -76,7 +76,9 @@ export const SliderModel = TileContentModel
         else return num
       }
 
-      if (self.multipleOf && self.increment) {
+      // The bounds check for date scale type happens in the use-slider-animation
+      // so we skip the bounds check if the axis is a date axis here
+      if (self.multipleOf && self.increment && self.scaleType !== "date") {
```
kswenson, line 81:
```suggestion
      if (self.multipleOf && self.increment) {
```
As we discussed in sync review, once `setValidatedValue()` is available, this should no longer be necessary.

v3/src/components/slider/slider-model.ts
```diff
@@ -76,6 +76,8 @@ export const SliderModel = TileContentModel
         else return num
       }
 
+      // The bounds check for date scale type happens in the use-slider-animation
+      // so we skip the bounds check if the axis is a date axis here
```
kswenson, line 80: Eliminate this comment as well, since it's related to the now-removed change.

kswenson (approved):
👍 LGTM -- just one stale comment to remove.

PR #1911, opened 2025-04-16
CODAP-601 Bar charts with computed heights

bfinzer:
[#CODAP-601] Feature: A formula for a bar chart computes the heights of the bars

* Introduce `BarChartFormulaAdapter` (modeled on `PlottedValueFormulaAdapter`) to bring about computation of bar heights.
* `BarChartModel` now has a volatile property `barSpecs` in which it stores the results of the formula computation for each bar. Iterating through these can produce min and max values that can be used to provide a default scale for the numeric axis.
* `barCompressionFactorForCase` can detect the need to use the computed value for a bar and number of cases for that bar and produce a compression factor that properly adjusts the thickness of case rectangles.
* `GraphDataConfigurationModel.graphCellKeyFromCaseID` yields the key we need to access the computed value for the bar to which a case belongs.
* `BarChartModel` needs to know about the secondary numeric axis when it has a formula so that it can set the bounds properly in reaction to changes in the min/max values generated for the bars. Accordingly, we pass a numeric axis accessor to the bar chart model that it can use when needed.
* Positioning the "cover" for a computed bar is best done by modifying the cover computed for an ordinary bar using the numeric scale to determine pixel values. 
* When the computed values are < 0 we need to adjust the axes so that the bars can extend below the numeric axis origin.


v3/src/components/graph/adornments/plotted-function/plotted-function-formula-adapter.ts
```diff
@@ -36,7 +36,7 @@ export class PlottedFunctionFormulaAdapter extends BaseGraphFormulaAdapter {
     super(PLOTTED_FUNCTION_FORMULA_ADAPTER, api)
   }
 
-  getAdornment(graphContentModel: IGraphContentModel) {
+  getFormulaSupportingObject(graphContentModel: IGraphContentModel) {
```
kswenson, line 39: As we discussed in slack, I think `getFormulaOwner` would be better.

v3/src/components/graph/adornments/utilities/base-graph-formula-adapter.ts
```diff
@@ -22,15 +20,16 @@ export const getDefaultArgument = (graphContentModel: IGraphContentModel) => {
   return defaultArgumentId ? localAttrIdToCanonical(defaultArgumentId) : undefined
 }
 
-interface IFormulaSupportingAdornment extends IAdornmentModel {
+interface IFormulaSupportingObject {
```
kswenson, line 23: As we discussed in slack, I think `IFormulaOwner` would be better.

kswenson (approved):
👍 Looks good -- I added a commit which adds the `PlotGraphApi` we discussed and tweaked a few other things. There are a few remaining suggestions, the main one being a rename to use `FormulaOwner`.

PR #1910, opened 2025-04-15
[CODAP-607]: Add 'primaryAxis' property to graphComponentHandler 'get' operation.

lublagg:
[CODAP-607](https://concord-consortium.atlassian.net/browse/CODAP-607?atlOrigin=eyJpIjoiNjVjODA2NWFhZjc3NGI1OWE5NmZiMTg0NjgwMzU0NmMiLCJwIjoiaiJ9)

This is being added in order to support the DAVAI plugin + any other plugins that will need to know which axis of the graph currently has the primary role.

PR #1913, opened 2025-04-17
feat: component-level notifications in API (CODAP-609)

emcelroy:
[[CODAP-609](https://concord-consortium.atlassian.net/browse/CODAP-609)]

These changes add V3 notifications for plugins when components are created or deleted. They also add notifications for certain component update actions. For all component, these include updates to the component title. For graphs, these also include changes to attributes on the axes. For numeric axes, these also include changes to domains.

The notifications match the V2 versions except that there is some additional information included. For graph attribute updates, the attribute ID is included as well as the attribute name. For axis updates, the parent component ID is included as well as the parent component type.

[CODAP-609]: https://concord-consortium.atlassian.net/browse/CODAP-609?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/v3/cypress/e2e/smoke/single-codap-smoke.ts
+++ b/v3/cypress/e2e/smoke/single-codap-smoke.ts
@@ -10,7 +10,8 @@ import { CfmElements as cfm } from "../../support/elements/cfm"
 import { AxisHelper as ah } from "../../support/helpers/axis-helper"
 
 
-context("codap single smoke test", () => {
+// TODO: Figure out why opening documents from the CFM is broken in this branch
+context.skip("codap single smoke test", () => {
   beforeEach(function () {
     const url = `${Cypress.config("index")}?mouseSensor&noEntryModal`
     cy.visit(url)
--- a/v3/cypress/e2e/smoke/single-codap-smoke.ts
+++ b/v3/cypress/e2e/smoke/single-codap-smoke.ts
@@ -10,8 +10,7 @@ import { CfmElements as cfm } from "../../support/elements/cfm"
 import { AxisHelper as ah } from "../../support/helpers/axis-helper"
 
 
-// TODO: Figure out why opening documents from the CFM is broken in this branch
-context.skip("codap single smoke test", () => {
+context("codap single smoke test", () => {
   beforeEach(function () {
     const url = `${Cypress.config("index")}?mouseSensor&noEntryModal`
     cy.visit(url)
--- a/v3/src/components/graph/utilities/graph-utils.ts
+++ b/v3/src/components/graph/utilities/graph-utils.ts
@@ -1,23 +1,19 @@
 import {format} from "d3"
 import * as PIXI from "pixi.js"
 import {IDataSet} from "../../../models/data/data-set"
-import { ITileModel } from "../../../models/tiles/tile-model"
-import { toV2Id } from "../../../utilities/codap-utils"
 import {
   defaultSelectedColor, defaultSelectedStroke, defaultSelectedStrokeWidth, defaultStrokeWidth
 } from "../../../utilities/color-utils"
 import { isFiniteNumber } from "../../../utilities/math-utils"
 import { t } from "../../../utilities/translation/translate"
-import {AxisOrientation, ScaleNumericBaseType} from "../../axis/axis-types"
+import {ScaleNumericBaseType} from "../../axis/axis-types"
 import {IAxisModel} from "../../axis/models/axis-model"
 import {isAnyNumericAxisModel} from "../../axis/models/numeric-axis-models"
-import { GraphPlace } from "../../axis-graph-shared"
 import {CaseData} from "../../data-display/d3-types"
 import {Point, PointDisplayType, transitionDuration} from "../../data-display/data-display-types"
 import {IDataConfigurationModel} from "../../data-display/models/data-configuration-model"
 import {IPixiPointMetadata, PixiPoints} from "../../data-display/pixi/pixi-points"
 import { IGraphDataConfigurationModel } from "../models/graph-data-configuration-model"
-import { IGraphContentModel } from "../models/graph-content-model"
 import { GraphLayout } from "../models/graph-layout"
 
 /**
@@ -679,50 +675,4 @@ export const curveBasis = (points: Point[]) => {
   path += pathBasis(p1, p2, p3, p3)
   path += pathBasis(p2, p3, p3, p3)
   return path
-}
-
-const axisPlaceToOrientationMap: Record<GraphPlace, AxisOrientation> = {
-  "bottom": "horizontal",
-  "top": "horizontal",
-  "left": "vertical",
-  "rightCat": "vertical",
-  "rightNumeric": "vertical",
-  "yPlus": "vertical",
-  "plot": "horizontal",
-  "legend": "horizontal"
-}
-
-export interface IUpdateGraphValues {
-  attributeId: string | number
-  attributeName?: string
-  axisOrientation?: string
-  plotType?: string,
-  primaryRole?: string
-}
-
-export const attrChangeNotificationValues = (
-  place: GraphPlace, attrId: string, attrName?: string, attrIdToRemove = "",
-  tile?: ITileModel
-) => {
-
-  const axisOrientation = axisPlaceToOrientationMap[place] || "axis"
-
-  const attributeName = attrIdToRemove
-    ? `Remove ${ axisOrientation === "horizontal" ? "X" : "Y" }: ${attrName}`
-    : attrName
-
-  const graphContent = tile?.content as IGraphContentModel
-  const plotType = graphContent?.plotType
-  const primaryRole = graphContent?.dataConfiguration.primaryRole || "x"
-
-  const values: IUpdateGraphValues = {
-    attributeId: toV2Id(attrId),
-    attributeName,
-    plotType,
-    primaryRole
-  }
-
-  if (place !== "legend") values.axisOrientation = axisOrientation
-
-  return values
 }
--- a/v3/src/components/graph/utilities/graph-utils.ts
+++ b/v3/src/components/graph/utilities/graph-utils.ts
@@ -1,19 +1,23 @@
 import {format} from "d3"
 import * as PIXI from "pixi.js"
 import {IDataSet} from "../../../models/data/data-set"
+import { ITileModel } from "../../../models/tiles/tile-model"
+import { toV2Id } from "../../../utilities/codap-utils"
 import {
   defaultSelectedColor, defaultSelectedStroke, defaultSelectedStrokeWidth, defaultStrokeWidth
 } from "../../../utilities/color-utils"
 import { isFiniteNumber } from "../../../utilities/math-utils"
 import { t } from "../../../utilities/translation/translate"
-import {ScaleNumericBaseType} from "../../axis/axis-types"
+import {AxisOrientation, ScaleNumericBaseType} from "../../axis/axis-types"
 import {IAxisModel} from "../../axis/models/axis-model"
 import {isAnyNumericAxisModel} from "../../axis/models/numeric-axis-models"
+import { GraphPlace } from "../../axis-graph-shared"
 import {CaseData} from "../../data-display/d3-types"
 import {Point, PointDisplayType, transitionDuration} from "../../data-display/data-display-types"
 import {IDataConfigurationModel} from "../../data-display/models/data-configuration-model"
 import {IPixiPointMetadata, PixiPoints} from "../../data-display/pixi/pixi-points"
 import { IGraphDataConfigurationModel } from "../models/graph-data-configuration-model"
+import { IGraphContentModel } from "../models/graph-content-model"
 import { GraphLayout } from "../models/graph-layout"
 
 /**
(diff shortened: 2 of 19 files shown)


v3/src/components/graph/utilities/graph-utils.ts
```diff
@@ -1,18 +1,22 @@
...
 import {IAxisModel, isBaseNumericAxisModel} from "../../axis/models/axis-model"
-import {ScaleNumericBaseType} from "../../axis/axis-types"
+import {AxisOrientation, ScaleNumericBaseType} from "../../axis/axis-types"
+import { GraphPlace } from "../../axis-graph-shared"
 import {CaseData} from "../../data-display/d3-types"
 import {Point, PointDisplayType, transitionDuration} from "../../data-display/data-display-types"
 import {IDataConfigurationModel} from "../../data-display/models/data-configuration-model"
 import {IPixiPointMetadata, PixiPoints} from "../../data-display/pixi/pixi-points"
 import { IGraphDataConfigurationModel } from "../models/graph-data-configuration-model"
+import { IGraphContentModel } from "../models/graph-content-model"
```
emcelroy, line 19: This is causing a circular dependency warning. I'm leaving it as is till I'm more certain of the overall structure of things.

v3/src/components/graph/utilities/graph-utils.ts
```diff
@@ -675,3 +679,49 @@ export const curveBasis = (points: Point[]) => {
...
+
+  const axisOrientation = axisPlaceToOrientationMap[place] || "axis"
+
+  const attributeName = attrIdToRemove
+    ? `Remove ${ axisOrientation === "horizontal" ? "X" : "Y" }: ${attrName}`
+    : attrName
+
+  const graphContent = tile?.content as IGraphContentModel
+  const plotType = graphContent?.plotType
+  const primaryRole = graphContent?.dataConfiguration.primaryRole || "x"
```
emcelroy, line 715: I'm especially unsure of including these values here. They're not included in V2 notifications about attribute changes. 

The information will be useful for the DAVAI plugin. I'm not sure how else to send notifications about changes to those things aside from including them here. I think the plugin could make a follow-up `get` request for that information after receiving an attribute change notification. From the plugin's perspective it'd be better to not have to do that. I'm not sure that's a good enough reason to add this information here, though.

v3/src/components/graph/utilities/graph-utils.ts
```diff
@@ -675,3 +679,49 @@ export const curveBasis = (points: Point[]) => {
...
+  plotType?: string,
+  primaryRole?: string
+}
+
+export const attrChangeNotificationValues = (
+  place: GraphPlace, attrId: string, attrName?: string, attrIdToRemove = "",
+  tile?: ITileModel
+) => {
+
+  const axisOrientation = axisPlaceToOrientationMap[place] || "axis"
```
tealefristoe, line 707: I'm surprised to see a default here that doesn't appear in `axisPlaceToOrientationMap`.

v3/src/data-interactive/notification-utils.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export const notification = (operation: string, values: any, tile: ITileModel, _callback?: (result: any) => void) => {
+  const resource = operation === "titleChange" ? `component[${toV2Id(tile.id)}]` : "component"
+  values.operation = operation
+  if (tile?.title && operation === "delete") values.title = tile.title
+  if (tile?.name && operation === "delete") values.name = tile.name
+  const callback = _callback ?? makeCallback(operation)
+
+  return { message: { action, resource, values }, callback }
+}
```
kswenson, line 20: Why is this separated out into its own file? Seems like it's only useful to component/tile notifications, so it should just live in the `tile-notifications.ts` module.

v3/src/components/graph/utilities/graph-utils.ts
```diff
@@ -675,3 +679,49 @@ export const curveBasis = (points: Point[]) => {
...
+
+  const attributeName = attrIdToRemove
+    ? `Remove ${ axisOrientation === "horizontal" ? "X" : "Y" }: ${attrName}`
+    : attrName
+
+  const graphContent = tile?.content as IGraphContentModel
```
tealefristoe, line 713: Could you use `isGraphContentModel` instead of casting here?

v3/src/components/graph/utilities/graph-utils.ts
tealefristoe, line 715: I'm not sure about whether it would be good to include this here or not. But I will say that in general, most plugins seem to ignore the content of notifications they receive, instead treating them as a signal that they need to send a `get` request to retrieve the information they care about.

v3/src/data-interactive/handlers/component-handler.ts
```diff
@@ -63,10 +66,10 @@ export const diComponentHandler: DIHandler = {
         (_type, createOrShowOptions) => document.content?.createOrShowTile(_type, createOrShowOptions)
       const _createOrShow = createOrShow ?? defaultCreateOrShow
       const _options = options ?? {}
+      const title = _title ?? name
+      const newTileOptions = { cannotClose, content, ...dimensions, name, title, ..._options }
+      const tile = _createOrShow(kComponentTypeV2ToV3Map[type], newTileOptions)
```
kswenson, line 71: The model changes should be made within the call to `applyModelChange()`, so these lines should move back where they were. You can use the same approach to getting access to the tile that is used elsewhere:
```
    let tile: Maybe<ITileModel>
```
and then set it.

v3/src/components/graph/utilities/graph-utils.ts
tealefristoe, line 19: Yeah, it seems like these additions should go into a new file, maybe `graph-notifications.ts` or `graph-notification-utils.ts`.

v3/src/components/graph/utilities/graph-utils.ts
kswenson: I don't have a problem with returning additional information if it's helpful. In #1910 the convention established is that the property returned is named `primaryAxis`, so we should maintain that convention here.

v3/src/components/tool-shelf/tool-shelf.tsx
```diff
@@ -119,10 +121,16 @@ export const ToolShelf = observer(function ToolShelf({ document }: IProps) {
   function handleTileButtonClick(tileType: string) {
     const tileInfo = getTileComponentInfo(tileType)
     const { undoStringKey = "", redoStringKey = "" } = tileInfo?.shelf || {}
+    let tile: ITileModel | undefined = undefined
     document?.content?.applyModelChange(() => {
-      document?.content?.createOrShowTile?.(tileType, { animateCreation: true })
-    }, { undoStringKey, redoStringKey, log: logMessageWithReplacement("Create component: %@",
-                                                {tileType}, "component")})
+      tile = document?.content?.createOrShowTile?.(tileType, { animateCreation: true })
+      return tile
```
tealefristoe, line 127: Just noting that the previous code didn't actually return a value.

v3/src/components/graph/utilities/graph-utils.ts
kswenson: Rather than putting `attrChangeNotificationValues` and friends in this module, I'd put them in a new `graph-notifications.ts` module, which should avoid the circular reference. This module is too low-level to contain a function that requires access to the `GraphContentModel`.

v3/src/components/graph/utilities/graph-utils.ts
kswenson, line 707: `"axis"` is not a valid value for orientation. I would think you would want to either leave it undefined or return a valid orientation.

v3/src/components/graph/utilities/graph-utils.ts
```diff
@@ -675,3 +679,49 @@ export const curveBasis = (points: Point[]) => {
...
+
+const axisPlaceToOrientationMap: Record<GraphPlace, AxisOrientation> = {
+  "bottom": "horizontal",
+  "top": "horizontal",
+  "left": "vertical",
+  "rightCat": "vertical",
+  "rightNumeric": "vertical",
+  "yPlus": "vertical",
+  "plot": "horizontal",
+  "legend": "horizontal"
```
kswenson, line 691: `"plot"` and `"legend"` don't really have an orientation, so these might be better left `undefined`.

v3/cypress/e2e/smoke/single-codap-smoke.ts
```diff
@@ -10,7 +10,8 @@ import { CfmElements as cfm } from "../../support/elements/cfm"
 import { AxisHelper as ah } from "../../support/helpers/axis-helper"
 
 
-context("codap single smoke test", () => {
+// TODO: Figure out why opening documents from the CFM is broken in this branch
+context.skip("codap single smoke test", () => {
```
kswenson, line 14: Is it broken for everyone on this branch or only on your machine? If it's only your machine this should not be included in the PR.

kswenson (changes requested):
👍 Generally looks good, but a number of suggestions.

v3/src/models/tiles/tile-notifications.ts
```diff
@@ -0,0 +1,45 @@
...
+}
+
+interface IUpdateTileNotificationValues {
+  attributeName?: string
+  axisOrientation?: string
+  id?: number
+  type?: string
+}
+
+export function updateTileNotification(updateType: string, tile?: ITileModel, values?: any) {
```
tealefristoe, line 33: Just want to point out the order of parameters is different between this function and `notification`. It would be good to keep them in sync if at all possible.

v3/src/data-interactive/notification-utils.ts
tealefristoe, line 20: As mentioned in the comment on the whole file, this function should be renamed `tileNotification`, moved to `tile-notifications.ts`, and it should call the generic `notification` function that you define in this file.

Additionally, you should make a couple of changes:
- You should define `id` and `type` here, then add them to `values`, instead of doing that separately in `create`, `delete`, and `updateTileNotification`.
- You should define `values.title` and `values.name` in `deleteTileNotification` instead of in this generic function.

v3/src/data-interactive/notification-utils.ts
tealefristoe, line 1: This is a good direction to go, but I think the abstraction is a bit off. This file should contain a very generic `notification` function, which just takes `operation`, `resource`, `_values` that defaults to `{}`, an optional `_callback`, and an `action` that defaults to `"notify"`. All this function should do is:
- Combine `operation` and `_values` into `values`.
- Sets `callback` to be either `_callback` or `makeCallback(operation)`.
- Returns the complete package: `{ message: { action, resource, values }, callback }`.

Then `notification` in `data-set-notifications.ts` (which should be renamed `dataSetNotification`) should call this function. And you should introduce a new `tileNotification` function in `tile-notifications.ts` that calls the generic `notification` as well.

tealefristoe (approved):
This is looking good 👍 

I left quite a few suggestions that I think are worth addressing. I also think you should figure out the test you're skipping before merging, as well as getting the test coverage up to the required amount.

I'm accepting since I'll be gone for a week and I don't want a request for changes to complicate merging the PR when it's ready to go.

v3/cypress/e2e/smoke/single-codap-smoke.ts
emcelroy: @kswenson It's very strange. There's a comment about it [in this Slack thread](https://concord-consortium.slack.com/archives/C035J6RDAK0/p1744921733378039?thread_ts=1744640961.384969&cid=C035J6RDAK0). It's broken when the Cypress tests run both here and on my machine. I skipped the tests for now so it would deploy. Oddly enough, I can open documents on the deployed branch even though I can't in my local instance of CODAP, on this branch or `main`.

emcelroy: Turns out the [Cypress test was failing](https://cloud.cypress.io/projects/msrfxa/runs/7095/test-results/111da057-1faa-4a29-ad75-d9d365f4180d?actions=%5B%5D&browsers=%5B%5D&groups=%5B%5D&isFlaky=%5B%5D&modificationDateRange=%7B%22startDate%22%3A%221970-01-01%22%2C%22endDate%22%3A%222038-01-19%22%7D&orderBy=EXECUTION_ORDER&oses=%5B%5D&specs=%5B%5D&statuses=%5B%7B%22value%22%3A%22FAILED%22%2C%22label%22%3A%22FAILED%22%7D%5D&testingTypesEnum=%5B%5D&utm_source=github) in a different way here than locally. I believe I've fixed the underlying problem.

v3/src/components/tool-shelf/tool-shelf.tsx
```diff
@@ -119,10 +121,15 @@ export const ToolShelf = observer(function ToolShelf({ document }: IProps) {
   function handleTileButtonClick(tileType: string) {
     const tileInfo = getTileComponentInfo(tileType)
     const { undoStringKey = "", redoStringKey = "" } = tileInfo?.shelf || {}
+    let tile: ITileModel | undefined = undefined
```
kswenson, line 124: Could use:
```suggestion
    let tile: Maybe<ITileModel>
```

kswenson (approved):
👍 LGTM -- just one small suggested tweak

PR #1914, opened 2025-04-18
CODAP-605 Date Slider API

scytacki:
This also adds support for updating sliders, and adds validation of many of the properties sent during create or update.

A question about the new fields is the use of strings for values. The existing fields of animationDirection and animationMode use indexes. The strings are consistent with the underlying modeling and in other parts of the CODAP api string are used.

CODAP-605


v3/src/components/slider/slider-component-handler.ts
```diff
@@ -0,0 +1,202 @@
...
+        return {error: errorResult(t("V3.DI.Error.noMultipleSliders", { vars: [globalValueName] }))}
+      }
+    }
+  }
+
+  if (dateMultipleOfUnit != null) {
+    if (!dateUnits.includes(dateMultipleOfUnit as any)) {
+      return {error: errorResult(t("V3.DI.Error.unsupportedDateUnit", { vars: [dateMultipleOfUnit] }))}
+    }
+    snapshot.dateMultipleOfUnit = dateMultipleOfUnit as ISliderSnapshot["dateMultipleOfUnit"]
```
tealefristoe, line 61: I feel like there must be a way to do this where it's not necessary to cast variables like this and in the conditional below. But maybe it's more work to set that up than it's worth.

v3/src/components/slider/slider-component-handler.ts
```diff
@@ -0,0 +1,202 @@
...
+    if (error) {
+      return error
+    }
+    const {
+      animationDirection,
+      animationMode,
+      dateMultipleOfUnit,
+      globalValue,
+      scaleType,
+    } = validatedValues as Partial<ISliderSnapshot>
```
tealefristoe, line 108: I think the only reason to cast `validatedValues` here is if it's `undefined`. So I think it would be better to throw an error if `validateValues` doesn't exist (which will probably never happen if there's no error but who knows what the future will bring). Then you wouldn't have to cast here. I think this will cause an error if `validatedValues` is ever `undefined`.

v3/src/components/slider/slider-component-handler.test.ts
```diff
@@ -0,0 +1,435 @@
...
+
+    const result = handler.create!({}, {
+      type: "slider",
+      globalValueName: "nonexistentGlobalValue",
+    })
+
+    expect(result.success).toBe(false)
+    if (result.success) {
+      throw new Error("Expected result to be unsuccessful")
+    }
```
tealefristoe, line 63: Are these extra checks necessary? Seems like if the result is anything other than `false`, the above check will fail and we'll bail on the test anyway.

tealefristoe (approved):
Looks good 👍 

I left a few suggestions, mostly related to typing, but I don't know if it's actually worth the time it would take to address them.

v3/src/components/slider/slider-component-handler.ts
kswenson: I think the way to clean this up, including the `any` a few lines above, is to implement an `isDateUnit()` type guard function which takes a string and return `value is DateUnit`.

v3/src/components/slider/slider-component-handler.test.ts
scytacki: They narrow the type of the result so the code below that calls `result.values!.error` is allowed by typescript. I could get rid of the `expect(result.success).toBe(false)` so at least it isn't redundant. However in that case if the code changes and the test fails, what gets printed out by Jest isn't as easy to understand.

v3/src/components/slider/slider-component-handler.ts
scytacki: Yeah I can do that. It adds more lines of code and still requires a cast to do the `includes` check inside of `isDateUnit()`, but it does look nicer. And perhaps the isDateUnit will come in handy some other time.

kswenson: True, but you can cast the other way: `(dateUnits as string[]).includes(value)` which seems nicer because it's a widening cast instead of a narrowing one, i.e. it assumes less.

v3/src/components/slider/slider-component-handler.test.ts
scytacki: This issue has more info on this: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/41179
But what isn't mentioned is the impact using the helper functions have on the Jest output when the tests fails.

scytacki:
Thank you both for the reviews.

v3/src/components/slider/slider-component-handler.ts
scytacki: I found a better way to do it, but I already merged the PR.

The `(dateUnits as string[]).includes(value)` isn't better because the user could pass almost anything as value since it is JSON object from a plugin in this case.

So the best approach seems to be:
```
export function isDateUnit(value: any): value is DateUnit {
  return dateUnits.includes(value)
}
```

When I ended up merging was:
```
export function isDateUnit(value: any): value is DateUnit {
  return dateUnits.includes(value as DateUnit) 
```
That `as DateUnit` is unnecessary.

PR #1918, opened 2025-04-21
CODAP-95 Graph display problems with changes in hierarchy

bfinzer:
[#CODAP-361] Bug fix: Graph legend incorrect after hierarchy change in table

[#CODAP-95] Bug fix: Legend colors not being shown, and graph not adapting to change in plotted attributes properly

* There are a number of bugs to fix here
  * In the Mammals sample document dragging **Diet** to the left to create a parent collection should be changing the number of cases in the **Diet** graph to 3, but it leaves all of them in place.
  * The data tip for points in graph where the cases are not childmost should not display values for childmost attributes; e.g. when there is a legend from an attribute that is more childmost than the axis attributes.
  * When the legend attribute in a graph happens to be more childmost than the axis attributes, nothing should be displayed in the category or numeric portion of the legend.

_Fixes_
* When we're adding `FilteredCases` to the `DataConfigurationModel` we should only be using attributes assigned to axes to determine the childmost collection.
* `DataConfigurationModel` `tipAttributes` should not include attributes that are more childMost than the axis attributes.
* The Legend component only displays if the legend attribute does not belong to a collection more childmost than the childmost collection determined from axis attributes.

### DIFF
--- a/v3/src/components/data-display/components/legend/legend.tsx
+++ b/v3/src/components/data-display/components/legend/legend.tsx
@@ -38,7 +38,7 @@ export const Legend = observer(function Legend({
   const dataConfiguration = useDataConfigurationContext(),
     legendID = dataConfiguration?.attributeID("legend"),
     legendRef = useRef() as React.RefObject<SVGSVGElement>
-    if (!dataConfiguration?.isAllowableNonAxisAttribute(legendID)) return null
+  if (!dataConfiguration?.isAttributeAllowedForNonAxisRole(legendID)) return null
   const attrType = dataConfiguration?.attributeType('legend'),
     LegendComponent = dataConfiguration && legendComponentManager.getLegendComponent(dataConfiguration)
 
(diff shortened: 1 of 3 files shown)


v3/src/components/data-display/components/legend/legend.tsx
```diff
@@ -36,9 +36,11 @@ export const Legend = observer(function Legend({
                                         layerIndex, setDesiredExtent, onDropAttribute
                                       }: ILegendProps) {
   const dataConfiguration = useDataConfigurationContext(),
-    attrType = dataConfiguration?.attributeType('legend'),
-    LegendComponent = dataConfiguration && legendComponentManager.getLegendComponent(dataConfiguration),
+    legendID = dataConfiguration?.attributeID("legend"),
     legendRef = useRef() as React.RefObject<SVGSVGElement>
+    if (!dataConfiguration?.isAllowableNonAxisAttribute(legendID)) return null
```
kswenson, line 41:
```suggestion
  if (!dataConfiguration?.isAllowableNonAxisAttribute(legendID)) return null
```

kswenson (changes requested):
👍 Looks good -- just a handful of suggestions.

kswenson (approved):
👍 LGTM

PR #1919, opened 2025-04-21
CODAP-86 Fix to ruler menu in presence of binned dot plot

bfinzer:
[#CODAP-86] Bug fix: The ruler menu for binned dot plots should only show Count and Percent

* This was fixed with a trivial change in adornment-ui-types.ts


kswenson (approved):
👍 LGTM

PR #1920, opened 2025-04-22
Change number of datasets for plugin created graphs

eireland:
Limits the dataset context added to a plugin created graph to the one specified by the plugin.

Having multiple datasets in the document with the NASA Earth Data plugin was causing problems when creating a graph from the plugin. The DI graph component handler was creating layers for each of the dataset, while CODAP only recognizes the first dataset in the data configuration model.

### DIFF
--- a/v3/src/components/graph/graph-component-handler.ts
+++ b/v3/src/components/graph/graph-component-handler.ts
@@ -22,7 +22,6 @@ import { IGraphDataConfigurationModel, kGraphDataConfigurationType } from "./mod
 import { GraphLayout } from "./models/graph-layout"
 import { syncModelWithAttributeConfiguration } from "./models/graph-model-utils"
 import { IGraphPointLayerModelSnapshot, kGraphPointLayerType } from "./models/graph-point-layer-model"
-import { a } from "react-dom-factories"
 
 interface AttributeInfo {
   id?: number | null
--- a/v3/src/components/graph/graph-component-handler.ts
+++ b/v3/src/components/graph/graph-component-handler.ts
@@ -22,6 +22,7 @@ import { IGraphDataConfigurationModel, kGraphDataConfigurationType } from "./mod
 import { GraphLayout } from "./models/graph-layout"
 import { syncModelWithAttributeConfiguration } from "./models/graph-model-utils"
 import { IGraphPointLayerModelSnapshot, kGraphPointLayerType } from "./models/graph-point-layer-model"
+import { a } from "react-dom-factories"
 
 interface AttributeInfo {
   id?: number | null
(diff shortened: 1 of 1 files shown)


v3/src/components/graph/graph-component-handler.ts
```diff
@@ -84,7 +84,12 @@ export const graphComponentHandler: DIComponentHandler = {
     const layers: Array<IGraphPointLayerModelSnapshot> = []
     let provisionalDataSet: IDataSet | undefined
     let provisionalMetadata: IDataSetMetadata | undefined
-    getSharedDataSets(appState.document).forEach(sharedDataSet => {
+    const sharedDataSets = getSharedDataSets(appState.document)
+    // We currently only support one dataset in a graph, so we find the one specified by plugin
+    const sharedDataSet = sharedDataSets.find(sd => {
+      return sd.dataSet.name === _dataContext
```
kswenson, line 90: We should match on id or title as well:
```suggestion
      return sd.dataSet.matchTitleOrNameOrId(_dataContext)
```

v3/src/components/graph/graph-component-handler.ts
```diff
@@ -84,7 +84,12 @@ export const graphComponentHandler: DIComponentHandler = {
     const layers: Array<IGraphPointLayerModelSnapshot> = []
     let provisionalDataSet: IDataSet | undefined
     let provisionalMetadata: IDataSetMetadata | undefined
-    getSharedDataSets(appState.document).forEach(sharedDataSet => {
+    const sharedDataSets = getSharedDataSets(appState.document)
+    // We currently only support one dataset in a graph, so we find the one specified by plugin
```
kswenson, line 88: Also, if no `DataSet` is specified in the API and there's only one `DataSet` in the document, we should default to it (as graphs created within CODAP by the user do).

kswenson (changes requested):
👍 Nice improvement -- couple of suggestions to make it even better.

v3/src/components/graph/graph-component-handler.ts
```diff
@@ -22,6 +22,7 @@ import { IGraphDataConfigurationModel, kGraphDataConfigurationType } from "./mod
 import { GraphLayout } from "./models/graph-layout"
 import { syncModelWithAttributeConfiguration } from "./models/graph-model-utils"
 import { IGraphPointLayerModelSnapshot, kGraphPointLayerType } from "./models/graph-point-layer-model"
+import { a } from "react-dom-factories"
```
kswenson, line 25: Remove extraneous auto-import.
```suggestion
```

kswenson (approved):
👍 LGTM -- just one stray import to remove.

PR #1923, opened 2025-04-24
Codap 7 - Adds connecting line plugin API affordances

eireland:
NEO Earth plugin required connecting lines to be the same color as the points as specified by the map pin color. This adds a check to see if legend attribute is of color type, and if it is use, that color for the connecting line.

This also adds the ability to update the graph to show the connecting lines through the API.

### DIFF
--- a/v3/src/components/graph/graph-component-handler.ts
+++ b/v3/src/components/graph/graph-component-handler.ts
@@ -220,6 +220,15 @@ export const graphComponentHandler: DIComponentHandler = {
         }
       })
     }
+
+    // TODO: There is a race condition that causes a d3 error when we try to
+    // add the connecting lines on graph creation. So disabling this feature
+    // for now, and we use update graph to enable it.
+    // Enable connecting lines after the graph model is fully initialized
+    // if (showConnectingLines !== undefined) {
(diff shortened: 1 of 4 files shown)


v3/src/components/graph/graph-component-handler.ts
```diff
@@ -84,7 +84,12 @@ export const graphComponentHandler: DIComponentHandler = {
     const layers: Array<IGraphPointLayerModelSnapshot> = []
     let provisionalDataSet: IDataSet | undefined
     let provisionalMetadata: IDataSetMetadata | undefined
-    getSharedDataSets(appState.document).forEach(sharedDataSet => {
+    const sharedDataSets = getSharedDataSets(appState.document)
+    // We currently only support one dataset in a graph, so we find the one specified by plugin
+    const sharedDataSet = sharedDataSets.find(sd => {
+      return sd.dataSet.name === _dataContext
+    })
+    if (sharedDataSet) {
```
kswenson, line 92: When you encounter the inevitable conflicts, take the version of the code from #1920 which matches on name, title, or id and defaults to an existing data set if there's only one.

kswenson (approved):
👍 LGTM -- where conflicts arise make sure you take the reviewed code from #1920.

PR #1926, opened 2025-04-27
CODAP-617 Even out the binned dot plot stacks

bfinzer:
[#CODAP-617] Bug fix: Binned dot plot stacks of points should be even

* The reason for the unevenness in the stacks of points was that the algorithm was adding points one column at a time until the points reached the end of the available space and then starting over again with a new column.
* The correct algorithm is similar to that used in displaying dot charts where we compute a number of points per row and add points essentially a row at a time.
* The code is complex, partly because it mixes together dot plots, binned dot plots and histograms. I considered a complete refactor but then found there was a reasonably straightforward way to modify the existing code.
* Fixed an `mstReaction` in `useBinnedPlotResponders` to get disposed when the binned plot component goes away.
* Fixed a problem whereby the lower/upper bounds of the binning axis was different than that specified by the bins
* By measuring the tick labels for a horizontal binning axis we can detect overlap and rotate the labels. And we have to respond to that by refreshing the points.
* There's some subtlety involved in making sure each of the bin boundaries has drag handlers since they come and go during a drag.

### DIFF
--- a/v3/src/components/axis/helper-models/numeric-axis-helper.ts
+++ b/v3/src/components/axis/helper-models/numeric-axis-helper.ts
@@ -74,7 +74,7 @@ export class NumericAxisHelper extends AxisHelper {
     } else if (!hasDraggableNumericAxis) {
       const formatter = (value: number) => this.multiScale?.formatValueForScale(value) ?? ""
       const {tickValues, tickLabels} = this.axisProvider.nonDraggableAxisTicks(formatter) ||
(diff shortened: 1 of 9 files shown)


kswenson (approved):
👍 Looks good! I added a couple of commits to fix some lint warnings and fix some white space issues. The one non-trivial change involved using a ref for one of the callbacks rather than a standard JS variable.

kswenson (approved):
👍 LGTM

PR #1927, opened 2025-04-29
CODAP-71 Multiple connecting lines colors match points

bfinzer:
[#CODAP-71] Bug fix: Connecting lines for multiple y-attributes should match point colors

* The fix of the quoted bug is a one-liner in use-connecting-lines.ts.
* But there are problems with the size of points whereby the points get scaled smaller more than once, and then, when connecting lines are removed, they get made too big.


kswenson (approved):
👍 LGTM

PR #1930, opened 2025-04-30
Adds tile name, and rescaleAxes to graph component handler

eireland:
NEO Plugin needs to rescale the graph component when data is updated.

### DIFF
--- a/v3/src/components/graph/graph-component-handler.ts
+++ b/v3/src/components/graph/graph-component-handler.ts
@@ -74,7 +74,7 @@ export const graphComponentHandler: DIComponentHandler = {
   create({ values }) {
     const {
       backgroundColor, dataContext: _dataContext, displayOnlySelectedCases, enableNumberToggle: showParentToggles,
-      filterFormula, hiddenCases: _hiddenCases, numberToggleLastMode: showOnlyLastCase, pointColor,
+      filterFormula, hiddenCases: _hiddenCases, name, numberToggleLastMode: showOnlyLastCase, pointColor,
       pointSize, showMeasuresForSelection, strokeColor, strokeSameAsFill, transparent,
       yAttributeID, yAttributeIDs, yAttributeName, yAttributeNames, yAttributeType
     } = values as V2Graph
@@ -120,8 +120,8 @@ export const graphComponentHandler: DIComponentHandler = {
               if (attribute) _yAttributeDescriptions.push({ attributeID: attribute.id, type: _yAttributeType })
             })
           } else if (yAttributeNames) {
-            yAttributeNames.forEach(name => {
-              const attribute = dataset.getAttributeByName(name)
+            yAttributeNames.forEach(aName => {
+              const attribute = dataset.getAttributeByName(aName)
               if (attribute) _yAttributeDescriptions.push({ attributeID: attribute.id, type: _yAttributeType })
             })
           } else {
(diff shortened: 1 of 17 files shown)


v3/src/components/graph/graph-component-handler.ts
```diff
@@ -437,6 +438,7 @@ export const graphComponentHandler: DIComponentHandler = {
     updateBounds("bottom", xLowerBound, xUpperBound)
     updateBounds("left", yLowerBound, yUpperBound)
     updateBounds("rightNumeric", y2LowerBound, y2UpperBound)
+    if (rescaleAxes != null) content.rescale()
```
tealefristoe, line 441: I think this will rescale if `false` is passed for `rescaleAxes`, which seems wrong. Maybe you should check for truthiness instead of non-null.

I also think it's worth considering if this should happen in an `update` request. This seems like the sort of thing that might make sense as a `notify` request, which generally work as commands rather than setting values like `update`.

tealefristoe (changes requested):
`create component` requests already handle `name`. You can test this yourself:
- Open mammals with the API tester: [https://codap3.concord.org/branch/main/#file=examples:Mammals
- Add the API tester to the document: ](https://codap3.concord.org/branch/main/?di=https%3A%2F%2Fconcord-consortium.github.io%2Fcodap-data-interactives%2FDataInteractiveAPITester%2Findex.html#file=examples:Mammals)
- Use the API tester to send the default `component create Create Graph` command.
- If you're still skeptical, try the default `component get Get Component` command.

v3/src/components/graph/graph-component-handler.ts
kswenson: >I also think it's worth considering if this should happen in an update request.

That sounds right to me now that you mention it.

tealefristoe:
@eireland I was wrong before--you can't set the `name` through an `update component` request. Sorry for the confusion. But there are still two things to consider here:
- Should it be possible to change the name? The way names are handled in Codap is inconsistent, but my understanding is that they are supposed to remain constant so they can always refer to the same thing. It's generally better to refer to components via id anyway, so you should try to do that in this plugin if possible.
- If it does make sense to allow changing names, this should happen in the general component handler (`component-handler.ts`), not in the graph specific component handler. It would make sense to follow the same pattern as is used for `title`.

kswenson:
> @eireland I was wrong before--you can't set the `name` through an `update component` request. Sorry for the confusion. But there are still two things to consider here:
> 
> * Should it be possible to change the name? The way names are handled in Codap is inconsistent, but my understanding is that they are supposed to remain constant so they can always refer to the same thing. It's generally better to refer to components via id anyway, so you should try to do that in this plugin if possible.
> * If it does make sense to allow changing names, this should happen in the general component handler (`component-handler.ts`), not in the graph specific component handler. It would make sense to follow the same pattern as is used for `title`.

My understanding is that `name`s should _not_ be changed after creation and so changing the `name` should not be supported by `update`. It _should_ be possible to set the `name` via `create` and it should be possible to set the `_title` (or perhaps `title` for API purposes) via `update`, and API requests should match on `id`, `name`, or `title`.

v3/src/data-interactive/handlers/component-handler.ts
```diff
@@ -141,9 +142,16 @@ export const diComponentHandler: DIHandler = {
...
-      uiState.setFocusedTile(component.id)
-    // } else if (request === "autoScale") { // TODO Handle autoScale requests
+    switch (request) {
+      case "select":
+        uiState.setFocusedTile(component.id)
+        break
+      case "autoScale":
+        if (component.content.type === "graph") {
+          const graphContent = component.content as IGraphContentModel
+          graphContent.rescale()
```
tealefristoe, line 152:
```suggestion
        const { content } = component
        if (isGraphContentModel(content) {
          content.rescale()
```

tealefristoe (approved):
You should use `isGraphContentModel` instead of checking for a hard coded string, but otherwise looks good. Glad to see the API getting filled out a bit with `autoScale`!

PR #1931, opened 2025-05-02
CODAP-654 Fixes for various graph undo/redo problems

bfinzer:
[#CODAP-654] Bug fix: Undo/Redo tool tips aren't showing/hiding connecting lines

* The clickHandler for the showConnectingLines checkbox should use an applyModelChange with appropriate notification, undo/redo strings and log message
* Using applyModelChange required passing in a ITileModel to getAdornmentMenuItems and down to getAdornmentsMenuItemsFromTheStore
* All the rest of the clickHandlers in adornments-store-utils.ts also required modification to use applyModelChange
* On another undo/redo front, the useEffect refreshInterceptLockChange in lsrl-adornment-component.tsx was setting up an mstReaction that was causing a model action as part of its accessor and thus causing another undo to be added to the history.
* When a document with a LSRL adornment was opened, undo actions were immediately added to the undo history. This was because the `lines` were a property of the model instead of volatile. Constructing them during document initialization was causing actions to be added to the undo history.
* In an aside, we added the `saveAsV2` debug property to README.md

### DIFF
--- a/v3/src/components/graph/adornments/store/adornments-store.test.ts
+++ b/v3/src/components/graph/adornments/store/adornments-store.test.ts
@@ -104,7 +104,7 @@ describe("AdornmentsStore", () => {
   })
   it("returns a list of adornment items for use in menus", () => {
     const adornmentsStore = AdornmentsStore.create()
-    const adornmentsMenuItems = adornmentsStore.getAdornmentsMenuItems("dotPlot", false)
+    const adornmentsMenuItems = adornmentsStore.getAdornmentsMenuItems(undefined,"dotPlot", false)
     expect(adornmentsMenuItems).toBeDefined()
     expect(adornmentsMenuItems?.length).toBeGreaterThan(0)
     expect(adornmentsMenuItems?.[0].title).toBe("DG.Inspector.graphCount")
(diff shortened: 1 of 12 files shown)


v3/src/components/graph/adornments/store/adornments-store.test.ts
```diff
@@ -104,7 +104,7 @@ describe("AdornmentsStore", () => {
   })
   it("returns a list of adornment items for use in menus", () => {
     const adornmentsStore = AdornmentsStore.create()
-    const adornmentsMenuItems = adornmentsStore.getAdornmentsMenuItems("dotPlot", false)
+    const adornmentsMenuItems = adornmentsStore.getAdornmentsMenuItems(undefined,"dotPlot", false)
```
kswenson, line 107:
```suggestion
    const adornmentsMenuItems = adornmentsStore.getAdornmentsMenuItems(undefined, "dotPlot", false)
```

kswenson (approved):
👍 Looks good -- just some straightforward suggestions around simplifying dependencies for `useCallback`/`useEffect`. Let me know if you want to discuss further.

PR #1932, opened 2025-05-02
CODAP-366-Fix parent toggles crash

bfinzer:
[#CODAP-366] Bug fix: Changing leftmost attribute while parent toggles are on crashes CODAP

* The problem was that a reference to the lastVisibleIndex was not being kept current with the actual number of parent toggle buttons, resulting in a crash when indexing beyond the bounds of the array of buttons. The fix is to reset the lastVisibleIndex whenever the array of buttons is reconstituted to make sure it is not greater than the array bounds.


kswenson (approved):
👍 LGTM

kswenson (changes requested):
👍 On further inspection, code looks fine but there are failing cypress tests that seem related.

kswenson (approved):
👍 Looks good -- nice win for the cypress tests.

PR #1933, opened 2025-05-02
Fix for invalid transition from bars to points

bfinzer:
[#CODAP-365] Bug fix: Bar chart point compression when changing graph axis

* A performance-oriented shortcut was getting in the way of properly matching the pixiPoint points with the current cases being plotted in the situation when the attribute change takes us from a child collection to a parent collection. By adding the requirement in the shortcut that the number of cases being plotted equals the current number of pixi points, we still allow the shortcut for the critical situations but bypass it when we really need to reconstitute the pixi points.


kswenson (approved):
👍 LGTM

PR #1934, opened 2025-05-07
CODAP-660 Fix binned dot plot display

bfinzer:
[#CODAP-660] Bug fix: Binning dot plot in fresh graph doesn't show all points

* The bug was due to the need for a rerender when the rotation of the bin labels changing, so we add a reference to primaryAxisMode.labelsAreRotated in the BinnedDotPlot component so that it can be observed and produce the desired rerender. It turns out that this fixes another, unrecorded bug, whereby points were disappearing off the top of a binned dot plot because overlap was not being correctly computed.


kswenson (approved):
👍 LGTM

PR #1935, opened 2025-05-08
CODAP-97 Case plots response to change in number of cases bug fix

bfinzer:
[#CODAP-97] Bug fix: New points in case plots appear on top of each other at top left

* Case plots were not paying attention to changes in the number of cases being plotted. We fix this by adding an mstReaction whose accessor is the data configuration's caseDataHash and whose effect is to randomly position any newly added cases.
* We also take this opportunity to remove a long-standing special case for case plot in use-plot.


kswenson (approved):
👍 LGTM

PR #1939, opened 2025-05-09
CODAP-51 Restriction of "percent" option in graph ruler menu

bfinzer:
[#CODAP-51] Bug fix: It should not be allowed to display percentages based on categorical attributes on top and right

* V2 doesn't allow computation of percents in a graph unless there is a categorical attribute on the bottom and/or the left. V3 was giving the user the option if there was a categorical attribute on the top and/or right. We choose to follow V2 in this regard to prevent even more cognitive load for the user (and the developer!)

### DIFF
--- a/v3/src/components/graph/models/graph-data-configuration-model.ts
+++ b/v3/src/components/graph/models/graph-data-configuration-model.ts
@@ -274,7 +274,7 @@ export const GraphDataConfigurationModel = DataConfigurationModel
     },
     get leftBottomCategoricalAttrCount() {
       const attrTypes = self.attrTypes
-      return (attrTypes.left === "categorical" ? 1 : 0) + (attrTypes.bottom === "categorical" ? 1 : 0)
+      return (isCategoricalAttributeType(attrTypes.left) ? 1 : 0) + (isCategoricalAttributeType(attrTypes.bottom) ? 1 : 0)
     }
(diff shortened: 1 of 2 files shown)


v3/src/components/graph/models/graph-data-configuration-model.ts
```diff
@@ -272,6 +272,10 @@ export const GraphDataConfigurationModel = DataConfigurationModel
       const yHasCategorical = attrTypes.left === "categorical"
       return (xHasCategorical && !yHasCategorical) || (!xHasCategorical && yHasCategorical)
     },
+    get leftBottomCategoricalAttrCount() {
+      const attrTypes = self.attrTypes
+      return (attrTypes.left === "categorical" ? 1 : 0) + (attrTypes.bottom === "categorical" ? 1 : 0)
```
kswenson, line 277: Shouldn't this be:
```suggestion
      return (isCategoricalAttributeType(attrTypes.left) ? 1 : 0) + (isCategoricalAttributeType(attrTypes.bottom) ? 1 : 0)
```
so that, for instance, `color` attributes are handled as well?

kswenson (changes requested):
👍 Looks good -- just one suggestion

kswenson (approved):
👍 LGTM

PR #1940, opened 2025-05-10
CODAP-136 Get the normal curve adornment to respond dynamically to point dragging

bfinzer:
[#CODAP-136] Bug fix: Normal curve doesn't respond dynamically to drag of points

* The problem was that the normal-curve-adornment-component wasn't behaving like other univariate adornment components that observe changes in model level computed properties. Instead, the model was being asked to compute the curve parameters every time the component rerendered. Since display of a normal curve requires both a sample mean and a sample standard deviation, we couldn't easily extend the single parameter adornment pattern. Instead, we created a CurveParamInstance to take the place of a MeasureInstance and cache sampleMean and sampleStdDev in these. The component asks for these cached values, thus observing them, and rerendering when they change.
* For another bug fix, we note that it was impossible to click or hover on points underneath the curve. This was fixed with a one line change to univariate-measure-adornment-base-component.scss.

### DIFF
--- a/v3/src/components/graph/adornments/univariate-measures/normal-curve/normal-curve-adornment-component.tsx
+++ b/v3/src/components/graph/adornments/univariate-measures/normal-curve/normal-curve-adornment-component.tsx
@@ -63,8 +63,7 @@ export const NormalCurveAdornmentComponent = observer(
     const computeMeanAndStdDev: () => ({ mean: number; stdDev: number, numStdErrs:number,
                                         stdError: number | undefined }) =
       useCallback(() => {
-        const sampleMean = dataConfig ? model.computeMean(numericAttrId, cellKey, dataConfig) : NaN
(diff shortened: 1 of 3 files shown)


v3/src/components/graph/adornments/univariate-measures/normal-curve/normal-curve-adornment-component.tsx
```diff
@@ -367,6 +366,10 @@ export const NormalCurveAdornmentComponent = observer(
       }
     }, [])
 
+    if (graphModel.plotType === "binnedDotPlot") {
```
kswenson, line 369: Now that we've got plot models, I think we should move away from explicit plot type tests in favor of delegating responsibility to the plot models. Could this be `graphModel.plot.isBinned`?

kswenson (changes requested):
👍 Looks good -- a couple of suggestions to consider.

kswenson (approved):
👍 LGTM

PR #1943, opened 2025-05-13
CODAP-670 Box plot responds dynamically to point drag

bfinzer:
[#CODAP-670] Bug fix: Box plot not responding properly to dragging poings

* The box plot component observes its model and re-renders when model properties change. But the only model property currently being computed dynamically is the median. Consequently the box plot doesn't re-render until the median changes. So, we reconfigure `BoxPlotAdornmentModel` to store all the numeric values required by the component as properties following the pattern set forth in univariate-measure, overriding `updateCategories` to recompute box plot values and store them.
* Also, we prevent a box plot from displaying if the plot is a binnedDotPlot or a linePlot.
* Finally, we add click handlers for the box plot elements to select cases in the corresponding range
* And we fix a lint error in graph-data-configuration-model.ts

### DIFF
--- a/v3/src/components/graph/adornments/univariate-measures/box-plot/box-plot-adornment-model.ts
+++ b/v3/src/components/graph/adornments/univariate-measures/box-plot/box-plot-adornment-model.ts
@@ -1,5 +1,5 @@
-import { observable } from "mobx"
-import { Instance, SnapshotIn, types } from "mobx-state-tree"
+import { comparer, observable, reaction } from "mobx"
+import { addDisposer, Instance, SnapshotIn, types } from "mobx-state-tree"
 import { median } from "mathjs"
 import { quantileOfSortedArray } from "../../../../../utilities/math-utils"
 import { UnivariateMeasureAdornmentModel } from "../univariate-measure-adornment-model"
@@ -193,6 +193,16 @@ export const BoxPlotAdornmentModel = UnivariateMeasureAdornmentModel
           self.updateBoxPlotParams(boxPlotParams, instanceKey)
         }
       })
+      self.setNeedsRecomputation(false)
+    },
+    afterCreate() {
+      addDisposer(self, reaction(
+        () => [self.showOutliers, self.showICI],
+        () => self.setNeedsRecomputation(true),
+        { name: "BoxPlotAdornmentModel.afterCreate.showOutliers", fireImmediately: true,
+          equals: comparer.structural}
+      ))
+
     }
   }))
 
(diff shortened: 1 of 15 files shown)


v3/src/components/graph/adornments/univariate-measures/box-plot/box-plot-adornment-model.ts
```diff
@@ -45,64 +99,91 @@ export const BoxPlotAdornmentModel = UnivariateMeasureAdornmentModel
...
+      const medianValue = Number(self.computeMeasureValue(attrId, cellKey, dataConfig))
       const interquartileRange = self.interquartileRange(caseValues)
-      const lowerQuartile = self.lowerQuartile(caseValues)
-      const min = lowerQuartile - 1.5 * interquartileRange
-      return caseValues.filter(v => v < min).sort((a, b) => a - b)
+      const ici = 1.5 * interquartileRange / Math.sqrt(caseValues.length)
+      return { min: medianValue - ici, max: medianValue + ici }
     },
-    upperOutliers(attrId: string, cellKey: Record<string, string>, dataConfig: IGraphDataConfigurationModel) {
+    computeBoxPlotParamValue(attrId: string, cellKey: Record<string, string>,
```
kswenson, line 114:
```suggestion
    computeBoxPlotParamValues(attrId: string, cellKey: Record<string, string>,
```

v3/src/components/graph/adornments/univariate-measures/box-plot/box-plot-adornment-model.ts
```diff
@@ -1,10 +1,39 @@
...
+  .actions(self => ({
+    setBoxPlotParams(boxPlotParams: BoxPlotParams) {
+      self.boxPlotParams = boxPlotParams
+      self.isValid = true
+    },
+    setIsValid(isValid: boolean) {
+      self.isValid = isValid
+    }
+  }))
+export interface IBoxPlotParamInstance extends Instance<typeof BoxPlotParamInstance> {}
```
kswenson, line 36: The following comment applies equally to the `BoxPlotParamInstance` and `boxPlotParams` volatile property below as well as to the `CurveParamInstance` and the `curveParams` volatile property in the `NormalCurveAdornmentModel`. The reasons to use an MST model to represent something are mainly that it has properties for which MST provides de/serialization and/or that it is part of a larger serializable tree, e.g. a document. None of that is true here since these objects aren't part of a tree and have no serializable properties. Essentially, MST is just being used as an elaborate wrapper around a MobX class with observable properties. In each case, the MST model is then placed into an `observable.map` initialized with `deep: false`, but since the contents of the maps are MST models, the internal properties are observable anyway. It seems like `BoxPlotParamInstance` and `CurveParamInstance` should be simple structures or classes, and then if we want them to have observable properties you can just _not_ pass `deep: false` to the `observable.map`. Unless I'm missing something, that should simplify the code and behave the same.

kswenson (approved):
👍 LGTM

PR #1944, opened 2025-05-13
CODAP-301 Prevent point dragging if attribute has formula

bfinzer:
[#CODAP-301] Bug fix: Should not be able to drag points whose coordinates are computed with formula

* Points can only be dragged in dot plots and scatter plots. So, in the drag routines for these plots we check to see if a relevant attribute has formula and, if so, zero out the delta in pixels so that the dragging doesn't affect the attribute values.
* It has been a long-standing irritation that points that are dragged don't necessarily go back to their original size. This is exacerbated by allowing the mouse to move far from the point during a drag. So we fix that by calling `handlePointerLeave` in `PixiPoints.onDragEnd`.


kswenson (approved):
👍 LGTM

PR #1945, opened 2025-05-13
Categorical legend selection sends notification to plugin

eireland:
Changes categorical legend selection to go through the selectCases notification path instead of directly to dataConfiguration.dataset so that plugins can get notified when there is a selection. This is needed by the NEOPlugin.

### DIFF
--- a/v3/src/components/data-display/components/legend/categorical-legend.tsx
+++ b/v3/src/components/data-display/components/legend/categorical-legend.tsx
@@ -2,7 +2,7 @@ import {drag, select} from "d3"
 import React, {useCallback, useEffect, useMemo, useRef, useState} from "react"
 import {mstReaction} from "../../../../utilities/mst-reaction"
 import { mstAutorun } from "../../../../utilities/mst-autorun"
-import { setSelectedCases } from "../../../../models/data/data-set-utils"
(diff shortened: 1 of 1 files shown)


bfinzer (approved):
👍🏻LGTM

v3/src/components/data-display/components/legend/categorical-legend.tsx
```diff
@@ -64,8 +65,8 @@ export const CategoricalLegend =
       if (caseIds) {
         // This is breaking the graph-legend cypress test
         // setOrExtendSelection(caseIds, dataConfiguration?.dataset, event.shiftKey)
-        if (event.shiftKey) dataConfiguration?.dataset?.selectCases(caseIds)
-        else dataConfiguration?.dataset?.setSelectedCases(caseIds)
+        if (event.shiftKey) setSelectedCases(caseIds, dataConfiguration?.dataset)
```
kswenson, line 68:
```suggestion
        if (event.shiftKey) selectedCases(caseIds, dataConfiguration?.dataset)
```
Otherwise, you'll set the selection rather than extending it.

kswenson (changes requested):

kswenson (approved):
👍 LGTM

PR #1946, opened 2025-05-14
CODAP-672 normalize pin longitude

scytacki:
This normalizes the pin longitude between -180 and 180. 
The pin cypress test was updated to verify this works.

One thing I noticed while tracking down issues with the cypress test was that the cypress helper `getComponentTile` was using a `then` which prevents re-tries. I updated it, it might allow us to remove some wait statements if it was taking a while for a tile to show up.

### DIFF
--- a/v3/src/components/map/components/map-pin-layer.tsx
+++ b/v3/src/components/map/components/map-pin-layer.tsx
@@ -101,8 +101,12 @@ export const MapPinLayer = observer(function MapPinLayer({ mapLayerModel }: IMap
       const layerBB = layerRef.current?.getBoundingClientRect()
       if (!layerBB || !dataset) return
       const { lat, lng } = map.containerPointToLatLng([e.clientX - layerBB.x, e.clientY - layerBB.y])
+      // Normalize the longitude to be between -180 and 180
+      // The first modulo ensures the value is between -360 and 360,
+      // the second modulo ensures the value is between 0 and 360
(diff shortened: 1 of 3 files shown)


v3/src/components/map/components/map-pin-layer.tsx
```diff
@@ -101,8 +101,12 @@ export const MapPinLayer = observer(function MapPinLayer({ mapLayerModel }: IMap
       const layerBB = layerRef.current?.getBoundingClientRect()
       if (!layerBB || !dataset) return
       const { lat, lng } = map.containerPointToLatLng([e.clientX - layerBB.x, e.clientY - layerBB.y])
+      // Normalize the longitude to be between -180 and 180
+      // The first modulo ensures the value is between -360 and 360,
+      // the second modulo ensures the value is between 0 and 360
+      const normalizedLng = ((lng + 180) % 360 + 360) % 360 - 180
```
tealefristoe, line 107: I'm surprised you need two modulos here, but I guess it's just one more quirk of javascript.

tealefristoe (approved):
Looks good 👍

v3/src/components/map/components/map-pin-layer.tsx
scytacki: Yeah I didn't realize it either, but javascript `%` is really a "remainder" operator instead of a "modulo" operator. So `-13 % 5` is `-3`. https://2ality.com/2019/08/remainder-vs-modulo.html

PR #1948, opened 2025-05-16
CODAP-658 Caption should not be a hidden attribute

bfinzer:
[#CODAP-658] Bug fix: Caption attributes in maps and graphs should not end up being hidden attributes

* Conveniently there is a routine in data-configuration-model.ts, `defaultCaptionAttributeID` whose job it is to compute the ID of the attribute to be used as a caption. So, we make sure the attribute so found is not hidden.


kswenson (approved):
👍 LGTM

PR #1950, opened 2025-05-17
CODAP-364 No data tips when pointer is down

bfinzer:
[#CODAP-364] Bug fix: The graph point tooltips appear unnecessarily when the Moveable Point is moved on top of them

* From now on, data tips for points and bars will not show when the mouse is down. This is desirable because the mouse being down indicates that the user is doing something unrelated to querying a data value. We bring this about by adding a singleton class, PointerState, whose job it is to monitor whether the mouse is down or not. We make use of it in PixiPoints handlePointerOver.
* The MovablePointAdornment was producing an error on load of a document with a movable point in a graph. We fix this by noticing that the computation where the error is produced is actually redundant and can be eliminated.


kswenson (approved):
👍 LGTM -- just one suggestion.

PR #1951, opened 2025-05-17
CODAP-112 Don't show data tips when something else is on top

bfinzer:
[#CODAP-112] Bug fix: Graph hover tip appears in graph that is behind the component where the mouse is

* In pixi-points.ts `handlePointerOver` we now check to see if there is a DOM element on top that is not the pixi canvas. If so, we bail without displaying any data tip.


kswenson (approved):
👍 LGTM

PR #1953, opened 2025-05-19
CODAP-678 Rescale with >1 y-axis attributes problems

bfinzer:
[#CODAP-678] Bug fix: Rescale not behaving properly when graph y-axis has more than one attribute

* There are two parts to the problem: (1) The **Rescale** button was not rescaling sufficiently to include the values of other than the first y-attribute. (2) The y-axis was not rescaling when an additional attribute was added to the y-axis.
* We fix the first problem by noting that data-configuration-model.ts was not equipped to deal with multiple y-attributes, which, after all, has to be done in graph-data-configuration-model.ts where multiple y-attributes are known about. So, we make it possible to override `numericValuesForAttrRole` and, in graph-data-configuration-model.ts, when the role is 'y', iterate through the attributes and include all the values found.
* For the second problem we note that graph-model-utils.ts `setupAxis` was using the values for a single attribute regardless of the role for the axis. We fix this by calling numericValuesForAttrRole.
* Because we've modified the routines for computing axis min and max we have to adjust the values in graph-component-handler.test.ts

### DIFF
--- a/v3/src/components/data-display/models/data-configuration-model.ts
+++ b/v3/src/components/data-display/models/data-configuration-model.ts
@@ -359,16 +359,20 @@ export const DataConfigurationModel = types
     })
   }))
   .views(self => ({
-    numericValuesForAttribute(attrID: string) {
-      const dataset = self.dataset
-      const allCaseIDs = Array.from(self.visibleCaseIds)
-      const allValues = attrID
-        ? allCaseIDs.map((anID: string) => {
-          const value = dataDisplayGetNumericValue(dataset, anID, attrID)
-          return isFiniteNumber(value) ? value : null
-        }) : []
-      return allValues.filter(aValue => aValue != null)
-    },
+    numericValuesForAttribute: cachedFnWithArgsFactory({
+      key: (attrID: string) => attrID,
+      calculate: (attrID: string) => {
+        const dataset = self.dataset
+        const allCaseIDs = Array.from(self.visibleCaseIds)
+        const allValues = attrID
+          ? allCaseIDs.map((anID: string) => {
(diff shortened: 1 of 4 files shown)


v3/src/components/data-display/models/data-configuration-model.ts
```diff
@@ -358,27 +358,39 @@ export const DataConfigurationModel = types
...
+    numericValuesForAttribute(attrID: string) {
+      const dataset = self.dataset
+      const allCaseIDs = Array.from(self.visibleCaseIds)
+      const allValues = attrID
+        ? allCaseIDs.map((anID: string) => {
+          const value = dataDisplayGetNumericValue(dataset, anID, attrID)
+          return isFiniteNumber(value) ? value : null
+        }) : []
+      return allValues.filter(aValue => aValue != null)
+    },
```
kswenson, line 371: Just noting that this function does no caching, so if it is called repeatedly it could be expensive.

v3/src/components/graph/models/graph-data-configuration-model.ts
```diff
@@ -142,7 +142,24 @@ export const GraphDataConfigurationModel = DataConfigurationModel
     },
     get primaryAttributeType(): AttributeType {
       return self.primaryRole && self.attributeType(self.primaryRole) || "categorical"
-    }
+    },
+    /**
+     * This is overridden in derived class where 'y' is a special role
```
kswenson, line 147:
```suggestion
     * This is overridden to handle multiple 'y' attributes
```

kswenson (approved):
👍 LGTM

PR #1956, opened 2025-05-20
CODAP-99 Improved handling of attribute menu for date attribute

bfinzer:
[#CODAP-99] Bug fix: The attribute menu for a date axis should show "Treat as Categorical" rather than "Treat as Numeric"

* It proved possible to take care of this entirely at the `AxisOrLegendAttributeMenu` level. We were also able to make the menu say "Treat as Date" to bring the axis back from categorical to date.


kswenson (approved):
👍 LGTM

PR #1958, opened 2025-05-21
CODAP-680 Prevent overlap of y-axis numeric axis numbers

bfinzer:
[#CODAP-680] Bug fix: Numeric axis numbers overlap and become illegible

* This was a simple fix since the problem was only for vertical axis and the height of the numbers is constant. So, just a matter of how many fit.


kswenson (changes requested):
👍 Code looks fine -- there are failing cypress tests that should be looked at.

kswenson (approved):
👍 LGTM

PR #1960, opened 2025-05-28
CODAP-702 Fix API response to create parent case to return correct caseID

bfinzer:
[#CODAP-702] Bug fix: Using API, create of parent case returns caseID of child

The problem arose as part of getting the Markov game plugin to work in V3. The existing response to creating a parent case in a 2-level dataset was returning the caseID of the *child* case created in association with the parent case instead of the caseID of the parent case which is required by Markov.

This fix impinges on performance critical code in which a plugin creates cases. I am not at all confident that the fix is as performant as it can/should be!

* In the `create` method of `case-handler.ts` we make sure to return the case id whose collection is the one specified in the request. ~~We also find its corresponding itemID because that appears to be part of the contract.~~

Finding the specific item id that corresponds to a particular case id in the appropriate collection turns out to be non-trivial, particularly to do so efficiently. Upon closer inspection, it doesn't appear that returning the item id was ever, actually, part of the contract. V2 didn't behave that way. The API documentation doesn't mention returning item ids. As near as I can tell, items ids were returned initially because at the time the handler was first written, item ids were stable but case ids were not. Later, when the behavior of case ids was fixed the case id(s) were added to the returned result, but the item ids were kept because why not? But the additional requirement to only return case ids for the requested collection complicates the resulting code and the decision here is that the extra complexity isn't worth it. If a client asks to create a case, we'll return the id of the created case, and if clients want to know about its items they can issue a separate request.

### DIFF
--- a/v3/src/data-interactive/handlers/case-handler.test.ts
+++ b/v3/src/data-interactive/handlers/case-handler.test.ts
@@ -1,4 +1,4 @@
-import { toV2Id, toV3CaseId, toV3ItemId } from "../../utilities/codap-utils"
+import { toV2Id, toV3CaseId } from "../../utilities/codap-utils"
 import { DISuccessResult, DIValues } from "../data-interactive-types"
 import { DINewCase } from "../data-interactive-data-set-types"
 import { diCaseHandler } from "./case-handler"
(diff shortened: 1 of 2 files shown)


kswenson (approved):
👍 LGTM, but now that I've worked on it I'm requesting a review from @tealefristoe.

tealefristoe (approved):
Looks good 👍

PR #1961, opened 2025-05-29
CODAP-704 Make sure api request to create scatterplot correctly computes axis bounds

bfinzer:
[#CODAP-704] Bug fix: Create graph scatterplot through plugin API chooses unusable axis bounds

* This seems to be an order of operations problem whereby the attempt to compute axis bounds is made before the data configuration's filtered cases have been set up. We fix by forcing the filtered cases computation early in the process.


kswenson (approved):
👍 LGTM

PR #1963, opened 2025-05-30
CODAP-703 Export to V2 includes plugin-created collection defaults

bfinzer:
[#CODAP-703] Bug fix: The dataset created by Markov, when exported to V2 needs have DG.GameContext as its type

* A prerequisite to getting the export to work is, during the creation of the plugin-created data context, to notice that a collection has defaults and to store those defaults in the collection metadata. Then, on export to V2, we can both include these defaults in the V2 collection and set the type of the V2 data context as "DG.GameContext".


kswenson (approved):
👍 LGTM

PR #1967, opened 2025-06-04
try using yarn

scytacki:
clean up mobx-state-tree alias
update github actions, and ignore correct yarn files


.github/workflows/v3-regression.yml
```diff
@@ -62,6 +62,21 @@ jobs:
           chrome-version: stable
       - name: Checkout
         uses: actions/checkout@v4
+      - name: Install Corepack
+        # This is so yarn can be used correctly.
+        # Note with Node <= 24 it is included by default and and
```
kswenson:
```suggestion
        # Note with Node <= 24 it is included by default and
```

.github/workflows/v3-regression.yml
```diff
@@ -62,6 +62,21 @@ jobs:
...
+        # `corepack enable` can be used instead, but the current approach
+        # should work will all Node versions >= 18
```
kswenson:
```suggestion
        # should work with all Node versions >= 18
```

.github/workflows/v3.yml
```diff
@@ -15,14 +15,24 @@ jobs:
       - uses: snow-actions/sparse-checkout@v1.2.0
         with:
           patterns: v3
+      - name: Install Corepack
+        # This is so yarn can be used correctly.
+        # Note with Node <= 24 it is included by default and and
```
kswenson:
```suggestion
        # Note with Node <= 24 it is included by default and
```

.github/workflows/v3.yml
```diff
@@ -15,14 +15,24 @@ jobs:
...
+        # `corepack enable` can be used instead, but the current approach
+        # should work will all Node versions >= 18
```
kswenson:
```suggestion
        # should work with all Node versions >= 18
```

kswenson: Note that the same comments/typos occur multiple times further down in this file and other files.

Also, ChatGPT says that corepack is included and enabled by default on node 16.10+ as well as 18+, so it seems like it shouldn't be necessary to install it explicitly.

v3/doc/modules.md
```diff
@@ -0,0 +1,114 @@
...
+  - `npm run build:webpack  186.00s user 14.68s system 134% cpu 2:29.65 total`
+  - second run `npm run build:webpack  19.84s user 1.59s system 209% cpu 10.251 total`
+- Note: if the user tries to run an npm command they will currently see a message about an unsupported protocol for the patch dependencies.
+- Updating gitignore doing some other steps I don't remember, I ran `yarn start` again and this time I got complaints about `flat` and `flatMap` from typescript with the recommendation that I update the `lib` setting from es2018 to es2019. Doing that did fix the problems.
+- In VSCode need to install the ZipFS extension. It is listed in the recommended extensions by this repository. This makes it possible to open the definitions of types from dependencies in VSCode
+- Cypress used to have issues with PnP during the runtime but some of those are fixed: https://github.com/cypress-io/cypress/issues/8008 However component testing is apparently still broken.
+- Cypress can't import modules in the config file when PnP is being used, that sounds like it will be fixed in Cypress 15: https://github.com/cypress-io/cypress/pull/31520
```
kswenson:
```suggestion
```
Duplicated below.

v3/doc/modules.md
```diff
@@ -0,0 +1,114 @@
...
+- Running `yarn start` seems to work.
+- `yarn build:webpack` is slightly slower than `npm run build:webpack`:
+  - `yarn build:webpack  201.05s user 15.74s system 136% cpu 2:39.38 total`
+  - second run `yarn build:webpack  23.55s user 1.68s system 215% cpu 11.686 total`
+  - `npm run build:webpack  186.00s user 14.68s system 134% cpu 2:29.65 total`
+  - second run `npm run build:webpack  19.84s user 1.59s system 209% cpu 10.251 total`
+- Note: if the user tries to run an npm command they will currently see a message about an unsupported protocol for the patch dependencies.
+- Updating gitignore doing some other steps I don't remember, I ran `yarn start` again and this time I got complaints about `flat` and `flatMap` from typescript with the recommendation that I update the `lib` setting from es2018 to es2019. Doing that did fix the problems.
```
kswenson, line 98: For the purposes of this PR it's probably best to be conservative, but I'd be inclined to push `target` and `lib` to ES2020, which is five years ago at this point.

v3/tsconfig.json
```diff
@@ -15,7 +15,7 @@
     "strict": true,
     "types": [],
     "jsx": "react",
-    "lib": ["dom", "dom.iterable", "es2018"],
+    "lib": ["dom", "dom.iterable", "es2019"],
```
kswenson, line 18: At this point my inclination would be to set `target` and `lib` to `es2020`, which is five years ago at this point, although I'd understand wanting to be conservative in this PR.

v3/doc/modules.md
```diff
@@ -0,0 +1,114 @@
...
+- ran `yarn set version stable` embed the version 4 of yarn in the project this way it will be used regardless of what version users have globally
+- switched to using yarn's built in patch support:
+  - removed patch package from dependencies and scripts
+  - for each patch in `/patches`:
+    - `yarn patch [package name]`
+    - `cd [temporary folder created by yarn]`
+    - `patch -p4 < [path to patch file from patch package in /patches]` Note: sometimes -p3 is needed. This depends on the number of path segments in the patch file you want to ignore when applying a patch. For organization packages like `@nedb/binary-search-tree`, `-p4` is needed. For top level packages like `react-data-grid`, `-p3` is needed.
+    - `cd -`
+    - `yarn patch-commit -s [temporary folder created by yarn]`
+  - updated references to "patch package" in the source code
```
kswenson, line 83: One question that could be answered here is what the patching process looks like moving forward. It used to be that you'd edit the files in the `node_modules` folder and then `patch_package` would compare the version in the `mode_modules` folder with the version on npm. What does the patch-development process look like using yarn's patch mechanism?

.github/workflows/v3.yml
scytacki: From what I could tell corepack is not enabled by default by any version of node. It is included from 16 to 24, but in 25 it will no longer be included. So this approach will be safe for the foreseeable future regardless of which version of node we use. I'll consolidate the comments into a doc and reference it.

v3/doc/modules.md
scytacki: This file isn't very good for documentation so I started a new one `yarn.md` where I included some basic steps for patching.

v3/tsconfig.json
scytacki: I'll leave it alone in this PR, but now that it is updated in main, I'll try to remember to update it when we come back to this.

PR #1968, opened 2025-06-05
CODAP-631 Case Card Performance

tealefristoe:
Jira Story: https://concord-consortium.atlassian.net/browse/CODAP-631

### DIFF
--- a/v3/src/models/data/attribute.ts
+++ b/v3/src/models/data/attribute.ts
@@ -242,7 +242,7 @@ export const Attribute = V2Model.named("Attribute").props({
     const numValue = self.numValues[index]
     return !isNaN(numValue) ? numValue : self.strValues[index]
   },
-  isNumeric(index: number) {
+  valueIsNumeric(index: number) {
     return !isNaN(self.numValues[index])
   },
   numValue(index: number) {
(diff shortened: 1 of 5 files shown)


v3/src/components/case-card/case-card-header.tsx
```diff
@@ -1,18 +1,80 @@
...
 
+interface ICaseCardControlsProps {
+  caseIndexText: string
+  cases: IGroupedCase[]
+  data?: IDataSet
+  displayedCaseIndex: number
+  isCollectionSummarized: boolean
+}
+
+function CaseCardControls({
```
kswenson: `CaseCardControls` could warrant its own file at this point, but I'm also fine with leaving it here if you prefer.

v3/src/models/data/attribute.ts
```diff
@@ -242,7 +242,7 @@ export const Attribute = V2Model.named("Attribute").props({
     const numValue = self.numValues[index]
     return !isNaN(numValue) ? numValue : self.strValues[index]
   },
-  isNumeric(index: number) {
+  valueIsNumeric(index: number) {
```
kswenson: Conventionally, a boolean function like this would be named `isValueNumeric`.

kswenson (approved):
👍 LGTM

PR #1969, opened 2025-06-06
chore: performance tweaks to reduce re-renders, etc.

kswenson:
[[CODAP-697](https://concord-consortium.atlassian.net/browse/CODAP-697)] Chromebook Performance

This PR includes a number of small tweaks that were identified with tools like WhyDidYouRender to reduce the number of components that re-render unnecessarily in situations like during a drag. Most of them are unlikely to have much effect individually but hopefully they have some effect cumulatively. The one potential exception is the wrapping of `rowClass` in `CollectionTable` in `useCallback`, which could potentially have resulted in the table doing much more work than necessary when re-rendering.

### DIFF
--- a/v3/src/components/case-table/collection-table.tsx
+++ b/v3/src/components/case-table/collection-table.tsx
@@ -188,7 +188,7 @@ export const CollectionTable = observer(function CollectionTable(props: IProps) 
       }
     }, [columns, columnWidths, caseTableModel])
 
-  const handleAddNewAttribute = () => {
+  const handleAddNewAttribute = useCallback(() => {
     let attribute: IAttribute | undefined
     data?.applyModelChange(() => {
       const newAttrName = uniqueName(t("DG.CaseTable.defaultAttrName"),
@@ -205,7 +205,7 @@ export const CollectionTable = observer(function CollectionTable(props: IProps) 
       log: logStringifiedObjectMessage("Create attribute: %@",
               {name: "newAttr", collection: data?.getCollection(collectionId)?.name, formula: ""}, "data")
     })
-  }
+  }, [collectionId, data])
 
   const showInputRow = !preventCollectionReorg(data, collectionId)
   const rows = useMemo(() => {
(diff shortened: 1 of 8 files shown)


v3/src/components/case-table/collection-table.tsx
```diff
@@ -421,8 +417,12 @@ export const CollectionTable = observer(function CollectionTable(props: IProps)
       "highlight-border-bottom": hasSelectedChild && !nextCaseHasSelectedChild,
       "last-child-case": isLastChild
     })
-  }
+  }, [collection?.caseIds, collectionId, data])
+
+  if (!data || !rows || !visibleAttributes.length) return null
```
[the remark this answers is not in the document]
kswenson: That would violate the rules of hooks.

v3/src/components/case-table/collection-table.tsx
```diff
@@ -421,8 +418,12 @@ export const CollectionTable = observer(function CollectionTable(props: IProps)
...
 
+  const dragId = String(active?.id)
```
tealefristoe, line 425: If `active` is not defined, then `dragId` is "undefined". I guess it will never contain `kInputRowKey` or `collectionId` in that case, but it feels a bit odd.

v3/src/components/case-tile-common/attribute-menu/attribute-menu-list.tsx
```diff
@@ -30,8 +30,7 @@ const AttributeMenuListComponent = forwardRef<HTMLDivElement, IProps>(
   const data = useDataSetContext()
   const metadata = useDataSetMetadata()
   const tableModel = useCaseTableModel()
-  // each use of useDisclosure() maintains its own state and callbacks so they can be used for independent dialogs
-  const propertiesModal = useDisclosure()
+  const { isOpen, onClose, onOpen } = useDisclosure()
```
tealefristoe, line 33: I have to admit I don't understand the difference here. Why does using a named variable cause excessive re-renders, but this method doesn't?

v3/src/components/map/components/inspector-panel/hide-show-menu-list.tsx
```diff
@@ -2,14 +2,14 @@ import {MenuItem, MenuList, useDisclosure} from "@chakra-ui/react"
...
+import { logMessageWithReplacement } from "../../../../lib/log-message"
 import {ITileContentModel} from "../../../../models/tiles/tile-content"
 import {ITileModel} from "../../../../models/tiles/tile-model"
 import {t} from "../../../../utilities/translation/translate"
-import {IMapContentModel, isMapContentModel} from "../../models/map-content-model"
-import { logMessageWithReplacement } from "../../../../lib/log-message"
-import { DataSetContext } from "../../../../hooks/use-data-set-context"
 import { EditFormulaModal } from "../../../common/edit-formula-modal"
-import { useInspectorFormulaString } from "../../../../hooks/use-inspector-formula-string"
+import {IMapContentModel, isMapContentModel} from "../../models/map-content-model"
```
tealefristoe:
```suggestion
import { IMapContentModel, isMapContentModel } from "../../models/map-content-model"
```

tealefristoe (approved):
Looks great 👍

v3/src/components/case-tile-common/attribute-menu/attribute-menu-list.tsx
kswenson: The Chakra code for `useDisclosure()` presumably looks something like this:

```typescript
const onClose = useCallback(/* ... */)
const onOpen = useCallback(/* ... */)
return { isOpen, onClose, onOpen, /* ... */ }
```

So `onClose` and `onOpen` are stable across renders, but the object returned by `useDisclosure()` is a new object on every render. Thus, anything that puts the object returned by `useDisclosure()` into a dependency array will itself change on every render. You can work around this by putting `[myModal.onClose, myModal.onOpen]` in the dependency array, but in most cases that's an anti-pattern because we prefer `[graphModel]` to `[graphModel.prop1, graphModel.prop2, /* ... */]` in our dependency arrays. So it's simpler to just adopt the convention of immediately dereferencing the object, which I think is the pattern Chakra intended when they implemented it that way.

v3/src/components/case-table/collection-table.tsx
kswenson: This code is unchanged from before, it just moved. But apparently the logic is that we only care about it during a drag, so it doesn't matter what it is when there isn't a drag.

PR #1970, opened 2025-06-06
CODAP-720 Fix layout of non-numeric axes

bfinzer:
[#CODAP-720] Bug fix: With categorical axes, the categories overlap the attribute label

* This problem was introduced when a change in use-axis.ts `computeDesiredExtent` inadvertently began using an axis model computed with `getNumericAxis` which could never be other than undefined or numeric and therefore never went through the portion of the code that computes extent for other than numeric axes.


kswenson (approved):
👍 LGTM

PR #1972, opened 2025-06-06
fix: make it easier to click case table scroll bar

kswenson:
The main issue was that our resize-border `<div>` was overlapping the scrollbar. The solution adopted here is to move the resize borders outside the content of the component. This eliminates the scrollbar conflict at the expense of possible being a bit confusing that you have to click outside the component to resize it. It would be less confusing if there were a visible border to click on, but I don't think we want an eight-pixel border around each component for esthetic reasons.

Architecturally, responsibility for resizing the component moves from `CodapComponent` (where it never really belonged) to the `FreeTileComponent`.

### DIFF
--- a/v3/src/components/component-resize-border.tsx
+++ b/v3/src/components/component-resize-border.tsx
@@ -10,7 +10,8 @@ interface IProps {
 }
 
 export function ComponentResizeBorder({ componentRef, containerRef, edge, onPointerDown }: IProps) {
-  console.log("ComponentResizeBorder", "edge:", edge)
+  const kResizeBorderSize = 8
+  const kResizeHandleSize = 22
 
   let top = 0
   let left = 0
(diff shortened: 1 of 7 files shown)


v3/src/components/component-resize-border.tsx
```diff
@@ -0,0 +1,51 @@
...
+  componentRef: React.RefObject<HTMLDivElement | null>
+  containerRef: React.RefObject<HTMLElement | null>
+  edge: "left" | "right" | "bottom"
+  onPointerDown?: (e: React.PointerEvent<HTMLDivElement>) => void
+}
+
+export function ComponentResizeBorder({ componentRef, containerRef, edge, onPointerDown }: IProps) {
+  const kOverlap = 2
+  const kResizeBorderSize = 8
+  const kResizeHandleSize = 22
```
tealefristoe: I'd put this in `constants.ts` next to `kTitleBarHeight`.

tealefristoe (approved):
Looks great 👍

PR #1973, opened 2025-06-10
feat: refactor tile dragging code to not use DnDKit

kswenson:
[[CODAP-718](https://concord-consortium.atlassian.net/browse/CODAP-718)] Performance: Dragging hierarchical table with many cases is laggy (Chromebook)
[[CODAP-719](https://concord-consortium.atlassian.net/browse/CODAP-719)] Difficulty grabbing/dragging graph tile (Chromebook)

We replace the DnD-Kit based implementation with a straight DOM event listener approach that mirrors the way resizing of tiles is handled.

[CODAP-718]: https://concord-consortium.atlassian.net/browse/CODAP-718?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CODAP-719]: https://concord-consortium.atlassian.net/browse/CODAP-719?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/v3/src/components/container/use-tile-drag.ts
+++ b/v3/src/components/container/use-tile-drag.ts
@@ -2,6 +2,7 @@ import { useCallback } from "react"
 import { logStringifiedObjectMessage } from "../../lib/log-message"
 import { IFreeTileLayout, IFreeTileRow } from "../../models/document/free-tile-row"
 import { ITileModel } from "../../models/tiles/tile-model"
+import { uiState } from "../../models/ui-state"
 import { IChangingTileStyle, kTileDragGridSize, kTitleBarHeight } from "../constants"
 
 interface IProps {
@@ -33,6 +34,7 @@ export function useTileDrag({ row, tile, tileLayout, setChangingTileStyle }: IPr
         // Require a minimum drag distance to start dragging
         if (!isDragging && Math.abs(xDelta) + Math.abs(yDelta) > kTileDragGridSize) {
           isDragging = true
+          uiState.setIsDraggingTile(true)
         }
 
         if (isDragging) {
@@ -69,6 +71,7 @@ export function useTileDrag({ row, tile, tileLayout, setChangingTileStyle }: IPr
           }
 
           setChangingTileStyle(undefined)
+          uiState.setIsDraggingTile(false)
           isDragging = false
         }
       }
--- a/v3/src/components/container/use-tile-drag.ts
+++ b/v3/src/components/container/use-tile-drag.ts
@@ -1,0 +1,57 @@
+import { useCallback } from "react"
+import { IFreeTileLayout, IFreeTileRow } from "../../models/document/free-tile-row"
+import { kTileDragGridSize } from "../constants"
+
+interface IProps {
+  containerRef: React.RefObject<HTMLElement | null>
+  row: IFreeTileRow
+  tileLayout?: IFreeTileLayout
+}
+
+export function useTileDrag({ row, tileLayout }: IProps) {
+
+  const handlePointerDown = useCallback((e: React.PointerEvent) => {
+    if (e.currentTarget && tileLayout) {
+      const targetElt = e.currentTarget
+      if (e.pointerId != null) {
+        targetElt.setPointerCapture(e.pointerId)
+      }
+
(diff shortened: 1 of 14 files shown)


v3/src/components/container/use-tile-drag.ts
```diff
@@ -0,0 +1,90 @@
+import { useCallback } from "react"
+import { IFreeTileLayout, IFreeTileRow } from "../../models/document/free-tile-row"
+import { IChangingTileStyle, kTileDragGridSize, kTitleBarHeight } from "../constants"
+import { logStringifiedObjectMessage } from "../../lib/log-message"
+import { ITileModel } from "../../models/tiles/tile-model"
```
tealefristoe: ⬆️

v3/src/components/container/use-tile-drag.ts
```diff
@@ -0,0 +1,90 @@
...
+      const { x: xStart, y: yStart } = tileLayout.position
+      const xPageStart = e.pageX
+      const yPageStart = e.pageY
+
+      const constrainToGrid = (n: number) => Math.ceil(n / kTileDragGridSize) * kTileDragGridSize
+      const constrainX = (dx: number) => constrainToGrid(xStart + dx)
+      const constrainY = (dy: number) => Math.max(0, constrainToGrid(yStart + dy))
+
+      function handleDragTile(event: Event) {
+        const ptrEvt = event as PointerEvent
```
tealefristoe, line 30: I'm guessing `event: PointerEvent` doesn't work for some reason?

v3/src/components/container/use-tile-drag.ts
```diff
@@ -0,0 +1,90 @@
...
+            })
+          }
+
+          setChangingTileStyle(undefined)
+          isDragging = false
+        }
+      }
+
+      targetElt.addEventListener("pointermove", handleDragTile)
+      targetElt.addEventListener("pointerup", handleDropTile)
```
tealefristoe: I'm guessing this needs to be `document.body.addEventListener`.

tealefristoe (changes requested):
Code looks good. But I tried the deployed branch, and it's really easy to lose the component you're dragging by moving your mouse quickly out of the title bar. I think some event listeners were added to that element rather than to the whole document.

kswenson:
> Code looks good. But I tried the deployed branch, and it's really easy to lose the component you're dragging by moving your mouse quickly out of the title bar. I think some event listeners were added to that element rather than to the whole document.

Pointer capture is supposed to handle that, but I'll take a look.

v3/src/components/container/use-tile-drag.ts
kswenson: Correct -- it's not the right signature for `add|removeEventListener`.

v3/src/components/container/use-tile-drag.ts
kswenson: I initially had the `setPointerCapture()` in `handlePointerDown()`, but that prevented other title bar interactions from working (e.g. editing the title), so I moved it to `handleDragTile()`, but apparently capturing only works when the pointer is down, so we skip it entirely and use the `document.body` approach. But I think it would have worked the other way if we were willing to capture on pointer down.

v3/src/components/web-view/web-view-drop-overlay.tsx
```diff
@@ -55,8 +59,8 @@ export function WebViewDropOverlay() {
     }
   })
 
-  // Only render the overlay if an attribute is being dragged
-  if (!dataSet || !attributeId) return null
+  // Only render the overlay if an attribute or tile is being dragged
+  if ((!dataSet || !attributeId) && !isDraggingTile) return null
```
tealefristoe:
```suggestion
  if (!isDraggingAttribute && !isDraggingTile) return null
```

tealefristoe (approved):
Nice 👍

PR #1974, opened 2025-06-10
CODAP-735 Fix display of negative mean and median

bfinzer:
[#CODAP-735] Bug fix: mean and median adornments not drawn correctly when negative

* The display value string for the adornment was formatted with "\u2212" for a negative sign, but then this didn't convert to a number for plotting since it's not a minus sign but an en-dash. We elect to continue to use the display value string, but first replace any en-dash with minus sign.

### DIFF
--- a/v3/src/components/graph/adornments/univariate-measures/univariate-measure-adornment-helper.ts
+++ b/v3/src/components/graph/adornments/univariate-measures/univariate-measure-adornment-helper.ts
@@ -226,7 +226,7 @@ export class UnivariateMeasureAdornmentHelper {
     cellCounts: Record<string, number>, secondaryAxisX=0, secondaryAxisY=0
   ) => {
     const displayValue = this.formatValueForScale(isVertical, value)
-    const plotValue = Number(displayValue.replace("\u2212", "-"))  // replace en-dash with minus sign
+    const plotValue = value
     const measureRange: IRange = attrId && this.model.hasRange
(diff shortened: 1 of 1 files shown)


v3/src/components/graph/adornments/univariate-measures/univariate-measure-adornment-helper.ts
```diff
@@ -226,7 +226,7 @@ export class UnivariateMeasureAdornmentHelper {
     cellCounts: Record<string, number>, secondaryAxisX=0, secondaryAxisY=0
   ) => {
     const displayValue = this.formatValueForScale(isVertical, value)
-    const plotValue = Number(displayValue)
+    const plotValue = Number(displayValue.replace("\u2212", "-"))  // replace en-dash with minus sign
```
kswenson: Seems to me we could just use:
```suggestion
    const plotValue = value
```
here. Is there a requirement that the numeric value returned be formatted for the screen? I would think that the actual numeric value would be more appropriate and clients that care about formatting for the screen would use the `displayValue`.

bfinzer: I was feeling cautious along the lines of "What if the displayValue shows one thing but the place along the axis that the adornment is plotted shows another?" But, on further reflection and looking at the one client that actually makes use of the `plotValue`, I think you're right.

PR #1975, opened 2025-06-11
fix: MST warning on dragging last attribute of collection

kswenson:
[[CODAP-715](https://concord-consortium.atlassian.net/browse/CODAP-715)] mobx-state-tree error when dragging an attribute down a collection

Dragging the last attribute from a collection removes that collection from the data set. In this scenario, a stale `parent` link to the removed collection was being accessed before being updated. We now call `syncCollectionLinks()` immediately upon removing a collection to prevent such stale accesses.

[CODAP-715]: https://concord-consortium.atlassian.net/browse/CODAP-715?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tealefristoe (approved):
Looks good 👍

PR #1976, opened 2025-06-11
CODAP-736-start-monorepo

scytacki:
The vast majority of the changed files are just the v2 files being moved into a v2 folder. 

A key change here is that the you should open the root CODAP folder in VSCode instead of the v3 folder. 

These are changes on top of the `yarn-experiment` branch. Compared to that branch:
- the .yarn folder was moved from v3 to the top level
- a docs/monorepo.md file was added with some information about this new configuration
- the .vscode folder was moved from v3 to the top level
- various source files were updated to fix linting errors. Some of these had been lingering from before, others seemed to newly crop up after the lint configuration was updated
- eslint-plugin-import was replaced with eslint-plugin-import-x, this new version is faster is more comparible with Yarn PnP, however it still needed to be patched to work with mono repos and Yarn PnP. I'll try to get this patch up stream. The maintainer is very responsive.
- the top level tsconfig refers to the v3 tsconfig using a "references" property. This helps VSCode know there is a sub configuration. It should also make it possible to run typescript on the whole repository, but that hasn't been tested.


.yarn/patches/eslint-plugin-import-x-npm-4.15.1-14cf13eac8.patch
```diff
@@ -0,0 +1,84 @@
...
+--- /dev/null
++++ b/patch.md
+@@ -0,0 +1,17 @@
++This patch fixes issues that come up with Yarn PnP when used in a monorepo. 
++
++This plugin uses node's `createRequire` in many places. `createRequire` is passed a filename. This filename represents the module that the created require function is relative to. For strict Yarn PnP if a require is used in a module that is part of a package which doesn't define a dependency on thing being required it will be rejected.
++
++When running Yarn PnP the require function returned by `createRequire` seems to look at both the package of the passed in filename and also the root package of the monorepo the code is running in. It does not look at the workspace package that the code is running in. 
++
++The import-x plugin is passing it's require.js as the filename for `createRequire` in many cases. This means that any dependency that is dynamically loaded has to either be defined on the mono repo root package or as a dependency of the import-x plugin itself. Specifically this was used when loading the parsers defined in the settings.
```
kswenson:
```suggestion
+The import-x plugin is passing its require.js as the filename for `createRequire` in many cases. This means that any dependency that is dynamically loaded has to either be defined on the mono repo root package or as a dependency of the import-x plugin itself. Specifically this was used when loading the parsers defined in the settings.
```

doc/monorepo.md
```diff
@@ -0,0 +1,35 @@
...
+# TS-Lib
+`ts-lib` is added to the root package.json to fix an issue with Jest and its transpiler. We have Jest transpile some of dependencies because they are distributed as ESM modules and Jest works with Common JS. When the transpiler tries to handle these it appears to be loading `ts-lib` as if it was required by one of the files in this package. These package don't depend on `ts-lib` so Yarn PnP fails this require. However because of the [top level fallback of Yarn PnP](https://yarnpkg.com/configuration/yarnrc#pnpFallbackMode) it will fallback to looking to see if the modules package is listed in the root package.json of the repository. This means that ts-lib has to be added to the root package.json to work around this issue. Hopefully Jest will fix this issue in the future.
+
+# TODO
+
+## Typescript Language Server
+When I run `yarn dlx @yarnpkg/sdks --verbose vscode` at the root I get Eslint and Typescript sdks installed but I also get:
+>  Typescript Language Server (dependency not found; skipped)
+
+I'm guessing this might be for something other than vcode like emacs perhaps, but if we have problems it might be worth looking into.
```
kswenson:
```suggestion
I'm guessing this might be for something other than vscode like emacs perhaps, but if we have problems it might be worth looking into.
```

doc/monorepo.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Linting root files
+Currently there is no linting of the files at the root of the workspace. We should lint some of them at least the `.github/workflow/*` files would be good to lint.
+
+With typescript it is possible to typecheck the whole repository. With the current eslint setup
```
kswenson: Incomplete comment?

v3/src/components/axis/hooks/use-axis.ts
```diff
@@ -142,7 +142,7 @@ export const useAxis = (axisPlace: AxisPlace) => {
           layout.setDesiredExtent(axisPlace, computeDesiredExtent())
         }, { name: "useAxis.axisDomainSync", equals: comparer.structural }, axisProvider)
     }
-  }, [axisPlace, axisProvider, isNumeric, multiScale])
+  }, [axisPlace, axisProvider, computeDesiredExtent, isNumeric, layout, multiScale])
```
kswenson: FYI: this change has already been made on `main`.

v3/src/test/setupTests.ts
```diff
@@ -73,3 +74,7 @@ declare global {
     }
   }
 }
+
+declare global {
+    const jsdom: JSDOM
```
kswenson:
```suggestion
  const jsdom: JSDOM
```

v3/src/test/setupTests.ts
```diff
@@ -2,6 +2,7 @@ import "@testing-library/jest-dom"
 import "jest-canvas-mock"
 import { enableFetchMocks } from "jest-fetch-mock"
 import "jest-webgl-canvas-mock"
+import type { JSDOM } from 'jsdom'
```
kswenson:
```suggestion
import type { JSDOM } from "jsdom"
```

v3/eslint.config.mjs
```diff
@@ -28,18 +28,19 @@ export default [{
...
+importX.flatConfigs.recommended,
+importX.flatConfigs.typescript,
+{
   plugins: {
     "@stylistic": stylisticEslintPlugin,
     "@typescript-eslint": fixupPluginRules(typescriptEslint),
     json: fixupPluginRules(json),
     react: fixupPluginRules(react),
-    "react-hooks": fixupPluginRules(reactHooks)
+    // "react-hooks": fixupPluginRules(reactHooks)
```
kswenson, line 43: Are we no longer linting React hooks or is this handled elsewhere somehow?

kswenson (approved):
👍 As with #1967, I'm not sure I've got my head wrapped around all of this, but it seems reasonable as far as I can tell. Just a handful of small comments/suggestions.

v3/eslint.config.mjs
scytacki: They are still enabled. I think they come from the line above `plugin:react-hooks/recommended`. When I run `yarn eslint --print-config src/components/codap-modal.tsx` it indicates `"react-hooks/rules-of-hooks": "error"` and `"react-hooks/exhaustive-deps": "warn"`. I think I commented it out just because it wasn't making any difference. I should just delete it, but I'm going leave it as is to reduce the number of merge conflicts.

PR #1977, opened 2025-06-11
Codap-103 Import/Export Cases from/to Clipboard

tealefristoe:
Jira stories:
https://concord-consortium.atlassian.net/browse/CODAP-287
https://concord-consortium.atlassian.net/browse/CODAP-103


v3/src/components/case-tile-common/inspector-panel/ruler-menu-list.tsx
```diff
@@ -61,14 +74,42 @@ export const RulerMenuList = observer(function RulerMenuList() {
...
+            setCopiedCasesString(`${data.itemIds.length} ${data.childCollection.title}`)
+            copiedAlert.onOpen()
+          }
+        }
+      }
     },
     {
-      itemKey: "DG.Inspector.getCaseDataFromClipboard"
+      itemKey: "DG.Inspector.getCaseDataFromClipboard",
+      handleClick: () => {
```
kswenson, line 87: Can we enable/disable the menu item based on the contents of the clipboard?

v3/src/components/case-tile-common/inspector-panel/ruler-menu-list.tsx
```diff
@@ -1,17 +1,30 @@
...
 import { t } from "../../../utilities/translation/translate"
+import { CopiedCasesAlert } from "../copied-cases-alert"
+import { CopyToClipboardModal } from "../copy-to-clipboard-modal"
 import { IMenuItem, StdMenuList } from "../std-menu-list"
 
 export const RulerMenuList = observer(function RulerMenuList() {
   const data = useDataSetContext()
+  const [copiedCasesString, setCopiedCasesString] = useState("")
+  const copyToClipboardModal = useDisclosure()
+  const copiedAlert = useDisclosure()
```
kswenson: This is the usage we're moving away from. Dereference the return immediately:
```suggestion
  const { isOpen: isCopyModalOpen, onClose: onCloseClipboardModal, onOpen: onOpenClipboardModal } = useDisclosure()
  const { isOpen: isCopiedAlertOpen, onClose: onCloseCopiedAlert, onOpen: onOpenCopiedAlert } = useDisclosure()
```

v3/src/components/case-tile-common/inspector-panel/ruler-menu-list.tsx
```diff
@@ -1,17 +1,30 @@
...
+
+  const handleCloseModal = () => {
+    copyToClipboardModal.onClose()
+    copiedAlert.onOpen()
```
kswenson: Should only show the copied alert on successful copy, not on Cancel.

v3/src/utilities/csv-export.ts
```diff
@@ -0,0 +1,48 @@
...
+  let csv = `# name: ${dataset.name}\n`
+
+  const attrs = collection?.attributesArray ?? dataset.attributes
+  attrs.forEach(attr => {
+    csv += `# attribute -- name: ${attr.name.replace(/,/g, "&comma;")}`
+    if (attr.description) csv += `, description: ${attr.description}`
+    if (attr.type) csv += `, type: ${attr.type}`
+    if (attr.units) csv += `, unit: ${attr.units}`
+    csv += `, editable: ${!metadata?.isEditProtected(attr.id)}`
+    // TODO: Are there other properties we should include?
```
kswenson: You can look at the `exportCaseData()` method of the `DG.DataContext` in the v2 code to see what metadata is exported there.

kswenson (changes requested):
👍  Generally looks good -- the multi-collection copy dialog looks like a bug, but I see it follows the "design" of v2's dialog, so we can save that for another day. See other suggestions below.

v3/src/components/case-tile-common/inspector-panel/ruler-menu-list.tsx
tealefristoe: Should I disable it if there are no contents? Or should I process the contents somehow to check? If so, how?

The plugin is happy to accept any text, but it just dumbly applies whatever is there to the dataset, which can't be undone, so it would be good to try to prevent unintentional imports.

tealefristoe:
@kswenson Do you mean that the multi-collection modal just looks bad, or is there some buggy behavior?

v3/src/components/case-tile-common/inspector-panel/ruler-menu-list.tsx
kswenson: I was thinking of disabling it if there is nothing on the clipboard or if it isn't text.

kswenson:
> @kswenson Do you mean that the multi-collection modal just looks bad, or is there some buggy behavior?

The layout looks like a bug with all that white space on the left, but that's the way v2 looked as well so 🤷‍♂️.

v3/src/components/case-tile-common/inspector-panel/ruler-menu-list.tsx
tealefristoe: So there are two complications here.
1. Reading the clipboard is asynchronous.
2. The clipboard changing won't cause a re-render.

What I can do is run a `useEffect` on every render that reads the clipboard and then sets some state, which will take care of the first complication. The second complication seems more serious to me--if the clipboard can't be imported, we'll disable the button, and then when the user copies text to the clipboard the button will still be disabled. The opposite can happen as well. Is there a way to deal with this problem?

v3/src/utilities/csv-export.ts
tealefristoe: I was able to flesh out the properties with that tip, thanks. Two follow up questions:
1. I need to check for the default precision (2). Is there a constant for this somewhere? I couldn't find one, so I added it to `constants.ts`, but it seems like it should already be somewhere else.
2. Dataset metadata is also included in the v2 export, but I couldn't figure out what's included in that. Any hint on where to find those properties?

v3/src/components/case-tile-common/inspector-panel/ruler-menu-list.tsx
kswenson: It was a thought that occurred to me on review. If it's not straightforward and/or it's not the way v2 behaves feel free to punt it.

v3/src/components/case-tile-common/inspector-panel/ruler-menu-list.tsx
```diff
@@ -61,14 +69,43 @@ export const RulerMenuList = observer(function RulerMenuList() {
...
   ]
 
   return (
-    <StdMenuList data-testid="ruler-menu-list" menuItems={menuItems} />
+    <>
+      <StdMenuList data-testid="ruler-menu-list" menuItems={menuItems} />
+      <CopyToClipboardModal
+        isOpen={isCopyModalOpen}
+        onClose={() => onCloseCopyModal()}
+        onComplete={() => onOpenCopiedAlert()}
```
kswenson:
```suggestion
        onClose={onCloseCopyModal}
        onComplete={onOpenCopiedAlert}
```
No need to convert a stable function reference to an unstable one.

v3/src/components/case-tile-common/inspector-panel/ruler-menu-list.tsx
```diff
@@ -61,14 +69,43 @@ export const RulerMenuList = observer(function RulerMenuList() {
...
+        setCopiedCasesString={setCopiedCasesString}
+      />
+      <CopiedCasesAlert
+        copiedCasesString={copiedCasesString}
+        isOpen={isCopiedAlertOpen}
+        onClose={() => onCloseCopiedAlert()}
```
kswenson:
```suggestion
        onClose={onCloseCopiedAlert}
```

v3/src/constants.ts
```diff
@@ -10,3 +10,10 @@ export const kRootGuideUrl = codapResourcesUrl("example-documents/guides")
 export function getPluginsRootUrl() {
   return urlParams.pluginURL || kRootPluginUrl
 }
+
+export const kImporterPluginUrl = "/Importer/index.html?lang=en-US"
+export function getImporterPluginUrl() {
+  return `${getPluginsRootUrl()}${kImporterPluginUrl}`
+}
+
+export const kDefaultPrecision = 2
```
kswenson: Use `kDefaultNumPrecision` in `attribute-types.ts`.

v3/src/utilities/csv-export.ts
```diff
@@ -0,0 +1,56 @@
+import { kDefaultPrecision } from "../constants"
```
kswenson: Use `kDefaultNumPrecision` in `attribute-types.ts`.

v3/src/utilities/csv-export.ts
kswenson: > I was able to flesh out the properties with that tip, thanks. Two follow up questions:
> 
> 1. I need to check for the default precision (2). Is there a constant for this somewhere? I couldn't find one, so I added it to `constants.ts`, but it seems like it should already be somewhere else.
> 2. Dataset metadata is also included in the v2 export, but I couldn't figure out what's included in that. Any hint on where to find those properties?

1. Use `kDefaultNumPrecision` in `attribute-types.ts`.
2. I presume this would correspond to:
```typescript
export interface ICodapV2DataContextMetadata {
  description?: string
  source?: string
  importDate?: string
  // Legacy (bug?) property imported as `importDate`
  "import date"?: string
}
```
which is represented in v3 `DataSetMetadata` properties:
```typescript
    // DataSet-wide metadata
    description: types.maybe(types.string),
    source: types.maybe(types.string),
    importDate: types.maybe(types.string),
```

PR #1978, opened 2025-06-11
CODAP-737 Fix attribute menu becoming inaccessible on component resize

bfinzer:
[#CODAP-737] Bug fix: Graph axis label becomes inaccessible

* Note: The bug also occurred for a map or graph legend attribute label. The cause of the bug was a lack of rerendering of the `AxisOrLegendAttributeMenu` component on resize of graph or map component.
* The fix is to pass the stringified layout bounds computed in attribute-label.tsx to the `AxisOrLegendAttributeMenu` to trigger rerender when it changes.


kswenson (approved):
👍 LGTM

PR #1985, opened 2025-06-21
feat: support Markov data game

kswenson:
- Manually copied the `Markov` and `Common` folders from the `data-games-plugins` repository to s3 in the `plugins/data-games/Markov` and `plugins/data-games/Common` directories.
- Updated the `processWebViewUrl` function to map the original Markov url to the new s3 Markov url.

Observations:
1. On initial loading of the Markov example document, there are two data sets named `Games/Turns` listed in the table menu.
2. After playing for a few turns, closing and reopening the graph results in a broken graph rendering which is not fixed by resizing the graph.
3. Cypress test failure due to opening/closing Markov example document bringing up a Do you want to save dialog.

Note that these observations are _not_ reasons to hold up this PR, which merely enables the Markov example document and the Markov plugin to be loaded successfully. The actual behavior of the plugin and its interaction with CODAP v3 via the API are separate issues.

@scytacki The review question for you is whether this seems like a reasonable migration strategy for migrating v3 plugins from GitHub Pages to the s3 bucket.

### DIFF
--- a/v3/cypress/e2e/import-v2.spec.ts
+++ b/v3/cypress/e2e/import-v2.spec.ts
@@ -7,10 +7,11 @@ context("import codap v2 documents", () => {
     cy.visit(url)
   })
   it('will load plugins from v2 documents', () => {
-    cfm.openExampleDocument("Markov Game")
-    cy.wait(1000)
-    webView.getIFrame().find(".welcome").should("contain.text",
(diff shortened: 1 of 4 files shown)


bfinzer (approved):
👍🏻LGTM

scytacki (approved):
This approach seems good to me. 👍

PR #1988, opened 2025-06-23
Improve performance of large .csv imports [CODAP-630]

kswenson:
[[CODAP-630](https://concord-consortium.atlassian.net/browse/CODAP-630)] Improve performance of large .csv imports

This PR addresses a number of performance bottlenecks in importing large .csv files. V2 (and now V3) invoke the Importer plugin to import .csv files. The Importer plugin offers several capabilities, notably including the ability to sub-sample from large files. It also batches up the imported items into blocks of 200 cases for sending to CODAP, so from a performance perspective it's really a large number of small incremental imports. For instance, the `us-cities.csv` file I used for testing has ~45K rows, which means ~225 blocks of items are communicated to CODAP. We will probably want to fold the features of the Importer plugin into CODAP itself at some point, but in the meantime this makes for a useful point of comparison. Also, adding items incrementally is a common use case in other scenarios, e.g. the Sampler, probes, etc., so it is a use case worth optimizing even if we eventually move away from the Importer plugin. Here are the total import times for importing the `us-cities.csv` file:

|Version|Time|
|-------|-----|
|V3 before this PR|3:45|
|V3 after this PR|0:25|
|V2|0:20|

Note that there still may be some performance gains to be had with further investigation, but this is a solid step in the right direction.

Here are some of the bottlenecks addressed in this PR:
- The code for grouping items into cases treats appended items as a special case for which it doesn't have to clear its internal caches and start the grouping process from scratch.
- Reduces work done to determine whether the type of an attribute can be inferred.
- Reduces work done to determine whether a given string can be interpreted as a boundary.
- The case table inspector minimizes the work done to determine which delete items to enable.

[CODAP-630]: https://concord-consortium.atlassian.net/browse/CODAP-630?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/v3/src/models/data/attribute.ts
+++ b/v3/src/models/data/attribute.ts
@@ -31,7 +31,7 @@ import { parseColor } from "../../utilities/color-utils"
 import { isDateString } from "../../utilities/date-parser"
 import { DatePrecision } from "../../utilities/date-utils"
 import { cachedFnFactory } from "../../utilities/mst-utils"
-import { isBoundaryValue, kPolygonNames } from "../boundaries/boundary-types"
+import { isBoundaryString, kPolygonNames } from "../boundaries/boundary-types"
 import { Formula, IFormula } from "../formula/formula"
 import { applyModelChange } from "../history/apply-model-change"
 import { withoutUndo } from "../history/without-undo"
@@ -133,7 +133,7 @@ export const Attribute = V2Model.named("Attribute").props({
     return self.strValues.reduce((prev, current) => isDateString(current) ? ++prev : prev, 0)
   }),
   getBoundaryCount: cachedFnFactory<number>(() => {
-    return self.strValues.reduce((prev, current) => isBoundaryValue(current) ? ++prev : prev, 0)
+    return self.strValues.reduce((prev, current) => isBoundaryString(current) ? ++prev : prev, 0)
   }),
   get hasFormula() {
     return !!self.formula && !self.formula.empty
--- a/v3/src/models/data/attribute.ts
+++ b/v3/src/models/data/attribute.ts
@@ -104,36 +104,53 @@ export const Attribute = V2Model.named("Attribute").props({
   get datePrecision() {
     return typeof self.precision === "string" ? self.precision : undefined
   },
-  getEmptyCount: cachedFnFactory<number>(() => {
-    // Note that `self.changeCount` is absolutely not necessary here. However, historically, this function used to be
-    // a MobX computed property, and `self.changeCount` was used to invalidate the cache. Also, there are tests
-    // (and possibly some features?) that depend on MobX reactivity. Hence, this is left here for now.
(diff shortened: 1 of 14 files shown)


v3/src/models/data/attribute.ts
```diff
@@ -104,36 +104,53 @@ export const Attribute = V2Model.named("Attribute").props({
...
-  getNumericCount: cachedFnFactory<number>(() => {
-    // Note that `self.changeCount` is absolutely not necessary here. However, historically, this function used to be
-    // a MobX computed property, and `self.changeCount` was used to invalidate the cache. Also, there are tests
-    // (and possibly some features?) that depend on MobX reactivity. Hence, this is left here for now.
-    self.changeCount // eslint-disable-line @typescript-eslint/no-unused-expressions
-    return self.numValues.reduce((prev, current) => isFinite(current) ? ++prev : prev, 0)
+  isInferredNumericType: cachedFnFactory<boolean>(() => {
+    let numCount = 0
+    const isEveryValueNumericOrEmpty = self.strValues.every((strValue, index) => {
+      if (strValue == null || strValue === "") return true
```
tealefristoe, line 110: Could make `strValue == null || strValue === ""` a function since it's used in a few places.

v3/src/models/data/data-set.ts
```diff
@@ -587,6 +589,25 @@ export const DataSet = V2Model.named("DataSet").props({
       })
       self.setValidCases()
     }
+  },
+  validateNewCases(itemIds: string[]) {
```
tealefristoe:
```suggestion
  validateNewCases(caseIds: string[]) {
```

v3/src/utilities/csv-import.ts
```diff
@@ -56,3 +56,21 @@ export function initiateImportFromCsv(csvContent: string, dataset: IDataSet) {
...
+      contentType: 'text/csv',
+      name: "Importer",
+      file
+    }
+  }
+  appState.document.content?.insertTileSnapshotInDefaultRow({
+    _title: "Importer",
+    content: webViewModelSnap
+  })
+}
```
tealefristoe: I'd abstract the common code from this and `initiateImportFromCsv` into a single function.

v3/src/models/data/data-set.ts
kswenson: The argument is actually items, but I renamed the function `validateCasesForNewItems` to make the relationships clearer.

v3/src/utilities/csv-import.ts
kswenson: Done

v3/src/models/data/attribute.ts
kswenson: Given the performance implications of these functions I'd rather not introduce a function call indirection for a pattern that I suspect can be trivially inlined.

PR #1989, opened 2025-06-24
CODAP-740-formula-library

scytacki:
Move formula system into its own package

As part of this a utilities package was created too for modules needed both by the main CODAP app (v3) and the formula system.

Several files from the formula system remain in v3 as just "links" to their implementation in the formula library. 
Many of the formula system tests remain in v3 because these tests use full CODAP documents to test the formulas.

CODAP-740


scytacki:
@kswenson I'm not sure what to do with this PR. I can't keep working on it. 

Here are two options:
- leave it open as a draft, at least until the boundary manager issue is fixed. I might be able to squeeze this in, in the next few weeks, but I'm not sure.
- merge it into monorepo

PR #1992, opened 2025-06-26
fix: parsing of unicode operators

kswenson:
The issue was that the grammar allowed certain unicode operators to be considered part of identifiers. We modify the grammar to exclude such operators.


tealefristoe (approved):
Looks good 👍

PR #1994, opened 2025-06-26
CODAP-753 Refresh the Sampler plugin URL

bfinzer:
[#CODAP-753] Bug fix: The Sampler plugin has a stale URL

* Changed URL to point to the current Sampler.

Note that the URL for the icon still points to /TP-Sampler/ (as it also does in V2)


kswenson (approved):
👍 LGTM

PR #1995, opened 2025-06-26
Codap-643 Default Dataset Title

tealefristoe:
Jira Story: https://concord-consortium.atlassian.net/browse/CODAP-643

This PR makes dataset titles follow the pattern set in v2. Specifically, if a dataset doesn't have an explicit title, its title is its collection names joined by "/"s. A dataset's name is only used if it has no collections, which I think can never happen.

This change broke a few tests which helped me uncover some minor bugs:
- Datasets created with the new table menu option now have titles. This matches v2 behavior.
- Updating the name via the Dataset Information form now updates the title instead of the name.
- The formula engine uses dataset names instead of titles for keys.


kswenson (approved):
👍 LGTM

PR #1997, opened 2025-06-26
CODAP-689 Box plot not displaying when it should

bfinzer:
[#CODAP-689] Bug fix: Graph: with box plots turned on, when changing the x-axis attribute, box plots were not displayed correctly (missing for some categories on the y axis)

* Simple logic error where we were allowing zero to masquerade as undefined


v3/src/components/graph/adornments/univariate-measures/box-plot/box-plot-adornment-model.ts
```diff
@@ -64,7 +64,8 @@ export const BoxPlotAdornmentModel = UnivariateMeasureAdornmentModel
     getQuantileValue(quantile: 25 | 75, caseValues: number[]) {
       const sortedCaseValues = caseValues.sort((a, b) => a - b)
       const quantileValue = quantileOfSortedArray(sortedCaseValues, quantile / 100)
-      return quantileValue || NaN
+      // eslint-disable-next-line eqeqeq
+      return quantileValue == undefined ? NaN : quantileValue
```
kswenson:
```suggestion
      return quantileValue == null ? NaN : quantileValue
```

kswenson (approved):
👍 LGTM -- note that you should use `== null` which will allow you to eliminate the `eslint-disable-next-line eqeqeq` comment (and also catches `undefined`), but doesn't require re-review.

PR #1999, opened 2025-06-27
CODAP-754 Treat as numeric works on graph axes

bfinzer:
[#CODAP-754] Bug fix: Treating an attribute as *numeric* should convert strings containing numbers to numeric values and plot them

* In the data configuration models when the user has "treat as numeric" and the  we attempt to *extract* a number from the string value.
* Note that in the various plots, if we're asking for a numeric value, we can assume that if the attribute type is not numeric, we should attempt the extraction.

### DIFF
--- a/v3/src/components/data-display/models/data-configuration-model.ts
+++ b/v3/src/components/data-display/models/data-configuration-model.ts
@@ -360,8 +360,8 @@ export const DataConfigurationModel = types
   }))
   .views(self => ({
     numericValuesForAttribute: cachedFnWithArgsFactory({
-      key: (attrID: string, type: AttributeType | undefined) => attrID + (type || ""),
-      calculate: (attrID: string, type: AttributeType | undefined) => {
+      key: (attrID: string, type?: AttributeType) => `${attrID}:${type || ""}`,
+      calculate: (attrID: string, type?: AttributeType) => {
         const dataset = self.dataset
         const allCaseIDs = Array.from(self.visibleCaseIds)
         const allValues = attrID
@@ -641,8 +641,7 @@ export const DataConfigurationModel = types
         return legendID
           ? self.getCaseDataArray(0).filter((aCaseData: CaseData) => {
             const value = dataDisplayGetNumericValue(dataset, aCaseData.caseID, legendID, typeIsNumeric)
-            // eslint-disable-next-line eqeqeq
-            return value != undefined && value >= min && value < max
+            return value != null && value >= min && value < max
           }).map((aCaseData: CaseData) => aCaseData.caseID)
           : []
 


v3/src/components/data-display/models/data-configuration-model.ts
```diff
@@ -360,13 +360,13 @@ export const DataConfigurationModel = types
   }))
   .views(self => ({
     numericValuesForAttribute: cachedFnWithArgsFactory({
-      key: (attrID: string) => attrID,
-      calculate: (attrID: string) => {
+      key: (attrID: string, type: AttributeType | undefined) => attrID + (type || ""),
```
kswenson: The intent would be clearer if this were something like:
```suggestion
      key: (attrID: string, type?: AttributeType) => `${attrID}:${type || ""}`,
```

v3/src/components/data-display/models/data-configuration-model.ts
```diff
@@ -636,10 +637,12 @@ export const DataConfigurationModel = types
...
         const dataset = self.dataset
         const legendID = self.attributeID('legend')
+        const typeIsNumeric = self.attributeType('legend') === 'numeric'
         return legendID
           ? self.getCaseDataArray(0).filter((aCaseData: CaseData) => {
-            const value = dataDisplayGetNumericValue(dataset, aCaseData.caseID, legendID)
-            return value !== undefined && value >= min && value < max
+            const value = dataDisplayGetNumericValue(dataset, aCaseData.caseID, legendID, typeIsNumeric)
+            // eslint-disable-next-line eqeqeq
+            return value != undefined && value >= min && value < max
```
kswenson: Banish`!= undefined` and the corresponding `eslint-disable` from your toolbox:
```suggestion
            return value != null && value >= min && value < max
```

v3/src/components/data-display/models/data-configuration-model.ts
```diff
@@ -360,13 +360,13 @@ export const DataConfigurationModel = types
   }))
   .views(self => ({
     numericValuesForAttribute: cachedFnWithArgsFactory({
-      key: (attrID: string) => attrID,
-      calculate: (attrID: string) => {
+      key: (attrID: string, type: AttributeType | undefined) => attrID + (type || ""),
+      calculate: (attrID: string, type: AttributeType | undefined) => {
```
kswenson:
```suggestion
      calculate: (attrID: string, type?: AttributeType) => {
```

kswenson (changes requested):
👍 Generally looks good -- I added a commit which simplified the `dataDisplayGetNumericValue()` function a bit. Also, several other small changes that should be addressed.

PR #2001, opened 2025-06-28
CODAP-681 Long categories need truncation on axes

bfinzer:
[#CODAP-681] Bug fix: Graph: sometimes the axis labels are so long in a graph that the graph itself cannot be displayed

* `GraphLayout.setAxisBounds` removed because it is not being called.
* Add `getDesiredExtent` to layout classes to use to determine how much to elide category labels
* Provide `elideStringToFit` for elision of category labels
* In `GraphLayout.setDesiredExtent` limit extent of axes so that categorical labels don't consume too much space
* in `use-axis.ts` make the axis responsive to changes in graph bounds

PR #2002, opened 2025-06-29
CODAP-757 Fix categorical axis labeling bug

bfinzer:
[#CODAP-757] Bug fix: Short category labels on graph axes only show up as "…"

* A previous commit made the mistaken assumption that all category labels on graph axes needed to be elided. But, in fact, only in those situations in which there is a "collision" of labels so that the labels are oriented perpendicular to the axis, does elision need to be considered. (Elision will only take place, of course, if there is insufficient room for the label.)


kswenson (approved):
👍 LGTM -- the original problem with #2001 was caught by the cypress tests, but only after I added the `Run Regression` label, which I did when I reviewed it, but then I didn't wait for the results. Process discussion is warranted.

PR #2003, opened 2025-06-30
CODAP-686 Binned dot plot split by categorical not displaying properly

bfinzer:
[#CODAP-686] Bug fix: Graph: selecting radio button “Group into Bins” (from “Points”) moved some nodes/dots to the left (expected) and some nodes/dots to the right, offscreen (unexpected)

* Remove unused `GraphLayout.getCategoricalScale`
* Fixed logic error in use-dot-plot.ts in computation of `secondaryPlace`
* The main problem was that the secondary axis scale was losing its categorical domain in the transition from dot plot to binned dot plot. We reconstitute it!

### DIFF
--- a/v3/src/components/graph/hooks/use-dot-plot.ts
+++ b/v3/src/components/graph/hooks/use-dot-plot.ts
@@ -1,7 +1,7 @@
 import { ScaleBand, ScaleLinear } from "d3"
+import { reaction } from "mobx"
 import { useCallback, useEffect } from "react"
 import { useMemo } from "use-memo-one"
-import { mstReaction } from "../../../utilities/mst-reaction"
 import { useDataSetContext } from "../../../hooks/use-data-set-context"
 import { AxisPlace } from "../../axis/axis-types"
 import { GraphAttrRole } from "../../data-display/data-display-types"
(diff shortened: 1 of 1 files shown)


v3/src/components/graph/hooks/use-dot-plot.ts
```diff
@@ -105,6 +105,10 @@ export const useDotPlot = (pixiPoints?: PixiPoints) => {
       minBinEdge, numExtraPrimaryBands, pointDiameter, primaryAttrID, primaryAxisScale, primaryIsBottom,
       secondaryBandwidth, totalNumberOfBins, numPointsInRow])
 
+  // The secondary axis scale can lose its domain in certain circumstances. Reconstitute it.
+  layout?.axisScales.get(secondaryPlace)?.setCategoricalDomain(
+    dataConfig?.categoryArrayForAttrRole(secondaryAttrRole) ?? [])
```
kswenson: This is a side effect that occurs on every render. 1. It should really be in a `useEffect()` and 2. Can we narrow the circumstances under which it must occur?

v3/src/components/graph/models/graph-layout.ts
```diff
@@ -101,10 +101,6 @@ export class GraphLayout extends DataDisplayLayout implements IAxisLayout {
     return this.axisScales.get(place)?.bandScale
   }
 
-  getCategoricalScale(place: AxisPlace) {
-    return this.axisScales.get(place)?.categoricalScale
-  }
```
kswenson: These functions (`getNumericScale()`, `getBandScale()`, `getCategoricalScale()`) mirror the corresponding properties in the `MultiScale`. I'm inclined to keep them in parallel unless you have an argument for why `getCategoricalScale()` would/should never be used. There are a bunch of callers to `getAxisScale()` which then cast to whatever they're expecting. These functions offer type-safe alternatives.

kswenson (approved):
👍 LGTM -- I added a commit which simplified the reaction and cleaned up the exports a bit.

PR #2005, opened 2025-06-30
Codap-696 Export Graphs as SVG

tealefristoe:
Jira Story: https://concord-consortium.atlassian.net/browse/CODAP-696

This PR enables exporting graphs as SVGs. The bulk of this PR breaks apart recent work to render graphs as PNGs, reusing much of it to render an SVG.

Unfortunately, the current approach for creating SVGs involves wrapping the tile's HTML in a `foreignObject` element, and several applications commonly used to display these SVGs (Word, Notes, etc) will not render `foreignObjects`. Despite this glaring inadequacy, we plan on moving forward with this approach for the time being with the hope of fixing this problem sometime in the future.

Additional improvements in this PR:
- Title information was removed from function calls to export tile images, since the new algorithm includes the tile's title naturally.
- A bug in the function that determines if an element should be included in the exported image has been fixed. This bug was making it so no elements were ever removed. I'm not sure how this bug was introduced. It would be easy to miss if I didn't call it out here, because it's buried in the code that was shifted around without many other changes.

### DIFF
--- a/v3/src/components/data-display/models/data-display-render-state.ts
+++ b/v3/src/components/data-display/models/data-display-render-state.ts
@@ -1,21 +1,18 @@
-import { graphSnapshot } from "../../graph/utilities/image-utils"
+import { graphSnapshot, IGraphSvgOptions } from "../../graph/utilities/image-utils"
 import { PixiPointsArray } from "../pixi/pixi-points"
 
 export class DataDisplayRenderState {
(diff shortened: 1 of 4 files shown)


kswenson (approved):
👍 LGTM -- note I didn't test it manually since others seem to have already done so.

PR #2006, opened 2025-07-01
Improve performance with sampler plugin [CODAP-36]

kswenson:
The major performance gains came with #1988. This PR incrementally improves performance further when running with the sampler plugin. Addresses three relatively small performance bottlenecks identified by the profiler:
- `CollectionModel` concatenates its own group key rather than calling `JSON.stringify()`
- `CollectionTableSpacer` uses `useEffect`s to retrieve/set information it requires rather than calling `getBoundingClientRect()` and `getStringCssVariable()` on every render.

The overall effect is relatively small, on the order of 15 to 14 seconds on importing 200 samples of 5 items in my test case. Unfortunately, v3 is still not as fast as v2 in this test, but this is a step in the right direction.

Also fixes an unrelated typo that I encountered along the way.

### DIFF
--- a/v3/src/components/case-table/collection-table-spacer.tsx
+++ b/v3/src/components/case-table/collection-table-spacer.tsx
@@ -106,7 +106,7 @@ export const CollectionTableSpacer = observer(function CollectionTableSpacer({
 
   // match relation colors to grid colors via CSS variables
   useEffect(() => {
-    const newRelationColors = { ...kDefaultRelationColors}
+    const newRelationColors = { ...kDefaultRelationColors }
     const relationSelectedFillColor = getStringCssVariable(gridElt, "--rdg-row-selected-background-color")
     if (relationSelectedFillColor) {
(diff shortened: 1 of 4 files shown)


v3/src/components/case-table/collection-table-spacer.tsx
```diff
@@ -85,10 +100,42 @@ export const CollectionTableSpacer = observer(function CollectionTableSpacer({
...
 
   const handleRef = (element: HTMLElement | null) => {
-    tableSpacerDivRef.current = element
+    setTableSpacerDiv(element)
     setNodeRef(element)
   }
 
+  // match relation colors to grid colors via CSS variables
+  useEffect(() => {
+    const newRelationColors = { ...kDefaultRelationColors}
```
tealefristoe:
```suggestion
    const newRelationColors = { ...kDefaultRelationColors }
```

PR #2013, opened 2025-07-08
CODAP-724 Fix Empty Case Card Bug

tealefristoe:
Jira Story: https://concord-consortium.atlassian.net/browse/CODAP-724

In a recent PR, one fix reverted a previous fix, resulting in the same incorrect behavior. This PR fixes the problem by disabling the generic case card attribute input field when there are no cases present.


kswenson (approved):
👍 LGTM

PR #2016, opened 2025-07-09
Codap-98 Axis Menus with Multiple Collections

tealefristoe:
Jira Story: https://concord-consortium.atlassian.net/browse/CODAP-98

This PR makes graph axis menus hierarchical when there are multiple collections in a document (either one dataset with multiple collections, or multiple datasets). I tried to replicate v2 behavior as much as I could:
- Submenus open when you hover over a collection item.
- Submenus close when you hover over another top level item, or if the top menu closes (like clicking the background or selecting an option).
- Clicking a top level collection item will not close the menu or modify the graph.

Remaining concerns:
- In v3, menu items are set up to drag the attribute assigned to the axis. This isn't how things work in v2, and I think we should remove this weird functionality. But I didn't do it as part of this PR.
- I'm accessing all of the datasets in the document using `gDataBroker`. Is there a better way of doing this?
- We're still using the `DataConfigurationContext` to find the dataset for the graph. This is then used to determine which attribute is assigned to the axis, as well as the buggy dragging behavior described above. This feels a bit weird to me, and prevents a graph from having attributes from different datasets on different axes. But it seems to reproduce v2 behavior.
- Keyboard controls don't work.

### DIFF
--- a/v3/src/components/axis/components/axis-or-legend-attribute-menu.tsx
+++ b/v3/src/components/axis/components/axis-or-legend-attribute-menu.tsx
@@ -39,6 +39,7 @@ export const AxisOrLegendAttributeMenu =
   const dataConfiguration = useDataConfigurationContext()
   const metadata = dataConfiguration?.metadata
   const data = dataConfiguration?.dataset
+  const collections = data?.collections || []
   const role = graphPlaceToAttrRole[place]
   const attrId = dataConfiguration?.attributeID(role) || ''
   const instanceId = useInstanceIdContext()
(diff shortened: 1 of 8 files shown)


v3/src/components/axis/components/axis-or-legend-attribute-menu.tsx
```diff
@@ -79,9 +164,45 @@ export const AxisOrLegendAttributeMenu =
...
+        <MenuItemsForCollection
+          collectionInfo={allCollectionInfo[0]}
+          onChangeAttribute={handleChangeAttribute}
+          place={place}
+        />
+      )
+    } else {
+      let dividerCount = 0
+      return allCollectionInfo.map(collectionInfo => {
+        if (typeof collectionInfo === 'string') {
```
kswenson: Perhaps a comment indicating that `divider` is the only supported string.

kswenson:
>I'm accessing all of the datasets in the document using gDataBroker. Is there a better way of doing this?

The `gDataBroker` is going to be phased out. The better way to access all the `DataSet`s is to use `getSharedDataSets()` in `shared-data-utils.ts`.

PR #2017, opened 2025-07-10
CODAP-741 Fix axis layout problems

bfinzer:
[#CODAP-741] Bug fix: Graph: Axis label covers the axis scale labels

I wasn’t able to reproduce the described bug, even on a Chromebook. But I found two new bugs and fixed them:
1. Putting age on the y-axis of a new graph caused the y-axis attribute label to overlap with the numeric values on the axis.
2. Putting age on the x-axis resulted in numeric labels that slightly collide with each other.

* The first bug required some machinations around getting the actual tick values that get displayed and use those to compute the desired axis extent.
* The second bug was fixed by adding a 4 pixel gap between numeric values on a horizontal axis in the collision detection algorithm.

### DIFF
--- a/v3/cypress/e2e/pixi-interaction/graph-pixi-interaction.spec.ts
+++ b/v3/cypress/e2e/pixi-interaction/graph-pixi-interaction.spec.ts
@@ -323,7 +323,7 @@ context("Graph UI with Pixi interaction", () => {
       gch.getGraphTileId().then((tileId: string) => {
         // Known inputs:
         const pointIndex = 3
-        const expectedX = 165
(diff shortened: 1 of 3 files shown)


kswenson (changes requested):
👍 Code looks fine, but there's a cypress test failure that should be addressed.

bfinzer:
The failure was due to a test relying on pixel coordinates for points, and these have changed because of improvements to the graph layout in this PR. I modified the expected point positions to correspond to observed values when running the test.

kswenson (approved):
👍 LGTM

PR #2019, opened 2025-07-11
feat: improve performance of computed value propagation

kswenson:
The essence of the problem was that computed values were being propagated to each attribute with a formula via a call to `DataSet.setCaseValues()`, which triggers a lot of MST overhead (e.g. action-tracking middleware) as well as potentially triggering clients like case tables and graphs to respond as well. Thus, when using the sampler to generate items, the items would be created, clients would respond to the `addCases` change, then the formula engine would respond by evaluating the formula and calling `setCaseValues` for each attribute with a formula, which caused additional rounds of MST middleware and client responses.

To address this, we add a new `DataSet.setAttributeValues` method which is classified as a `view` rather than an `action` for MST purposes, so that it doesn't trigger action-tracking middleware. Furthermore, `DataSet.setAttributeValues` makes a single call to `Attribute.setValues` rather than calling `Attribute.setValue` for every item, which results in further performance gains. Essentially what we're saying here is that formula evaluation and caching is under-the-hood shuffling of values that should not be treated the same way user-triggered changes are treated (which is what `DataSet.setCaseValues` is used for).

While adding jest tests for the new code and checking test coverage, I noticed that there were other aspects of the `DataSet`, particularly involving hierarchical data sets, that were lacking in test coverage so I added some additional tests there as well. While adding those tests, I found some code that was unclear, some methods that were poorly named, etc., and so I made a few other unrelated changes in the interest of general code hygiene. Specifically, this PR:

- adds `setAttributeValues` as a `view` method of the `DataSet`.
- calls `DataSet.setAttributeValues` from the `AttributeFormulaAdapter`.
- adds explicit invalidation when re-randomizing formulas (so clients can respond).
- adds explicit invalidation when an attribute formula depends on a global/slider value.
- renames `DataSet.getParentCase` to `DataSet.getParentCaseInfo` because it returns metadata about the case rather than the case itself.
- makes `DataSet.childCases` a `get`ter.
- fixes `DataSet.removeCases` so it handles removing parent cases properly.

### DIFF
--- a/v3/cypress/e2e/table.spec.ts
+++ b/v3/cypress/e2e/table.spec.ts
@@ -543,7 +543,7 @@ context("case table ui", () => {
       table.getGridCell(2, 5).then(cell => {
         random2 = +cell.text()
         expect(random2 >= 0).to.eq(true)
-        expect(random2 < 1).to.eq(true)
+        expect(random2 <= 1).to.eq(true)
         expect(random2).not.to.eq(random1)
       })
       // Delete formula, verify values remain
@@ -552,7 +552,7 @@ context("case table ui", () => {
       table.getGridCell(2, 5).then(cell => {
         const value = +cell.text()
         expect(value >= 0).to.eq(true)
-        expect(value < 1).to.eq(true)
+        expect(value <= 1).to.eq(true)
         expect(value).to.eq(random2)
       })
       // verify that formula was deleted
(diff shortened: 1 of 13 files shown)


kswenson:
> I haven't looked at the code but I identified a couple of problems with this approach:
> 
> 1. If you have a chain of formulas `x`, `y = x + 1`, `z = y + 1`. And then you edit a value of x, the z value doesn't update.
> 2. When the formula is edited for z, the values also don't update.

A larger refactor is required to fix this issue properly. I was hoping that this more limited approach would pay some dividends without breaking anything, but apparently half-measures won't work. Closing for now to await the larger refactor.

PR #2020, opened 2025-07-11
CODAP-769 Improve category drag animation

bfinzer:
[#CODAP-769] Bug fix: Graph: Better animate category reordering in graphs

* The poor animations in which points disappeared and then reappeared was due to the fact that each pixi point has a mask assigned to it. When the categories are reordered, the points that are going to move are suddenly not in the "correct" sub-plot and are thus masked.
* The solution is to, before we start the animation, remove all masks and allow the points to be seen as they cross from one sub-plot to another. When the animation has finished, we reinstate the masks.
* This required 
  * adding a routine to pixi-points.ts to `removeMasks`
  * Passing an optional callback function parameter to `DataDisplayContentModel.startAnimation` to be invoked at the end of an animation
  * In use-plot.ts `respondToCategorySetChanges`, before starting the animation, removing the cell masks and passing a callback to startAnimation that will update the masks after the animation completes.

### DIFF
--- a/v3/src/components/data-display/models/data-display-content-model.ts
+++ b/v3/src/components/data-display/models/data-display-content-model.ts
@@ -124,9 +124,12 @@ export const DataDisplayContentModel = TileContentModel
         clearTimeout(self.animationTimerId)
       }
     },
-    startAnimation() {
+    startAnimation(stoppingCallback?: () => void) {
       if (self.animationTimerId) clearTimeout(self.animationTimerId)
-      self.animationTimerId = window.setTimeout(() => this.stopAnimation(), 2000)
+      self.animationTimerId = window.setTimeout(() => {
+        this.stopAnimation()
+        if (stoppingCallback) stoppingCallback()
+      }, 2000)
     },
     stopAnimation() {
       self.animationTimerId = 0
--- a/v3/src/components/data-display/models/data-display-content-model.ts
+++ b/v3/src/components/data-display/models/data-display-content-model.ts
@@ -124,11 +124,11 @@ export const DataDisplayContentModel = TileContentModel
         clearTimeout(self.animationTimerId)
       }
     },
-    startAnimation(stoppingCallback?: () => void) {
+    startAnimation(onComplete?: () => void) {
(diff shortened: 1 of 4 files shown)


v3/src/components/graph/utilities/graph-utils.ts
```diff
@@ -331,6 +331,10 @@ export function updateCellMasks({ dataConfig, layout, pixiPoints }: IUpdateCellM
   pixiPoints?.setPointsMask(dataConfig.caseDataWithSubPlot)
 }
 
+export function removeCellMasks(pixiPoints?: PixiPoints) {
+  pixiPoints?.removeMasks()
+}
+
```
kswenson: This appears to not be used:
```suggestion
```

v3/src/components/data-display/models/data-display-content-model.ts
```diff
@@ -124,9 +124,12 @@ export const DataDisplayContentModel = TileContentModel
         clearTimeout(self.animationTimerId)
       }
     },
-    startAnimation() {
+    startAnimation(stoppingCallback?: () => void) {
```
kswenson: I would call this `onComplete`:
```suggestion
    startAnimation(onComplete?: () => void) {
```

v3/src/components/data-display/models/data-display-content-model.ts
```diff
@@ -124,9 +124,12 @@ export const DataDisplayContentModel = TileContentModel
...
       if (self.animationTimerId) clearTimeout(self.animationTimerId)
-      self.animationTimerId = window.setTimeout(() => this.stopAnimation(), 2000)
+      self.animationTimerId = window.setTimeout(() => {
+        this.stopAnimation()
+        if (stoppingCallback) stoppingCallback()
```
kswenson: With the name change suggested above, this can be simplified to:
```suggestion
        onComplete?.()
```

bfinzer:
Noted. And Dan, who reported the difference, is quite OK with this less than optimal behavior.

PR #2021, opened 2025-07-14
Codap-98 Fix Axis Menus for Graphs with No Data Configuration

tealefristoe:
Jira story: https://concord-consortium.atlassian.net/browse/CODAP-280

This PR fixes a bug that Bill found while reviewing another story. The immediate issue was that I was using `dataConfiguration.dataSet` as the MST model to pass to `getSharedDataSets()`, but in some cases a graph had no `dataConfiguration`, so I wasn't finding any datasets in the document. The solution I use here is to use the `tileModel` as the MST model instead.

It seems to me like reaching this state in the first place is a bug, but I don't fully understand how the `dataConfiguration` is intended to work. To reach this state:
- Open any document, like mammals.
- Add a new table.
- Add a new graph.
The new graph has no `dataConfiguration` and will show no dots.

### DIFF
--- a/v3/build_number.json
+++ b/v3/build_number.json
@@ -1,1 +1,1 @@
-{"buildNumber":2365}
+{"buildNumber":2370}
--- a/v3/cypress/e2e/axis.spec.ts
+++ b/v3/cypress/e2e/axis.spec.ts
@@ -551,7 +551,7 @@ context("Test date axes with multiple y-axes", () => {
   it("will create a time series graph with multiple y-axes using date x-axis", () => {
     // Create graph with date on x-axis and chlorophyll on y-axis
     ah.openAxisAttributeMenu("bottom")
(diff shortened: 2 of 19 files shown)


kswenson:
> It seems to me like reaching this state in the first place is a bug, but I don't fully understand how the `dataConfiguration` is intended to work. To reach this state:
> 
> * Open any document, like mammals.
> * Add a new table.
> * Add a new graph.

I thought v3 was following v2 behavior in which a graph created when there is only one data set auto-connects to that data set, but a graph created when there is more than one data set will not auto-connect to any data set. I would still expect there to be a data configuration in that case, however.

tealefristoe:
@kswenson It does look like the `dataConfiguration` is available in either situation, so that can be used as the MST model to ground `getSharedDataSets`.

kswenson (approved):
👍 LGTM

PR #2026, opened 2025-07-15
CODAP-692 Bunching of category labels under drag

bfinzer:
[#CODAP-692] Bug fix: Graph: y-axis labels bunched up along the bottom half when this axis was scaled, with no labels along the upper half of the y axis

* The bunching of axis category labels was mainly caused by the fact that under certain methods of opening the document (dragging the file onto CODAP with the Open dialog box present) the document was being opened twice and the graph, under rehydration was emerging with a kind of a ghost data configuration model in addition to the most recently created one. In the (long) process of attempting to debug what was going on, a few, more minor things in the code got improved.
* And a new bug surfaced, CODAP-786, namely, that on the second attempt to drag categories of the graph in this enclosed document, the dragging doesn’t work.
* The bottom line is that we’ll call this story done when the changes are merged and leave the remaining problem to the new story.


v3/src/components/axis/hooks/use-sub-axis.ts
```diff
@@ -87,6 +87,15 @@ export const useSubAxis = ({
     categoriesSelectionRef = useRef<Selection<SVGGElement | BaseType, CatObject, SVGGElement, any>>(),
     categoriesRef = useRef<string[]>([]),
 
+    getCatArray = useCallback(() => {
+      const catArray = dataConfig?.categoryArrayForAttrRole(axisPlaceToAttrRole[axisPlace]).slice() ?? []
+      // console.log('categoryArray.length =', catArray.length, 'for axisPlace', axisPlace)
```
kswenson:
```suggestion
```

v3/src/components/axis/hooks/use-sub-axis.ts
```diff
@@ -87,6 +87,15 @@ export const useSubAxis = ({
     categoriesSelectionRef = useRef<Selection<SVGGElement | BaseType, CatObject, SVGGElement, any>>(),
     categoriesRef = useRef<string[]>([]),
 
+    getCatArray = useCallback(() => {
```
kswenson:
```suggestion
    getCategoryArray = useCallback(() => {
```
Using the more descriptive name doesn't take up substantially more space.

v3/src/components/graph/models/graph-data-configuration-model.ts
```diff
@@ -903,7 +902,6 @@ export const GraphDataConfigurationModel = DataConfigurationModel
           () => self.clearCasesCache(),
           { name: "GraphDataConfigurationModel getCellKeys reaction", equals: comparer.structural }
         ))
-        baseAfterCreate()
```
kswenson, line 906: This should not be removed. `DataConfiguration.afterCreate` sets up a number of important listeners. The removal of those listeners could potentially explain some of the odd behavior we've been seeing. 🤷‍♂️

kswenson (changes requested):
A couple of minor suggestions and one more substantive one that could adversely affect behavior.

v3/src/components/graph/models/graph-data-configuration-model.ts
bfinzer: I removed this because I found that `DataConfiguration.afterCreate` was being called without it.

kswenson: 🤯 👍 I stand corrected. I just double-checked and it appears that you are right -- apparently MST calls `DataConfigurationModel.afterCreate` on its own. At the very least, I think a comment here indicating that the base model `afterCreate` is called automatically would be useful. It appears that `DataConfigurationModel.afterCreate` is called before `GraphDataConfigurationModel.afterCreate`, so the comment would be more appropriate at the top of the function than here at the bottom.

PR #2027, opened 2025-07-15
fix: double import when file dropped on user entry dialog

kswenson:
To allow file drops to occur on the document window, the `useDropHandler()` hook attaches DOM event handlers for the `dragover` and `drop` events to the `codap-app-id` div. To support file drops when the user entry dialog is showing, the handlers are attached to the overlay div behind it that darkens the background instead. When the user entry dialog is shown or hidden the handlers are first disposed from their current div before being installed on the other one. The bug was occurring when the `drop` event was being handled once by the overlay and once by the application, even though those handlers are never installed at the same time. In any case, the fix is to call `stopPropagation()` so that the event is only handled once.


bfinzer (approved):
👍🏻LGTM

PR #2028, opened 2025-07-16
CODAP-786 2nd drag of categories on graph not working

bfinzer:
[#CODAP-786] Bug fix: Second drag of categories of a graph in a particular document doesn't work

* This mysterious bug *appeared* to be restricted to a particular document, but that was not true. It was restricted to graphs using a categorical attribute *that had not yet acquired a non-provisional categorySet*. The drag code was using a newly created provisional categorySet for the first drag and holding onto that for subsequent drags even though it was no longer the operative categorySet, having been replaced by a non-provisional one.
* So the fix is to dynamically compute the categorySet so we're always sure of getting the operative one.


kswenson (approved):
👍 LGTM -- since there's been a lot of confusion in this area, it would be good to have a cypress test for category dragging, but we should write that as a new story rather than holding up this one.

PR #2032, opened 2025-07-17
Codap-767 Case Card Paging

tealefristoe:
Jira Story: https://concord-consortium.atlassian.net/browse/CODAP-767

This PR changes how case cards page to match v2 behavior. The Jira story was ambiguous about what should be displayed on the card ("[current] of [number selected]" vs "[current] of [total]", which is closer to v2), so I left that as is.

Other changes:
- If multiple cases are selected, `getDisplayCaseIndex` in `case-card-header.tsx` now returns `-1` to make it clear that no single case is being displayed.
- The inner `CaseView` in `case-view.tsx` has been renamed `InnerCaseView`. This component is recursive, and I found that the internal reference to `CaseView` was referencing the inner `CaseView` when the names were the same. I'm not sure if this was causing problems, but the inner `CaseView` is not an observer, so it seemed safer to make sure the outer `CaseView` was being referenced.

### DIFF
--- a/v3/src/models/data/data-set.ts
+++ b/v3/src/models/data/data-set.ts
@@ -912,9 +912,7 @@ export const DataSet = V2Model.named("DataSet").props({
                 ? group.childItemIds.every(id => self.selection.has(id))
                 : self.selection.has(caseId)
       },
-      isCaseLooselySelected(caseId: string) {
-        // a case is loosely selected if any of its child items are selected
-        // This function is used by the case card.
+      isAnyChildItemSelected(caseId: string) {
         const group = self.caseInfoMap.get(caseId)
         return group
           ? group.childItemIds.some(id => self.selection.has(id))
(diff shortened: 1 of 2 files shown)


v3/src/models/data/data-set.ts
```diff
@@ -906,12 +906,20 @@ export const DataSet = V2Model.named("DataSet").props({
...
       isCaseSelected(caseId: string) {
-        // a pseudo-case is selected if all of its individual cases are selected
+        // a case is selected if all of its child items are selected
         const group = self.caseInfoMap.get(caseId)
         return group
                 ? group.childItemIds.every(id => self.selection.has(id))
                 : self.selection.has(caseId)
       },
+      isCaseLooselySelected(caseId: string) {
+        // a case is loosely selected if any of its child items are selected
```
kswenson: Could we call this `isAnyChildItemSelected()` and then the comment wouldn't be necessary?

kswenson (approved):
👍 Looks good

PR #2034, opened 2025-07-21
CODAP-794 wordListMatches Dataset and Attribute Lookup

tealefristoe:
Jira Story: https://concord-consortium.atlassian.net/browse/CODAP-794

### DIFF
--- a/v3/src/models/formula/functions/string-functions.ts
+++ b/v3/src/models/formula/functions/string-functions.ts
@@ -279,12 +279,13 @@ export const stringFunctions = {
       const [textRefArg] = args as TWordListMatchesArgs
       const { dataSetId, attrId, otherAttrId } = stringFunctions.wordListMatches.getDependency(args)
       const text = String(evaluateNode(textRefArg, scope))
-      const dataSet = scope.getDataSet(dataSetId)
+      const dataSet = scope.getDataSet(dataSetId) ?? scope.getDataSetByTitle(dataSetId)
       if (!dataSet) {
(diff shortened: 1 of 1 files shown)


v3/src/models/formula/utils/name-mapping-utils.ts
```diff
@@ -69,17 +69,19 @@ export const getDisplayNameMap = (options: IDisplayNameMapOptions, useSafeSymbol
...
-      attribute: {}
+    if (dataSet.title) {
+      // No LOCAL_ATTR prefix is necessary for external attributes. They always need to be resolved manually by custom
+      // mathjs functions (like "lookupByIndex"). Also, it's never necessary to use safe names, as these names
+      // are string constants, not symbols, so MathJS will not care about special characters there.
+      const dataSetKey = key(dataSet.title, false)
+      displayNameMap.dataSet[dataSetKey] = {
+        id: idToCanonical(dataSet.id),
+        attribute: {}
+      }
```
kswenson: This seems fine to fix the bug, but since we're here I think it would make sense to include the `name` in the `displayNameMap` as well, something like:
```suggestion
      displayNameMap.dataSet[key(dataSet.title, false)] = {
        id: idToCanonical(dataSet.id),
        attribute: {}
      }
      if (dataSet.name !== dataSet.title) {
        displayNameMap.dataSet[key(dataSet.name, false)] = {
          id: idToCanonical(dataSet.id),
          attribute: {}
        }
      }
```
So we're following the general principle that we should match on either wherever possible.

kswenson (approved):
👍 Looks good -- just one addition to consider.

PR #2036, opened 2025-07-21
fix: Getting Started with CODAP 2 example document

kswenson:
The URL used for the `Getting Started with CODAP` document (`../../../../extn/plugins/onboarding/`), is a substring match for the url used for the `Getting Started with CODAP 2` document (`../../../../extn/plugins/onboarding/onboarding_2.html`), so we have to be careful about how the string-matching works when converting old urls to new ones. Toward that end, with this PR we switch from simple string-matching to regex-matching and support both full- and partial-string matching.


tealefristoe (approved):
Looks good 👍

PR #2038, opened 2025-07-22
Codap-768 Card Swipe Animation

tealefristoe:
Jira story: https://concord-consortium.atlassian.net/browse/CODAP-768

This PR adds a swipe animation to the case card.
- The animation is moderated by volatile properties in `CaseCardModel`.
- Most of the animation code lives in `case-view.tsx`.
  - `SingleCaseView` has been separated from `CaseView`. This allows it to be rendered multiple times, once as the actual case card and once as a dummy case card that's used for the animation.
  - A `useEffect` in `CaseView` monitors changes to the selection and triggers the animation when appropriate.
  - Only the highest level collection animates.
  - When an animation triggers, the moving cards are placed in their initial position in one frame (the actual card to the right or left, and the dummy card in the middle taking its place). In the next frame the animation actually starts via css transition.
- Unfortunately, I wasn't able to get the dummy card to maintain the previous values very well, despite spending a decent amount of time trying. I think I've already sunk too much time into trying, so I'm giving up for now. Since the cards move very quickly, I doubt many people will notice.

Other improvements:
- Use proper functions for selection in `card-view.tsx`.
- A `FreeTileLayoutContext` was added, since `CaseView` now needs access to the `freeTileLayout`. `ComponentResizeWidgets` now accesses `freeTileLayout` via the context instead of props.
- In `case-attrs-view.ts`, a key was improved to avoid stale state when creating a new case.
- `selectedCaseIds` and `selectedCaseIndices` were added to `DataSet`. These getters return arrays of selected cases for each collection in the dataset. These functions are used in several places where `CaseCardModel.caseLineage` (which has been removed) was being used previously.
- `CaseCardModel.addNewCase` has been greatly simplified, and its behavior is more like v2 (admittedly in a very subtle way: previously, adding a case to a collection immediately below a summarized collection would acquire any single cases in collections higher up the hierarchy; in v2, a completely blank case would be created in this situation).

### DIFF
--- a/v3/src/components/case-card/case-card.scss
+++ b/v3/src/components/case-card/case-card.scss
@@ -3,7 +3,8 @@
 
 .case-card.react-data-card {
   height: calc(100% - vars.$title-bar-height);
-  overflow: auto;
+  overflow-x: hidden;
+  overflow-y: auto;
 }
 
 .codap-case-card {


v3/src/components/case-card/case-card-model.ts
```diff
@@ -32,38 +37,29 @@ export const CaseCardModel = TileContentModel
...
+        collections.forEach((collection, index) => {
+          if (index < collections.length - 1) {
+            const cases = self.data?.getCasesForCollection(collection.id) ?? []
+            const collectionSelectedCaseIds = self.data?.selectedCaseIds[index]
+            if (cases.length > 1 && collectionSelectedCaseIds?.size !== 1) {
+              collectionIdsToSummarize.add(collection.id)
+            }
+          } else {
+            // always summarize the last collection
+              collectionIdsToSummarize.add(collection.id)
```
kswenson:
```suggestion
            collectionIdsToSummarize.add(collection.id)
```

v3/src/components/case-card/case-card-model.ts
```diff
@@ -16,6 +16,11 @@ export const CaseCardModel = TileContentModel
     // key is collection id; value is percentage width
     attributeColumnWidths: types.map(types.number),
   })
+  .volatile(() => ({
+    animationLevel: Infinity,
+    animationDirection: "right" as "left" | "right",
+    animationTimeout: undefined as ReturnType<typeof setTimeout> | undefined
```
kswenson, line 22: I was going to comment on the use of `NodeJS.Timeout`, but Copilot beat me to it. 😉

v3/src/components/case-card/case-view.tsx
```diff
@@ -1,46 +1,112 @@
...
-  const displayedCase = initialSelectedCase ?? cases[0]
-  const displayedCaseId = displayedCase?.__id__
+  const selectedCaseIndices = data?.selectedCaseIndices ?? []
+  const collectionSelectedCaseIndices = selectedCaseIndices[level] ?? []
+  const displayedCaseIndex = collectionSelectedCaseIndices[0] ?? -1
+  const previousSelectedCaseIndex = useRef<number>(displayedCaseIndex)
+  const displayedCaseId = collection?.caseIds[displayedCaseIndex] ?? ""
+  const displayedCase = data?.caseInfoMap.get(displayedCaseId)?.groupedCase ?? cases[0]
+
+  const disableAnimation = urlParams.noComponentAnimation != null
```
kswenson, line 57: As I mentioned in slack, I don't think it's necessary for `noComponentAnimation` to control this internal animation, but I'm also not opposed to it, so it's fine for now.

v3/src/components/case-card/case-view.tsx
```diff
@@ -1,46 +1,112 @@
...
+        )
+        // Reset the animation after the proper duration
+        cardModel.setAnimationTimeout(setTimeout(() => {
+          cardModel.setAnimationLevel(Infinity)
+          setAnimationStarted(false)
+        }, animationDuration))
+
+        // We need one frame to set up the animation, so we actually start the animation in a timeout
+        setTimeout(() => {
+          // We need to make sure a lower level did not supercede this one
```
kswenson:
```suggestion
          // We need to make sure a lower level did not supersede this one
```

v3/src/components/case-card/case-view.tsx
```diff
@@ -60,41 +126,107 @@ export const CaseView = observer(function InnerCaseView(props: ICaseViewProps) {
...
-            displayedCaseLineage={displayedCaseLineage}
-            onNewCollectionDrop={handleNewCollectionDrop}
-          />
-        </CollectionContext.Provider>
-      </ParentCollectionContext.Provider>
-    )
+  const classes = clsx(
+    "case-card-view",
+    colorCycleClass(level, collectionCount),
+    { "animating": animationStarted && !dummy }
```
kswenson:
```suggestion
    { animating: animationStarted && !dummy }
```

v3/src/components/case-card/case-view.tsx
```diff
@@ -60,41 +126,107 @@ export const CaseView = observer(function InnerCaseView(props: ICaseViewProps) {
...
+  )
+  const tileWidth = tileLayout?.width ?? 0
+  const leftLeft = `-${tileWidth}px`
+  const rightLeft = `${tileWidth}px`
+  const left = !isAnimating || animationStarted ? 0 : isFlippingRight ? leftLeft : rightLeft
+  const style = { left }
+  const otherLeft = !animationStarted ? 0 : isFlippingRight ? rightLeft : leftLeft
+  const otherStyle: React.CSSProperties = {
+    left: otherLeft,
```
kswenson: It seems like this applies to the dummy instance, in which case `dummyLeft` and `dummyStyle` would be clearer.

v3/src/hooks/use-free-tile-layout-context.ts
```diff
@@ -0,0 +1,8 @@
+import { createContext, useContext } from "react"
+import { IFreeTileLayout } from "../models/document/free-tile-row"
+
+export const FreeTileLayoutContext = createContext<IFreeTileLayout | undefined>(undefined)
+
+export const useFreeTileLayoutContext = () => {
+  return useContext(FreeTileLayoutContext)
+}
```
kswenson, line 8: Nice

v3/src/models/data/data-set.ts
```diff
@@ -618,9 +618,44 @@ export const DataSet = V2Model.named("DataSet").props({
     Array.from(self.childCollection.caseGroupMap.values()).forEach(caseGroup => {
       self.itemIdChildCaseMap.set(caseGroup.childItemIds[0] ?? caseGroup.hiddenChildItemIds[0], caseGroup)
     })
+  },
+  // The returned array is in collection order.
+  // Includes cases with ANY child items selected, not just those with all items selected.
+  get selectedCaseIds(): Set<string>[] {
```
kswenson: The naming of these methods is misleading. In data set parlance, a case is selected if _all_ of its child cases/items are selected. Taking Copilot's suggestion into account as well, perhaps `partiallySelectedCaseIdsByCollection`?

v3/src/models/data/data-set.ts
```diff
@@ -618,9 +618,44 @@ export const DataSet = V2Model.named("DataSet").props({
...
+    })
+
+    return selectedCaseIds
   }
 }))
 .views(self => ({
+  // The returned array includes arrays of the sorted indices of selected cases in each collection
+  // So selectedCaseIndices[i] contains an array of the indices of the cases selected in data.collections[i]
+  // Includes cases with ANY child items selected, not just those with all items selected.
+  get selectedCaseIndices(): number[][] {
```
kswenson: Following the logic of the previous comment, this would be `partiallySelectedCaseIndicesByCollection`.

v3/src/models/data/data-set.ts
```diff
@@ -618,9 +618,44 @@ export const DataSet = V2Model.named("DataSet").props({
...
+        selectedCaseIds[index].add(caseId)
+      })
+    })
+
+    return selectedCaseIds
   }
 }))
 .views(self => ({
+  // The returned array includes arrays of the sorted indices of selected cases in each collection
+  // So selectedCaseIndices[i] contains an array of the indices of the cases selected in data.collections[i]
```
kswenson:
```suggestion
  // The returned array includes arrays of the sorted indices of partially selected cases in each collection.
  // So selectedCaseIndices[i] contains an array of the indices of the cases partially selected in data.collections[i].
```

kswenson (changes requested):
👍 Looks great overall -- a few minor suggestions and one about naming in the `DataSet`. There's also a bug:

1. Open https://codap3.concord.org/branch/card-swipe/?sample=mammals
2. Switch to the case card
3. Click the right arrow twice to get to the second case
4. Click the left arrow to go back to the first case

Note that there is a horizontal scroll bar that allows scrolling the case card completely out of view. Presumably, it's scrolling to the no-longer-visible but apparently still rendered dummy. There are other ways to get into the same state but this is one of the simplest.

v3/src/components/case-card/case-view.tsx
tealefristoe: Ah yeah, I was going to comment about this on slack but I never got around to it. I had already implemented this before you posted--it's only three lines or so. And it turns out it is helpful in getting one of the cypress tests to pass that was proving to be a pain to fix without it.

tealefristoe:
@kswenson I ended up disabling horizontal scrolling, which seems to fix the issue you found. For what it's worth, that's also how the case card works in v2.

kswenson (approved):
👍 LGTM

>I ended up disabling horizontal scrolling, which seems to fix the issue you found. For what it's worth, that's also how the case card works in v2.

There are potential issues with non-scrollability now that the case card is wider, e.g. the behavior of the `Browse Individual Cases` button, but I think this is fine for now and we can address the scrollability question down the road.

PR #2039, opened 2025-07-23
Codap 118 case table font restyle

eireland:
Restyles the case table content font style.


v3/src/components/case-table/cell-text-editor.tsx
```diff
@@ -57,7 +58,7 @@ export default function CellTextEditor({ row, column, onRowChange, onClose }: TR
   return (
     <input
       data-testid="cell-text-editor"
-      className={textEditorClassname}
+      className={`${textEditorClassname} cell-text-editor ${isNumeric ? "numeric-format" : ""}`}
```
tealefristoe:
```suggestion
      className={clsx(textEditorClassname, "cell-text-editor", { "numeric-format": isNumeric }}
```

v3/src/components/case-table/case-table.scss
```diff
@@ -56,15 +56,16 @@ $table-body-font-size: 8pt;
           .collection-title {
             position: absolute;
             text-align: center;
-            font-family: "museo-sans", sans-serif;
-            font-size: 12px;
+            font-family: "Noto Sans", sans-serif;
+            font-size: 14px !important;
+            font-weight: 500 !important;
```
tealefristoe:
```suggestion
            font-size: 14px;
            font-weight: 500;
```
I tried removing the `!importants` here, and the text seemed to still be 14px with 500 weight. Is there a reason you had the `!importants`?

tealefristoe (changes requested):
Looks good overall 👍 

The main thing is to remove those `!importants` if it's safe, which I think it is. Please let me know if I'm mistaken about that.

v3/src/components/case-table/cell-text-editor.tsx
```diff
@@ -7,6 +7,7 @@ import { selectAllCases } from "../../models/data/data-set-utils"
 import { uiState } from "../../models/ui-state"
 import { blockAPIRequestsWhileEditing } from "../../utilities/plugin-utils"
 import { TRenderEditCellProps } from "./case-table-types"
+import clsx from "clsx"
```
tealefristoe: ⬆️ Import order

tealefristoe (approved):
Nice 👍

v3/src/components/case-table/cell-text-editor.tsx
kswenson: ... and replace default import with named import:
```suggestion
import { clsx } from "clsx"
```

v3/src/components/case-table/case-table.scss
```diff
@@ -2,7 +2,7 @@
 @use "./case-table-shared.scss" as shared;
 @use "../case-tile-common/case-tile-vars" as caseTileVars;
 
-$table-body-font-size: 8pt;
+$table-body-font-size: 12px;
```
kswenson: As discussed in the meeting yesterday, the font size changes are tied to a corresponding row height change, which we decided was not part of this PR. The new cell font size is clearly too large for the existing row height, so we should use something smaller (e.g. 10px or 11px) for now and include increasing the font size in the new row height story. I believe @mtirenin was tasked with choosing an appropriate font size for the current row height in yesterday's meeting.

v3/src/components/case-table/case-table.scss
```diff
@@ -56,15 +56,16 @@ $table-body-font-size: 8pt;
...
+            font-size: 14px;
+            font-weight: 500;
             text-overflow: ellipsis;
             cursor: text;
             vertical-align: middle;
             white-space: nowrap;
             user-select: none;
             font-style: italic;
-            color: #0f2633; // Not used anywhere else at the moment, so we're leaving it unnamed
+            color: black;
```
kswenson: My reading of the spec has the color of the title bar text as `#222`, not black. Probably worth a named constant since it occurs elsewhere.

kswenson (changes requested):
👍 Generally looks good -- a couple of minor comments, but the major one is that we need to choose a cell font size that is commensurate with the current row height.

v3/src/components/case-table/case-table.scss
eireland: This is for the collection title not the component title, which is black in the spec.

kswenson: Oops -- right you are.

PR #2040, opened 2025-07-23
CODAP-83 add recover deleted formula

scytacki:
This should work the same as CODAPv2.  I tested opening a v2 doc with a deleted formula and could recover it in v3. I also tested saving a doc with a deleted formula in v3 and opening in v2 and I was able to recover the formula.

It uses the dataset metadata to store the deleted formula which was already setup and not being used.
There was a bug in the import code for these deleted formulas which I fixed. I changed the typing of the setDeletedFormula function to prevent that kind of mistake in the future.

### DIFF
--- a/v3/src/components/case-tile-common/attribute-menu/attribute-menu-list.tsx
+++ b/v3/src/components/case-tile-common/attribute-menu/attribute-menu-list.tsx
@@ -103,7 +103,11 @@ const AttributeMenuListComponent = forwardRef<HTMLDivElement, IProps>(
       isEnabled: () => !metadata?.isEditProtected(attributeId) && !!attribute?.hasFormula,
       handleClick: () => {
         data?.applyModelChange(() => {
-          attribute?.clearFormula()
+          if (!metadata) {
+            console.warn(`Metadata not found for data set: ${data?.id}`)
+            return
+          }
+          metadata.deleteAttributeFormula(attributeId)
         }, {
           // TODO Should also broadcast notify component edit formula notification
           undoStringKey: "DG.Undo.caseTable.editAttributeFormula",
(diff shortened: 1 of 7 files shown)


v3/src/components/case-tile-common/attribute-menu/attribute-menu-list.tsx
```diff
@@ -113,7 +117,23 @@ const AttributeMenuListComponent = forwardRef<HTMLDivElement, IProps>(
...
+        data?.applyModelChange(() => {
+          if (!metadata) {
+            console.warn(`Metadata not found for data set: ${data?.id}`)
+            return
+          }
+          metadata.recoverAttributeFormula(attributeId)
+        }, {
+          undoStringKey: "DG.Undo.caseTable.editAttributeFormula",
+          redoStringKey: "DG.Undo.caseTable.editAttributeFormula",
+          log: logMessageWithReplacement("Recover formula for attribute %@", { name: attribute?.name })
```
kswenson, line 134: Out of curiosity, did you check to see whether this is the message v2 issued? We've been trying to maintain compatibility in case there are clients out there that care.

kswenson (approved):
👍 Looks good -- just one question to consider.

v3/src/components/case-tile-common/attribute-menu/attribute-menu-list.tsx
scytacki: No the log messages do not seem to match v2. I looked around and it does seem like most of the other log messages do match. So I'll try to update these.

This is what is used in v2 for recovering the formula
```
          var action = "attributeEditFormula";
          this.log = "%@: { name: '%@', collection: '%@', formula: '%@' }".fmt(
              action, tAttrName, tCollection.get('name'), tFormula);
```
where `tFormula` is the deleted formula.

And this is what is used when deleting the formula
```
          var action = "attributeEditFormula";
          this.log = "%@: { name: '%@', collection: '%@', formula: '%@' }".fmt(
              action, tAttrName, tCollection.get('name'), tFormula);
```
where `tFormula` is `''`

scytacki: I will update this PR so the log messages match for the main actions deleting a formula and recovering a formula. I'm not going to bother with the log messages that happen during Undo and Redo. In v2 these messages are customized for these actions as well.

scytacki: Aligning these messages would probably result in changes to several other log messages so that work has been punted to a new story: https://concord-consortium.atlassian.net/browse/CODAP-808

PR #2041, opened 2025-07-23
Codap-694 Position New Tiles Within Viewport

tealefristoe:
Jira story: https://concord-consortium.atlassian.net/browse/CODAP-694

This PR ensures that new tiles are added to the viewport, making them always visible to users.

I took the opportunity to clean up the `getPositionOfNewComponent` function a bit. There are still some very funky things about this function, like `iOffset`, which is always `0` for both x and y but is sprinkled throughout the function.

### DIFF
--- a/v3/src/utilities/view-utils.ts
+++ b/v3/src/utilities/view-utils.ts
@@ -4,26 +4,35 @@ const kCodapAppHeader = 95
 const kTitleBarHeight = 25
 const kGap = 5 // Also used to increment during search
 
-export const getPositionOfNewComponent = (iViewRect: {width?: number, height?: number}, iPosition = "top") => {
+export const getPositionOfNewComponent = (componentSize: { width?: number, height?: number }, iPosition = "top") => {
   // note that this will only work when the document content is a single FreeTileRow
   const parentEl = document.getElementsByClassName("free-tile-row")
   if (!parentEl?.[0]) return { x: 100, y: 100 }
-  const rowRect = parentEl[0].getBoundingClientRect()
+  const viewportLeft = parentEl[0].scrollLeft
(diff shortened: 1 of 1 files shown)


v3/src/utilities/view-utils.ts
```diff
@@ -4,26 +4,35 @@ const kCodapAppHeader = 95
...
   const existingRects = Array.from(existingEls).map(el => el.getBoundingClientRect())
-  const viewWidth = iViewRect.width || kDefaultTileWidth
-  const viewHeight = iViewRect.height || kDefaultTileHeight
+    .map(rect => ({ height: rect.height, width: rect.width, x: rect.x + viewportLeft, y: rect.y + viewportTop }))
+  const componentWidth = componentSize.width || kDefaultTileWidth
+  const componentHeight = componentSize.height || kDefaultTileHeight
 
+  const iOffset = { x: 0, y: 0 }
+  const initialX = kGap + viewportLeft + iOffset.x
+    const tStartAtBottom = (iPosition === 'bottom')
```
kswenson:
```suggestion
  const tStartAtBottom = (iPosition === 'bottom')
```

v3/src/utilities/view-utils.ts
```diff
@@ -4,26 +4,35 @@ const kCodapAppHeader = 95
...
+  const initialY = (tStartAtBottom ? viewportBottom - componentHeight - kGap : kGap) + iOffset.y
```
kswenson: Need to account for `viewportTop` here, something like:
```suggestion
  const initialY = (tStartAtBottom ? viewportBottom - componentHeight - kGap : viewportTop + kGap) + iOffset.y
```

PR #2042, opened 2025-07-24
Codap-797 Beta Button

tealefristoe:
Jira story: https://concord-consortium.atlassian.net/browse/CODAP-797

This PR adds a button to Codap that allows users to submit feedback on the v3 beta. The button only shows when the `release=beta` url parameter is present.


v3/src/components/beta/beta-button.tsx
```diff
@@ -0,0 +1,27 @@
+import React from "react"
+import { urlParams } from "../../utilities/url-params"
+
+import FeedbackIcon from "../../assets/icons/beta/feedback-icon.svg"
+
+import "./beta-button.scss"
+
+export function BetaButton() {
+  if (urlParams.release !== "beta") return null
```
kswenson: We also want to show this when the current url contains `codap3.concord.org/version/beta`, which doesn't exist yet. You can test it locally, however, by including it in a list with `codap3.concord.org/branch/beta-button`, which would mean that it could be tested on the branch without the `release=beta` url parameter.

v3/src/index.html
```diff
@@ -7,7 +7,9 @@
     <link rel="preconnect" href="https://fonts.googleapis.com">
     <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
     <link href="https://fonts.googleapis.com/css2?family=Noto+Sans+Mono:ital,wght@0,100..900;1,100..900&family=Lato:ital,wght@0,100;0,300;0,400;0,700;0,900;1,100;1,300;1,400;1,700;1,900&display=swap" rel="stylesheet">
-    <link href="https://fonts.googleapis.com/css2?family=Noto+Sans:ital,wght@0,100..900;1,100..900&family=Roboto:ital,wght@0,100..900;1,100..900&display=swap" rel="stylesheet">    <!--
+    <link href="https://fonts.googleapis.com/css2?family=Noto+Sans:ital,wght@0,100..900;1,100..900&family=Roboto:ital,wght@0,100..900;1,100..900&display=swap" rel="stylesheet">
+    <link href="https://fonts.googleapis.com/css2?family=Gloria+Hallelujah&family=Montserrat:ital,wght@0,100..900;1,100..900&family=Noto+Color+Emoji&display=swap" rel="stylesheet">
```
kswenson, line 12: I wish there were some way to conditionalize this so that we don't end up shipping the Gloria Hallelujah font when we don't need it. At the very least a comment would be good to hopefully remind us to remove it once we're out of beta.

v3/src/components/beta/beta-button.tsx
kswenson: > We also want to show this when the current url contains `codap3.concord.org/version/beta`, which doesn't exist yet. You can test it locally, however, by including it in a list with `codap3.concord.org/branch/beta-button`, which would mean that it could be tested on the branch without the `release=beta` url parameter.

On second thought, I think the beta url should be `codap3.concord.org/version/3.0.0-beta`.

tealefristoe: I ended up testing for "https://codap3.concord.org/beta" and "https://codap3.concord.org/branch/beta". Right now the beta button is the only beta feature, but if there are any others we want to add, with the set up they can be tested on their deployment branch if it starts with "beta" without jumping through any other hoops.

I wasn't sure where to put the function that checks if we're in beta mode, so I added a new `version-utils.ts`. But let me know if there's a better place to put this.

kswenson (approved):
👍 LGTM

PR #2045, opened 2025-07-25
add Dutch and update all other languagues

ddamelin:
This is not a real pull request (or at least probably isn't). I'm trying to get a test build of v2 that includes the new Dutch language, so I can test it with the other branch builds I have for SageModeler

https://sagemodeler.concord.org/branch/dutch-translation/app/
https://sage.concord.org/branch/dutch-translation/sage.html
https://cloud-file-manager.concord.org/branch/dutch-translation/

The only thing I'm missing is a link to a CODAP build (doesn't need to be public) that includes the new translation.

I tried for a while to get CODAP building on my local machine but just could not get around the issues on my own. (I did have it running last year...)

I figured creating a branch, making the edits I think need to be made to include the new translation (and updating the other languages), and making a PR would be the way I could possibly be helpful in moving this forward.

It does seem that perhaps we have changed some things about how translation strings are stored and referred to since the Developer Guide instructions were written, so this might be overkill, or I might have guessed wrong about making an nl.proj directory and copying the strings file from the v3 folder that got downloaded with the script that retrieves files from POEditor.

Anyway, feel free to ignore this if it is not helpful.


kswenson:
Superseded by #2391.

PR #2047, opened 2025-07-26
Codap-81 Unsaved Changes Warning

tealefristoe:
Jira story: https://concord-consortium.atlassian.net/browse/CODAP-81

This PR displays a warning when the user tries to navigate away from a document with unsaved changes.

This behavior was breaking every cypress test, so there's also a new url parameter that disables this behavior, which has been turned on for all cypress tests. This change resulted in the vast majority of line changes in this PR.


v3/src/components/app.tsx
```diff
@@ -168,10 +169,16 @@ export const App = observer(function App() {
 
       appState.enableDocumentMonitoring()
       Logger.initializeLogger(appState.document)
+
+      window.onbeforeunload = function() {
+        if (!urlParams.suppressUnsavedWarning && cfm.client.state.dirty) {
```
kswenson: In general the boolean url parameters that are declared as `propName?: string | null` are designed to be used without an argument, e.g. `?propName` is considered true. For that to work, the test needs to be `!== undefined` because a property that is present without a value is parsed as `null` while a property that is not present at all is treated as `undefined`. So my inclination would be to change this to:
```suggestion
        if (urlParams.suppressUnsavedWarning !== undefined && cfm.client.state.dirty) {
```
You _could_ then also change the cypress tests to not have the `=true`, but that's not necessary.

While looking into this I also noticed that the `noComponentAnimation` flag is sometimes used in this way, but the test for it in `case-view.tsx` is `!= null` -- this should really be changed to `!== undefined` although it's not directly related to this PR, of course.

kswenson (approved):
👍 Looks good -- one suggestion about using `!== undefined`.

PR #2048, opened 2025-07-26
Codap-768 Card Swipe Followup

tealefristoe:
Jira story: https://concord-consortium.atlassian.net/browse/CODAP-768

This PR switches the direction cards flip.

It also runs `validateCases` before accessing the selected cases within collections, ensuring that the accessed information is up to date. The views in `data-set.ts` were shifted so `validateCases` is defined, but the only actual change to the code is the call to `validateCases`.

### DIFF
--- a/v3/package-lock.json
+++ b/v3/package-lock.json
@@ -36280,7 +36280,8 @@
       }
     },
     "domhandler": {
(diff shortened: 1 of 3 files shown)


kswenson (approved):
👍 LGTM

PR #2049, opened 2025-07-26
Codap-774 Focus Flipped Tile

tealefristoe:
Jira story: https://concord-consortium.atlassian.net/browse/CODAP-774

This PR makes the case card or case table tile become focused once the other tile is switched to it, bringing it to the front.


kswenson (approved):
👍 LGTM

PR #2050, opened 2025-07-28
feat: v2 import/export for plugins

kswenson:
[[CODAP-318](https://concord-consortium.atlassian.net/browse/CODAP-318)] Plugins v2 import/export

- import/export plugin-settable properties of `WebViewModel`, e.g. `preventAttributeDeletion`, etc.
- import/export `DG.GameContext` (derived class of `DG.DataContext`)
- import/export link between `WebViewModel` and data context
- import/export `userSetTitle` property of components
  - previous efforts assumed that if `_title` ≠ `name` then it must have been user-set, but plugins can set the title as well, so we need to explicitly track whether the user has set the title.
- import/export default attributes for a collection (often set by plugins)

[CODAP-318]: https://concord-consortium.atlassian.net/browse/CODAP-318?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/v3/src/models/tiles/tile-model.ts
+++ b/v3/src/models/tiles/tile-model.ts
@@ -4,7 +4,7 @@ import { getParent, getSnapshot, getType,
 import { applyModelChange } from "../history/apply-model-change"
 import { StringBuilder } from "../../utilities/string-builder"
 import { v3Id, typeV3Id } from "../../utilities/codap-utils"
-import { V2Model } from "../data/v2-model"
+import { V2UserTitleModel } from "../data/v2-user-title-model"
 import { DisplayUserTypeEnum } from "../stores/user-types"
 import { ITileContentModel } from "./tile-content"
 import { getTileContentInfo, ITileExportOptions } from "./tile-content-info"
@@ -48,15 +48,7 @@ export function getTileModel(tileContentModel: ITileContentModel) {
   }
 }
 
-export function getTileTitleFromContent(tileContentModel: ITileContentModel) {
-  return getTileModel(tileContentModel)?.title
-}
-
-export function setTileTitleFromContent(tileContentModel: ITileContentModel, title: string) {
-  getTileModel(tileContentModel)?.setTitle(title)
-}
-
-export const TileModel = V2Model.named("TileModel")
+export const TileModel = V2UserTitleModel.named("TileModel")
   .props({
     // if not provided, will be generated
     id: typeV3Id("TILE"),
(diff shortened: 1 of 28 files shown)


v3/src/models/tiles/tile-model.ts
```diff
@@ -48,15 +48,7 @@ export function getTileModel(tileContentModel: ITileContentModel) {
   }
 }
 
-export function getTileTitleFromContent(tileContentModel: ITileContentModel) {
-  return getTileModel(tileContentModel)?.title
-}
-
-export function setTileTitleFromContent(tileContentModel: ITileContentModel, title: string) {
-  getTileModel(tileContentModel)?.setTitle(title)
-}
```
kswenson, line 57: These functions were unused.

v3/src/v2/codap-v2-data-context-types.ts
```diff
@@ -205,7 +208,7 @@ export interface ICodapV2GameContextStorage extends ICodapV2DataContextStorage {
 }
 
 export type CodapV2Context = ICodapV2DataContext | ICodapV2GameContext | ICodapV2ExternalContext
-export const isV2ExternalContext = (context: CodapV2Context): context is ICodapV2ExternalContext =>
-  !("type" in context) && ("externalDocumentId" in context)
-export const isV2InternalContext = (context: CodapV2Context): context is ICodapV2DataContext | ICodapV2GameContext =>
-  ("type" in context)
+export const isV2ExternalContext = (context?: CodapV2Context): context is ICodapV2ExternalContext =>
```
scytacki:
```suggestion
export const isV2ExternalContext = (context: Maybe<CodapV2Context>): context is ICodapV2ExternalContext =>
```
Using maybe here seems better because it make sure the user passes some argument to function. Calling it without an argument seems like an error.

v3/src/v2/codap-v2-data-context-types.ts
```diff
@@ -205,7 +208,7 @@ export interface ICodapV2GameContextStorage extends ICodapV2DataContextStorage {
...
+  !!context && !("type" in context) && ("externalDocumentId" in context)
+export const isV2InternalContext = (context?: CodapV2Context): context is ICodapV2DataContext | ICodapV2GameContext =>
```
scytacki:
```suggestion
export const isV2InternalContext = (context: Maybe<CodapV2Context>): context is ICodapV2DataContext | ICodapV2GameContext =>
```
Same as above.

PR #2051, opened 2025-07-28
Codap-694 Stop Scrolling when Creating New Table

tealefristoe:
Jira story: https://concord-consortium.atlassian.net/browse/CODAP-694

This PR disables animation when creating a new table, which was causing the document to scroll to the origin, even if the table would ultimately be far from the origin. A TODO has been added to turn the animation back on and fix the underlying issue, which is that the document scrolls in the first place.


kswenson (approved):
👍 LGTM

PR #2053, opened 2025-07-29
Codap-803 Inspector Panel Restyle

tealefristoe:
Jira story: https://concord-consortium.atlassian.net/browse/CODAP-803

This PR updates all of the inspector panel styling (except for the text tile).

Additionally, the tile title bar color has been updated, because the old colors do not work with the updated fonts.


v3/src/components/map/components/map-inspector.tsx
```diff
@@ -1,21 +1,22 @@
...
-import {isMapPointLayerModel} from "../models/map-point-layer-model"
-import {HideShowMenuList} from "./inspector-panel/hide-show-menu-list"
-import {SaveImageMenuList} from "./inspector-panel/save-image-menu-list"
-import {MapMeasurePalette} from "./inspector-panel/map-measure-palette"
-import {MapLayersPalette} from "./inspector-panel/map-layers-palette"
+import React, { useRef, useEffect, useState } from "react"
+import { observer } from "mobx-react-lite"
+import { InspectorButton, InspectorMenu, InspectorPanel } from "../../inspector-panel"
+import { t } from "../../../utilities/translation/translate"
+import { useDndContext } from "@dnd-kit/core"
```
kswenson: Move import ⬆️

v3/src/components/inspector-panel.tsx
```diff
@@ -11,63 +11,82 @@ interface IProps {
...
 
-export const InspectorPanel = forwardRef(({ component, show, setShowPalette, children }: IProps, ref) => {
+export const InspectorPanel = forwardRef(({ component, show, setShowPalette, children, width }: IProps, ref) => {
   useOutsidePointerDown({
     ref: ref as unknown as RefObject<HTMLElement>,
     handler: ()=> setShowPalette?.(undefined),
     enabled: !!(show && ref && setShowPalette),
     info: {name: "InspectorPanel", component}
   })
+  const classes = clsx("inspector-panel", component ?? "", width ?? "normal")
```
kswenson:
```suggestion
  const classes = clsx("inspector-panel", component, width ?? "normal")
```
I believe `clsx` is smart enough to ignore `null`/`undefined`.

kswenson (approved):
👍 Looks great! Nicely (and quickly) done! Just a couple of nits to pick.

Since you were so quick with this, one other related thing that could be done as part of this PR and would be worth doing for the beta is that we now have two different ways of indicating that an inspector menu item is unimplemented. More recently, we've simply been disabling the menu item and adding the 🚧 icon. Before that we used Chakra to pop up a "toast" which says something like "Menu Item clicked!" The inconsistency is annoying and people tend to get confused by the toasts -- "It's a bug!" -- so I think it would be worth eliminating the remaining toasts and just disabling all of the appropriate menu items and adding the 🚧 icon. If you'd prefer to treat that as a separate PR, that's fine, too, but it is right in this neighborhood and I was reminded of it while testing this PR.

tealefristoe:
@kswenson Switching all the toasts to grayed out under construction options seems like a good thing to do before the beta release. I'll take care of that in a separate PR tomorrow.

PR #2055, opened 2025-07-30
CODAP-37 only enable autosave if the document was saved by v3

scytacki:
This disables autoSave if the loaded document is in v2 format and the appVersion does not start with `3.`. 
The exception is when the `interactiveApi` parameter is used. In this case autoSave remains enabled. This exception is because the file menu is hidden in this case so there would be no way for the user to save the document.

Notes:
- Even if autoSave is disabled, inside of the AP the CODAP document is saved when the user changes pages. So disabling autoSave really didn't make sense without additional work.
- After the user saves a document that has autoSave disabled, autoSave is enabled. We might want to do that to help the user not lose work and to have a consistent experience with other cases.
- This PR does not change default saving format of CODAPv3. It will still save in the v3 format unless the debug flag `saveAsV2` is set.

Other changes:
- fix the "user entry box" which was staying open when reloading a local storage file.
- remove unused errorBoundaryErrors list
- a document was added explaining how to setup google drive for local testing

### DIFF
--- a/v3/doc/google-drive.md
+++ b/v3/doc/google-drive.md
@@ -15,7 +15,7 @@ These values can be found in the Google Cloud interface. The cloud project is ca
 Find the project dashboard: If you click on the hamburger menu on the top left and then "Cloud overview", then select "Dashboard. Now copy the "Project number". This is the AppId that you need.
 
 ### ClientId
-In the cloud project's API & Services section, go to the Credentials section. Then in the OAuth 2.0 Client IDs section choose the open the "codap-dev" and copy the Client ID. This client has been configured for local development, look at its Authorized JavaScript origins and make sure you are running CODAP locally from a matching origin.
+In the cloud project's API & Services section, go to the Credentials section. Then in the OAuth 2.0 Client IDs section open the "codap-dev" and copy the Client ID. This client has been configured for local development. Look at its Authorized JavaScript origins and make sure you are running CODAP locally from a matching origin.
 
 ### ApiKey
 In the cloud project's API & Services section, go to the Credentials section. Then in the API Keys section is a list of api keys (currently there is only one). Open the "Browser key 1" key, then on the page that comes up click the "Show key" button.
--- a/v3/src/lib/cfm/handle-cfm-event.ts
+++ b/v3/src/lib/cfm/handle-cfm-event.ts
@@ -12,7 +12,7 @@ import { hideSplashScreen } from "./splash-screen"
 import { isCodapV2Document } from "../../v2/codap-v2-types"
 
 // -1 is used to disable autosave because the CFM's client.autoSave function only takes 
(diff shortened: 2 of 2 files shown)


v3/doc/google-drive.md
```diff
@@ -0,0 +1,21 @@
...
+GOOGLE_DRIVE_API_KEY=[ApiKey]
+```
+
+These values can be found in the Google Cloud interface. The cloud project is called "CODAP". It is **not** inside of the Concord Consortium organization.
+
+### AppId
+Find the project dashboard: If you click on the hamburger menu on the top left and then "Cloud overview", then select "Dashboard. Now copy the "Project number". This is the AppId that you need.
+
+### ClientId
+In the cloud project's API & Services section, go to the Credentials section. Then in the OAuth 2.0 Client IDs section choose the open the "codap-dev" and copy the Client ID. This client has been configured for local development, look at its Authorized JavaScript origins and make sure you are running CODAP locally from a matching origin.
```
kswenson:
```suggestion
In the cloud project's API & Services section, go to the Credentials section. Then in the OAuth 2.0 Client IDs section open the "codap-dev" and copy the Client ID. This client has been configured for local development. Look at its Authorized JavaScript origins and make sure you are running CODAP locally from a matching origin.
```

v3/src/lib/cfm/handle-cfm-event.ts
```diff
@@ -5,10 +5,17 @@ import pkg from "../../../package.json"
...
 import { removeDevUrlParams, urlParams } from "../../utilities/url-params"
-import { DEBUG_CFM_EVENTS } from "../debug"
+import { DEBUG_CFM_EVENTS, DEBUG_CFM_NO_AUTO_SAVE } from "../debug"
 import { wrapCfmCallback } from "./cfm-utils"
 import { resolveDocument } from "./resolve-document"
 import { hideSplashScreen } from "./splash-screen"
+import { isCodapV2Document } from "../../v2/codap-v2-types"
+
+// -1 is used to disable autosave because the CFM's client.autoSave function only takes 
+// numbers and -1 it is more clear than 0. Also if the autoSaveIntervale is falsy, then
```
kswenson:
```suggestion
// numbers and -1 it is more clear than 0. Also if the autoSaveInterval is falsy, then 
```

v3/src/lib/cfm/handle-cfm-event.ts
```diff
@@ -116,6 +161,7 @@ export async function handleCFMEvent(cfmClient: CloudFileManagerClient, event: C
       break
     }
     case "savedFile": {
+      // TODO: We might will want to enable autoSave if we had disabled it previously.
```
kswenson:
```suggestion
      // TODO: We might want to enable autoSave if we had disabled it previously.
```

PR #2056, opened 2025-07-30
Codap-775 Fix Dropping on Overlapping Tiles

tealefristoe:
Jira story: https://concord-consortium.atlassian.net/browse/CODAP-775

This PR fixes a bug where certain tiles would steal an attribute drop from other tiles, even when below them.

There were two problems with the existing code:
- Previously, the containers were sorted by tile z-index before checking for collisions. But this did not guarantee that the resulting collisions were in the same order. Now we sort the collisions after they are found.
- Previously, we'd use the first tile-specific collision handler we could find. That means that we'd ignore tiles that didn't have a special handler. Now, we determine all collisions for the top tile, then only look for a special handler for it. If we don't find one, we just return the collisions in tile order.

### DIFF
--- a/v3/src/lib/dnd-kit/dnd-detect-collision.ts
+++ b/v3/src/lib/dnd-kit/dnd-detect-collision.ts
@@ -53,7 +53,7 @@ export const dndDetectCollision: CollisionDetection = (args) => {
           // filter the drop zones to those appropriate for the relevant tile
           const containers =
             args.droppableContainers.filter(({id: containerId}) => droppableRegex.test(`${containerId}`))
-          // apply the collection detection function specified by the tile
+          // apply the collision detection function specified by the tile
           return detect({ ...args, droppableContainers: containers })
         }
       }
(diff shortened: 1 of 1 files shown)


v3/src/lib/dnd-kit/dnd-detect-collision.ts
```diff
@@ -38,18 +39,26 @@ export const dndDetectCollision: CollisionDetection = (_args) => {
...
+      const tileCollisions = sortedCollisions.filter(c => `${c.id}`.startsWith(prefix))
+      for (const collision of tileCollisions) {
+        const { id: collisionId } = collision
+        const handler = gTileCollisionDetectionRegistry.find(({overlayRegex}) => overlayRegex.test(`${collisionId}`))
+        if (handler) {
+          const { droppableRegex, detect } = handler
+          // filter the drop zones to those appropriate for the relevant tile
+          const containers =
+            args.droppableContainers.filter(({id: containerId}) => droppableRegex.test(`${containerId}`))
+          // apply the collection detection function specified by the tile
```
kswenson:
```suggestion
          // apply the collision detection function specified by the tile
```
Pre-existing condition, but may as well fix it since you're in the neighborhood. 😉

kswenson (approved):
👍 Looks great

PR #2057, opened 2025-07-30
Convert Unimplemented Menu Items from Toast to Under Construction

tealefristoe:
This PR makes menu items that haven't been implemented very obvious by graying them out and adding an icon, rather than making them look usable but only displaying a toast when clicked.

### DIFF
--- a/v3/src/components/beta/unimplemented-menu-item.tsx
+++ b/v3/src/components/beta/unimplemented-menu-item.tsx
@@ -1,0 +1,18 @@
+import { MenuItem } from "@chakra-ui/react"
+import React from "react"
+
+import ConstructionIcon from "../../assets/icons/icon-alert.svg"
+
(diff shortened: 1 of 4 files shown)


v3/src/components/beta/unimplemented-menu-item.tsx
```diff
@@ -0,0 +1,14 @@
...
+import React from "react"
+
+interface IUnimplementedMenuItemProps {
+  label: string
+  testId?: string
+}
+export function UnimplementedMenuItem({ label, testId }: IUnimplementedMenuItemProps) {
+  return (
+    <MenuItem isDisabled data-testid={testId}>
+      {`🚧 ${label}`}
```
kswenson:
```suggestion
      {`${label} 🚧`}
```
The convention elsewhere is to put the under construction symbol at the end of the label.

PR #2059, opened 2025-07-30
CODAP-792 Make rescale button for bar chart work

bfinzer:
[#CODAP-792] Bug fix: The rescale button does nothing when the plot is a bar chart

* We need to special case the barChart in `GraphContentModel.rescale` so that we use the counts, percentages, or formula results of all the bars to determine the maximum for the `Count` axis. So we move responsibility for computing the numeric values appropriate to a given attribute role to the plot so that the `BarChartModel` can handle its breakdown types.

### DIFF
--- a/v3/src/components/graph/plots/bar-chart/bar-chart-model.ts
+++ b/v3/src/components/graph/plots/bar-chart/bar-chart-model.ts
@@ -211,7 +211,7 @@ export const BarChartModel = DotChartModel
         case "percent":
           return [0, dataConfiguration.maxPercentAllCells(epRole, esRole)]
         case "formula":
-          return this.getMinMaxOfFormulaValues().concat(0)
+          return [0, ...this.getMinMaxOfFormulaValues()]
         default:
           return []
       }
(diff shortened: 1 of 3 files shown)


v3/src/components/graph/plots/bar-chart/bar-chart-model.ts
```diff
@@ -198,6 +198,23 @@ export const BarChartModel = DotChartModel
...
+      if (!dataConfiguration) return []
+      const epRole = dataConfiguration.primaryRole === 'x' ? 'topSplit' : 'rightSplit'
+      const esRole = dataConfiguration.primaryRole === 'x' ? 'rightSplit' : 'topSplit'
+      switch (self.breakdownType) {
+        case "count":
+          return [0, dataConfiguration.maxOverAllCells(epRole, esRole)]
+        case "percent":
+          return [0, dataConfiguration.maxPercentAllCells(epRole, esRole)]
+        case "formula":
+          return this.getMinMaxOfFormulaValues().concat(0)
```
kswenson:
```suggestion
          return [0, ...this.getMinMaxOfFormulaValues()]
```

kswenson: Code is more consistent with the other options and perhaps marginally more efficient.

PR #2060, opened 2025-07-30
Codap-101 Import Dataset Metadata

tealefristoe:
Jira story: https://concord-consortium.atlassian.net/browse/CODAP-101

This PR adds support for dataset metadata (`source`, `importDate`, and `description`) when handling `create dataContext` API requests.

### DIFF
--- a/v3/cypress/e2e/table.spec.ts
+++ b/v3/cypress/e2e/table.spec.ts
@@ -105,7 +105,7 @@ context("case table ui", () => {
     it("should open dataset information button and make changes", () => {
       const newInfoName = "Animals",
         newSource = "The Internet",
-        importDate = "May 4",
+        importDate = "May 4, 2021",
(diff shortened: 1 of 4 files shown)


kswenson (approved):
👍 LGTM

tealefristoe:
@kswenson Maybe not worth worrying about while we're pushing for the beta, but I realized that the previous implementation where the formatted string was saved was incorrect, so I'm requesting another review. Now, an ISO timestamp is saved when possible, and a formatted version is displayed. If an ISO timestamp can't be interpreted from the string submitted by a user, the raw string is saved. This isn't quite how v2 works--there, dates that could be interpreted as an ISO timestamp are saved as the raw string. I think the current behavior is better, but if we do want to make it more like v2, I think it would be best to create a new story for it, since this is already getting pretty far removed from the initial intent of this one.

PR #2061, opened 2025-07-30
CODAP-807 enable logging when served from a host ending in .concord.org

scytacki:
this should also enable analytics when on a matching host


kswenson (approved):
👍 LGTM

PR #2064, opened 2025-07-31
Codap-814 Update Tile Minimize and Close Icons

tealefristoe:
Jira story: https://concord-consortium.atlassian.net/browse/CODAP-814

This PR updates the minimize and close icons for tile title bars and updates the other related styling.


kswenson (approved):
👍 Looks good -- ideally, named constants would be used for the colors, particularly ones that are likely to recur, but that can wait for another day if you don't want to deal with it now.

tealefristoe:
@kswenson I think we'll have to revisit these colors anyway. They should be transparent but I don't know how to get transparent colors from zeplin. Not important enough to hold up the beta.

PR #2068, opened 2025-07-31
Fix Connecting Line Position

tealefristoe:
This PR fixes a bug where the lines connecting parent and child cases in the table were offset from their correct position.


kswenson (approved):
👍 LGTM

PR #2071, opened 2025-08-01
fix: case names for collapsed cases in table [CODAP-363]

kswenson:
[[CODAP-363](https://concord-consortium.atlassian.net/browse/CODAP-363)] The displayed text for collapsed cases in the case table should match v2

While fixing the cypress tests, I also increased the size of the default case table when using `?sample=...`. A number of our cypress tests rely on dragging `Diet` and `Habitat` with `?sample=mammals`, but those attributes aren't visible with the default width of the table.

[CODAP-363]: https://concord-consortium.atlassian.net/browse/CODAP-363?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/v3/cypress/fixtures/hierarchical.json
+++ b/v3/cypress/fixtures/hierarchical.json
@@ -13,7 +13,7 @@
               "move": "newCollection"
             }
           ],
-          "childCases": [ "9 cases", "11 cases", "7 cases"],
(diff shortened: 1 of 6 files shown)


tealefristoe (approved):
Looks good 👍

PR #2072, opened 2025-08-02
Codap-788 Hierarchical Plugin Menu

tealefristoe:
Jira story: https://concord-consortium.atlassian.net/browse/CODAP-788


v3/src/components/tool-shelf/debug-plugins.json
```diff
@@ -1,13 +1,13 @@
...
-    "subMenu": [
+    "plugins": [
       {
         "title": "Plugin API Tester",
         "description": "Exercise the CODAP Plugin API and see the results.",
         "width": 420,
         "height": 540,
         "path": "https://concord-consortium.github.io/codap-data-interactives/DataInteractiveAPITester/index.html",
-        "icon": "/TP-Sampler/icon-sampler.svg",
+        "icon": "SamplerIcon",
```
kswenson: Should get a better icon for this at some point. Do we have a generic plugin icon?

v3/src/components/tool-shelf/standard-plugins.json
```diff
@@ -159,6 +230,17 @@
         "categories": [
           "Tools"
         ]
+      },
+      {
+        "title": "Testimate",
+        "description": "Test hypotheses using classical inferential methods.",
+        "path": "https://codap.concord.org/releases/latest/extn/plugins/eepsmedia/plugins/testimate/index.html",
```
kswenson, line 237: This is an example of depending on the v2 build that we should really avoid, but that can be left for a follow-up PR.

v3/src/components/tool-shelf/tool-shelf.scss
```diff
@@ -112,127 +112,121 @@ $tool-shelf-border-width: 1px;
...
+    &.top-menu {
+      margin-top: -13px;
     }
 
-    &:active {
-      background-color: #e9e9e9;
+    &.tool-shelf-menu-list:focus,
+    &.tool-shelf-menu-list[tabindex]:focus,
+    &.tool-shelf-menu-list:active {
+      // TODO: This is necessary because of the following line in index.scss, which should be removed:
```
kswenson:
```suggestion
      // TODO: This `!important` is necessary because of the following line in index.scss, which should be removed:
```

v3/src/utilities/translation/translate.ts
```diff
@@ -50,3 +50,11 @@ export function translate (key: string, options?: ITranslateOptions) {
 }
 
 export const t = translate
+
+// Non-roman character languages require special styling so we add a class to
+// components that use these languages.
+export const getSpecialLangFontClassName = (languageCode?: string) => {
```
kswenson: Could also use:
```suggestion
export const getSpecialLangFontClassName = (languageCode = gLocale.current) => {
```

kswenson (approved):
👍 Looks great -- just a smattering of small suggestions.

v3/src/components/tool-shelf/debug-plugins.json
tealefristoe: I'll use the plugins button icon for now. Seems more appropriate than the sampler icon.

PR #2073, opened 2025-08-02
CODAP-828 fix render warning in table

scytacki:
The issue was that a list of `<>` components was being returned. If `<Fragment>` is used then a key can be added to component.


kswenson (approved):
👍 LGTM -- thx for figuring that out!

PR #2075, opened 2025-08-04
CODAP-831 Flip histogram rescale crash

bfinzer:
[#CODAP-831] Bug fix: Crash after flipping histogram axes and rescaling count axis

* In numeric-axis-drag-rects.tsx we need to clean up the drag behaviors when the component goes away so we're not hanging onto axis models that are no longer part of the state tree.
* Also make sure the drag rects are removed from the dom.

### DIFF
--- a/v3/src/components/axis/components/numeric-axis-drag-rects.tsx
+++ b/v3/src/components/axis/components/numeric-axis-drag-rects.tsx
@@ -1,7 +1,7 @@
 import {observer} from "mobx-react-lite"
 import React, {useEffect, useRef} from "react"
 import { comparer, reaction } from "mobx"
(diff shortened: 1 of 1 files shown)


v3/src/components/axis/components/numeric-axis-drag-rects.tsx
```diff
@@ -35,7 +35,10 @@ export const NumericAxisDragRects = observer(
       layout = useAxisLayoutContext(),
       displayModel = useDataDisplayModelContextMaybe()
 
+    console.log("NumericAxisDragRects render", axisModel.type, "lockZero =", lockZero, "place =", place)
+
```
kswenson:
```suggestion
```

PR #2076, opened 2025-08-05
CODAP-826 left & right-numeric scales problem

bfinzer:
[#CODAP-826] Bug fix: The graph rescale button doesn't work properly with two numeric y axes

* `GraphDataConfigurationModel.numericValuesForAttrRole` was wrongly including values for a right numeric attribute in the values for a y attribute. We fix so that it only includes values for the attributes on the left axis.
* Adjusted graph-component-handler.test.ts because left y axis gets different bounds


kswenson (approved):
👍 LGTM

PR #2077, opened 2025-08-05
Codap-830 Tile Restyle

tealefristoe:
Jira story: https://concord-consortium.atlassian.net/browse/CODAP-830

This PR addresses many updates to styling for general tiles.
- Tiles have outlines that differ based on whether the tile is focused or not.
- The resize tile icon has been updated.
- The title display and input has been changed and is more like CLUE now.
- The icons for switching between case card and table have been updated.
- Minimized tiles are now the correct size and animate in both directions.
  - I had to make one change to make this work which might have broken something. I left a `FIXME` in `codap-component.tsx` to indicate this spot.
  
Bugs fixed along the way:
- The `onPointerDown` handler moved from the minimize icon to its containing button.
- `ComponentResizeWidgets` is now wrapped in `observer` so the resize icon will disappear when the tile is no longer focused.
  
I also cleaned up css and removed more `!importants` whenever I could.

There were a lot of changes requested, and I'm sure I got some wrong and missed some. Here are a couple I know I missed, which I think aren't worth the effort or are problematic:
- The spec suggests hovering over a tile in the tiles menu should be the tile to the front, even if it isn't focused. Right now, the focused tile is always in the front. I can understand why it might be helpful to show where a tile is, but this seemed like it would create some unwanted situations where tiles are shifting around too much, and also might cause some bugs if the focused tile is not always on top.
- The spec suggests that the inspector panel should be duplicated, one on top of the tile and one below it, so drop shadows will show up properly. This seems like overkill to me for a very subtle effect.


v3/src/components/codap-component.tsx
```diff
@@ -71,14 +72,19 @@ export const CodapComponent = observer(function CodapComponent(props: IProps) {
...
-  const classes = clsx("codap-component", tileEltClass, { minimized: isMinimized },
-                    { shadowed: uiState.isFocusedTile(tile.id) || uiState.isHoveredTile(tile.id) })
+  const classes = clsx("codap-component", tileEltClass, { focused, minimized: isMinimized })
   return (
     <TileModelContext.Provider value={tile}>
       <TileSelectionContext.Provider value={tileSelection}>
         <CodapComponentContext.Provider value={codapComponentRef}>
-          <div className={classes} ref={codapComponentRef} key={tile.id} data-testid={tileEltClass}
-            onFocus={handleFocusEvent} onPointerDownCapture={handleFocusEvent}>
+          {/* FIXME: Is it ok to remove the onFocus handler and change onPointerDownCapture to onPointerDown? */}
```
kswenson: I think not. This means that clicking on a tile will only focus the tile for tiles that don't swallow the event (which is probably what's happening in the case of the attribute headers in the table). My proposal for this PR would be to put this back the way it was and then make a new story for special-casing the behavior of the minimize button.

kswenson:
> * Minimizing or maximizing a tile no longer focuses the tile

I can understand why minimizing the tile shouldn't focus it, but it seems to me that maximizing should.

kswenson (changes requested):
👍 Looks great overall -- I think the tile focus changes should be reverted for the purpose of this PR and addressed later. Focusing the tile when clicking on it (e.g. an attribute header) is more important than _not_ focusing it when clicking on the minimize button. We can figure out how to achieve both at a later date.

kswenson (approved):
👍 Looks great!

PR #2080, opened 2025-08-11
CODAP 816 Use V2 document background

bfinzer:
[#CODAP-816] Feature: Return to document background used in V2

We no longer need the slight differentiation provided by the V3 background we were using.
* Changed `container.scss` to refer to `bg-grid-grey.png` as in V2


kswenson (approved):
👍 LGTM

PR #2081, opened 2025-08-11
CODAP-833 Failure to swap scatterplot axes properly

bfinzer:
[#CODAP-833] Bug: Swapping attributes on graph axes not working in this document

* The attributes on the x and y axes had value such that the bounds for the two axes were the same, so one possible reaction that is often triggered wasn't being triggered.
* The mstReaction set up to detect changes in attribute assignment was calling `syncModelWithAttributeConfiguration` and only refreshing point positions if this function returned true, which in this situation it wasn't because there was no change in model required. Removing dependence on the result of this call fixes the bug.
* There was a bit of fussing to do with returned boolean values whereby I first thought I could remove them but found out otherwise by running jest tests.

### DIFF
--- a/v3/src/components/graph/hooks/use-plot.ts
+++ b/v3/src/components/graph/hooks/use-plot.ts
@@ -137,10 +137,9 @@ export const usePlotResponders = (props: IPlotResponderProps) => {
     const disposer = mstReaction(
       () => GraphAttrRoles.map((aRole) => dataConfiguration?.attributeID(aRole)),
       () => {
-        if (syncModelWithAttributeConfiguration(graphModel, layout)) {
-          startAnimation()
-          callRefreshPointPositions()
-        }
+        syncModelWithAttributeConfiguration(graphModel, layout)
+        startAnimation()
+        callRefreshPointPositions()
       }, {name: "usePlot [attribute assignment]"}, dataConfiguration
     )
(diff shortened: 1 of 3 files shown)


v3/src/components/graph/hooks/use-plot.ts
```diff
@@ -137,10 +137,9 @@ export const usePlotResponders = (props: IPlotResponderProps) => {
     const disposer = mstReaction(
       () => GraphAttrRoles.map((aRole) => dataConfiguration?.attributeID(aRole)),
       () => {
-        if (syncModelWithAttributeConfiguration(graphModel, layout)) {
-          startAnimation()
-          callRefreshPointPositions()
-        }
+        syncModelWithAttributeConfiguration(graphModel, layout)
+        startAnimation()
+        callRefreshPointPositions()
```
kswenson, line 142: The intended purpose of the earlier code was to avoid doing work when it's not necessary (or perhaps desirable in the case of animation). So either 1) it is always necessary to do this work (as you have now coded it) or 2) it is sometimes appropriate to skip these steps but the previous attempt to determine when to do so was inadequate. The boolean value in the previous code seemed to be tracking whether the plot type changed. Is it the case that we always want to refresh points and animate when any attribute change occurs?

PR #2082, opened 2025-08-12
CODAP-841 Fix to allow graph/map attribute menus to render outside component boundaries

bfinzer:
[#CODAP-841] Bug fix: Unable to access graph attribute menu

* The previous merge to main of the commenting out of `overflow: hidden` was an error in that it should have been held off to be part of this commit.
* We add a `todo` explaining that the `overflow: hidden` can be reinstated, thus producing rounded corners for components, once attribute menus are implemented to render in portals so that they are not constrained by component boundaries.


kswenson (approved):
👍 LGTM

PR #2083, opened 2025-08-13
CODAP-57 Make tile transitionComplete a volatile property

scytacki:
Additionally, transitionComplete is now set to true immediately if the tile is not animated into its final size.

This change required changing the way the map rescaling code worked. The is a new map.md file describing how it should work. The problematic case was how the code was rescaling the map to fit the points of a dataset. This was relying on transitionComplete being a serialized property. transitionComplete was used to prevent a map created with a center and zoom from being rescaled. It was also used to prevent a map restored from serialized state from being rescaled again which would override any changes made by the user.

The fix for these issues is to have the code that automatically calls rescale to make sure the map doesn't already have a set zoom before rescaling. zoom is a serialized property and after a successful rescale it will be set.

Additionally the code which automatically rescales was moved into useMapModel in a MobX `when`.

Other Fixes:
- when a map is created by the plugin api in a document with a dataset with latitude and longitude attributes it will now automatically fit this data like CODAPv2 does. That is unless the plugin sets a zoom value for the new map.
- when a dataset has a single mappable point the map would zoom in too far and not have any map tiles to show at this zoom level. This was fixed by adding a maxZoom setting of 12. The value of 12 is kind of arbitrary. At this level the user can probably still see some cities or roads that might provide visual context. But that depends on where in the world the point is located.
- when the noComponentAnimation url param is used the text tile used to not automatically get focused and also not be focused when the tile was selected.
- when the text tile was created by the api it would also not get automatically focused and also not be focused when the tile was selected.

Other minor changes:
- update usage of useTileModelContext to be consistent
- when tile is created you can no longer specify transitionComplete. This required transitionComplete to be part of the serialized properties, and was only used by the map code.

### DIFF
--- a/v3/doc/map.md
+++ b/v3/doc/map.md
@@ -1,0 +1,69 @@
+# Creation behaviors
+
+## Map created from application toolbar
+
+When a map is created by the user clicking on the map toolbar button:
+- if there is no dataset with latitude longitude attributes, the map will automatically center on the user's current location.
+- if there is a dataset with latitude longitude attributes, the new map will fit the bounds of the points of this dataset.
+
+## Map created by a plugin
+When a map is created by the plugin api:
+- the plugin can specify a center and zoom for the map, this will override any changes below
+- if there is no dataset with latitude longitude attributes, the map will automatically center on the user's current location.
+- if there is a dataset with latitude longitude attributes, the new map will fit the bounds of the points of this dataset.
+
+# Other Map behaviors
+
+If a new layer is added to the map, this should rescale the map to fit the bounds of the points of this layer plus the points any existing layers. It isn't clear if this is working properly.
+
+If a new point is added to a dataset being shown on the map, the map will not automatically rescale to show this new point.
+
+Once the map has been created the map bounds should be preserved when it is reloaded or new points are added to it.
+
+# Current implementation
+
+There are 3 places the bounds of a map are stored:
+- `MapContentModel` is the MST object which is serialized into the document to save the map state.
+- `LeafletMapState` this proxies the actual Leaflet object as a MobX object, so changes to Leaflet can be observed by components and reactions.
+- `LeafletMap` this is the actual Leaflet object provided by the Leaflet library.
+
+## MapContentModel (map-content-model.ts)
+This has reactions for syncing the center and zoom from itself to the `LeafletMapState` and from the `LeafletMapState` to itself.
+
+A `rescale` action which should compute the bounds of the all of the points of all of the datasets shown on the map, and calls `LeafletMapState.adjustMapView` to update Leaflet itself. This update of Leaflet itself triggers the listeners added by `LeafletMapState` to `LeafletMap`. The `LeafletMapState` listeners then update the `LeafletMapState` properties. The sync method in `MapContentModel` then updates its properties.
+
+A reaction listens for changes to the layers and calls `rescale` when the layers change.
+
+## LeafletMapState (leaflet-map-state.ts)
+
+In addition to providing an observable MobX proxy of the LeafletMap, this object also handles creating undo-able actions when the user uses Leaflet's UI to pan and zoom the map.
+
+Because LeafletMap sends events as the user is panning or zooming the map, this would result in many undo-able actions for single pan or zoom operation. This is handled by using Leaflet's `movestart` `moveend`, and `zoomstart` `zoomend` events. The incremental `zoom` and `move` events do update the `zoom` and `center` properties of `LeafletMapState`. But these are not sync'd to the `MapContentModel` immediately because the syncing reaction in `MapContentModel` only updates itself when `LeafletMapState.isChanging` is false. This `isChanging` view becomes true when a `zoomstart` or `movestart` happens, and it goes back to false when the corresponding `zoomend` and `moveend` happen.
+
+The `LeafletMapState` is a way to separate out the proxy of the listener based Leaflet api to an observable MobX api from the map content model. While this approach reduces the code in the map content model, it does complicate things because there are now 3 places where the map state is stored: leaflet, contentModel, and leafletMapState.
+
+In most cases the React components use the `LeafletMapState` to make changes. The plugin api uses the `MapContentModel` to make changes. When methods on the `LeafletMapState` are called to change Leaflet, it does not update the properties in `LeafletMapState` directly. Instead the methods update the LeafletMap object and then events are fired by the LeafletMap object which listeners in `LeafletMapState` handle and update the `LeafletMapState` properties. So the `LeafletMap` is the source of truth.
+
+## MapInterior (map-interior.tsx)
+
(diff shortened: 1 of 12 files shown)


v3/doc/map.md
```diff
@@ -0,0 +1,79 @@
...
+
+There are 3 places the bounds of a map are stored:
+- `MapContentModel` is the MST object which is serialized into the document to save the map state.
+- `LeafletMapState` this proxies the actual Leaflet object as a MobX object, so changes to Leaflet can be observed by components and reactions.
+- `LeafletMap` this is the actual Leaflet object provided by the Leaflet library.
+
+## MapContentModel (map-content-model.ts)
+This has reactions for syncing the center and zoom from itself to the `LeafletMapState` and from the `LeafletMapState` to itself.
+
+A `rescale` action which computes the bounds of the all of the points of all of the datasets shown on the map, and calls `LeafletMapState.adjustMapView` to update Leaflet itself. This update of Leaflet itself triggers the listeners added by `LeafletMapState` to `LeafletMap`. The `LeafletMapState` listeners then update the `LeafletMapState` properties. The sync method in `MapContentModel` then updates its properties.
```
kswenson:
```suggestion
A `rescale` action which computes the bounds of all of the points of all of the datasets shown on the map, and calls `LeafletMapState.adjustMapView` to update Leaflet itself. This update of Leaflet itself triggers the listeners added by `LeafletMapState` to `LeafletMap`. The `LeafletMapState` listeners then update the `LeafletMapState` properties. The sync method in `MapContentModel` then updates its properties.
```

v3/doc/map.md
```diff
@@ -0,0 +1,79 @@
...
+
+Additionally it includes a debouncing feature. When `LeafletMapState.adjustMapView` is called a debounced `completeMapViewChange` method created. Then the actual function runs it removes itself again. The function is called at the end of `adjustMapView`. The `LeafletMapState.isChanging` monitors the presence of this method. The effect of this is that when `adjustMapView` is first called `isChanging` becomes true. And then if `adjustMapView` is called again within the debounced time `isChanging` stays true. When there are no more `adjustMapView` calls one after another then finally `isChanging` will become false as long as the `zoomend` and `moveend` have also been called. This approach prevents lots of changes from being synced to the `MapContentModel`. When the map tile is animated into place `adjustMapView` is called many times in succession, so the debouncing above handles this case and generally the changes are not sync'd to the `MapContentModel` until the animation is done.
```
kswenson:
```suggestion
Additionally it includes a debouncing feature. When `LeafletMapState.adjustMapView` is called a debounced `completeMapViewChange` method is created. Then the actual function runs and it removes itself again. The function is called at the end of `adjustMapView`. The `LeafletMapState.isChanging` monitors the presence of this method. The effect of this is that when `adjustMapView` is first called `isChanging` becomes true. And then if `adjustMapView` is called again within the debounced time `isChanging` stays true. When there are no more `adjustMapView` calls one after another then finally `isChanging` will become false as long as the `zoomend` and `moveend` have also been called. This approach prevents lots of changes from being synced to the `MapContentModel`. When the map tile is animated into place `adjustMapView` is called many times in succession, so the debouncing above handles this case and generally the changes are not sync'd to the `MapContentModel` until the animation is done.
```

v3/doc/map.md
```diff
@@ -0,0 +1,79 @@
...
+
+This same debouncing approach is used with `zoomstart` and `movestart`. In this case the method is `completeLeafletInteraction`. This way if Leaflet triggers multiple `zoomstart`s and `movestart`s they will grouped into one change synced to the `MapContentModel`.
```
kswenson:
```suggestion
This same debouncing approach is used with `zoomstart` and `movestart`. In this case the method is `completeLeafletInteraction`. This way if Leaflet triggers multiple `zoomstart`s and `movestart`s they will be grouped into one change synced to the `MapContentModel`.
```

v3/doc/map.md
```diff
@@ -0,0 +1,79 @@
...
...ender and use the new size. On top of this, while the map tile is animating into place `LeafletMapState.adjustMapView({invalidateSize: true})` is being called each time the `CodapMap` re-renders during the animation. This is done so Leaflet re-renders in its new space. However each of these calls to `invalidateSize` causes Leaflet to fire `move` events which slightly adjust the center of the map. This eventually causes the center and zoom properties of `LeafletMapState` to be sync'd to the `MapContentModel`. Because of the debouncing described above this syncing might not actually take effect until after the animation completes but if there are are slow downs in the browser the syncing might happen during the animation. The result of all of this is that when the animation is done, sometimes the zoom of the `MapContentModel` would remain the default `-1` value and sometimes it would be set to `0`. That makes it hard to know if we should rescale the map at the end of the animation or not. It is possible to work around these nuances, but this is more complicated than the temporary leaflet approach. Additionally the temporary leaflet approach should result in a nicer looking animation.
```
kswenson, line 74:
```suggestion
Originally this problem was solved by trying to wait for the animation to finish before calling `fitBounds`. It is very difficult to make that work correctly. We know when the CSS transition ends, but the actual size of the Leaflet container isn't updated until a resize observer sees the new tile size and updates the data configuration layout, which triggers the CodapMap to re-render and use the new size. On top of this, while the map tile is animating into place `LeafletMapState.adjustMapView({invalidateSize: true})` is being called each time the `CodapMap` re-renders during the animation. This is done so Leaflet re-renders in its new space. However each of these calls to `invalidateSize` causes Leaflet to fire `move` events which slightly adjust the center of the map. This eventually causes the center and zoom properties of `LeafletMapState` to be sync'd to the `MapContentModel`. Because of the debouncing described above this syncing might not actually take effect until after the animation completes but if there are slow downs in the browser the syncing might happen during the animation. The result of all of this is that when the animation is done, sometimes the zoom of the `MapContentModel` would remain the default `-1` value and sometimes it would be set to `0`. That makes it hard to know if we should rescale the map at the end of the animation or not. It is possible to work around these nuances, but this is more complicated than the temporary leaflet approach. Additionally the temporary leaflet approach should result in a nicer looking animation.
```

v3/doc/map.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Undo Redo
+The `LeafletMapState` object stores a undoStringKey and redoStringKey. Whenever the component calls the actions on `LeafletMapState` it will set these keys if the change should be undo-able. Then when the reaction in the map content model observes a change in `LeafletMapState` it uses these keys to record the change in the content model as undo able.
```
kswenson:
```suggestion
The `LeafletMapState` object stores an undoStringKey and a redoStringKey. Whenever the component calls the actions on `LeafletMapState` it will set these keys if the change should be undo-able. Then when the reaction in the map content model observes a change in `LeafletMapState` it uses these keys to record the change in the content model as undo-able.
```

kswenson (approved):
👍 Looks great -- nice work! Just a few suggestions on the comments.

PR #2084, opened 2025-08-14
CODAP-43 Getting Story Builder to work in V3

bfinzer:
[#CODAP-43] Feature: Story Builder works in V3

Story Builder requires the data interactive api to support `get` and `update` of a `document` resource. The get passes a V2 JSON representation of the current document state while the update accepts such a representation and modifies the current document state to conform to it.

With this PR, the Story Builder plugin works to a first approximation, but there remain many problems, the most egregious of which are rehydration bugs in the graph. These problems will be listed in separate Jira stories.

* Created document-handler.ts. Register it.
* Added types for document as a data interactive type
* Implemented the 'get' so that Story Builder can request the current document state.
* Roughly following the pattern from V2 in document_helper.js `updateDocument`
* After importing all the data contexts and creating a temporary SharedModelManager.
* So that Story Builder doesn't mark the currently selected moment as changed on load, we don't notify in the component-handler.ts update if the requesting frame is the one being updated.
* Provided a calculator importer in v2-calculator-importer.ts. But it isn't actually hooked up yet
* Dealt with need for document "get" and "update" to be asynchronous
* There are many bugs that will be logged as separate stories.


v3/src/data-interactive/handlers/document-handler.ts
```diff
@@ -0,0 +1,223 @@
...
+    reinstateDatasets().then(() => {
+      reinstateComponents()
+    })
+
+    // The 500ms timeout here gives the document time to settle down.
+    // Especially any changes to center and zoom of a map
+    setTimeout(function() {
+      // DG.UndoHistory.clearUndoRedoHistory();
+      notifyDocumentChange('updateDocumentEnded')
+    }, 500)
```
kswenson: Should probably put this after `reinstateComponents()` inside the `.then()` so it's guaranteed to happen after components are reinstated.

PR #2093, opened 2025-08-21
CODAP-862 Fix for movable line causing invalid document

bfinzer:
[#CODAP-862] Bug fix: Switching bottom axis to categorical after adding a moveable line causes an invalid document

* The switch to categorical was causing the movable-line-adornment-model.ts in `updateCategories` to add specious lines. We fix by detecting that the data configuration does not have numeric attributes for both x and y axes and bailing if that is so.

Note: In CLUE a patch was added through a `preProcessSnapshot` for its `MovableLineInstance` to take care of possible specious lines. We do not follow this path for CODAP because the specious lines should not appear.


scytacki (approved):
Looks good. 👍

PR #2095, opened 2025-08-22
CODAP-837 Splitting a histogram wasn't resulting in binned dot plot as it should

bfinzer:
[#CODAP-837] Bug fix: Transition from a histogram to split binned plot fails

* `GraphContentModel.syncPlotWithAttributeConfiguration` wasn't catching the case in which there was a histogram and the user was attempting to split it with a categorical attribute on the secondary axis.


kswenson (approved):
👍 LGTM

PR #2096, opened 2025-08-22
CODAP=845 Displaying small numbers in equations

bfinzer:
[#CODAP-845] Bug fix: Nearly flat least squares lines slope display as 0

* The problem lay in graph-utils.ts `formatEquationValue` which was not dealing correctly with very small numbers, rounding them to zero instead of expressing them in scientific notation with the desired precision.
  * It was surprisingly difficult to get this right. In the end ChatGPT5 was able to come up with a general solution, but it had to think deeply to do so.

### DIFF
--- a/v3/src/components/graph/utilities/graph-utils.ts
+++ b/v3/src/components/graph/utilities/graph-utils.ts
@@ -225,11 +225,78 @@ export function findNeededFractionDigits(slope: number, intercept: number, layou
   return { slopeDigits, interceptDigits }
 }
 
-function formatEquationValue(value: number, digits: number, units = "", parenthesizeUnits = false) {
-  const exponent = `e${digits}`
-  const roundedValue = Math.round(parseFloat(`${value}${exponent}`))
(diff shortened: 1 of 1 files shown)


v3/src/components/graph/utilities/graph-utils.ts
```diff
@@ -225,11 +225,77 @@ export function findNeededFractionDigits(slope: number, intercept: number, layou
...
+ * Requirements:
+ * - see formatEquationValue below for requirements
+ *
+ * @param value - The numeric value to format
+ * @param digits - Number of decimal places to use for rounding
+ * @returns Formatted string representation of the value
+ */
+export function formatValue(value: number, digits: number): string {
+  if (Number.isNaN(value)) return "NaN"
+  if (!Number.isFinite(value)) return value > 0 ? "Infinity" : `${kMinus}Infinity`
```
kswenson: Should we use `♾️` rather than `Infinity`? We generally avoid `NaN` elsewhere. Is there something else we should use?

PR #2097, opened 2025-08-22
CODAP-850 Edit text should notify

bfinzer:
[#CODAP-950] Bug fix: Editing text in a text component should notify plugins

* The notification was missing in text-tile.tsx `handleBlur`.

### DIFF
--- a/v3/src/components/text/text-tile.tsx
+++ b/v3/src/components/text/text-tile.tsx
@@ -83,7 +83,7 @@ export const TextTile = observer(function TextTile({ tile }: ITileBaseProps) {
         log: textDidChange ? () => `Edited text component: ${textModel.textContent}` : undefined,
         undoStringKey: "DG.Undo.textComponent.edit",
         redoStringKey: "DG.Redo.textComponent.edit",
-        notify: updateTileNotification("edit text", {}, tile)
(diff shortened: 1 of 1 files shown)


v3/src/components/text/text-tile.tsx
```diff
@@ -81,7 +82,8 @@ export const TextTile = observer(function TextTile({ tile }: ITileBaseProps) {
         // For now, we log just the text content, not the full JSON-stringified slate value.
         log: textDidChange ? () => `Edited text component: ${textModel.textContent}` : undefined,
         undoStringKey: "DG.Undo.textComponent.edit",
-        redoStringKey: "DG.Redo.textComponent.edit"
+        redoStringKey: "DG.Redo.textComponent.edit",
+        notify: updateTileNotification("edit text", {}, tile)
```
kswenson: Seems like this should follow the logging example of only notifying when the text actually changed:
```suggestion
        notify: textDidChange ? () => updateTileNotification("edit text", {}, tile) : undefined
```
Also, is this compatible with the v2 notification, i.e. without any content in the second argument?

kswenson (changes requested):
👍 looks good, but see suggestions below

kswenson (approved):
👍 LGTM

PR #2099, opened 2025-08-26
CODAP-850 continued —Text tile notification

bfinzer:
Refinement of handleBlur for text tile

* The "fix" for CODAP-850 got merged but did not get approved. A test for whether the text had been changed was added in the final commit bug but didn't work and is not necessary since the blur only occurs if there has been a change. The `log` function also would not have been called.
  * Note that a change of documented behavior is that notification and logging will occur when the style changes, not just when the text changes.

### DIFF
--- a/v3/src/components/text/text-tile.tsx
+++ b/v3/src/components/text/text-tile.tsx
@@ -73,7 +73,7 @@ export const TextTile = observer(function TextTile({ tile }: ITileBaseProps) {
   function handleBlur() {
     // update the model on blur, not on every change (e.g. keystroke)
     if (textModel && !textModel.isEquivalent(editor.children)) {
-      // handleBlur only gets called if the text or style has changed
+      // We only get here if the text or style has changed
       textModel?.applyModelChange(() => {
         textModel.setValueFromEditor(editor.children)
       }, {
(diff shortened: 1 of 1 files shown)


kswenson (commented):
>A test for whether the text had been changed was added in the final commit bug but didn't work and is not necessary since the blur only occurs if there has been a change.

This is not true -- `handleBlur` is called whenever the text editor loses focus, regardless of whether there were any changes or not. What _is_ true is that the `textModel.isEquivalent(editor.children)` test limits the response to circumstances in which _something_ has changed, but the purpose of the `textDidChange` test was to prevent logging and notifying when the text didn't actually change. Can you say more about what didn't work about it and/or are you suggesting that we _should_ notify and log when style changes occur or merely that it's acceptable to notify and log when styles change because it's too hard to determine whether the text has changed?

bfinzer:
@kswenson Yes, sorry, I misspoke. `handleBlur` is called whether or not anything has changed. I was ignoring the test for whether something has changed. What didn't work was the computation of `textDidChange` as in `const textDidChange = textOnFocus.current !== textModel.textContent`. This was always resulting in false and so neither logging or notification would occur. I'm suggesting that it's *acceptable* to notify and log when styles change *and* I didn't want to take the time to figure out whether the text had changed.
I'm changing the comment and pushing that commit. I'll re-request a review.

PR #2103, opened 2025-08-27
CODAP-849 Graph rehydration bugs

bfinzer:
[#CODAP-849] Bug fix: Graph rehydration bugs evidenced in Story Builder

This PR has significant problems having to do with restoring datasets. But it works pretty well for rehydrating graphs.

* The main problem was that applySnapshot was having to recreate the graph's layer model because it was an element of an array and the snapshot did not contain the model's ID. Solution: Models that have arrays of models where those model elements have volatile state that needs to be preserved now register a postImportSnapshotProcessor that will be called for existing tiles. The job of such a processor is to assign IDs to the model elements where the IDs are pulled from the existing models.
* The above also applies to maps. So maps and graphs now register post-import-snapshot-processor functions that carry out the assigning of IDs where appropriate.
* `v2MapExporter` was creating a storage object with center and zoom as unserializable proxies. Fixed this using `toJS`.


kswenson (approved):
👍 LGTM -- I added a commit which tweaked a few things.

PR #2104, opened 2025-08-28
CODAP-874 Fix dynamic response of LSRL

bfinzer:
[#CODAP-874] Bug fix: LSRL isn't updating during drag of points

* Probably during some performance optimization back in early May, response to point dragging broke for the LSRL. The fix is to introduce a volatile `changeCount` in the `LSRLAdornmentModel` which gets incremented when it recomputes. This gives the lsrl adornment component something to react to.
* We also take this opportunity to fix a lint error in dot-line-plot.tsx


kswenson (approved):
👍 LGTM

PR #2109, opened 2025-08-29
CODAP-839 Fix parent visibility toggles for case plot

bfinzer:
[#CODAP-839] Bug fix: Parent visibility toggles not consistently toggling

* This problem was only occurring when there were no attributes assigned to the graph and the cause was an assumption in `Dataset.getCollectionIndexForAttribute` that if no attribute id were passed in, the result should be undefined. We alter that to return the index of the childmost collection instead.


kswenson (approved):
👍 LGTM

PR #2110, opened 2025-09-03
feat: tool palette aligns with top of the component (CODAP-846)

emcelroy:
[CODAP-846](https://concord-consortium.atlassian.net/browse/CODAP-846)

This change will align the top of components' tool palette panels with the top of the components' title bars.

I'm a little on the fence about whether adding a variable is justified, but I think it makes sense and may save some trouble if the border width changes.

[CODAP-846]: https://concord-consortium.atlassian.net/browse/CODAP-846?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/v3/src/components/inspector-panel.scss
+++ b/v3/src/components/inspector-panel.scss
@@ -9,7 +9,7 @@ $inspector-panel-border-width: 2px;
   align-items: center;
   background-color: white;
   border: $inspector-panel-border-width solid vars.$focus-border-color;
-  border-radius: 0 5px 5px 5px;
+  border-radius: 5px;
   box-shadow: rgba(0, 0, 0, 0.25) 0px 1px 3px 0px;
   display: flex;
   flex-direction: column;
@@ -66,6 +66,7 @@ $inspector-panel-border-width: 2px;
     }
 
     &.top {
+      border-top-left-radius: 4px;
       border-top-right-radius: 4px;
       height: 56px;
       padding-top: 8px;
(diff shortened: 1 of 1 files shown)


v3/src/components/inspector-panel.scss
```diff
@@ -3,18 +3,19 @@ $inspector-panel-width-very-narrow: 48px;
...
 $inspector-panel-width-medium: 72px;
 $inspector-panel-width-wide: 80px;
+$inspector-panel-border-width: 2px;
 
 .inspector-panel {
   align-items: center;
   background-color: white;
-  border: 2px solid vars.$focus-border-color;
+  border: $inspector-panel-border-width solid vars.$focus-border-color;
   border-radius: 0 5px 5px 5px;
```
tealefristoe:
```suggestion
  border: $inspector-panel-border-width solid vars.$focus-border-color;
  border-radius: 5px;
```

tealefristoe (approved):
Looks good 👍 

I think adding a variable makes sense. One thing I noticed is that the top left corner of the inspector panel should now be rounded like the other three corners.

v3/src/components/inspector-panel.scss
emcelroy: @tealefristoe Thanks. Is there a spec for the new CODAP design that shows that? If so, can you provide a link?

I was going by the image in the Jira story, which doesn't have a rounded top left corner. (I definitely think it'd look better with one, though, FWIW.)
image

emcelroy: @tealefristoe Sorry, I found the spec without much effort. You're right -- the top left corner should be rounded. I've made that change in the latest commit.

PR #2112, opened 2025-09-04
fix: Connecting lines don't appear in each subplot (CODAP-868)

emcelroy:
[CODAP-868](https://concord-consortium.atlassian.net/browse/CODAP-868)

These changes fix a bug where connecting lines only appeared in the last subplot of a subdivided plot.

Previously, when `ScatterPlot` called `renderConnectingLines` multiple times for multiple subplots, each call cleared the previous subplot's lines because `prepareConnectingLines` ran:

```
connectingLinesArea.selectAll("path").remove()
```

This change moves that removal of lines to `refreshConnectingLines` (in both `ScatterPlot` and `MapPointLayer`), just before `renderConnectingLines` is called. That way, duplicates are still prevented (during tile resizing, for example) without removing lines from other subplots.

[CODAP-868]: https://concord-consortium.atlassian.net/browse/CODAP-868?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/v3/cypress/e2e/bivariate-adornments.spec.ts
+++ b/v3/cypress/e2e/bivariate-adornments.spec.ts
@@ -261,6 +261,7 @@ context("Graph adornments", () => {
     c.selectTile("graph", 0)
     cy.dragAttributeToTarget("table", "Sleep", "bottom")
     cy.dragAttributeToTarget("table", "Speed", "left")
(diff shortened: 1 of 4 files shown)


kswenson (approved):
👍 LGTM

PR #2113, opened 2025-09-04
Codap-776 Case Card Styling

tealefristoe:
Jira story: https://concord-consortium.atlassian.net/browse/CODAP-776

This PR makes case cards respond better to width changes. There are also several other minor stylistic tweaks to improve the case card.

I also removed a bunch of commented out code. After a year, it seemed like we wouldn't need it any more.


emcelroy (approved):
Looks good 👍

Nice to see all that dead code removed.

PR #2114, opened 2025-09-04
fix: Leaving URL blank in "Web Page" component crashes CODAP (CODAP-870)

emcelroy:
[CODAP-870](https://concord-consortium.atlassian.net/browse/CODAP-870)

These changes fix a bug where if a user clicked the Web Page URL modal's OK button without entering a URL, CODAP would crash. Now the URL modal's OK button is disabled until a value is entered.

This could probably be improved by adding a check for a valid URL value, but I think that's outside the scope of the story. (Note that the [CODAP-870](https://concord-consortium.atlassian.net/browse/CODAP-870) description has already been updated to include more functionality than it originally did, so I'm wary of adding more features.)

Additionally, if no URL value is entered and the user clicks the Cancel button, the Web Page component is automatically removed.

[CODAP-870]: https://concord-consortium.atlassian.net/browse/CODAP-870?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/v3/src/components/web-view/web-view-inspector.tsx
+++ b/v3/src/components/web-view/web-view-inspector.tsx
@@ -35,10 +35,6 @@ export const WebViewInspector = observer(function WebViewInspector({tile, show}:
     if (tile) {
       documentContent?.applyModelChange(() => {
         documentContent?.deleteTile(tile.id)
-      }, {
-        undoStringKey: "V3.Undo.webView.remove",
-        redoStringKey: "V3.Redo.webView.remove",
-        log: logMessageWithReplacement("Remove empty web view: %@", {tileId: tile.id})
       })
     }
   }
--- a/v3/src/utilities/translation/lang/en-US.json5
+++ b/v3/src/utilities/translation/lang/en-US.json5
@@ -131,8 +131,6 @@
     "V3.Redo.webView.show": "Redo adding web page component",
     "V3.Undo.webView.changeUrl": "Undo changing web page url",
     "V3.Redo.webView.changeUrl": "Redo changing web page url",
-    "V3.Undo.webView.remove": "Undo remove web page component",
-    "V3.Redo.webView.remove": "Redo remove web page component",
 
     // V3 plugin strings that are not present in V2 and will eventually require translation.
     "V3.Undo.plugin.create": "Undo adding %@ plugin",


v3/cypress/e2e/plugin.spec.ts
```diff
@@ -496,4 +496,20 @@ context("codap plugins", () => {
...
+    webView.getUrlModalOkButton().should("be.enabled")
+  })
+
+  it("removes empty web view when URL modal is cancelled", () => {
+    c.clickIconFromToolShelf("web page")
+    webView.getUrlModal().should("be.visible")
+    webView.cancelUrlModal()
+    webView.getUrlModal().should("not.exist")
+    webView.verifyWebViewRemoved()
+  })
```
emcelroy: Would it make sense to move these to a new, separate web-view.spec.ts file?

tealefristoe: You could move these into `tool-shelf.spec.ts`, where there's already one test about adding a web view and specifying a url or two for it.

v3/src/components/web-view/web-view-inspector.tsx
```diff
@@ -31,6 +31,18 @@ export const WebViewInspector = observer(function WebViewInspector({tile, show}:
...
   }
 
+  const handleRemoveEmptyWebView = () => {
+    if (tile) {
+      documentContent?.applyModelChange(() => {
+        documentContent?.deleteTile(tile.id)
+      }, {
+        undoStringKey: "V3.Undo.webView.remove",
+        redoStringKey: "V3.Redo.webView.remove",
+        log: logMessageWithReplacement("Remove empty web view: %@", {tileId: tile.id})
```
tealefristoe: I don't think you want this to be undoable. In fact, what you really want is for the original tile creation action to not be undoable if no url is specified. This might be tricky to do, but I think it's possible. Make the original tile creation action not undoable, then make setting the url undoable.

emcelroy: As discussed elsewhere, since making this work correctly isn't straightforward, we've created a [new story for addressing that](https://concord-consortium.atlassian.net/browse/CODAP-895).

PR #2121, opened 2025-09-08
Codap-879 Dynamic Toolbar Position Menu Item

tealefristoe:

This PR makes use of the latest version of `cloud-file-manager` to update the wording of the toolbar position menu item based on the current selection.


v3/src/utilities/translation/lang/en-US.json5
```diff
@@ -218,7 +218,7 @@
     "V3.Undo.errorTester.create": "Undo adding error tester",
     "V3.Redo.errorTester.create": "Redo adding error tester",
 
-    "V3.AppController.optionMenuItems.positionToolShelf": "Toggle Toolbar Position",
+    "V3.AppController.optionMenuItems.positionToolShelf": "Toolbar Position:",
```
kswenson: Should use either variable replacement or two separate strings here rather than string concatenation because the order may be different in other languages. In this case I think two separate strings is simpler:
```suggestion
    "V3.AppController.optionMenuItems.toolbarPositionLeft": "Toolbar Position: Left",
    "V3.AppController.optionMenuItems.toolbarPositionTop": "Toolbar Position: Top",
```

v3/src/lib/cfm/use-cloud-file-manager.ts
```diff
@@ -182,6 +182,79 @@ function getHelpUrl() {
...
+        ]
+      },
+      {
+        className: "settings-menu",
+        menuAnchorIcon: SettingsIcon,
+        menuAnchorName: t("DG.ToolButtonData.optionMenu.title"),
+        menu: [
+          {
+            icon: ToolbarPositionIcon,
+            name: `${t("V3.AppController.optionMenuItems.positionToolShelf")} ${persistentState.toolbarPosition}`,
```
kswenson: Should use either variable replacement or two separate strings here rather than string concatenation because the order may be different in other languages. In this case I think two separate strings is simpler:
```suggestion
            name: t(`V3.AppController.optionMenuItems.toolbarPosition${persistentState.toolbarPosition}`,
```

kswenson (approved):
👍 Looks good -- just one suggested change for better localizability of the strings.

PR #2124, opened 2025-09-08
fix: import of computed bar chart

kswenson:
[[CODAP-610](https://concord-consortium.atlassian.net/browse/CODAP-610)] Bar chart with computed bar heights isn't importing from V2 properly

[CODAP-610]: https://concord-consortium.atlassian.net/browse/CODAP-610?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/v3/src/models/app-state.ts
+++ b/v3/src/models/app-state.ts
@@ -8,7 +8,7 @@
  */
 import { CloudFileManager } from "@concord-consortium/cloud-file-manager"
 import { action, autorun, computed, makeObservable, observable, reaction, flow } from "mobx"
-import { getSnapshot } from "mobx-state-tree"
+import { destroy, getSnapshot } from "mobx-state-tree"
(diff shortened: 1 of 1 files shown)


kswenson:
From @bfinzer via slack:

>Hi. After I approved PR#2124 I noticed in the console the following:
```
[mobx] Encountered an uncaught exception that was thrown by a reaction or observer component, in: 'Reaction[FormulaManager.registerActiveFormulas.reaction]'
```
>Probably you should take a look at that before merging.

It turns out that the error was being thrown by a temporary document that is made on the way to making the real document. Simply disposing of the temporary document once we're done with it is sufficient to keep it out of trouble.

PR #2126, opened 2025-09-08
Codap-776 Make case card columns equal widths

tealefristoe:
Jira story: https://concord-consortium.atlassian.net/browse/CODAP-776

This PR makes the case card name and value columns equal widths, matching v2 behavior.


emcelroy (approved):
Looks good 👍

PR #2127, opened 2025-09-08
CODAP-835 Implement keyboard shortcuts for undo & redo

bfinzer:
[#CODAP-835] Feature: Undo / redo keyboard shortcuts (Command/Control Z and Y)

* The npm `tinyKeys` is well supported and work easily with React, so we're using that.
* Created a hook `useKeyboardShortcuts` to encapsulate the keyboard shortcut functionality in a way that can be easily extended 
* Also removed the tool tips giving keyboard shortcuts for other toolshelf tools.


v3/src/hooks/use-keyboard-shortcuts.ts
```diff
@@ -0,0 +1,39 @@
...
+    const isEditable = (el: Element | null) => {
+      if (!el) return false
+      if (el instanceof HTMLInputElement ||
+        el instanceof HTMLTextAreaElement ||
+        (el instanceof HTMLElement && el.isContentEditable)) {
+        return true
+      }
+    }
+
+    const okToUndoRedo = (e: KeyboardEvent, test:boolean) => {
```
kswenson:
```suggestion
    const okToUndoRedo = (e: KeyboardEvent, test: boolean) => {
```
Also, this code would benefit from a comment indicating why editability matters.

v3/src/hooks/use-keyboard-shortcuts.ts
```diff
@@ -0,0 +1,39 @@
...
+      const target = e.target as Element | null
+      if (isEditable(target)) return false
+      e.preventDefault()
```
kswenson: 1. You probably want to `stopPropagation()` as well as `preventDefault()`, at least if we're handling the event.
2. I'm wondering whether we should only `preventDefault()` and `stopPropagation()` if `test` is `true`. 🤔  The argument _for_ doing so is that if we're not going to handle the event perhaps we should let the browser (or someone else) have a crack at it. On the other hand, we're not using a capturing handler and we're attached to the window, so we're getting the event fairly late in the process and maybe we don't want anyone behind us to make use of one of our keyboard shortcuts anyway. 🤷‍♂️

kswenson (approved):
👍 Looks good -- just a couple of thoughts/suggestions.

PR #2130, opened 2025-09-09
fix: Long component title overflow (CODAP-873)

emcelroy:
[CODAP-873](https://concord-consortium.atlassian.net/browse/CODAP-873)

These changes prevent a bug that allowed tile title text and the tile title text input field to extend beyond the boundaries of a tile's title bar. ~~Included are some adjustments to styling that follow the [new spec](https://app.zeplin.io/project/5e4baae7fb685faac9bf4a0a/screen/6889160d875b3fae20982a54) except that we use a narrower minimum tile width.~~ Now when the title text is longer than the space available in the title bar, the text gets truncated.

As per discussion elsewhere, these changes also drop the minimum tile width value to 50 to match CODAP V2. This means that like in V2, at very narrow tile widths the title bar buttons and title text will overlap. That is understood and accepted by the project team.

There are some additional small changes related to issues uncovered in this work:

- Prevent tile dragging during title edit (in `ComponentTitleBar`).
- Prevent triggering of title edit mode after dragging tile by title text (in `ComponentTitleBar`).
- Enforce minimum tile width value on tile resize (in `FreeTileComponent`).
- Consolidate two `kDefaultMinWidth` variables. There was one in `v3/src/models/tiles/tile-layout.ts` and another in `v3/src/models/tiles/tile-model.ts`. Both were set to 60. 

[CODAP-873]: https://concord-consortium.atlassian.net/browse/CODAP-873?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/v3/src/components/component-title-bar.tsx
+++ b/v3/src/components/component-title-bar.tsx
@@ -30,6 +30,7 @@ export const ComponentTitleBar = observer(function ComponentTitleBar(props: ITil
   const [isHovering, setIsHovering] = useState(false)
   const blankTitle = "_____"
   const hasDraggedRef = useRef(false)
+  const pointerStart = useRef<{x: number, y: number} | null>(null);
 
   // Input sizing is based on https://stackoverflow.com/questions/8100770/auto-scaling-inputtype-text-to-width-of-value
   const [inputWidth, setInputWidth] = useState(2 * kInputPadding)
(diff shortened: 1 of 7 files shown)


kswenson (commented):
My first test was opening the Mammals example document in both v2 and v3:

v2:
Image

v3 (main):
Image

v3 (this PR):
Image

Minimum tile width aside, I think the goal should be to show as much of the title as possible and only elide when necessary. I'll hold off on a detailed code review for now.

v3/src/components/component-title-bar.tsx
```diff
@@ -74,14 +75,26 @@ export const ComponentTitleBar = observer(function ComponentTitleBar(props: ITil
     setInputWidth(measureRef.current.offsetWidth + 2 * kInputPadding)
   }
 
+  const handleTitlePointerDown = () => {
+    hasDraggedRef.current = false
+  }
+
+  const handleTitlePointerMove = () => {
+    hasDraggedRef.current = true
```
kswenson: In situations like this a threshold of ~3 pixels is often used so as not to be confused by incidental movements of a pixel or two that can occur when clicking. Copilot suggests the following:
```typescript
const pointerStart = useRef<{x: number, y: number} | null>(null);

const handleTitlePointerDown = (e: React.PointerEvent) => {
  hasDraggedRef.current = false;
  pointerStart.current = { x: e.clientX, y: e.clientY };
};

const handleTitlePointerMove = (e: React.PointerEvent) => {
  if (pointerStart.current) {
    const dx = Math.abs(e.clientX - pointerStart.current.x);
    const dy = Math.abs(e.clientY - pointerStart.current.y);
    if (dx > 3 || dy > 3) { // 3px threshold
      hasDraggedRef.current = true;
    }
  }
};
```

kswenson (approved):
👍 Looks good -- just one suggestion to use a threshold for the drag/click test.

v3/src/components/component-title-bar.tsx
emcelroy: @kswenson Ah, makes sense. Thanks 👍

PR #2131, opened 2025-09-09
fix: formula editor button styling

kswenson:
[[CODAP-823](https://concord-consortium.atlassian.net/browse/CODAP-823)] The formula editor buttons for Insert Value and Insert Function have lost their formatting

Note that there is a larger re-styling discussion happening around the formula editor. This is not that. I took a few of the colors from the in-progress spec, but the goal here is just to get the buttons to look and act like buttons.


tealefristoe (approved):
Looks good 👍

PR #2133, opened 2025-09-09
Codap-879 Switch Toolbar Position Labels

tealefristoe:

This PR switches the labels of the toolbar position menu item, so the label indicates what will happen if it's clicked, rather than the current state of things.


kswenson (approved):
👍 LGTM

PR #2134, opened 2025-09-09
CODAP-898 Box plot and hover tips placement 

bfinzer:
[#CODAP-898] Bug fix: Box plot placement not correct when graph is split on both top and right

* The cited bug is fixed by noting that the plotWidth and plotHeight passed in refer to that of the subplot which was then being divided by the number of columns/rows. Instead we get the plotWidth/Height from the graph layout.
* A second problem that manifests when the box plot tips are close to the element is that pointer events are being intercepted by the tip and thus we get a flickering of the tip. We fix this by specifying pointer-events as "none" for the box plot tips. And this seems to require "!important" because of class nesting with style that specifies pointer-events as "all".

### DIFF
--- a/v3/src/components/graph/adornments/univariate-measures/box-plot/box-plot-adornment-component.scss
+++ b/v3/src/components/graph/adornments/univariate-measures/box-plot/box-plot-adornment-component.scss
@@ -75,10 +75,11 @@
   height: 100%;
   top: 0;
   width: 100%;
+}
 
-  &.box-plot-label {
-    background: #ffc;
-    box-shadow: 1px 1px 1px rgba(51, 51, 51, .7);
-    padding: 2px 4px;
-  }
+.box-plot-label {
+  background: #ffc;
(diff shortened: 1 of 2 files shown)


v3/src/components/graph/adornments/univariate-measures/box-plot/box-plot-adornment-component.scss
```diff
@@ -76,9 +76,10 @@
   top: 0;
   width: 100%;
 
-  &.box-plot-label {
+  .box-plot-label {
     background: #ffc;
     box-shadow: 1px 1px 1px rgba(51, 51, 51, .7);
     padding: 2px 4px;
+    pointer-events: none;
   }
```
kswenson, line 84: I added a commit which puts this back inside the `.measure-labels` block (without the ampersand) which allowed me to remove the `!important`.

kswenson (approved):
👍 LGTM after adding a commit which removed the `!important`.

PR #2135, opened 2025-09-09
Codap-840 Restrict Dropping Images

tealefristoe:
Jira story: https://concord-consortium.atlassian.net/browse/CODAP-840

This PR prevents Codap from opening web views when image files are dropped onto it.


kswenson:
Tabling this for now.

PR #2137, opened 2025-09-10
Codap-879 Different Menu Item Icons

tealefristoe:

Just when you thought this story was done...

This PR makes the icon for Toolbar Position: Top and Left different. The difference is a matter of applying a simple rotation, but since I wasn't sure how to do that with an icon being rendered by the `cfm`, I just duplicated the icon file and applied the rotation directly.


kswenson (approved):
👍 LGTM

PR #2140, opened 2025-09-11
CODAP-844 Add notification for tile minimization

bfinzer:
[#CODAP-844] Bug fix: Minimize and close buttons should notify plugins

* The close button was already notifying.
* Added applyModelChange to free-tile-component.tsx `handleMinimizeTile` with full suite of options

### DIFF
--- a/v3/src/components/container/free-tile-component.tsx
+++ b/v3/src/components/container/free-tile-component.tsx
@@ -45,12 +45,13 @@ export const FreeTileComponent = observer(function FreeTileComponent({ row, tile
   }, [])
 
   const handleMinimizeTile = useCallback(() => {
+    const logString = isMinimized ? "Expanded component" : "Minimized component"
     if (setMinimized) {
       tile.applyModelChange(() => {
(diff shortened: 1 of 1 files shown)


v3/src/components/container/free-tile-component.tsx
```diff
@@ -45,8 +45,18 @@ export const FreeTileComponent = observer(function FreeTileComponent({ row, tile
...
 
   const handleMinimizeTile = useCallback(() => {
-    setMinimized?.(!isMinimized)
-  }, [isMinimized, setMinimized])
+    if (setMinimized) {
+      tile.applyModelChange(() => {
+        setMinimized(!isMinimized)
+      }, {
+        notify: updateTileNotification("toggle minimize component", {}, tile),
+        log: logMessageWithReplacement("Component minimized: %@", {isMinimized}, "component"),
```
kswenson: Did you check the content of this log message? It seems like it would log the current state rather than the new state.

PR #2141, opened 2025-09-11
Codap-779 Offset Tooltips

tealefristoe:
Jira story: https://concord-consortium.atlassian.net/browse/CODAP-779

This PR makes tooltips for dropping attributes follow the cursor, even when the user has scrolled during the drag.


v3/src/hooks/use-scrollable-container.ts
```diff
@@ -0,0 +1,26 @@
...
+  })
+}
+
+// Returns the difference between the current scroll and the scroll when the drag started
+export function getScrollDifference() {
+  return {
+    left: (scrollableContainer?.scrollLeft ?? 0) - initialScroll.left,
+    top: (scrollableContainer?.scrollTop ?? 0) - initialScroll.top
+  }
+}
```
kswenson: It seems like you could avoid the globals with something like:

```typescript
export function useContainerDragScrollOffset() {
  const containerRef = useTileContainerContext()
  const initialScrollRef = useRef<{ left: number, top: number }>({ left: 0, top: 0 })

  useDndMonitor({
    onDragStart() {
      if (!containerRef.current) return

      initialScrollRef.current.left = element.scrollLeft ?? 0
      initialScrollRef.current.top = element.scrollTop ?? 0
    }
  })

  return {
    left: (containerRef.current?.scrollLeft ?? 0) - initialScrollRef.current.left,
    top: (containerRef.current?.scrollTop ?? 0) - initialScrollRef.current.top
  }
}
```

tealefristoe: Thanks, that's a lot cleaner 👍

kswenson (approved):
👍 LGTM

PR #2142, opened 2025-09-11
CODAP-865 Group into bins not undoable

bfinzer:
[#CODAP-865] Bug fix: "Group into bins" is not undoable

* Added the undo/redo strings to `handleDisplayTypeChange`
* There were 5 missing notifications in display-config-palette.tsx
* Also missing was a notification for changing a formula for a computed bar chart. V2 didn't have this either. The place to add it is in bar-chart.tsx `handleEditExpressionClose`, but I couldn't figure out how to get a numeric id.

### DIFF
--- a/v3/src/components/axis/helper-models/axis-helper.test.ts
+++ b/v3/src/components/axis/helper-models/axis-helper.test.ts
@@ -1,10 +1,10 @@
 // Tests for the AxisHelper class
 import { select } from "d3"
-import { AxisHelper, IAxisHelperArgs } from "./axis-helper"
-import { IAxisModel } from "../models/axis-model"
 import { IDataDisplayContentModel } from "../../data-display/models/data-display-content-model"
 import { IAxisLayout } from "../models/axis-layout-context"
-import { IAxisProvider } from "../hooks/use-axis-provider-context"
+import { IAxisModel } from "../models/axis-model"
+import { IAxisProvider } from "../models/axis-provider"
+import { AxisHelper, IAxisHelperArgs } from "./axis-helper"
 
 jest.mock("d3", () => ({
   select: jest.fn().mockReturnValue({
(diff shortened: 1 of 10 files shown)


v3/src/components/graph/plots/bar-chart/bar-chart.tsx
```diff
@@ -204,14 +204,23 @@ export const BarChart = observer(function BarChart({ abovePointsGroupRef, pixiPo
...
     handleCloseModal()
     const expression = barChartModel.formula?.display ?? ""
+    const notification = {
+      message: {
+        action: 'notify',
+        resource: "component",
+        // todo: id is supposed to be numeric, but graphModel.id is a string
+        values: { type: "graph", operation: "change formula", id: 0 }
+      }
+      }
```
kswenson: I think what you want is something like:
```suggestion
    const notification = tileNotification("change formula", {}, tile)
```
The `tile` can be retrieved with `useTileModelContext()`.

bfinzer:
@kswenson I've run into a puzzle. (First you'll notice I've made some additional changes to `display-config-palette.tsx`.`handleDisplayTypeChange`.)
When I go from binned dot plot to **Bar for Each Point** and then **Undo** that takes us through a reaction on line 273 of graph-content-model.ts which invokes an action on `GraphContentModel`. I believe that as a result the **Redo** button is disabled and the **Undo** button is enabled with the hover tip "Undo Last Action". I'm not sure how to fix this, especially since the offending action is invoked by a reaction installed in the `GraphContentModel`'s `afterCreate`.

kswenson (approved):
👍 LGTM -- I added a commit which addresses the undo/redo issue raised in [this comment](https://github.com/concord-consortium/codap/pull/2142#issuecomment-3288892655).

PR #2146, opened 2025-09-16
feat: Tiles menu item hover brings component forward (CODAP-894)

emcelroy:
[CODAP-894](https://concord-consortium.atlassian.net/browse/CODAP-894)

Update `handleFocus` in `TilesListShelfButton` to set focus on associated tile and scroll tile into view if its not fully visible in the browser window. Adds a new `scrollTileIntoView` utility function for the latter behavior.

[CODAP-894]: https://concord-consortium.atlassian.net/browse/CODAP-894?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


v3/src/utilities/dom-utils.ts
```diff
@@ -0,0 +1,16 @@
...
+    if (!el) return
+
+    const rect = el.getBoundingClientRect()
+    const vw = window.innerWidth
+    const vh = window.innerHeight
+    const fullyVisible = rect.top >= 0 && rect.left >= 0 && rect.bottom <= vh && rect.right <= vw
+
+    if (!fullyVisible) {
+      el.scrollIntoView({ block: "center", inline: "center", behavior: "smooth" })
+    }
```
kswenson: I believe this scroll is too simplistic. The scroll should be the minimum required to bring the tile into view, so if the tile is off the bottom then the bottom of the tile should be aligned with the bottom of the viewport, etc. You can see this in the way v2 behaves. I think there are also some edge cases to consider. For instance, if the document is larger than the tile such that only part of the tile is shown, then not scroll should occur even though the tile is not fully visible.

kswenson (changes requested):
👍 Generally looks good, but I think the scroll algorithm needs some additional work.

v3/src/utilities/dom-utils.test.ts
```diff
@@ -0,0 +1,121 @@
+import { scrollTileIntoView } from "./dom-utils"
+
+describe("scrollTileIntoView", () => {
+  let getElementByIdSpy: jest.SpyInstance
+  let requestAnimationFrameSpy: jest.SpyInstance
+  let scrollToSpy: jest.SpyInstance
+  let originalInnerWidth: number
+  let originalInnerHeight: number
+
+
```
kswenson:
```suggestion
```

kswenson (approved):
👍 Looks good -- while testing I came across one situation where it behaved in a way I didn't expect, but I can't remember what it was so let's just merge this and if we encounter any oddities we can fix them down the road.

PR #2149, opened 2025-09-18
CODAP-915 Fix crash on marquee select in maps

bfinzer:
[#CODAP-915] Bug fix: Crash—Marquee select tool in a map produces a display error on selection

* The cause of the crash was that the `Background` component, in order to deal with background images, needed to `useGraphContentModelContext`, but this is not possible for `Map` components. The solution is to move `plotBackgroundImage` and `plotBackgroundImageLockInfo` from `GraphContentModel` `DataDisplayContentModel` following a prior pattern for three other properties related to graph backgrounds.
* The change required a skanky type coercion to `GraphContentModel`in an `mstReaction` in background.tsx that requires access to axes. The type coercion could be eliminated by an equally skanky inclusion of `getAxis` in `DataDisplayContentModel`.


kswenson (approved):
👍 LGTM

PR #2150, opened 2025-09-18
CODAP-914 Don't show html formatting in hover tips

bfinzer:
[#CODAP-914] Bug fix: Hover labels for univariate measures like mean are including html tags

* The problem was caused by adding formatting to the units in measure labels. Unfortunately, univariate measure hover tips were using the same string but, since they're rendered with svg, we can't do the formatting. The fix is to distinguish between the string used for labels and the string used for hover tips.


kswenson (approved):
👍 LGTM

PR #2151, opened 2025-09-23
CODAP-900 use full document applySnapshot in di doc update

scytacki:
Use a full document applySnapshot to update the current document when data interactive document handler update API is used. This approach fixes several problems with the story builder plugin in v3.

This approach is tricky for a few few reasons:
- document state is round tripped through v2 format so several MST ids are lost
- because models and components are reused, volatile state in the models and components has to be synchronized after the applySnapshot.
- MST destroyed or detached errors

## v2 format round trip
The document state is sent in v2 format to story builder, and story builder returns this state in v2 format. This makes things more complicated. Several ids/keys are not preserved doing this round trip:
- the shared model map entry keys are lost. During import the shared model keys are matched up and updated to equal the existing shared models using ids that are preserved inside of the shared models. 
- the ids of graph and map content models are lost. Additionally their child objects such as layers and data configurations have ids that are lost. In this PR the graph and map content model ids are not updated to match. There is code that is left in which can do this matching incase we want to revisit this. See the comment in the code for the reasoning behind this.
- formula ids are not round tripped, see the formula section.

## Volatile State
An `afterApplySnapshot` method was added to tiles and shared models so they can update their state after the applySnapshot. Commonly the initialization of volatile state also needs to be done by an `afterCreate` action. So in many cases an `initializeVolatileState` was added, and is called by `afterApplySnapshot` and `afterCreate`.

Here are some of the updated models:
- Attribute volatile properties were reviewed and I believe handled correctly by `initializeVolatileState`.
- Collection volatile properties and reviewed and I believe handled correctly by `initializeVolatileState`. Collections maintain several properties for quickly accessing cases. These properties are all cleared out when the snapshot is applied to be safe. In most cases this is not really necessary, but without it might result in residual entries in these properties which might case hard to find problems. 
- DataSet volatile properties were reviewed and most were handled correctly. The one known volatile property not dealt with is the `managingControllerId`.

The formulas are re-run after the applySnapshot, see below.

The `afterAttachToDocument` method of tile components is used to update volatile state in a few tiles. The calling of this was updated so that it would run after `applySnapshot` if the content model of the tile changed when the snapshot was applied.

There are many more volatile properties in other models that haven't been carefully reviewed. It would be best to review them systematically. However if things are working well enough we could just fix issues as we identify them rather than going through the code. 

## MST Errors
Because applySnapshot removes model objects from the tree if their id is different from the incoming snapshot, this will destroy these objects. CODAP has lots of reactions that are monitoring objects like this. Some of these are the React components themselves. Others are explicit reactions. We have patterns for making sure that both these types of reactions can be cleaned up without an error when the models they are observing are destroyed. However these patterns require checking for the liveliness of the models, or enumerating them as dependencies. 

There is at least one case that we don't have a good pattern for. It is in the attribute label component. This component works with several different model types. It adds a reaction that runs an superclass method on these model types. Within each model type it might use different child objects when implementing this method. Because of this the attribute label component reaction can't easily identify which model objects should be monitored for being destroyed and then dispose the reaction. The solution in this PR is to make sure the top level object is destroyed. This way the reaction will see this and dispose of itself. An alternative solution that might also work is to have the methods in the model types simply check if their child objects are destroyed. If one of them is destroyed it should trigger the reaction to run again, so the new version of the child object would be used when it is run again. So most likely just checking if they are alive will be enough to short circuit MobX from actually checking the properties of the stale object. 

One source of MST errors was caused by the FormulaManager instance that is created during the import of a v2 document. These error were fixed by destroying the temporary v3 document created during the import process. The reason this prevents the errors is just because the document is destroyed before the FormulaManager has time to run the problematic code. So this fix is fragile, if the timing is changed the problems might return. See below for more about this temporary FormulaManager.

This PR includes instructions for tracking down the cause of these errors when they happen. And it includes some fixes of the errors:
- some of the graph reactions were not checking if the dataConfiguration they were using was destroyed
- the graph component itself needed to check if the graph model was still alive before using it
- specify the graphModel.dataConfiguration as a mstReaction dependency when possible because in the future the graphModel might remain the same but the dataConfiguration will be updated

### Web View Tiles (plugins)
It is very likely that web view tile (plugin) state is not updated when the snapshot is applied. The snapshot might have changes to plugin state. Probably this state should be sent to the plugin so it can reflect the moment recorded by story builder. This can be complicated however because it is very likely that some plugins make assumptions that their state will match the CODAP document. And it is also likely that these plugins will have volatile state that won't get updated when the state of the plugin is updated. I'm not even sure there is a method for CODAP to update the state of a plugin after it has been initially loaded by CODAP.

Note: that story builder is a plugin itself, so we would not want to send this plugin its state. The data interactive document-handler already handles this when it is computing the snapshot to apply. It doesn't look for the story builder plugin explicitly. Instead it looks for plugins that are listening to document updates. That same logic could be used to prevent sending the state.

## Formulas
The formulas are now re-run after the applySnapshot. A method was added to the FormulaManager to do that. This was added because the snapshot might change a dataset value, and if that is the only thing that changes, the formulas would not recompute. This approach isn't great though. In the current code, if any of the changes made by the applySnapshot are things observed by the FormulaManager this will also trigger the formulas to recalculate. From what I can tell this happens before the `afterApplySnaphot` is called. That timing can be too soon. For example the `strValues` of attributes do not get correctly setup until `afterApplySnapshot`.  This timing issue might be fixed by bundling both the `applySnapshot` and `afterApplySnapshot` in a single action.

There is another problem with this though: even if the timing was fixed the formulas would be recomputed at least twice. There are other cases where formulas are likely computed more often than they need to be. The formula observer system could be updated so each observer has two parts. One part would go in the MobX reaction "data" function, and the other part would go in the "effect" function. Then the formula manager could have a single reaction to handle all of these observers. There would need to be some unification of the effects though so the recomputing is optimized.

With this design above, the manual recomputing triggered by afterApplySnapshot could be implemented as a property that is observed by the common reaction. This way it would be grouped with any other triggers caused by the applySnapshot itself.

Currently the formula ids are not round tripped through the v2 format. This formula id is something observed by the formula manager. We could fix these ids so they are round tripped, but there are also other things in the snapshot that can trigger the formulas to update, so fixing these ids is not a general solution for the issues above. 

Here is a list of formula types:
- attribute-formula-adapter
- filter-formula-adapter
- data-display-filter-formula-adapter
- bar-chart-formula-adapter: this stores bars in the bar-chart volatile state, so it need to be recomputed
- plotted-value-formula-adapter: this stores result in MeasureInstance state, so it shouldn't need to be recomputed
- plotted-function-formula-adapter: this stores a function result in volatile state when the function is recalculated, so if the function uses item values perhaps with a lookup, then it might need to be recomputed if the item values change.

## Additional changes:
- typescript type guards were added for snapshot types and model types to make this matching of ids type safe.
- real MST identifiers are now used for DataDisplayContentModels, this way MST recreates them when the snapshot is applied, and this eliminates several stale reaction errors
- delete a dataset volatile property `attrNameMap`. There is a view with the same name that computes this map now, and it looks like the volatile property was not cleaned up.
- an id is added to the FormulaManager instance to help debug issues caused by multiple FormulaManagers

## Unresolved Questions
There is a  method `document.setContent` which also calls applySnapshot. Perhaps this should also use the afterApplySnapshot logic. 

Every time a v2 document is converted to v3, a new formula manager is created. Because this conversation happens when each moment is loaded, that can be a lot of formula managers. These formula managers do setup observers/reactions, however they should only be observing a MST models that are destroyed. It might be better to not create a formula manager when converting a v2 document. But the v2 importing code might need this formula manager to exist in order to work properly.

### DIFF
--- a/v3/src/models/data/data-set.ts
+++ b/v3/src/models/data/data-set.ts
@@ -1402,7 +1402,7 @@ export const DataSet = V2UserTitleModel.named("DataSet").props({
     }
 
     // TODO: replace this with initializeVolatileState()
-    // However we still need to add the initial collection, hopefully the order of
+    // However we still need to add the initial collection, hopefully the order
     // of doing that doesn't matter.
     self.syncCollectionLinks()
 
@@ -1481,7 +1481,7 @@ export const DataSet = V2UserTitleModel.named("DataSet").props({
             // TODO: It looks like the mirror array is used just so we can figure out what the itemId
             // was that was just removed, this could also be done by looking at the second
             // argument to the onPatch callback which is the reversePatch.
-            // The reversePath will have the value that is being removed
+            // The reversePatch will have the value that is being removed
             const itemId = self.setAsideItemIdsMirror[index]
             self.setAsideItemIdsMirror.splice(index, 1)
             self.setAsideItemIdsSet.delete(itemId)
--- a/v3/src/models/formula/formula-manager.ts
+++ b/v3/src/models/formula/formula-manager.ts
@@ -1,7 +1,7 @@
 import { comparer, makeObservable, observable, reaction, action } from "mobx"
 import { addDisposer } from "mobx-state-tree"
 import { parse } from "mathjs"
-import { nanoid } from "nanoid"
+import { uniqueId } from "../../utilities/js-utils"
 import { boundaryManager, BoundaryManager } from "../boundaries/boundary-manager"
 import { IDataSet } from "../data/data-set"
 import { ICase } from "../data/data-set-types"
@@ -34,7 +34,7 @@ export class FormulaManager implements IFormulaManager {
   areAdaptersInitialized = false
 
   // Unique identifier of this formula manager instance, useful for debugging.
-  id = nanoid()
+  id = uniqueId()
 
   constructor() {
     makeObservable(this)
--- a/v3/src/models/codap/create-codap-document.test.ts
+++ b/v3/src/models/codap/create-codap-document.test.ts
@@ -8,6 +8,7 @@ import "../shared/shared-data-set-registration"
 // eslint-disable-next-line no-var
 var mockNodeIdCount = 0
 jest.mock("../../utilities/js-utils", () => ({
+  ...jest.requireActual("../../utilities/js-utils"),
   hashOrderedStringSet: () => 12345678,
   hashStringSet: () => 12345678,
   typedId: () => `test-${++mockNodeIdCount}`,


v3/src/components/app.tsx
```diff
@@ -44,6 +45,11 @@ import "../models/shared/shared-data-set-registration"
 
 import "./app.scss"
 
+// Uncomment this to help track down MST errors like:
+// "You are trying to read or write to an object that is no longer part of a state tree"
+// See mst-detached-error.md for more information.
+// setLivelinessChecking("error")
```
kswenson, line 51: If there were any more such configuration changes, they could be packaged into a custom hook.

v3/src/models/data/data-set.ts
```diff
@@ -1390,6 +1478,10 @@ export const DataSet = V2UserTitleModel.named("DataSet").props({
             self.setAsideItemIdsSet.add(itemId)
           }
           else {
+            // TODO: It looks like the mirror array is used just so we can figure out what the itemId
+            // was that was just removed, this could also be done by looking at the second
+            // argument to the onPatch callback which is the reversePatch.
+            // The reversePath will have the value that is being removed
```
kswenson:
```suggestion
            // The reversePatch will have the value that is being removed.
```

v3/src/models/data/data-set.ts
```diff
@@ -1318,6 +1401,9 @@ export const DataSet = V2UserTitleModel.named("DataSet").props({
       invalidate: () => self.invalidateCases()
     }
 
+    // TODO: replace this with initializeVolatileState()
+    // However we still need to add the initial collection, hopefully the order of
```
kswenson:
```suggestion
    // However we still need to add the initial collection, hopefully the order
```

v3/src/models/formula/formula-manager.ts
```diff
@@ -32,9 +33,11 @@ export class FormulaManager implements IFormulaManager {
   @observable
   areAdaptersInitialized = false
 
+  // Unique identifier of this formula manager instance, useful for debugging.
+  id = nanoid()
```
kswenson: Would be more conventional to call `uniqueId()` or `typedId()` in `js-utils` rather than calling `nanoid()` directly.

v3/src/models/document/document.ts
```diff
@@ -117,6 +117,7 @@ export const DocumentModel = Tree.named("Document")
     setContent(snapshot: IDocumentContentSnapshotIn) {
       if (self.content) {
         applySnapshot(self.content, snapshot)
+        // TODO: should we run afterApplySnapshot here?
```
kswenson, line 120: Clearly `afterApplySnapshot()` should be called somewhere. Is the question whether it should be called here instead of somewhere else or whether it should be called here as well as elsewhere?

v3/src/models/formula/formula-manager.ts
scytacki: I initially used typedId but that broke tests because there are tests that mock typedId to return a counter based id so the ids can be checked in mock data. I'll give uniqueId a try...

v3/src/models/document/document.ts
scytacki: I didn't track down where setContent was called from, so I wasn't sure if it was necessary to update the volatile state in these places where it is used. So the question is whether it should be called here at all. If it does need to be called, here seems as good a place as any.

PR #2152, opened 2025-09-23
CODAP-822 Guide button functionality

bfinzer:
[#CODAP-822] Feature: When a Guide is present in a document the Guide button should be present, enabled and functional

* Made it so the Guide button is present and enabled when a guide exists
* Defined a `GuideButton` component and gave it the ability to show the guide and set the page from those available
* Moved handleSelectTile from tiles-list-button.tsx to guide-button.tsx
* Made `ITileContentInfo.hideOnClose` a function so that the guide subType of `WebViewModel` could override the default.
* Required `Container.handleCloseTile` to go through `DocumentContentModel.deleteOrHideTile` so that a guide is hidden rather than deleted.

### DIFF
--- a/v3/src/components/tool-shelf/tool-shelf-utilities.tsx
+++ b/v3/src/components/tool-shelf/tool-shelf-utilities.tsx
@@ -24,13 +24,14 @@ export const handleSelectTile = (tileId: string, documentContent?: IDocumentCont
   const tileRow = documentContent?.findRowContainingTile(tileId)
   const tileLayout = tileRow?.getTileLayout(tileId)
   if (isFreeTileLayout(tileLayout) && (tileLayout.isHidden || tileLayout.isMinimized)) {
+    const type = documentContent?.getTile(tileId)?.content.type || "tile"
     documentContent?.applyModelChange(() => {
       tileLayout.setHidden(false)
       tileLayout.setMinimized(false)
     }, {
-      undoStringKey: "DG.Undo.guide.show",
-      redoStringKey: "DG.Redo.guide.show",
-      log: logMessageWithReplacement("Show Guide", {}, "document")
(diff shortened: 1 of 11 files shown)


v3/src/components/tool-shelf/tool-shelf.tsx
```diff
@@ -30,9 +31,10 @@ export function isShelfTileComponent(info?: ITileComponentInfo): info is IShelfT
   return !!info && "shelf" in info && info.shelf != null
 }
 
-function hasGuideTile(document: IDocumentModel): boolean {
+function hasGuideTile(document: IDocumentModel) {
```
kswenson: The name `hasGuideTile` implies a boolean return. Perhaps `getGuideTileId` instead?

v3/src/components/tool-shelf/tool-shelf-utilities.tsx
```diff
@@ -16,3 +18,20 @@ export const showWebView = (url: string, documentContent?: IDocumentContentModel
...
+  const tileRow = documentContent?.findRowContainingTile(tileId)
+  const tileLayout = tileRow?.getTileLayout(tileId)
+  if (isFreeTileLayout(tileLayout) && (tileLayout.isHidden || tileLayout.isMinimized)) {
+    documentContent?.applyModelChange(() => {
+      tileLayout.setHidden(false)
+      tileLayout.setMinimized(false)
+    }, {
+      undoStringKey: "DG.Undo.guide.show",
+      redoStringKey: "DG.Redo.guide.show",
+      log: logMessageWithReplacement("Show Guide", {}, "document")
```
kswenson: It seems like `handleSelectTile` can be called for any hidden or singleton component (e.g. Calculator), so the undo/redo and log message strings either need to be more generic or customized based on tile type.

PR #2154, opened 2025-09-29
feat: first pass at text inspector

kswenson:
Full client-side text toolbar (inspector) implementation including dialogs
- updates to latest version of slate-editor library (0.12.0) with improved external toolbar support
- uses icons from Michael's spec
- uses existing CODAP color picker for text color
- uses new client-side link dialog
  - single-click on link in text launches link url (after delay sufficient to determine that it's not part of a double-click)
  - double-click on link in text opens link dialog
- uses new client-side image dialog
  - single-click on image selects image
  - double-click on image in text opens image dialog

Note: After discussion with @bfinzer, this PR does not attempt to precisely match the spec. The goal is to approximate the spec and achieve the basic functionality. A follow-up PR can address any remaining tweaks.


v3/src/components/text/text-tile.tsx
```diff
@@ -28,26 +38,68 @@ export const TextTile = observer(function TextTile({ tile }: ITileBaseProps) {
...
+      }
+    })
+    textModel?.setEditor(_editor)
+    return _editor
+  }, [initialValue, openImageDialog, openLinkDialog, textModel])
+
+  const focusEditor = useCallback(() => {
+    if (animationFrame.current) cancelAnimationFrame(animationFrame.current)
+    animationFrame.current = requestAnimationFrame(() => ReactEditor.focus(editor))
+  }, [editor])
```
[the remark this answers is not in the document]
kswenson: `animationFrame` is stable. `animationFrame.current` need not be in the dependency array.

v3/src/components/text/text-tile.tsx
```diff
@@ -28,26 +38,68 @@ export const TextTile = observer(function TextTile({ tile }: ITileBaseProps) {
...
 
   useEffect(() => {
-    let animationFrame: number
-    const disposer = tile && mstReaction(
+    return () => {
+      if (animationFrame.current) cancelAnimationFrame(animationFrame.current)
+    }
+  }, [])
```
tealefristoe, line 73: Why do we need to cancel the animation frame on first render?

v3/src/components/text/format-image-button.tsx
```diff
@@ -0,0 +1,282 @@
...
+    }
+  }, [editImage, isOpen])
+
+  const inputBorderProps = {
+    border: "1px solid #006c8e",
+    borderRadius: "4px"
+  }
+  const placementLabel =
+    placement === "left" ? "Float Left" :
+    placement === "right" ? "Float Right" : "Inline"
```
tealefristoe:
```suggestion
  const placementLabels = {
    "left": "Float Left",
    "right": "Float Right",
    "inline": "Inline"
  }
  const placementLabel = placementLabels[placement]
```
Then you can use `placementLabels` in lines 246-254 to avoid duplicating the display strings.

v3/src/components/text/text-inspector.scss
```diff
@@ -0,0 +1,16 @@
...
+  grid-template-rows: repeat(9, 34px);
+  grid-auto-flow: column;
+  place-items: center;
+
+  .inspector-tool-button {
+    min-width: 34px !important;
+    max-width: 34px !important;
+    width: 34px !important;
+    height: 34px !important;
+    padding: 5px !important;
```
tealefristoe: I don't think any of these `!important`s are necessary because `.inspector-panel.text .inspector-tool-button` will supersede `.inspector-panel .inspector-tool-button`. If there's some case where these are necessary, I bet you could get around it by adding `.text` to the button or wrapping all this in `.free-tile-component`. Maybe a bit hacky feeling, but better than `!important`s in my opinion.

v3/src/components/text/text-tile.tsx
kswenson: We're returning a disposer which will cancel the animation frame before unmounting the component.

v3/src/components/text/text-inspector.scss
kswenson: Good catch -- I added those initially just to bypass any specificity issues during the initial investigation and never came back to double-check.

PR #2157, opened 2025-10-01
fix: V2 doc won't open in V3 (group key import bug)

kswenson:
[[CODAP-941](https://concord-consortium.atlassian.net/browse/CODAP-941)] V2 doc won't open in V3

There are some subtleties here. To maintain persistent case ids, we must store the necessary mappings between item ids and values and their corresponding case ids. Specifically, for child collections item ids map directly to case ids, while for parent collections a set of values for the attributes in the collection is mapped to a case id. Internally, this is called the group key, and the serialized form of parent collection group keys has changed over time. Initially, it was just a `JSON.stringify`ed array of attribute values. In #2006 it was changed to be a tab-delimited "array" of values for performance reasons. There were two problems with the implementation in #2006, however. The first is that the choice of `[` and `]` to delimit the group key is a poor one given that it makes it ambiguous whether this is a `JSON.stringify`ed array or a newer group key. The second is that the code to import v2 documents was not updated to use the new group key format, resulting in v2 documents going through the legacy import path unnecessarily. With this PR we switch to using `\r` to delimit the beginning/end of parent group keys, and switch v2 import to use the new group key format. There is `preProcessSnapshot` code in the `CollectionModel` which handles the various legacy formats.

[CODAP-941]: https://concord-consortium.atlassian.net/browse/CODAP-941?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/v3/src/models/data/collection.test.ts
+++ b/v3/src/models/data/collection.test.ts
@@ -1,7 +1,7 @@
 import { destroy, getSnapshot, isAlive, types } from "mobx-state-tree"
 import { Attribute, IAttribute } from "./attribute"
 import {
-  CollectionModel, ICollectionModel, IItemData, defaultItemData, isCollectionModel, syncCollectionLinks
+  CollectionModel, ICollectionModel, IItemData, defaultItemData, isCollectionModel, makeGroupKey, syncCollectionLinks
 } from "./collection"
 
 const Tree = types.model("Tree", {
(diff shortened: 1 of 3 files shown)


bfinzer (approved):
👍🏻LGTM
Nice work! I'm glad we caught and fixed this one.

PR #2158, opened 2025-10-01
fix: boolean evaluation in if() function and ternary operator

kswenson:
[[CODAP-931](https://concord-consortium.atlassian.net/browse/CODAP-931)] Formula testing boolean value not yielding correct result

- fix the `if()` function to perform boolean evaluation using our logic
- fix the ternary operator by replacing it with the `if()` function


bfinzer (approved):
👍🏻LGTM

PR #2159, opened 2025-10-02
CODAP-949 New plugin api for fusing dots into rectangles

bfinzer:
[#CODAP-949] Feature: Plugin API allows "Fuse Dots into Rectangles"

* We add `pointesAreFusedIntoBars` to `V2Graph` so that this becomes a property specifiable in the update of a graph component.
* In graph-component-handler.ts `update` we detect the presence of this property and pass it to the `GraphContentModel` using the existing `fusePointsIntoBars`. If the property is true, we preflight by calling the new method `pointsCanBeFusedIntoBars` and return an error if this is not true.
* Also started a branch, `V3-changes-to-wiki` in the `codap.wiki` project to document this new api.

### DIFF
--- a/v3/src/components/graph/models/graph-content-model.ts
+++ b/v3/src/components/graph/models/graph-content-model.ts
@@ -564,6 +564,14 @@ export const GraphContentModel = DataDisplayContentModel
       }
       self.setPlotType(newPlotType)
     },
+    pointsCanBeFusedIntoBars(): boolean {
+      const primaryRole = self.dataConfiguration.primaryRole
+      const primaryType = primaryRole ? self.dataConfiguration.attributeType(primaryRole) : undefined
+      const secondaryRole = self.dataConfiguration.secondaryRole
+      const secondaryType = secondaryRole ? self.dataConfiguration.attributeType(secondaryRole) : undefined
+      return self.plot.displayType !== "bars" && secondaryType === undefined &&
+        (isCategoricalAttributeType(primaryType) || isNumericAttributeType(primaryType))
+    },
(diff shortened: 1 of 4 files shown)


v3/src/components/graph/models/graph-content-model.ts
```diff
@@ -564,6 +564,14 @@ export const GraphContentModel = DataDisplayContentModel
       }
       self.setPlotType(newPlotType)
     },
+    pointsCanBeFusedIntoBars(): boolean {
+      const primaryRole = self.dataConfiguration.primaryRole
+      const primaryType = primaryRole ? self.dataConfiguration.attributeType(primaryRole) : undefined
+      const secondaryRole = self.dataConfiguration.secondaryRole
+      const secondaryType = secondaryRole ? self.dataConfiguration.attributeType(secondaryRole) : undefined
+      return self.plot.displayType !== "bars" && secondaryType === undefined &&
```
kswenson: I would use `!secondaryType` in this case, but that's a personal preference.

v3/src/components/graph/models/graph-content-model.ts
```diff
@@ -564,6 +564,14 @@ export const GraphContentModel = DataDisplayContentModel
...
+        (isCategoricalAttributeType(primaryType) || isNumericAttributeType(primaryType))
+    },
```
kswenson: Since this doesn't change the model, it seems like this should be in `views` rather than `actions`, where it could also be a getter (i.e. `get pointsCanBeFusedIntoBars()`.

kswenson (approved):
👍 LGTM -- just a couple of suggestions.

PR #2165, opened 2025-10-03
CODAP-912 Use correct color for map boundary strokes

bfinzer:
[#CODAP-912] Bug fix: Map boundaries not being painted correctly when map has legend

* `refreshPolygonStyles` in map-polygon-layer.tsx was always returning the legend color for the case as the stroke color instead of only doing that when `itemStrokeSameAsFill` was true.


kswenson (approved):
👍 LGTM

PR #2528, opened 2026-04-24
CODAP-1255: keep points visible during category reorder

bfinzer:
## Summary
- Fixes CODAP-1255: points disappeared for ~2 s when a category was dragged to
  reorder on a graph axis. Two reaction paths were re-applying masks against
  stale `state.subPlotNum` values during the animation window that
  `respondToCategorySetChanges` intentionally leaves mask-free.
- Track a `categorySetChangeInFlightRef` so the debounced
  `callRefreshPointPositions` skips `updateCellMasks` while the delayed
  callback owns the mask reapply; heal the flag if the animation timer is
  cancelled externally (e.g., by a point-drag's `stopAnimation`).
- Gate the renderer-change `useEffect`'s `updateCellMasks` on
  `rendererChanged` (matching the adjacent `callMatchCirclesToData`); without
  this, dot-chart's `SubPlotCells` recreation on axis-scale `changeCount`
  flipped `refreshPointPositions` identity mid-reorder and reapplied masks
  against stale state.

## Test plan
- [ ] Open Mammals, Diet graph: drag "both" to swap with "meat" — points stay
      continuously visible throughout the animation.
- [ ] Same as above on a dot plot (Diet vs. a numeric attribute) — points stay
      visible.
- [ ] Regression: hide cases in a bar chart with formula-computed heights —
      bars rescale correctly (CODAP-810 behavior preserved).
- [ ] Regression: bar chart category drag — bars don't disappear (CODAP-1214
      CODAP-769 behavior preserved).
- [ ] CI: lint, type-check, unit tests pass; Cypress regression (via label).

### DIFF
--- a/v3/src/components/graph/hooks/use-plot.ts
+++ b/v3/src/components/graph/hooks/use-plot.ts
@@ -110,9 +110,27 @@ export const usePlotResponders = (props: IPlotResponderProps) => {
     metadata = dataConfiguration?.metadata,
     instanceId = useInstanceIdContext()
 
+  // Suppresses updateMasks while a category-set change is in flight. The delayed
+  // callback in respondToCategorySetChanges owns the mask reapply in that case; running
+  // updateCellMasks sooner would reapply masks against stale state.subPlotNum values,
+  // clipping sprites to their old cells as they animate to their new positions.
+  const categorySetChangeInFlightRef = useRef(false)
+
   const callRefreshPointPositions = useAccumulatingDebouncedCallback((_props: IRefreshProps) => {
     const { selectedOnly = false, updateMasks = false } = _props
-    if (updateMasks) {
+    // If the flag is set but no animation is actually running, the delayed cleanup
+    // was cancelled (e.g., by a point-drag calling stopAnimation). Heal it here so
+    // we don't suppress mask updates indefinitely. Sync subPlotNums and reapply masks
+    // together so rendered sprite masks match state — syncing one without the other
+    // leaves a stale pairing that later mask reapplications (e.g., on resize) would bake in.
+    if (categorySetChangeInFlightRef.current && !graphModel.animationTimerId) {
+      categorySetChangeInFlightRef.current = false
+      if (renderer) {
+        renderer.updateSubPlotNums(dataConfiguration.caseDataWithSubPlot)
+        updateCellMasks({ dataConfig: dataConfiguration, layout, renderer })
+      }
+    }
+    if (updateMasks && !categorySetChangeInFlightRef.current) {
       updateCellMasks({ dataConfig: dataConfiguration, layout, renderer })
     }
     refreshPointPositions(selectedOnly)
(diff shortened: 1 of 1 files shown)


kswenson (approved):

## PR Review Summary — #2528 (CODAP-1255)

**Changes:** 1 file, +25 / -3 lines (`v3/src/components/graph/hooks/use-plot.ts`)

### What it does

Fixes a visual regression where, on graphs with a categorical axis, dragging a
category to reorder caused points to disappear for roughly 2 seconds (the
animation window) before reappearing in their new cells. The fix ensures that
all paths that might call `updateCellMasks` during the animation window are
suppressed, so masks aren't re-applied against stale `subPlotNum` values while
sprites are mid-animation.

### The approach

The animation flow in `respondToCategorySetChanges` (use-plot.ts:213) removes
masks, starts a 2-second animation, and schedules a delayed callback to
`updateSubPlotNums` + `updateCellMasks` with fresh state. The bug was that
other reactions were calling `updateCellMasks` during that window. Since
`doResize` in pixi-point-renderer.ts:229 always rebuilds masks and re-applies
them to sprites via `reapplyMasksFromState` (which reads `state.subPlotNum`),
any mid-animation call clipped sprites to their old cells — so as they animated
to new positions they disappeared behind the stale masks.

The fix plugs the two paths that fire during a categorical reorder:

1. **Debounced `callRefreshPointPositions` (use-plot.ts:119)** — a new
   `categorySetChangeInFlightRef` is set in `respondToCategorySetChanges` and
   cleared in the delayed animation callback. While set, the debounced path
   skips `updateCellMasks`. This guards the `[axis scale domain]` reaction
   (line 272), `respondToCasesChange` (246), and `[legendColorChange]` (336),
   all of which dispatch `callRefreshPointPositions({ updateMasks: true })`.

2. **Renderer-change `useEffect` (use-plot.ts:162)** — `updateCellMasks` is
   moved inside the `if (rendererChanged)` block, matching the adjacent
   `callMatchCirclesToData`. In dot-chart/bar-chart, `subPlotCells` is
   recreated on axis-scale `changeCount` (use-chart-dots.ts:29), which flips
   `refreshPointPositions` identity; before the fix, this re-ran the effect
   and re-applied masks against stale state even when the renderer hadn't
   actually changed.

A **heal path** in the debounced callback covers the edge case where the
animation is cancelled externally (e.g., a point-drag calling `stopAnimation`).
The flag would otherwise stay set indefinitely. On the next
`callRefreshPointPositions`, if the flag is set but `animationTimerId === 0`,
the hook syncs `subPlotNums` and reapplies masks together — keeping the
rendered state consistent and unblocking future mask updates.

### Assessment

The analysis of the two reaction paths matches what's in the code. Key facts
I verified:

- `doResize` does always rebuild masks + call `reapplyMasksFromState` using
  `state.subPlotNum` (pixi-point-renderer.ts:288), so the stale-state concern is
  real and the gate placement is necessary.
- `graph.tsx`'s `handleSubPlotsUpdate` (graph.tsx:170) watches
  `categoricalAttrsWithChangeCounts`, which reflects attribute-value change
  counts — NOT category order. Reorder operates on `CategorySet.values` via
  `moveValueToIndex` without bumping `attr.changeCount`, so this reaction does
  not fire on reorder and doesn't need a gate. Correct targeting.
- `respondToCategorySetChanges` also watches `allCategoriesForRoles`, which
  DOES change on reorder — that's the reaction doing the mask removal and
  animation orchestration.
- Moving `updateCellMasks` inside `rendererChanged` is safe: layout resize is
  covered by `graph.tsx:139` (which calls `updateCellMasks` directly when
  `plotWidth/Height` change) and by `[axis range]` (which calls
  `callRefreshPointPositions` without `updateMasks`, so mask rebuild wasn't
  happening from this path anyway). No regression expected.
- Regression coverage for CODAP-810 and CODAP-1214 is preserved: those flows
  don't set `categorySetChangeInFlightRef`, so the gate is open and
  `updateCellMasks` continues to run normally.
- CI is all green, including the full Cypress regression suite (365 tests) and
  Jest unit tests.

### Issues

No concerns. The fix is focused, well-commented, and correctly targets the two
reaction paths that were re-applying masks during the animation window. The
heal path for externally-cancelled animations is a nice defensive touch.

PR #2539, opened 2026-04-29
CODAP-141: skip save-before-language-change in CFM

kswenson:
## Summary

- Fixes CODAP-141 — the "[object Object]" alert that appeared when changing
  languages with a saved document open (notably in Activity Player embeds).
- Bumps `@concord-consortium/cloud-file-manager` from `~2.2.9` to `~2.2.10`,
  picking up the upstream callback-shape fix in CFM's `changeLanguage()`.
- Wires up the new `saveOnLanguageChange` appOption (added in CFM 2.2.10) and
  sets it to `false` for CODAP v3. CODAP v3 does not reload the page on
  language change, so the save-before-change flow was unnecessary anyway.

Fixes CODAP-141.

## Test plan

- [ ] CI green (lint, build, limited Cypress)
- [ ] Manual: open a document from Google Drive, switch languages via the
      Languages menu, confirm the language changes with no dialog and no
      `[object Object]` alert.
- [ ] Regression: change languages with no document loaded and with a local
      document loaded — both should remain seamless.
- [ ] Activity Player embed (the original repro): confirm the `[object Object]`
      modal no longer appears after selecting a different language.

### DIFF
--- a/v3/package-lock.json
+++ b/v3/package-lock.json
@@ -16,7 +16,7 @@
         "@codemirror/commands": "^6.10.3",
         "@codemirror/language": "^6.12.3",
         "@codemirror/view": "^6.41.1",
-        "@concord-consortium/cloud-file-manager": "~2.2.9",
+        "@concord-consortium/cloud-file-manager": "~2.2.10",
         "@concord-consortium/log-monitor": "^1.0.0",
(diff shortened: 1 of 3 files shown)


kswenson:
Merging without review as it's a trivial application of the CFM change that was reviewed in [CFM #425](https://github.com/concord-consortium/cloud-file-manager/pull/425).

PR #2545, opened 2026-05-01
CODAP-1285: fix percentile() argument to use [0,1] range

kswenson:
## Summary

- The `percentile()` formula function was passing `p / 100` to `quantileOfSortedArray`, treating the argument as if it were in the range [0, 100]. The user-facing docs and the underlying utility expect a value in [0, 1].
- Drop the spurious `/ 100` and update the tests to use the correct range.
- Brings v3 to parity with v2.

Fixes CODAP-1285.

Follow-up: filed CODAP-1287 to address asymmetric out-of-range handling (`p > 1` silently clamps to max) — out of scope for this fix.

## Test plan

- [ ] CI lint + build pass
- [ ] `univariate-stats-functions.test.ts` passes (verified locally)
- [ ] Spot-check `percentile(Speed, 0.5)` returns the median in a real document


tealefristoe (approved):
Looks good 👍

PR #2555, opened 2026-05-05
CODAP-1295: fix plugin create notifications carrying wrong tile id

kswenson:
## Summary

Extracts two plugin-API bug fixes from PR #2532 (CODAP-1231) so they can ship in V3.0.0 ahead of the stacked tour API rewrite, which targets a post-3.0.0 `cisco-tours` branch.

Both bugs caused plugin `{ component: <id> }` lookups to fail because the plugin received a tile id that didn't resolve to a workspace component:

- **Bug 1** (`case-table-tool-shelf-button.tsx`): `handleCreateNewCaseTable` was broadcasting the orphan tile id (used only as a `providerId` for `gDataBroker.addDataSet`) instead of the visible tile's id. The orphan tile is never inserted into the document, so plugin lookups via `document.getElementById` returned null. Fix: notify with `sharedMetadata.caseTableTileId`.
- **Bug 2** (`component-handler.ts`): `notify: createTileNotification(tile)` was evaluated immediately when the options object literal was constructed — `tile` was still `undefined` inside the `applyModelChange` actionFn closure, so `createTileNotification(undefined)` short-circuited and no create notification was emitted at all. Fix: wrap in a thunk so it runs after the actionFn assigns `tile`.

Includes a regression test for Bug 2 (`component-handler.test.ts`). Bug 1's regression test was deliberately not extracted from PR #2532 to minimize merge conflicts when the tour API rewrite eventually lands.

Fixes CODAP-1295.

## Test plan

- [ ] CI passes (lint, build, jest, cypress)
- [ ] Verify plugin highlight via `{ component: <id> }` works against a newly-created case table from the tool shelf
- [ ] Verify plugin highlight via `{ component: <id> }` works against a plugin-created component (graph, table, etc.)

### DIFF
--- a/v3/src/components/case-table/case-table-tool-shelf-button.tsx
+++ b/v3/src/components/case-table/case-table-tool-shelf-button.tsx
@@ -62,12 +62,27 @@ const CaseTableToolShelfMenuList = observer(
       // the new tile is created at 0,0. The correct fix for this issue is to prevent scrolling to show the new table.
       // const options: INewTileOptions = { animateCreation: true, markNewlyCreated: true }
       const options: INewTileOptions = { markNewlyCreated: true }
-      tile = createDefaultTileOfType(kCaseTableTileType, options)
-      if (!tile) return
-      const { sharedData, sharedMetadata } = gDataBroker.addDataSet(ds, tile.id)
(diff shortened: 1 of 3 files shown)


dougmartin (approved):
Looks good 👍

PR #2556, opened 2026-05-05
CODAP-1290: case card and formula observer reactivity

kswenson:
## Summary

Fixes CODAP-1290 — case card values (both per-case and summary) didn't update when values were changed externally (e.g. by dragging a slider that drives a formula). Two distinct issues were uncovered during investigation, fixed in two commits:

- **Case card reactivity** (commit 1): the per-element value accessors on `Attribute` (`strValue`, `value`, `numValue`) read volatile arrays whose mutations are not deeply observable. The case card relied on `observer()` to refresh, so reads silently registered no MobX dependency. Touched `attr.changeCount` in `CaseAttrView` and `summarizedValues` so the existing mutation counter (already tracked by `length` and `type`) drives re-evaluation. Avoids imposing tracking overhead on hot graph / map / collection paths that share the accessors.

- **Formula observer registration** (commit 2): if a formula referenced a global value by name before that global existed (e.g. formula authored before a slider was created), the formula's static dependency list was captured empty at registration time. Adding the global later correctly re-resolved the canonical form and triggered a one-shot recalc, but the dependency-keyed observers were never refreshed — subsequent value changes were silently missed. `setupFormulaObservers`'s `updateDisplay` closure now disposes and re-registers observers when the canonical form changes.

## Test plan

- [x] Unit tests added for both fixes (`case-card-model.test.ts`; `formula-manager.test.ts` new "global value referenced by name added after the formula" case)
- [x] `npm test -- formula` (460/460 pass) and `npm test -- case-card` (9/9 pass)
- [x] `npm run lint` and `npm run build:tsc` clean
- [ ] Manual verification in dev server: open a case card on a slider-driven formula attribute, drag the slider, confirm both the displayed value and (where applicable) the summary update
- [ ] Manual verification of the formula-after-slider scenario: create slider first → formula should update on drag; then create formula first → slider afterward → formula should still update on drag (the second commit's fix)
- [ ] Cypress regression run via `run regression` label

### DIFF
--- a/v3/src/components/case-card/case-attr-view.tsx
+++ b/v3/src/components/case-card/case-attr-view.tsx
@@ -36,6 +36,11 @@ export const CaseAttrView = observer(function CaseAttrView (props: ICaseAttrView
   const caseId = groupedCase?.__id__ ?? ""
   const cardModel = useCaseCardModel()
   const { data, metadata } = cardModel || {}
+  // Establish a MobX dependency on the attribute's mutation counter so this observer
+  // re-renders when values change (e.g. slider-driven formula recomputes). The data
+  // accessors below read non-observable volatile arrays. (CODAP-1290)
+  // eslint-disable-next-line @typescript-eslint/no-unused-expressions
+  attr?.changeCount
   const cellStrValue = isCollectionSummarized
                         ? cardModel?.summarizedValues(attr, collection)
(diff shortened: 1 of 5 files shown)


bfinzer (approved):
👍🏻LGTM
Nice explanation of the problem and solution

PR #2557, opened 2026-05-05
CODAP-1296: fix calculator display when result matches prior value

kswenson:
## Summary
- Calculator's display string is synced from the model via an autorun, which fires only when the model value changes. When a new expression evaluates to the same string already in the model (e.g. `0*0` after `2*0=0`), the autorun didn't fire and the display kept showing the typed expression instead of collapsing to the result.
- Set the local display state explicitly in the evaluate path, matching the existing pattern in `clearValue`.
- Adds a Cypress test that reproduces the bug.

Fixes [CODAP-1296](https://concord-consortium.atlassian.net/browse/CODAP-1296).

## Test plan
- [ ] Open the Calculator
- [ ] Type `2*0` then Enter — display collapses to `0`
- [ ] Type `*0` then Enter — display collapses to `0` (was: stuck on `0*0`)
- [ ] Cypress `calculator.spec.ts` passes (locally: 17/17 ✓)

[CODAP-1296]: https://concord-consortium.atlassian.net/browse/CODAP-1296?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/v3/cypress/e2e/calculator.spec.ts
+++ b/v3/cypress/e2e/calculator.spec.ts
@@ -104,6 +104,12 @@ context("Calculator", () => {
     calc.typeExpression("+5{enter}")
     calc.checkCalculatorDisplay("15")
   })
+  it("collapses expression when new result equals the previous result", () => {
+    calc.typeExpression("2*0{enter}")
+    calc.checkCalculatorDisplay("0")
+    calc.typeExpression("*0{enter}")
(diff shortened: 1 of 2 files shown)


bfinzer (approved):
👍🏻LGTM

PR #2561, opened 2026-05-06
CODAP-1298: keep movable line pivot stable across vertical

bfinzer:
## Summary
- The movable-line rotation pivot was recomputed on every drag tick and fell back to `pointsOnAxes.current.pt2` (or `pt1`) when no saved pivot existed. Because `lineToAxisIntercepts` sorts the two intercepts by x, `pt2` swapped which physical end of the line it represented the moment the slope crossed vertical and went negative — the pinned end teleported and the line jumped around.
- Capture the rotation pivot once at the start of a drag into a ref and reuse it for the duration; reset on drag end.

Fixes CODAP-1298.

## Test plan
- [ ] Open Mammals → graph with two numeric attributes → add a moveable line
- [ ] Drag the bottom handle to the right past vertical: the top end stays pinned and the line rotates smoothly into a negative slope
- [ ] Drag the top handle the other way past vertical: the bottom stays pinned
- [ ] Sanity-check normal positive-slope rotation, the middle-handle translate, and the intercept-locked variant

### DIFF
--- a/v3/src/components/graph/adornments/movable-line/movable-line-adornment-component.tsx
+++ b/v3/src/components/graph/adornments/movable-line/movable-line-adornment-component.tsx
@@ -66,6 +66,7 @@ export const MovableLineAdornment = observer(function MovableLineAdornment(props
   const lineRef = useRef() as React.RefObject<SVGSVGElement>
   const [lineObject, setLineObject] = useState<ILine>({})
   const pointsOnAxes = useRef<IAxisIntercepts>({pt1: {x: 0, y: 0}, pt2: {x: 0, y: 0}})
+  const dragPivotRef = useRef<Point | null>(null)
   const xScaleRef = useRef(xScale.copy())
   const yScaleRef = useRef(yScale.copy())
   const logFn = useRef<Maybe<LogMessageFn>>()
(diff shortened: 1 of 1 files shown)


kswenson (approved):

## PR Review Summary — #2561

**Changes:** 1 file, +22 / -13 lines

### What it does

Fixes CODAP-1298: when rotating a movable line past vertical, the pinned end of the line would teleport and the line would jump. The fix is to capture the rotation pivot once at the start of a drag and reuse it for the duration of that drag.

### The approach

Adds `dragPivotRef` (a `useRef<Point | null>`) at the top of `MovableLineAdornment`. Inside `handleRotation`, the pivot is computed and stored in the ref the first time per drag (when the ref is null), and reused on every subsequent drag tick. The ref is cleared on `isFinished` after `applyModelChange` runs (so the next drag sees the freshly-saved pivot from the model).

The pivot is stored as a plain `Point` POJO (`{x: ..., y: ...}`) rather than as a reference to the MST `PointModel` instance. This is the right call — capturing a snapshot value isolates the cached pivot from any subsequent model mutations during the drag.

The bug-trigger documented in the comment is real: `lineToAxisIntercepts` in `graph-utils.ts:88-97` explicitly sorts pt1/pt2 so `pt1.x < pt2.x`. Since `pointsOnAxes.current` is recomputed every drag tick from the new slope/intercept (line 307), the previous fallback path (`lineSection === "lower" ? pt2 : pt1`) physically swapped which end of the line it referred to the moment the slope crossed vertical. That's the teleport.

### Assessment

Root-cause analysis is correct, fix is minimal, and the comment explaining *why* the pivot is captured once is exactly the kind of non-obvious WHY that warrants a comment.

A few things I confirmed:

- The bug primarily showed when the saved model pivot was invalid (NaN — the default `kInfinitePoint` for fresh lines that haven't been dragged yet. After the first drag completes,  saves valid /, so the model-pivot path used by subsequent drags was already stable. The new code makes both paths stable — strict improvement.
- The reset of  is inside the  branch right after the model write, so ordering is fine.
- d3-drag fires  reliably, so the ref gets cleared on every normal drag completion.
- Each  instance has its own ref, so multiple subplots are unaffected.

No unit tests for the React component exist (only model/handler tests), consistent with the rest of the file. Manual testing per the PR's test plan is appropriate. The  label is already applied.

### Issues

No concerns. This is a clean, focused, well-explained fix with a useful comment capturing the non-obvious invariant (/ sort order) that caused the bug.
EOF
)

kswenson (approved):

## PR Review Summary — #2561

**Changes:** 1 file, +22 / -13 lines

### What it does

Fixes CODAP-1298: when rotating a movable line past vertical, the pinned end of the line would teleport and the line would jump. The fix is to capture the rotation pivot once at the start of a drag and reuse it for the duration of that drag.

### The approach

Adds `dragPivotRef` (a `useRef<Point | null>`) at the top of `MovableLineAdornment`. Inside `handleRotation`, the pivot is computed and stored in the ref the first time per drag (when the ref is null), and reused on every subsequent drag tick. The ref is cleared on `isFinished` after `applyModelChange` runs (so the next drag sees the freshly-saved pivot from the model).

The pivot is stored as a plain `Point` POJO (`{x: ..., y: ...}`) rather than as a reference to the MST `PointModel` instance. This is the right call — capturing a snapshot value isolates the cached pivot from any subsequent model mutations during the drag.

The bug-trigger documented in the comment is real: `lineToAxisIntercepts` in `graph-utils.ts:88-97` explicitly sorts pt1/pt2 so `pt1.x < pt2.x`. Since `pointsOnAxes.current` is recomputed every drag tick from the new slope/intercept (line 307), the previous fallback path (`lineSection === "lower" ? pt2 : pt1`) physically swapped which end of the line it referred to the moment the slope crossed vertical. That's the teleport.

### Assessment

Root-cause analysis is correct, fix is minimal, and the comment explaining *why* the pivot is captured once is exactly the kind of non-obvious WHY that warrants a comment.

A few things I confirmed:

- The bug primarily showed when the saved model pivot was invalid (NaN — the default `kInfinitePoint` for fresh lines that haven't been dragged yet). After the first drag completes, `setLine` saves valid `pivot1`/`pivot2`, so the model-pivot path used by subsequent drags was already stable. The new code makes both paths stable — strict improvement.
- The reset of `dragPivotRef.current = null` is inside the `applyModelChange` branch right after the model write, so ordering is fine.
- d3-drag fires `on("end")` reliably, so the ref gets cleared on every normal drag completion.
- Each `MovableLineAdornment` instance has its own ref, so multiple subplots are unaffected.

No unit tests for the React component exist (only model/handler tests), consistent with the rest of the file. Manual testing per the PR's test plan is appropriate. The `run regression` label is already applied.

### Issues

No concerns. This is a clean, focused, well-explained fix with a useful comment capturing the non-obvious invariant (`pt1`/`pt2` sort order) that caused the bug.

PR #2570, opened 2026-05-09
CODAP-1308: don't dirty the document on leaflet→model sync

bfinzer:
## Summary

Fixes CODAP-1308: V2 documents were reported as "Unsaved" on open whenever they contained a map tile. Leaflet emits small center/zoom adjustments after the map renders (floating-point precision around the projection), and `MapContentModel.syncCenterAndZoomFromMapWithoutUndo` was using `withoutUndo()` to skip undo but not `{ noDirty: true }` to skip dirtying — so the round-trip sync would bump `revisionId` and mark the doc dirty.

The image-vs-no-image variant noted in the bug rejection was incidental: the image affected surrounding layout enough that Leaflet rounded to different values, producing (or not) the sync patch.

## Changes

- `MapContentModel.syncCenterAndZoomFromMapWithoutUndo`: pass `{ noDirty: true }` to `withoutUndo()`. The action still records in history for traceability but does not dirty the document.
- New unit test that wires a real `TreeManager`/`TreeMonitor` under a map tile, drives the sync with a mock Leaflet returning non-default center/zoom, and asserts `revisionId` is unchanged. Verified the test fails on the unfixed code.

## Test plan

- [ ] CI passes (lint, build, unit tests)
- [ ] Open "Birgit Animated V2.codap" (attached to CODAP-1308): expect "Saved", not "Unsaved"
- [ ] Open another V2 doc with a map but no image: expect "Saved"
- [ ] Open a V3 doc with a map: still "Saved" on open
- [ ] Pan/zoom the map: doc transitions to "Unsaved" (real user changes still dirty)
- [ ] Undo/redo of map pan/zoom still works as before

### DIFF
--- a/v3/src/components/map/models/map-content-model.test.ts
+++ b/v3/src/components/map/models/map-content-model.test.ts
@@ -1,10 +1,12 @@
 import { Map as LeafletMap } from "leaflet"
 import { DataSet } from "../../../models/data/data-set"
+import { createCodapDocument } from "../../../models/codap/create-codap-document"
 import { DocumentContentModel, IDocumentContentModel } from "../../../models/document/document-content"
 import { FreeTileRow } from "../../../models/document/free-tile-row"
 import {
   ISharedModelDocumentManager, SharedModelDocumentManager
 } from "../../../models/document/shared-model-document-manager"
+import { TreeManagerType } from "../../../models/history/tree-manager"
 import { SharedDataSet } from "../../../models/shared/shared-data-set"
 import { DataSetMetadata } from "../../../models/shared/data-set-metadata"
 import { kMapTileType } from "../map-defs"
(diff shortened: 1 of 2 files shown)


kswenson (approved):

## PR Review Summary

**Changes:** 2 files, +55 / -1 lines

### What it does

Fixes CODAP-1308: V2 documents containing a map tile were reported as "Unsaved" immediately on open. After Leaflet renders, it emits small center/zoom adjustments (floating-point precision around the projection) which are mirrored back into the model via the leaflet→model sync reaction. That action used `withoutUndo()` (keeping it off the undo stack) but did not pass `{ noDirty: true }`, so the resulting history entry still bumped `treeManager.revisionId` and the dirty-state reaction in `app-state.ts` flipped the doc to "Unsaved".

### The approach

One-line fix: `syncCenterAndZoomFromMapWithoutUndo` now calls `withoutUndo({ noDirty: true })` (map-content-model.ts:232). The action still records in history (useful for traceability) but no longer bumps `revisionId` (tree-manager.ts:183 — `if (!entry.noDirty) self.revisionId = entry.id`). This is consistent with the other two `withoutUndo({ noDirty: true })` sites already in this same file (`setLeafletMap` at L461 and `setHasBeenInitialized` at L467) and with the `noDirty` pattern used widely elsewhere (attribute.ts, data-set.ts, slider, plotted-function adornment, etc.).

The test (map-content-model.test.ts:362-409) wires up a real `TreeManager`/`TreeMonitor` via `createCodapDocument()` (the file's existing `beforeEach` uses a bare `DocumentContentModel` with no history tracking — too lean to exercise this code path), enables monitoring, captures the `revisionId` baseline *after* `setLeafletMap` (defensive), drives the sync with non-default center/zoom, then asserts both that the model state was updated and that `revisionId` is unchanged. PR description says the test was verified to fail on the unfixed code — confirmed locally: reverting just the one-line change makes the new test fail with the expected revisionId mismatch, then re-applying the fix makes it pass again.

### Assessment

This is a clean, narrowly-scoped fix that addresses a real user-visible regression and complements the prior CODAP-1308 work in #2562 (which cleared the post-load dirty baseline; this one prevents a specific dirty source after that baseline). The fix doesn't change any user-facing behavior other than the dirty flag — real user pan/zoom goes through the other branch of the reaction (the `applyModelChange` path with undo strings, L292-300) which is untouched, so undo/redo and the user-edit dirty transition still work as before.

The test is well-isolated, uses real dirty-tracking machinery rather than mocks, and the comments accurately explain what's being verified. `npm test`, `npm run build:tsc`, and `npm run lint` all pass on the changed files.

### Issues

No concerns. This is a tight, well-tested fix.

### Follow-up

While reviewing, we discussed the *root* cause of the round-trip: Leaflet emits sub-pixel center/zoom adjustments after rendering, which the leaflet→model sync action then writes back into the model — creating a no-op history entry every time, even though (with this PR) it no longer dirties the doc. The fix here addresses the consequence (don't dirty); the underlying clutter could also be addressed by thresholding the reaction call site so the action doesn't run for sub-pixel drift (using `map.latLngToContainerPoint` to measure pixel deltas at the current zoom). Filed as **CODAP-1311** for follow-up — low priority since it has no user-visible consequence once this PR lands.

PR #2574, opened 2026-05-14
CODAP-1320: render movable line on date-time x-axis

bfinzer:
## Summary
- Movable line now appears when added to a scatterplot with a date-time x-axis. Previously the model gated initial-line creation on `xAndYAreNumeric`, which excluded date axes, so nothing rendered except an empty (yellow-backgrounded) equation `<p>`.
- Aligns the gate with the `isAnyNumericAxisModel` check already used by movable-point and movable-value adornments.
- Removed the orphan `xAndYAreNumeric` view; movable-line was its only caller.

Fixes [CODAP-1320](https://concord-consortium.atlassian.net/browse/CODAP-1320).

## Dependency
The slope-only equation rendering on date-time x-axes lands separately in [#2572 (CODAP-1314)](https://github.com/concord-consortium/codap/pull/2572). If this PR merges first, the line will draw correctly but the equation will briefly show the meaningless `y = slope·x + intercept` form until #2572 also lands. No file overlap between the two PRs — merges in either order are clean.

## Test plan
- [x] `npm test -- movable-line graph-data-configuration` passes (29 tests).
- [x] `npm run build:tsc` clean.
- [x] Lint clean.
- [x] Verified in browser with the `Movable Line with Date.codap` attachment — line draws.
- [ ] CI lint, type-check, and Cypress pass.

[CODAP-1320]: https://concord-consortium.atlassian.net/browse/CODAP-1320?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/v3/src/components/graph/adornments/movable-line/movable-line-adornment-model.ts
+++ b/v3/src/components/graph/adornments/movable-line/movable-line-adornment-model.ts
@@ -1,6 +1,7 @@
 import { Instance, SnapshotIn, types } from "mobx-state-tree"
 import { JsonNumber } from "../../../../utilities/json-number"
 import { IAxisModel } from "../../../axis/models/axis-model"
+import { isAnyNumericAxisModel } from "../../../axis/models/numeric-axis-models"
 import { Point } from "../../../data-display/data-display-types"
 import { migrateInstanceKeyMap, stringToCellKey } from "../../utilities/cell-key-utils"
 import { computeSlopeAndIntercept } from "../../utilities/graph-utils"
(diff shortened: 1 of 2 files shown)


kswenson (approved):
👍 LGTM -- the following review was developed in conjunction with Claude Code 🤖

## What it does

Allows the movable-line adornment to render on a scatter plot whose x-axis is date-time. Previously the model gated initial-line creation on `dataConfig.xAndYAreNumeric`, which required both `attrTypes` to be exactly `"numeric"`. A date attribute has type `"date"`, so the gate was false on a date-time scatterplot — the lines map stayed empty, the refresh autorun bailed early, and only the (empty, yellow-backgrounded) equation `<p>` appeared.

## The approach

In `movable-line-adornment-model.ts:104`, the gate is replaced with `isAnyNumericAxisModel(xAxis) && isAnyNumericAxisModel(yAxis)`. This is the axis-level check that also accepts count, percent, and date axes. `computeSlopeAndIntercept` (`graph-utils.ts:509`) already uses this same predicate. The orphan `xAndYAreNumeric` view is removed from `graph-data-configuration-model.ts` (verified: no remaining callers). Equation rendering for the slope-only form on date-time axes lands separately in PR #2572 (CODAP-1314), already on `main`.

## Assessment

Correct, minimal, well-scoped. The choice of `isAnyNumericAxisModel` aligns with how the other movable adornments and shared utilities reason about axes. Movable-line is only registered for `scatterPlot`, so the quirk that `isAnyNumericAxisModel` also returns true for `qualitative` doesn't apply here. All CI passes: build, lint, Jest (29 tests), full Cypress regression (368 tests). Copilot found no issues.

### Optional follow-up (non-blocking)

A small unit test exercising `updateCategories` with a date axis would prevent silent reintroduction of the old gate. Author's call.

PR #2581, opened 2026-05-19
CODAP-1340: fix 2x image scaling for map Open in Draw Tool

bfinzer:
## Summary
- Fixes the ~2x image scaling that the Draw Tool would show when launched from a map (passes pixelRatio=1 through compositeMapPng to html-to-image, and forwards the map's bounding rect as the dimensions hint to openInDrawTool).
- Leaves the "Open in Draw Tool" menu item disabled for maps: the Draw Tool plugin has severe issues in Safari that are out of scope here. The handler is kept wired up so the feature can be re-enabled by removing isDisabled once Safari support is sorted.

Fixes CODAP-1340. Depends on #2580 (CODAP-1318); targets the CODAP-1318 branch and will retarget to `main` after that PR merges.

## Test plan
- [ ] Temporarily remove `isDisabled` from the menu item, open the Roller Coasters example, make a map with points + polygons + a point legend, choose "Copy as Image" — the Draw Tool tile matches the map's size and the points are present.
- [ ] Verify the same with multi-layer, heatmap, polygons-only, and points-only maps.
- [ ] "Export PNG Image" (file download path) still produces a full-resolution image on retina displays.

### DIFF
--- a/v3/src/components/map/components/inspector-panel/save-image-menu-list.tsx
+++ b/v3/src/components/map/components/inspector-panel/save-image-menu-list.tsx
@@ -111,6 +111,10 @@ export const SaveImageMenuList = ({tile}: IProps) => {
     return undefined
   }
 
+  // The menu item below is disabled because the Draw Tool plugin itself has severe issues
+  // in Safari (CODAP-1340 covered the scaling fix; the underlying Safari blocker remains).
+  // The handler is kept wired up so the feature can be re-enabled by removing `isDisabled`
+  // once Draw Tool works in all supported browsers.
   const handleOpenInDrawTool = async () => {
     if (!tile || !mapModel?.renderState) return
     const { displayElement } = mapModel.renderState
@@ -156,8 +160,8 @@ export const SaveImageMenuList = ({tile}: IProps) => {
 
   return (
     <InspectorMenuContent data-testid="save-image-menu-list">
-      <MenuItem data-testid="open-in-draw-tool" onAction={handleOpenInDrawTool}>
-        {t("DG.DataDisplayMenu.copyAsImage")}
+      <MenuItem data-testid="open-in-draw-tool" onAction={handleOpenInDrawTool} isDisabled={true}>
+        {t("DG.DataDisplayMenu.copyAsImage")}<span aria-hidden="true"> 🚧</span>
       </MenuItem>
       <MenuItem data-testid="export-png-image" onAction={handleExportPNG}>
         {t("DG.DataDisplayMenu.exportPngImage")}
--- a/v3/src/components/map/components/inspector-panel/save-image-menu-list.tsx
+++ b/v3/src/components/map/components/inspector-panel/save-image-menu-list.tsx
@@ -29,9 +29,9 @@ function loadImage(src: string): Promise<HTMLImageElement> {
 // its output. Composite each visible WebGL point layer's static snapshot on top of the
 // html-to-image base (basemap, polygons, connecting-lines SVG, heatmap).
 async function compositeMapPng(
-  displayElement: HTMLElement, rendererArray: PointRendererArray
+  displayElement: HTMLElement, rendererArray: PointRendererArray, pixelRatio?: number
 ): Promise<string> {
-  const baseDataUri = await toPng(displayElement)
+  const baseDataUri = await toPng(displayElement, { pixelRatio })
   const baseImage = await loadImage(baseDataUri)
 
   const canvas = document.createElement("canvas")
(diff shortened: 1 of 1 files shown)


kswenson (changes requested):
👍 Bug-fix looks good. Re-enabling draw tool export requires further consideration. See details below in review developed in conjunction with Claude Code 🤖.

The bug-fix half of this PR is well-done: the `pixelRatio=1` and dimensions-hint changes mirror the graph's pattern in `camera-menu-list.tsx` cleanly, the default file-download path is preserved at full retina resolution, and the change is appropriately scoped. Happy to approve those changes.

I'd like to discuss the re-enable separately before approving it. When this work began, two questions were on the table:

(a) Was map PNG export itself sufficiently broken that the whole "save image" submenu should be disabled on the map, alongside the draw-tool entry?

(b) The draw-tool path specifically was already disabled by CODAP-1303.

#2580 + this PR resolve (a) — the map's PNG export pipeline is in good shape now (points included, attribute background not black, scaling correct), so we don't need to disable the submenu.

They don't resolve (b). The draw-tool-specific bugs that prompted the original disable still apply to the map's path:
- [CODAP-1275](https://concord-consortium.atlassian.net/browse/CODAP-1275) — Camera button inside the draw tool doesn't work. Plugin-side; affects every image opened in the draw tool, not just graphs.
- [CODAP-1175](https://concord-consortium.atlassian.net/browse/CODAP-1175) — "Open in Draw Tool" produces a blank tile in Safari. `openInDrawTool` uses the same `setTimeout(500)` + `broadcastMessage` handoff for both graphs and maps, so almost certainly affects maps too.

Both of these currently affect the graph side, so any decision to re-enable on the map ought to be made jointly with whether to keep/disable the graph's draw-tool path — consistency between the two tiles seems worth a team conversation.

Could you revert the `isDisabled={true}` + 🚧-marker removal in `save-image-menu-list.tsx` and the corresponding `handleOpenInDrawTool` refactor for now? The `compositeMapPng(pixelRatio?)` plumbing can stay — it's harmless when unused and ready when we do re-enable. Once the team weighs in (with the graph side considered too), the re-enable can land as a small follow-up.

kswenson (approved):
👍 Reduced scope looks good. The following review was developed in conjunction with Claude Code 🤖.

## PR Review Summary (round 2)

**Changes:** 2 files, +80 / −12 lines on the branch. The actual CODAP-1340 delta is much smaller (~20 net lines in `save-image-menu-list.tsx`) — the rest is duplicate commits from #2580 that already merged to `main`.

Scope has narrowed since round 1: the menu item now stays `isDisabled={true}` (deferring the broader Safari issues), so this PR is now just the scaling/dimensions fix wired up for a future re-enable.

### What it does

Fixes the ~2× image scaling that the Draw Tool would show when launched from a map on retina displays, and sizes the new Draw tile to match the source map's bounding rect instead of falling back to the 400×300 default. Both fixes target the dead handler — the menu item itself remains disabled, with an updated comment explaining the remaining Safari blocker is out of scope.

### The approach

Plumbs an optional `pixelRatio` parameter through `getImageDataUri → compositeMapPng → toPng({ pixelRatio })`. The file-download path (`handleExportPNG`) keeps calling without a pixelRatio so html-to-image continues to use `window.devicePixelRatio` for full-resolution exports. The new `handleOpenInDrawTool` path passes `1` to force logical-pixel scaling.

`handleOpenInDrawTool` also pulls `displayElement.getBoundingClientRect()` and forwards `{ width, height }` to `openInDrawTool` (which uses it for tile positioning, defaulting to 400×300 when omitted — see `data-display-image-utils.ts:15-18`). This mirrors the graph's pattern in `camera-menu-list.tsx:92-101`.

The CODAP-1303 comment is rewritten to reference CODAP-1340 and the Safari blocker.

### Assessment

The CODAP-1340 fix is small, focused, and mirrors the equivalent graph code path. Logic checks out: when `pixelRatio` is undefined, html-to-image falls back to `devicePixelRatio`; when `1`, it produces a logical-size image. The bounding-rect dimensions feed `openInDrawTool`'s existing optional hint correctly.

### Issues

1. **~~Branch needs a non-trivial rebase before merge.~~ RESOLVED during review.** Rebased via `git rebase --onto origin/main 4bd74acaf` — dropped the duplicate CODAP-1318 commits and replayed the two CODAP-1340 commits cleanly on top of current `main`. Force-pushed; PR is now `MERGEABLE`.

2. **~~Keep the CODAP-1303 reference in the disabled-MenuItem comment.~~ RESOLVED during review.** Restored in commit `2286386ba`.

3. **Non-blocking: could mirror the graph's `getImageSnapshot` shape.** Left for the author — out of scope for this scaling-only PR.

PR #2588, opened 2026-05-21
CODAP-1345: restore case card inspector

bfinzer:
## Summary
- Removes `overflow: auto hidden` from `.codap-case-card`, which was clipping the inspector panel (positioned at `right: -64px` relative to the tile).
- The inner `.case-card.react-data-card` already handles its own overflow, so the outer rule was redundant.

## Background
The inspector was previously rendered as a sibling outside `.codap-component`, so the clipping had no effect. #2507 (CODAP-1164, build 2832) moved the inspector inside the tile so the keyboard tab-trap could see it, exposing this latent bug. Regression range: builds 2805 → 2832.

The case table is unaffected because it scopes `overflow: auto hidden` to the inner `.case-table-content`, not to the outer `.codap-case-table`.

Fixes CODAP-1345

## Test plan
- [ ] Open a document with a case card; confirm the inspector panel is visible to the right of the tile when focused
- [ ] Confirm each inspector button (info, hide/show, ruler, trash) opens
- [ ] Confirm horizontal/vertical scrolling inside the card still works
- [ ] Confirm attribute drag preview inside the card still works
- [ ] Confirm case table inspector is unchanged

### DIFF
--- a/v3/src/components/case-card/case-card.scss
+++ b/v3/src/components/case-card/case-card.scss
@@ -9,7 +9,6 @@
 
 .codap-case-card {
   min-width: 150px;
-  overflow: auto hidden;
   position: relative;
 
   .attribute-drag-overlay {


kswenson (approved):
👍 Looks good! Thx for finding this and fixing it so promptly!

The following review was developed in conjunction with Claude Code 🤖.

## Review

**Changes:** 1 file, +0 / -1 — `v3/src/components/case-card/case-card.scss`

### What it does

Removes `overflow: auto hidden` from `.codap-case-card`, which was clipping the inspector panel — an absolutely-positioned sibling of the tile component, offset to `right: calc(-1 * width)` so it intentionally sticks out past the tile's right edge.

### Assessment

Verified every claim in the description against the code:

- **Inner div handles overflow** — confirmed: `.case-card.react-data-card` (`case-card.scss:4-8`) scrolls vertically and clips horizontally on its own.
- **Case table unaffected** — confirmed: `case-table.scss:30` scopes `overflow: auto hidden` to the inner `.case-table-content`, with a comment explicitly preserving horizontal auto-scroll; the outer `.codap-case-table` has no such rule.
- **Regression origin** — confirmed: #2507 / CODAP-1164 (`e96177ba7`) moved the inspector inside the tile, while `overflow: auto hidden` itself predates it (Sept 2024, #1453) — a latent bug newly exposed, matching the stated builds 2805 → 2832 range.

Minimal, well-targeted, and the diff is exactly the one line it should be. No accompanying automated test, which is appropriate: the bug is visibility-without-scrolling, which Cypress can't meaningfully assert (it auto-scrolls clipped elements into view — also why this slipped past CI), and existing inspector functional tests in `case-card.spec.ts:373-492` already cover behavior.

### One minor nit (no change needed)

The description calls the outer rule "redundant" because the inner div handles overflow. Strictly, the two rules govern *opposite* axes — the removed outer was `overflow-x: auto; overflow-y: hidden`, while the inner is `overflow-x: hidden; overflow-y: auto`. The outer's `overflow-x: auto` never produced useful card scrolling, though (the inner div fills the outer exactly and percentage-width resizable columns always fit), so its only practical effect was clipping the inspector. Removal is safe — just a precision note.

PR #2591, opened 2026-05-22
CODAP-1348: recompute formulas for datasets re-added to a document

kswenson:
## Summary

When a Story Builder "moment" is restored, the document handler applies the restored state with `applySnapshot()`. If the moment contains a dataset that isn't currently in the document, `applySnapshot` creates a new `DataSet` instance — but nothing registered it with the `FormulaManager`. `addDataSet()` was only called at document creation, by the Data Interactive `create dataContext` handler, and by the case table tool-shelf button — never on the snapshot-application path.

An unregistered dataset is invisible to the formula adapters, so its formula attributes are never recomputed. Because formula-computed values are intentionally not serialized, formula columns such as NOAA Weather's `when` came back empty after a moment restore. Undo of a dataset deletion has the same gap, leaving the formula stale.

## Changes

- `createDocumentModel` installs a reaction that keeps the `FormulaManager`'s dataset registry in sync with the document's shared datasets, replacing the one-time registration. Datasets re-added by `applySnapshot` (moment restore) or by undo are registered and their formulas recomputed.
- `FormulaManager.addDataSet` is made idempotent so the reaction and the existing explicit callers can't double-register or leak disposers.
- New tests: `create-document-model.test.ts` (formula recompute for a dataset added after document creation, and after a deletion is undone); `v2-document-round-trip.test.ts` (raw attribute values survive a v2 save/import round-trip).

## Notes

On the webpack dev server, restoring a dataset can additionally drop raw (non-formula) values — a separate, dev-build-only artifact (the undo/snapshot machinery records MST snapshots without `prepareSnapshot()`, and `Attribute.values` is `undefined` in development). It does not occur in a production build and is out of scope for this PR; it will be tracked as a separate lower-priority issue ([CODAP-1350](https://codap3.concord.org/branch/sync-formula-datasets/)).

Fixes CODAP-1348

[CODAP-1350]: https://concord-consortium.atlassian.net/browse/CODAP-1350?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/v3/src/models/document/create-document-model.test.ts
+++ b/v3/src/models/document/create-document-model.test.ts
@@ -1,0 +1,119 @@
+import { when } from "mobx"
+import { Instance } from "mobx-state-tree"
+import { createCodapDocument } from "../codap/create-codap-document"
+import { createDataSet } from "../data/data-set-conversion"
+import { AttributeFormulaAdapter } from "../formula/attribute-formula-adapter"
+import { FormulaManager } from "../formula/formula-manager"
+import { TreeManager } from "../history/tree-manager"
+import { kSharedDataSetType, SharedDataSet } from "../shared/shared-data-set"
+import { getFormulaManager, getSharedModelManager } from "../tiles/tile-environment"
+
+// register the attribute formula adapter so createFormulaAdapters() installs it
+// (in the app this happens via app.tsx)
+beforeAll(() => {
+  AttributeFormulaAdapter.register()
+})
+
+// safety net: restore any console spy even if a test fails before restoring it itself
+afterEach(() => {
+  jest.restoreAllMocks()
+})
+
+// CODAP-1348: The FormulaManager only computes formulas for datasets in its `dataSets`
+// registry. Datasets can be added to a document after it is created by paths that bypass
+// the explicit addDataSet() calls -- notably applySnapshot() when a Story Builder "moment"
+// is restored, and undo of a dataset deletion. createDocumentModel installs a reaction that
+// keeps the registry in sync with the document's shared datasets, so formula attributes in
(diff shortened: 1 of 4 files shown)


bfinzer (changes requested):
@kswenson Though the changes look good when I test the branch it doesn't compute the values for the `when` attribute and the non-computed values are also not restored.
[Here](http://localhost:8080/#shared=https%3A%2F%2Fcfm-shared.concord.org%2FofmTHIJq6w4xHFgBr9LM%2Ffile.json) is a shared document with which you can observe the problem.

kswenson:
> @kswenson Though the changes look good when I test the branch it doesn't compute the values for the `when` attribute and the non-computed values are also not restored. [Here](http://localhost:8080/#shared=https%3A%2F%2Fcfm-shared.concord.org%2FofmTHIJq6w4xHFgBr9LM%2Ffile.json) is a shared document with which you can observe the problem.

It looks like you were testing on localhost. The fix doesn't work on localhost, as is mentioned in the Notes section. We deferred fixing the dev server side of it to [CODAP-1350](https://concord-consortium.atlassian.net/browse/CODAP-1350). Run your test on the deployed branch: https://codap3.concord.org/branch/sync-formula-datasets/.

[CODAP-1350]: https://concord-consortium.atlassian.net/browse/CODAP-1350?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

bfinzer:
Sorry I missed the note. Approved!

PR #2599, opened 2026-05-25
CODAP-1299: White space on right side of table should allow unselect

kswenson:
## Summary
- Lifts `useWhiteSpaceClick` from `collection-table.tsx` up to `case-table.tsx` so a single per-tile hook instance owns the focus-state tracking, then threads `handleWhiteSpaceClick` down to each `<CollectionTable>` as a new `onWhiteSpaceClick` prop. Pure refactor — drops the unused `gridRef` prop from the hook signature in the process.
- Adds an `onClick` handler on `.case-table-content` (guarded by `event.target === contentRef.current` plus `isTileSelected()`) so clicks landing on the bare flex-container area to the right of the rightmost collection clear the case selection, matching the existing left-spacer and below-rows behavior.
- The `isTileSelected()` guard keeps the new path symmetric with the in-grid background-click handler in `collection-table.tsx`. Also restores a blank line between import groups in `collection-table.tsx` that was lost when the local `useWhiteSpaceClick` import was removed in the refactor commit.

Fixes CODAP-1299

## Test plan
- [x] \`npm run build:tsc\` — clean
- [x] \`npm run lint\` — clean
- [x] \`npm test -- case-table\` — 21 passed, 1 pre-existing skip
- [x] Browser: small 5-row dataset in a wide tile, click right whitespace with a case selected → cases deselect
- [x] Browser: regression sweep — left-spacer, below-rows-in-grid, cell click (no spurious deselect), first-click-focuses-without-deselecting, second-click-deselects — all behave correctly
- [ ] CI to confirm lint/build/limited Cypress
- [ ] After CI passes, apply \`run regression\` label for full Cypress suite

### DIFF
--- a/v3/src/components/case-table/case-table.tsx
+++ b/v3/src/components/case-table/case-table.tsx
@@ -24,6 +24,7 @@ import { CollectionTable } from "./collection-table"
 import { CaseTableAnnounceContext } from "./use-case-table-announce"
 import { useCaseTableModel } from "./use-case-table-model"
 import { useSyncScrolling } from "./use-sync-scrolling"
+import { useWhiteSpaceClick } from "./use-white-space-click"
 
 import "./case-table.scss"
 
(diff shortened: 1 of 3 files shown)


bfinzer (approved):
👍🏻LGTM

PR #2612, opened 2026-06-02
CODAP-1373: fix least squares line disappearing with numeric categorical legend

kswenson:
## Summary
- `getCaseValues` in the LSRL adornment compared a legend value from
  `dataset.getValue` (which returns a number for numeric attributes) against
  the category label from `categoryArrayForAttrRole` (always a string). When
  the legend attribute was numeric used as categorical, no cases ever matched
  and the regression collapsed.
- Switched the legend lookup to `getStrValue` so both sides of the comparison
  are strings, matching how the category labels are produced.
- Added a unit test that drives `getCaseValues` with a numeric legend attribute.

Fixes CODAP-1373.

## Test plan
- [x] `npm test -- lsrl-adornment-model` (new test fails before fix, passes after)
- [x] `npm test -- src/components/graph/adornments` (244/244 pass)
- [x] `npm run build:tsc`
- [x] `npm run lint`
- [x] Manual: Four Seals → Distance vs Depth → Least Squares Line → drag
  `animal_id` to legend → confirm one line per animal id

### DIFF
--- a/v3/src/components/graph/adornments/lsrl/lsrl-adornment-model.test.ts
+++ b/v3/src/components/graph/adornments/lsrl/lsrl-adornment-model.test.ts
@@ -262,6 +262,45 @@ describe("LSRLAdornmentModel", () => {
       expect(lSRL.labels.has(instanceKey3)).toBe(false)
     })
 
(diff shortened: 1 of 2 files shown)


bfinzer (approved):
👍🏻LGTM

PR #2617, opened 2026-06-04
CODAP-1323: document re-mirroring the v2→v3 redirect clone after prod config drift

dougmartin:
## Summary

Adds **Recipe 6 — "Prod config changed: re-mirror the clone"** to the v2→v3 redirect
PLAYBOOK. The clone distribution (`codap2to3.concord.org`) is a one-time snapshot of the
production CloudFront distribution; when prod is edited afterward, the clone goes stale.
Neither `clone-distribution.sh` (skips create when the clone already exists) nor
`modify-clone.sh` (transforms the clone's own config, not prod's) pulls those upstream
changes in — and `verify-clone.sh` only *detects* drift, it doesn't fix it. This recipe
closes that gap.

The recipe covers:
- Scoping drift with `verify-clone.sh` and the captured `artifacts/*-norm.json` configs.
- Checking the blast radius (WordPress default behavior vs. app-serving behaviors).
- Choosing between a surgical in-place update (isolated drift) and a full clone recreate
  (broad/structural drift).
- Whether the change also needs to be folded into `modify-clone.sh` / `expected-diff.md`.

The worked example is the recent modernization of prod's WordPress default cache behavior
(legacy `ForwardedValues` + TTLs → `codap_concord_org_cache` cache policy +
`Managed-AllViewer` origin-request policy), which was mirrored onto the clone and
re-validated (`verify-clone.sh` back to zero unexpected diffs).

Docs-only — no code changes.

Refs CODAP-1323.

### DIFF
--- a/devops/cloudfront-functions/v2-v3-redirect/PLAYBOOK.md
+++ b/devops/cloudfront-functions/v2-v3-redirect/PLAYBOOK.md
@@ -149,6 +149,91 @@ firing.
 
 ---
 
+## Recipe 6 -- Prod config changed: re-mirror the clone
+
+The clone (`$CLONE_DIST_ID`) was a one-time snapshot of prod (`$PROD_DIST_ID`,
+`E3H9X49AG3GYSO`) taken by `clone-distribution.sh`. When someone edits the LIVE prod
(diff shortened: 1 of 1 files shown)


kswenson (approved):
👍 LGTM

PR #2623, opened 2026-06-05
CODAP-1399: wrap destination key in backticks in lookupByKey join formula

bfinzer:
## Summary
- When joining datasets by dragging a source key onto a destination
  attribute whose name contains a space (or other character that isn't
  valid in a bare formula identifier), the generated `lookupByKey`
  formula left the destination key bare and the parser rejected it.
- Wrap the destination key in backticks when needed, escaping any
  embedded backticks or backslashes via `escapeBacktickString`.
- Added three unit tests covering the wrap-needed, escape-needed, and
  safe-name paths.

Fixes CODAP-1399.

A follow-up bug — CODAP-1400 — was filed for the related issue that
the first three (string-constant) arguments aren't escaped for embedded
double-quote characters.

## Test plan
- [ ] `npm test -- join-datasets` passes (8/8)
- [ ] In the Demo Join Bug document attached to CODAP-1399, drag `Key`
      from **Source** onto `Attribute Name` in **Destination**; the new
      `value` attribute populates from Source instead of erroring.
- [ ] Drag a key onto a normally-named destination attribute and
      confirm the generated formula is unchanged (no spurious backticks).

### DIFF
--- a/v3/src/models/data/join-datasets.test.ts
+++ b/v3/src/models/data/join-datasets.test.ts
@@ -93,6 +93,83 @@ describe("joinSourceToDestCollection", () => {
     expect(dataAttr?.formula?.display).toBe('lookupByKey("Source", "Data", "Key", Key)')
   })
 
+  it("wraps the destination key attribute in backticks when its name needs escaping", () => {
+    const sourceData = createDataSetWithHistory("Source")
+    sourceData.addAttribute({ name: "Key" })
+    sourceData.addAttribute({ name: "value" })
+
+    const destData = createDataSetWithHistory("Destination")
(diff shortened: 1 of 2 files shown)


kswenson (approved):
👍 Looks good — clean, focused fix. Wrapping the 4th `lookupByKey` argument in backticks via `asFormulaIdentifier` is the right call, and using `safeSymbolName` as the wrap-detection signal keeps it consistent with the rest of the formula system. Nice reuse of `escapeBacktickString` (correct inverse of the parser's `unescapeBacktickString`), and the three tests cover the meaningful paths. Good call deferring the string-constant double-quote escaping to CODAP-1400.

PR #2630, opened 2026-06-09
CODAP-1385: numeric legend transparency interpolates across quintiles

bfinzer:
## Summary
- `interpolateColors` now carries the alpha channel through, so the five-stop gradient used by numeric legends (graph and map) fades transparency smoothly between the low and high colors instead of clamping every interior stop to fully opaque. Closes the original repro from CODAP-1385.
- `MapPolygonLayer` now observes `dataConfiguration.choroplethColors` and refreshes polygon styles when the gradient changes, so polygons refill live while the user drags the color/transparency sliders (previously only the legend bar updated; polygons lagged until a pan/zoom/resize).
- Added unit tests for `interpolateColors` and `getChoroplethColors` covering both the opaque and alpha-fading cases.

Fixes CODAP-1385.

## Test plan
- [ ] Open `Problem controlling numeric legend transparency.codap` (attached to CODAP-1385).
- [ ] Map → Layers → click rightmost States legend color → More → drag the transparency slider. All five legend quintiles should fade together, and the map polygons should refill in sync.
- [ ] Repeat on the graph's numeric legend — the gradient should fade smoothly across all five stops.
- [ ] Verify the existing point/heatmap dynamic update and the graph's dynamic legend update still work.

### DIFF
--- a/v3/src/components/map/components/map-polygon-layer.tsx
+++ b/v3/src/components/map/components/map-polygon-layer.tsx
@@ -246,6 +246,18 @@ export const MapPolygonLayer = function MapPolygonLayer(props: {
     )
   }, [refreshPolygonStyles, mapLayerModel])
 
+  // The numeric-legend gradient (low/high colors set via the Layers palette) is observed by the
+  // legend itself but not by the polygon fills, so dragging the color/transparency slider would
+  // update the legend bar without refilling the polygons until a pan/zoom/resize forced a redraw.
+  useEffect(function respondToLegendColorRangeChange() {
+    return mstReaction(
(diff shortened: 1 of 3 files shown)


kswenson (approved):
👍 Looks good — clean, focused fix. The alpha interpolation is correct (colord's `toRgb()` already carries alpha through; the old code was just discarding it), the new `choroplethColors` reaction is the right missing dependency and follows the existing reaction patterns in the file, and the unit tests cover both the opaque and alpha-fading cases. Verified the new tests pass locally and CI is green. Approving.

PR #2637, opened 2026-06-15
CODAP-1415: replace formula lookbehind regex with linear scanner

kswenson:
## Summary

Fixes CODAP-1415.

On Safari < 16.4 the formula engine crashes on load. The regex literal in
`makeDisplayNamesSafe` (`canonicalization-utils.ts`) used a negative
lookbehind `(?<!\\)`, which JavaScriptCore did not support until Safari
16.4. The module throws a `SyntaxError` on evaluation — a hard failure,
not a degraded formula. The Safari 15.6 floor includes iPad Air 2 / mini 4
and 2013 iMacs, which are system-WebKit-capped and common in classrooms.

This replaces the regex with a small linear scanner that tracks escape
state. It is not just a compat fix:

- **Fixes the crash** — no lookbehind, so JavaScriptCore no longer rejects
  the module.
- **Faster** — a single linear pass with no regex backtracking.
- **More correct** — an escaped backslash (`\\`) followed by a backtick now
  correctly opens a delimited name; the one-char lookbehind misread that
  backtick as escaped and left the name unprocessed.

This was the only regex lookbehind in `src`, and lookbehind was the only
ES2018+ regex feature Safari 15.6 lacks, so this closes the entire
regex-syntax exposure on the floor.

## Tests

Added cases to the `makeDisplayNamesSafe` suite, pinning: a name at string
start, adjacent names, an ignored escaped backtick, an unterminated
backtick left untouched, and the `\\`+backtick improvement. Full formula
suite (467 tests) passes; lint clean.

### DIFF
--- a/v3/src/models/formula/utils/canonicalization-utils.test.ts
+++ b/v3/src/models/formula/utils/canonicalization-utils.test.ts
@@ -60,6 +60,12 @@ describe("makeDisplayNamesSafe", () => {
   it("leaves an unterminated backtick untouched", () => {
     expect(makeDisplayNamesSafe("a + `unterminated")).toEqual("a + `unterminated")
   })
+  it("leaves an empty backtick pair untouched", () => {
+    expect(makeDisplayNamesSafe("a + `` + b")).toEqual("a + `` + b")
+  })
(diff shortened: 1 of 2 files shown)


scytacki (approved):
looks good

PR #2639, opened 2026-06-18
CODAP-1424: re-establish plugin connection for interactives added by drag/drop or URL import

kswenson:
## Summary

Fixes an interactive added to CODAP by **drag/drop or URL import** (e.g. a Lab
interactive from lab.concord.org) failing to detect it's embedded in CODAP, so
it never showed its "send data to CODAP" controls. The same interactive worked
via share link, Google Drive reload, or "make a copy".

## Root cause

A dropped/imported URL is created as a *generic* web view (`subType` undefined,
`isPluginCandidate` false), so `needsLocaleReload` is false and the iframe first
loads the raw URL. When the iframe-phone handshake succeeds,
`setPluginIsCommunicating()` promotes the tile to a plugin
(`setSubType("plugin")`), flipping `needsLocaleReload` to true. That makes
`web-view.tsx` recompute the iframe `src` to append `?lang=…&locale=…`, forcing
a second iframe load. Because the model `url` (a dependency of the connection
effect) is unchanged, the effect did not re-run, so CODAP never re-sent
`codap-present` to the reloaded iframe — the surviving load never learned it was
in CODAP.

Paths that classify as a plugin before first render don't hit this: the Plugins
menu sets `setPluginCandidate(true)`, and saved docs / "make a copy" carry
`subType: "plugin"` in the snapshot, so they load exactly once.

## Fix

Add `needsLocaleReload` to the connection effect's dependency array in
`use-data-interactive-controller.ts`, so the iframe-phone connection is
re-established (and `codap-present` re-sent) when the runtime plugin-promotion
flips it and reloads the iframe. This reuses the re-establish mechanism the
locale-change path already relies on.

## Why fix the reload path instead of forcing a single load?

The obvious alternative is to mark dropped/imported URLs as a plugin candidate up
front (as the Plugins menu does via `setPluginCandidate(true)`), so
`needsLocaleReload` is true from the first render and the iframe loads once with
locale params — no second load. We deliberately did **not** do this:

- Drag/drop and URL import are also how plain **web pages** (and images) are
  added, and there's no way to know at drop time whether a URL is a plugin or an
  ordinary web page.
- Treating every dropped URL as a plugin candidate would append `?lang=…&locale=…`
  to arbitrary web pages. Some sites interpret those params (changing language or
  behavior), it would force those pages to reload on every CODAP locale change
  (losing scroll/form state), and it extends a plugin-only feature to non-plugins
  — diverging from the existing design (and from v2).

Keeping locale params plugin-only and making the already-intended reload robust
is the lower-risk, more faithful fix. (The cost is one extra iframe load for a
dropped interactive at the moment it's recognized as a plugin — the same reload
the locale feature already performs — with no impact on web pages.)

## Testing

- New regression test in `use-data-interactive-controller.test.ts` (verified it
  fails without the added dependency).
- Manually reproduced and verified in the running app with `debug=plugins`: after
  promotion, a second connection + `codap-present` reaches the reloaded iframe and
  the interactive's controls appear.
- `npm run build:tsc`, lint, and the full web-view jest suite pass.

Fixes CODAP-1424.

### DIFF
--- a/v3/src/components/web-view/use-data-interactive-controller.test.ts
+++ b/v3/src/components/web-view/use-data-interactive-controller.test.ts
@@ -1,4 +1,28 @@
-import { extractOrigin } from "./use-data-interactive-controller"
+import { act, renderHook } from "@testing-library/react"
+import iframePhone from "iframe-phone"
+import { DIComponentInfo } from "../../data-interactive/data-interactive-types"
+import { diComponentHandler } from "../../data-interactive/handlers/component-handler"
+import { appState } from "../../models/app-state"
+import { toV3Id } from "../../utilities/codap-utils"
+import { extractOrigin, useDataInteractiveController } from "./use-data-interactive-controller"
+import { IWebViewModel } from "./web-view-model"
+import { kWebViewIdPrefix } from "./web-view-registration"
+
+// Replace iframe-phone with lightweight mocks so the hook doesn't attempt real postMessage
+// handshakes; we only care about how many times a connection (ParentEndpoint) is established.
+jest.mock("iframe-phone", () => ({
+  __esModule: true,
+  default: {
+    ParentEndpoint: jest.fn(() => ({ disconnect: jest.fn() })),
+    IframePhoneRpcEndpoint: jest.fn(() => ({ call: jest.fn(), disconnect: jest.fn() }))
+  }
+}))
+
+jest.mock("../../hooks/use-cfm-context", () => ({ useCfmContext: () => undefined }))
+
+jest.mock("../../data-interactive/data-interactive-request-processor", () => ({
+  setupRequestQueueProcessor: jest.fn(() => jest.fn())
+}))
 
 describe("extractOrigin", () => {
   it("extracts origin from absolute https URLs", () => {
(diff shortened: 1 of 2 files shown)


scytacki (approved):
Looks good to me.
I was concerned that this might also be a problem when the language was changed, but I tested that on main and it worked fine. And I retested that on this branch and it continued to work fine.

kswenson:
> Looks good to me. I was concerned that this might also be a problem when the language was changed, but I tested that on main and it worked fine. And I retested that on this branch and it continued to work fine.

Good thought -- this probably fixes that case.

kswenson:
> Good thought -- this probably fixes that case.

Correction to my earlier reply: on reflection this PR doesn't actually touch the language-change path — that path was already working on `main`, which matches what you found in testing.

A language change for an established plugin is handled by the pre-existing `localeVersion` counter: the locale reaction bumps `localeVersion` (already in the connection effect's dependency array on `main`), which re-establishes the connection and re-sends `codap-present` to the reloaded iframe.

The new `needsLocaleReload` dependency only fires on a *value change*, and `needsLocaleReload` (`isPlugin || isPluginCandidate`) depends on `subType`, not locale — so it doesn't change during a language change. What it catches is the one-time `false → true` flip when a dropped/imported generic web view is *promoted* to a plugin after its handshake. That's not a locale change, so no locale reaction fires and `localeVersion` isn't bumped — which is exactly why that transition was missed on `main` and needed this dependency.

So the two paths are distinct: locale changes → `localeVersion` (already worked); promotion → `needsLocaleReload` (the bug fixed here).

PR #2641, opened 2026-06-18
CODAP-1419: restore "+" on y-axis drop zone and fix Safari drag flicker

bfinzer:
## Summary
- Restores the v2 "+" indicator inside the y-axis multi-attribute (yPlus) drop zone so users can see that dragging a numeric attribute to the top of a scatterplot adds a second y attribute.
- Reshapes the yPlus drop zone — full-width across the top normally, narrowed to the y-axis column when a top axis (categorical split) is already in place — and trims the left-axis drop zone so it doesn't overlap the strip.
- Fixes a cross-tile drop-zone flicker that was most visible in Safari: when dragging from one tile (e.g., a case table) over a graph, case-table droppables would intermittently "win" collision detection at certain cursor positions, causing the yellow drop highlight to flash off. `dndDetectCollision` now uses `document.elementFromPoint` to identify the visually-topmost tile and prefer its prefix.

Fixes CODAP-1419

## Test plan
- [ ] Open Roller Coasters; put Top_Speed on x and Max_Height on y. Drag Drop to the top of the plot — a yellow strip appears with a small yellow "+" over the y-axis column.
- [ ] Same plot, in Safari: drag slowly across the strip — yellow highlight stays steady.
- [ ] Split the plot by adding Type to the top axis. Drag Drop to the top — the yPlus zone is now narrowed to just the y-axis column; the top-axis zone stays plot-bounded.
- [ ] Drag a numeric attribute to the left axis — replace works; left-axis drop zone extends the full plot height when there's a top axis.
- [ ] Sanity-check drag-and-drop interactions in case table and other tiles for any cross-tile regressions from the dndDetectCollision change.

### DIFF
--- a/v3/src/components/graph/components/droppable-add-attribute.tsx
+++ b/v3/src/components/graph/components/droppable-add-attribute.tsx
@@ -16,10 +16,11 @@ interface IAddAttributeProps {
   place: GraphPlace
   plotType: PlotType
   hasRightDropZone?: boolean
+  hasTopAxis?: boolean
   onDrop: (dataSet: IDataSet, attributeId: string) => void
 }
 
-export const DroppableAddAttribute = ({place, hasRightDropZone, onDrop}: IAddAttributeProps) => {
+export const DroppableAddAttribute = ({place, hasRightDropZone, hasTopAxis, onDrop}: IAddAttributeProps) => {
   const layout = useGraphLayoutContext()
   const graphID = useInstanceIdContext(),
     dataConfiguration = useGraphDataConfigurationContext(),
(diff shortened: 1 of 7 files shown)


kswenson (approved):

## PR Review Summary — #2641

**Changes:** 6 files, +120 / -39 lines

### What it does

Restores the v2 "+" indicator inside the y-axis multi-attribute (`yPlus`) drop zone on a scatterplot, so users get a visual cue that dragging a numeric attribute to the top of the plot adds a *second y* attribute. Reshapes that drop zone (full-width strip normally, narrowed to the y-axis column when a top/categorical axis already occupies the top), and trims the adjacent left-axis drop zone so the two don't overlap. Separately, fixes a cross-tile drop-zone flicker (most visible in Safari) where, while dragging from one tile (e.g. a case table) over a graph, the source tile's droppables could intermittently "win" collision detection and flash the yellow highlight off.

Fixes CODAP-1419.

### The approach

**Drop-zone geometry (`graph.tsx`, `droppable-add-attribute.tsx`, `droppable-axis.tsx`, `graph-axis.tsx`):**
- `graph.tsx` now computes the drop-zone flags/insets *before* `renderGraphAxes`, unifying the top overlap into a single `topStripOverlap = max(0, kDropZoneSize + kDropZoneGap - plot.top)`. Because `plot.top` and `left-axis.top` are equal, that one value drives both the plot-drop inset and the new `leftAxisTopInset`. When a top axis pushes the boundary below the strip, overlap is 0 and nothing shrinks. (graph.tsx:431-446)
- `DroppableAddAttribute` splits the old combined `top`/`yPlus` branch: `yPlus` spans the full width incl. the y-axis column (with the "+"), or narrows to just the y-axis column when `hasTopAxis`; `top` stays plot-bounded. The "+" is an inline-sized SVG, `aria-hidden`. (droppable-add-attribute.tsx:67-97)
- `DroppableAxis`/`GraphAxis` gain an optional `topInset` that shrinks the left-axis droppable rect from the top to avoid overlapping the yPlus strip. Default 0 → no behavior change for other axes. (droppable-axis.tsx:30-33)

**Collision flicker (`dnd-detect-collision.ts`):** The existing code picked the prefix of the highest-z-index collision. The subtlety: z-index ordering is already *correct* — the drag source tile (e.g. a case table) is raised to the top z-index when focused at drag start, so it legitimately sorts above the destination it's dragged onto. `pointerWithin` is rect-based, so at cursor positions where the source's droppable *rect* extends past its painted area into the destination, the source's droppables win even though the user sees the destination — the flicker. The fix adds `document.elementFromPoint(x, y)` to find the *actually-painted* topmost tile and try its droppable prefix *first*, with the old first-prefix appended as a fallback. The loop tries candidates in order, so if the elementFromPoint-derived tile has no matching handler it degrades to the prior behavior.

### Assessment

Solid, well-reasoned change. Strengths:
- **No regression risk in the collision fix.** The new `candidatePrefixes` list always appends the original `firstPrefix`, so worst case it behaves exactly as before. In jsdom (tests) `elementFromPoint` returns null → falls straight through to old behavior, so no test breakage.
- `elementFromPoint` is already an established pattern in this codebase (pixi-point-renderer, canvas-point-renderer), so no new perf/portability concern.
- `topInset` defaults to 0 and is only passed for `place === 'left'`, so other axes and tiles are untouched.
- The geometry refactor is clean — collapsing the two overlap computations into one shared `topStripOverlap` removes the prior duplication and the comments explain the invariant (`plot.top === left-axis.top`).
- `tsc` and `eslint` both pass locally on the changed files.

### Changes made during review (pushed to this branch)

- **Reworded the `dnd-detect-collision.ts` comment** that justified the `elementFromPoint` heuristic. The original attributed the flicker to "separate CSS stacking contexts," but verification showed all free tiles are siblings in one parent stacking context and `data-tile-z-index` (read live from the DOM, same element/render as `style.zIndex`) orders them correctly by paint. The real cause is geometric (as described above) — comment now states that.
- **Tokenized the "+" indicator color** — moved the hardcoded `#e6b800` from `graph.scss` into `$drop-zone-plus-color` in `vars.scss`, next to the related `$drop-highlight-color` it contrasts against. SCSS compiles to the identical value; no visual change.

### Minor notes (non-blocking)

1. **No automated test coverage.** +120 lines, 0 test changes. The collision-detection change is genuinely hard to unit-test (needs real layout for `elementFromPoint`), and the geometry is visual — so the all-manual test plan is defensible. Worth a sentence in the PR acknowledging the deliberate gap.
2. **`elementFromPoint` runs on every collision pass during a drag.** Gated behind `collisions.length > 0` and only does a single call per pass, so likely negligible — but it does force style/layout. Acceptable given the same pattern is used in the renderers.

PR #2646, opened 2026-06-22
CODAP-1421: preserve plugin-set axis bounds across data updates

bfinzer:
## Summary
- Plugins can now set xLowerBound/xUpperBound/yLowerBound/yUpperBound/y2LowerBound/y2UpperBound on a graph and have those bounds survive subsequent data additions, matching V2.
- These bounds are now also honored at graph create time, not just on update.
- Fixes a visual flash where the d3 scale briefly widened past the user-set bound before settling back.

Fixes CODAP-1421.

## Test plan
- [ ] Create a graph via the api-tester plugin with xLowerBound: 0; verify x-axis lower bound is 0.
- [ ] Update an existing graph with xLowerBound: 0; add cases via plugin; verify lower bound stays at 0 with no flash.
- [ ] Use the inspector's auto-rescale button on a graph with a plugin-set bound; verify rescale still tightens bounds.
- [ ] Regression tests (`run regression`) pass on CI.

### DIFF
--- a/v3/src/components/axis/axis-domain-utils.test.ts
+++ b/v3/src/components/axis/axis-domain-utils.test.ts
@@ -1,0 +1,41 @@
+import { setNiceDomain } from "./axis-domain-utils"
+import { NumericAxisModel } from "./models/numeric-axis-models"
+
+describe("setNiceDomain", () => {
+  it("does nothing when given empty values", () => {
+    const axis = NumericAxisModel.create({ place: "bottom", min: 0, max: 10 })
+    setNiceDomain([], axis)
+    expect(axis.min).toBe(0)
+    expect(axis.max).toBe(10)
+  })
+
+  it("preserves bounds when data fits within them (CODAP-1421)", () => {
+    const axis = NumericAxisModel.create({ place: "bottom", min: 0, max: 10 })
+    setNiceDomain([1, 5, 9], axis)
+    expect(axis.min).toBe(0)
+    expect(axis.max).toBe(10)
+  })
+
+  it("expands bounds when data extends beyond the current upper bound", () => {
+    const axis = NumericAxisModel.create({ place: "bottom", min: 0, max: 10 })
+    setNiceDomain([1, 15], axis)
+    expect(axis.min).toBe(0)
+    expect(axis.max).toBeGreaterThanOrEqual(15)
+  })
+
+  it("recomputes bounds when allowRangeToShrink is set (autoScale path)", () => {
+    const axis = NumericAxisModel.create({ place: "bottom", min: 0, max: 100 })
+    axis.setAllowRangeToShrink(true)
+    setNiceDomain([3, 7], axis)
+    // Bounds should have tightened around the data.
+    expect(axis.max).toBeLessThan(100)
+  })
+
+  it("recomputes bounds when clampPosMinAtZero option is set", () => {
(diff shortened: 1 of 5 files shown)


kswenson (approved):

## PR Review Summary

**Changes:** 5 files, +100 / −4 lines

### What it does

Plugins can set explicit axis bounds on a graph (`xLowerBound`/`xUpperBound`/`yLowerBound`/`yUpperBound`/`y2LowerBound`/`y2UpperBound`). Previously those bounds were only honored transiently — adding cases via the plugin API caused the axis to re-"nice" itself and widen past the user-set bound. This PR makes plugin-set bounds **(a)** honored at graph *create* time, not just on update, and **(b)** survive subsequent data additions, matching V2's `setDataMinAndMax`. It also removes a visual flash where the d3 scale briefly widened before settling back.

### The approach

Three coordinated production changes plus tests:

1. **`setNiceDomain` (`axis-domain-utils.ts`)** — new `else if (!allowRangeToShrink)` branch: when data already fits inside a current bound, keep that bound instead of widening to the computed nice bound; return early if neither side changes. Mirrors V2.
2. **`use-sub-axis.ts`** — same no-widen logic applied to the d3 `multiScale` domain *before* `setNiceDomain` runs on the model. This is what eliminates the flash.
3. **`graph-component-handler.ts`** — `applyBoundsAtCreate` calls `setMinimum`/`setMaximum` per supplied bound after model sync, so create-time bounds are exact. Persists only because #1 stops the first render from re-widening.
4. **Tests** — new `axis-domain-utils.test.ts` (5 cases) + a create-time-bounds case in `graph-component-handler.test.ts`.

### Assessment

Clean, focused, well-commented, and well-tested. The three production changes are coherent — create-time bounds only persist because both the data-add path and the d3-scale path were taught not to re-widen. Both new unit suites pass locally (axis-domain-utils 5/5, graph-component-handler 2/2), and the full Cypress regression suite is green.

### V2 compatibility verification (checked against `origin/master`)

Both V2-parity claims check out:

- **Data-add preservation** mirrors V2's `cell_linear_axis_model.js` `setDataMinAndMax`, which bails when `iDataMin >= lowerBound && iDataMax <= upperBound`, and V2's `dataDidChange` only rescales when a new case lies outside the current bounds. (Nuance: on a one-sided overflow V2 recomputes *both* bounds from the data extent while this PR preserves the fitting side exactly — arguably more faithful to plugin intent.)
- **Create-time bounds** mirror V2's document round-trip: `graph_controller.js` `createComponentStorage` writes `{x,y,y2}LowerBound`/`UpperBound` and `restoreComponentStorage` applies them via `setLowerAndUpperBounds`. This PR's use of the direct `setMinimum`/`setMaximum` setters (no nice-ing/snapping) aligns with that restore mechanism.

### Changes made during review

- Standardized `setNiceDomain`'s no-widen check on `axisModel.domain` rather than the static `min`/`max` (commit `9a4202d13`). `domain` accounts for dynamic bounds (e.g. during a drag) and matches the reference already used in `use-sub-axis` and in `setDomain`'s own clamp logic; behavior is unchanged when no drag is in progress.

### Issues

No blocking issues. One low-priority, pre-existing note: on a one-sided expansion `setNiceDomain` still calls `setDomain`, whose snap-to-zero logic could nudge a plugin-set lower bound within 1/100 of the (larger) max down to 0. Marginal, matches existing V3 snapping — not worth changing here.

PR #2649, opened 2026-06-23
[Do Not Merge] CODAP-1429: fix stale CategorySet for formula attributes in loaded documents

bfinzer:
This PR is a temporary stop-gap. #2650 is the proper fix.

## Summary

Fixes CODAP-1429. In a copied or saved document, dragging a formula-driven
attribute (e.g. `randomPick("A","B")`) to a graph axis produced points
bunched at the upper-left corner instead of a proper categorical dot chart.

## Root cause

`CategorySet`'s `mstReaction` on `attribute.changeCount` is installed inside
`afterCreate` — an MST action. MobX defers a reaction's *initial* accessor
invocation until the outermost batch ends. When CategorySet creation, the
initial empty `refresh()`, and the formula adapter's `setComputedValues`
all land in the same outer batch (which is the natural pattern during
document load), the deferred first invocation captures the post-mutation
`changeCount` as its baseline. No "change" is detected, the effect never
fires, and the empty pre-formula cache sticks around — producing NaN
positions and the bunched-at-origin symptom.

A fresh document dodges this because the user adds the formula and creates
the graph in separate actions; the CategorySet is constructed only after
the formula's batch has drained, so its first `refresh()` sees populated
`strValues` directly.

This has the same fingerprint as the bug fixed in CODAP-1280, which was
a different way of reaching the same stale-categories state.

## Fix

`fireImmediately: true` makes the effect run on the first accessor
invocation regardless of whether mobx detects a "change" — so the
(potentially stale) cache is invalidated on the first chance, and the
next read recomputes from current `strValues`.

A follow-up cleanup PR will replace the manual cache + reaction with a
pure computed view, which eliminates the race-prone pattern entirely.

## Test plan

- [x] New regression test (`category-set.test.ts`) wraps CategorySet
      creation, initial refresh, and `setComputedValues` in a single
      `runInAction` — fails without the fix, passes with it
- [x] Existing CategorySet tests still pass (8/8)
- [x] Full jest suite passes (3276/3276)
- [x] Lint clean
- [x] tsc clean
- [ ] Manual verification in browser of the original bug repro

### DIFF
--- a/v3/src/models/data/category-set.test.ts
+++ b/v3/src/models/data/category-set.test.ts
@@ -1,3 +1,4 @@
+import { runInAction } from "mobx"
 import { applySnapshot, getSnapshot, types } from "mobx-state-tree"
 import { kellyColors } from "../../utilities/color-utils"
 import { Attribute, IAttribute } from "./attribute"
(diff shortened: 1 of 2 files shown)


kswenson:
Closing -- replaced by #2650.

PR #2655, opened 2026-06-26
CODAP-1434: fix map boundaries missing when data crosses the antimeridian

kswenson:
## Summary

Map boundary (polygon) layers rendered blank whenever the dataset included a region that crosses the antimeridian — most commonly **Alaska**, whose Aleutian Islands span ~172°E to ~−130°W. This was reported against the NASS plugin (boundaries showed in the case table but not the map) and appeared to also affect the Microdata plugin.

## Root cause

The failure is purely in **map viewport fitting**, not in boundary data or formulas — every polygon parses and is added to Leaflet. When the data spans the antimeridian, `computeBoundsFromCoordinates` returns a compact dateline-crossing arc whose center longitude lands in the **+1 world copy** (e.g. ~`[172, 293]`, centered near 232°). The map auto-fit then centers there, but polygon vector features render once at their **raw canonical longitudes** (world copy 0), so the viewport is parked a full world copy east of the polygons → blank map. Point layers avoid this because they already shift each point into the viewport via `shiftLongitudeIntoView`; polygons never shift.

(Datasets without an antimeridian-crossing region — e.g. California alone, or the contiguous states — were unaffected, which matched the original intermittent reports.)

## Fix

- Add `wrapBoundsToCanonicalCenter()` (`map-utils.ts`): shifts fitted bounds by whole 360° world copies so the center longitude lands in `[−180, 180]` — i.e. the world copy where polygons are actually drawn. No-op for already-canonical bounds.
- Apply it at the two map-fit sites: `MapContentModel.rescale()` and the map's initial auto-fit (`use-map-model.ts`).
- Point-layer rendering is untouched.

## Tests

4 unit tests for `wrapBoundsToCanonicalCenter` (canonical no-op, the Alaska `[172, 293]→` canonical case, span preservation, generic >+180° wrap). Full `map-utils` suite passes (30/30); typecheck and lint clean. Manually verified the all-50-states NASS repro now renders.

Fixes CODAP-1434

### DIFF
--- a/v3/src/components/map/hooks/use-map-model.ts
+++ b/v3/src/components/map/hooks/use-map-model.ts
@@ -7,7 +7,7 @@ import { getTileInfo } from "../../../models/document/tile-utils"
 import { useDataDisplayLayout } from "../../data-display/hooks/use-data-display-layout"
 import { kTitleBarHeight } from "../../constants"
 import { kDefaultMapZoomForGeoLocation, kMapBoundsExtensionFactor, kMaxZoomForFitBounds } from "../map-types"
-import { expandLatLngBounds } from "../utilities/map-utils"
+import { expandLatLngBounds, wrapBoundsToCanonicalCenter } from "../utilities/map-utils"
 import { useMapModelContext } from "./use-map-model-context"
 
 export function useMapModel() {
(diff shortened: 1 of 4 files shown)


bfinzer (approved):
👍🏻LGTM

PR #2666, opened 2026-07-22
CODAP-1448: add a feature-flag system for gating in-development features

kswenson:
Adds a single system for gating in-development features to a limited audience, with a path to general availability and a path to retirement, replacing the ad hoc mix of URL params, `DEBUG_` flags, and a dormant `isBeta()`.

Fixes CODAP-1448

## How it works

Flags are declared in a typed registry, so a typo is a compile error and deleting an entry turns every stale call site into one. Reads go through `isFeatureEnabled()` on a MobX singleton that imports nothing from `app-state` or `document-content` — the open document's grants are pushed in from `app-state` via an autorun, keeping the manager a leaf that models and components can both call.

Resolution order, first match wins:

1. Server config says `off` → **off**, absolutely
2. URL says on or off → that value
3. Server config says `on` → on
4. The open document grants it → on
5. Otherwise → off

`?features=residualPlot,-someOther` is the pilot channel. The server config graduates a feature or kills it without a release, and is fetched at module load (racing the document load rather than waiting for React) from `codap-resources.concord.org` directly, because the `codap.concord.org` path caches at the CloudFront edge for a day and ignores `no-cache`. The fetch carries an `AbortController` timeout and fails open.

Documents record only the flags this session enabled by URL, unioned in the save path so loading never dirties an untouched document.

## Two things worth a reviewer's attention

- **`featureFlags` is `types.maybe(types.array(...))`.** A plain `types.array` writes `"featureFlags": []` into every saved CODAP document; the test suite caught it.
- **v2 saves drop grants.** `debug=saveAsV2` is developer-only so this seemed acceptable, but it's a real gap — happy to close it if you'd rather.

## Deferred

`localStorage` caching of server `off` and an ESLint rule for the affordance rule (both open questions in the story), v2 round-trip, and migrating `gaussianFit`/`ICI` (out of scope per the story).

## Testing

30 new unit tests covering the full resolution matrix, reactivity, fail-open fetch behavior, and the save-path union rules. Full suite: 3380 passing, tsc and lint clean.

Per the discussion on the ticket, `docs/feature-flags.md` documents the affordance rule and scopes the flag-off regression test to flags that gate document-writing state.

### DIFF
--- a/v3/docs/feature-flags.md
+++ b/v3/docs/feature-flags.md
@@ -16,7 +16,7 @@ export const kFeatureFlags = {
     description: "Residual plots on the graph",
     owner: "ESTEEM",
     added: "2026-07-22",
-    expires: "2027-01-22"
+    expires: "2027-07-31"
   }
 } as const satisfies Record<string, IFeatureFlagInfo>
 ```
--- a/v3/docs/feature-flags.md
+++ b/v3/docs/feature-flags.md
@@ -1,0 +1,165 @@
+# Feature flags
+
+A feature flag gates an in-development feature so it can be shown to a limited
+audience before general availability, graduated without a code change, and turned
+off in a hurry if it misbehaves.
+
+Everything lives in `src/models/feature-flags/`.
+
+## Using a flag
+
+Declare it in the registry (`feature-flag-registry.ts`):
+
+```ts
+export const kFeatureFlags = {
+  residualPlot: {
+    description: "Residual plots on the graph",
+    owner: "ESTEEM",
+    added: "2026-07-22",
+    expires: "2027-01-22"
+  }
+} as const satisfies Record<string, IFeatureFlagInfo>
+```
+
+Read it where the affordance renders:
+
+```ts
+import { isFeatureEnabled } from "../../models/feature-flags/feature-flag-manager"
+
+{isFeatureEnabled("residualPlot") && <ResidualPlotMenuItem />}
+```
+
+The flag name is typed, so a typo is a compile error and deleting a registry entry
+turns every stale call site into a compile error too. That is what makes retiring a
+flag mechanical rather than archaeological.
+
+## The affordance rule
+
+**Gate affordances, not registrations.** A gated feature registers its tile,
+adornment, or handler unconditionally; only the things a user clicks — the tool
+shelf button, the menu item, the inspector control — are wrapped in
+`isFeatureEnabled`. Never read a flag at module scope.
+
+Three reasons:
+
+- A document containing artifacts of the feature still opens when the flag is
+  off, instead of degrading to `unknown-content` because someone dropped a query
+  string.
+- Every gate is a reactive read, so a server directive that arrives after render
+  takes effect without a reload.
+- It costs nothing in bundle size. Conditional registration doesn't exclude a
+  module from the bundle; it only skips a few registry insertions.
+
+`error-tester-registration.ts` conditionally registers at module scope and is
+**not** the template here. `errorTester` is a developer tool, not a piloted
+feature, and it stays as it is.
+
+### When a flag needs a flag-off test
+
+The affordance rule keeps a flag-off document loadable; it does not by itself make
+the document *usable*. If a flag gates something that writes document state the
+ungated app can't otherwise reach, add a test that enables the flag, modifies the
+document through the gated affordance, disables the flag, and exercises what could
+break.
+
+Flags that only gate UI over state the document already models — an adornment, a
+legend setting — don't need this; a flag-off document is inert rather than broken.
+
+## Turning a flag on
+
+### By URL — the pilot channel
+
+```
+?features=residualPlot            enable
+?features=-residualPlot           disable
+?features=residualPlot,-someOther comma-separated for several
+```
+
+The setting lives in the URL and nowhere else; closing the tab ends it. This is
+deliberately not a sticky per-browser opt-in, so "what is this user actually
+running?" always has an answer — which matters the first time a pilot teacher
+files a bug report.
+
+### By server config — general availability and the kill switch
+
+`https://codap-resources.concord.org/config/v3-feature-flags.json`
+
+```json
+{ "residualPlot": "on" }
+```
+
+Marking a flag `on` graduates it to everyone without a release. Marking it `off`
+takes it away from everyone, including people holding pilot URLs, without waiting
+for a release.
+
+The URL is fetched from `codap-resources.concord.org` directly, **not** through
+`codap.concord.org/codap-resources/*`, which caches at the CloudFront edge for a
+day and ignores `Cache-Control: no-cache`. A kill switch that takes 24 hours to
+propagate is not a kill switch.
+
+### By document — handing a pilot to a group
+
+A saved document records the flags its author had enabled by URL, and grants them
+to whoever opens it. A researcher can hand twelve teachers a pilot document rather
+than a query string.
+
+## Resolution order
+
+First rule that matches wins:
+
+1. Server config says `off` → **off**. Nothing overrides this.
+2. URL parameter says on or off → **that value**.
+3. Server config says `on` → **on**.
+4. The open document grants it → **on**.
+5. Otherwise → **off**.
+
+In a sentence: the server can always take a feature away; short of that, the URL
+decides; short of that, the server can give it to everyone; short of that, the
+document can give it to you.
+
+The asymmetry is deliberate. `off` is a kill switch and must reach every session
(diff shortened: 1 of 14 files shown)


v3/src/models/feature-flags/feature-flag-manager.ts
```diff
@@ -0,0 +1,93 @@
...
+  /*
+   * Loads the server configuration. Called at module load rather than from a
+   * React effect so that it races the CFM document load rather than waiting for
+   * React to mount. Document loading is reliably slower than a small json fetch,
+   * so the server directives win that race in practice — which is what makes
+   * "server off is absolute" a real guarantee rather than an aspiration.
+   */
+  async loadServerConfig(fetchConfig: () => Promise<FeatureFlagConfig>) {
+    this.setServerConfig(await fetchConfig())
+  }
```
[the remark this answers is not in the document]
kswenson: Good catch. `loadServerConfig` now wraps the fetch in try/catch and applies an empty config on error, so a rejecting loader can no longer surface as an unhandled rejection at module load. Added a test covering the reject-fails-open path.

v3/src/models/feature-flags/start-feature-flag-load.ts
```diff
@@ -0,0 +1,10 @@
+import { fetchFeatureFlagConfig } from "./feature-flag-config"
+import { featureFlagManager } from "./feature-flag-manager"
+
+/*
+ * Imported for its side effect from the application entry point so that the
+ * config fetch begins at module load, racing the CFM document load rather than
+ * waiting for React to mount. Kept separate from the manager so that importing
+ * the manager — in tests, or from a model — never starts a network request.
+ */
+featureFlagManager.loadServerConfig(fetchFeatureFlagConfig)
```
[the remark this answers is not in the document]
kswenson: Done — prefixed the module-load call with `void` to make the fire-and-forget explicit.

v3/src/models/app-state.test.ts
```diff
@@ -72,3 +73,20 @@ describe("AppState", () => {
     })
   })
 })
+
+describe("AppState feature flags", () => {
+  it("grants the open document's feature flags", async () => {
```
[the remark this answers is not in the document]
kswenson: Good catch. Added an `afterEach` to the feature-flag block that calls `appState.disableDocumentMonitoring()`, mirroring the cleanup in the dirty-state block above, so the observers enabled by `setDocument` no longer leak into later tests.

v3/docs/feature-flags.md
```diff
@@ -0,0 +1,165 @@
...
+
+- On save, a document's `featureFlags` become the union of what it already had and
+  what this session enabled by URL.
+- Flags on because of a server `on` are **never** persisted. They're on for
+  everybody anyway, and recording them would bake a grant into every document
+  saved during the general-availability window, resurrecting the feature for those
+  documents after it's later turned off.
+- An explicit `?features=-foo` does **not** strip an existing grant. Disabling is
+  session-scoped, and the document may still contain artifacts of the feature.
+- So grants are a one-way door. The exits are the server kill switch and eventual
```
scytacki: If I'm reading this right, there is another partial exit. If the server says on, but the URL says off. It is only a partial exit because as stated above if the document says on the URL doesn't override it. 

I can see how technically this isn't an "exit" because the the server on check happens after the URL check, but it seems useful to call out. And it also points out a flaw in the compact "First rule that matches wins" summary. URL off is listed as a match, but it doesn't always win.

v3/docs/feature-flags.md
```diff
@@ -0,0 +1,165 @@
...
+can be reminded when a flag is overdue.
+
+## Not built yet
+
+- Caching server `off` directives in `localStorage` so the kill switch survives a
+  failed fetch.
+- A lint rule enforcing the affordance rule.
+- Round-tripping `featureFlags` through v2 documents. Saving as v2 is a developer
+  debug option (`debug=saveAsV2`), so v2 saves currently drop grants.
+- A developer escape hatch for overriding a server `off` locally.
```
scytacki: Is there a way a developer can override a document `on` locally? One way with this design is to copy the doc and change the value in the document, but that doesn't seem very convenient.

v3/src/models/feature-flags/feature-flag-manager.ts
```diff
@@ -0,0 +1,103 @@
...
+    this.urlFlags.forEach((enabled, name) => {
+      if (enabled && this.serverFlags.get(name) !== "off") flags.push(name)
+    })
+    return flags
+  }
+
+  isFeatureEnabled(name: FeatureFlagName): boolean {
+    if (this.serverFlags.get(name) === "off") return false
+    const urlValue = this.urlFlags.get(name)
+    if (urlValue !== undefined) return urlValue
```
scytacki, line 86: This contradicts my understanding of feature-flags.md
In this code if the urlValue is `false` which I think cooresponds to `-flag` in the URL this will override what is in the document.

In the documents section of feature-flags.md:
> An explicit `?features=-foo` does **not** strip an existing grant.

v3/src/models/feature-flags/feature-flag-registry.ts
```diff
@@ -0,0 +1,43 @@
...
+  owner: string
+  // ISO date the flag was added
+  added: string
+  // ISO date by which the flag should have graduated or been retired
+  expires: string
+}
+
+/*
+ * Keep entries alphabetical. The registry must never be empty; `keyof {}` is
+ * `never`, which would make every call to isFeatureEnabled() a type error.
```
scytacki, line 26: This is interesting. It would be correct to make calls to isFeatureEnabled errors if there are no features. But I'm guessing there are some tests or something that will always call it. We could add a sentinel value that is always there to mitigate this so it is explicit that it can't be empty.

scytacki (approved):
Looks good. 

I think I found an inconsistency between the documentation and the code regarding the handling of negative url flags and document flags.

From my point of view the way the code works seems OK to me. If a URL sets a negative flag it overrides the positive flag in the document.

v3/docs/feature-flags.md
kswenson: You're right that a server-`on` (or document-granted) feature can be turned off for a session with `?features=-foo`, and I've called that out. But it's a *session escape*, not an exit from the persisted grant — the grant stays in the document. That's the distinction the "one-way door" is about, and I've reworded the bullet to make it explicit.

On the compact summary: it actually is correct. `-foo` (URL off) does beat a document grant — resolution rule 2 wins over rule 4 — so "first rule that matches wins" holds. The confusion came from the persistence wording below, which I've now separated from runtime resolution.

v3/docs/feature-flags.md
kswenson: Yes — `?features=-residualPlot` disables it for your session even when the open document grants it. That's already built; it's overriding a server `off` that isn't (server off is absolute by design). I've added a parenthetical to the "Not built yet" bullet to make that split clear.

v3/src/models/feature-flags/feature-flag-manager.ts
kswenson: Good catch on the contradiction — it's in the doc wording, not the code. The code is intentional and matches the resolution order: URL off (rule 2) beats a document grant (rule 4), so `-foo` disables the feature for the session. What the doc meant by "does not strip an existing grant" was the *save path* — the persisted grant isn't removed — but it read like a runtime claim. Reworded that bullet to split session-runtime behavior from persistence, so the two no longer appear to disagree.

v3/src/models/feature-flags/feature-flag-registry.ts
kswenson: Agreed the empty registry is a real edge, and I traced it through. Product and runtime handle it fine: a graduating feature deletes both its registry entry and its `isFeatureEnabled(...)` gate, so the clean end-state is zero entries and zero call sites, and `never` (uncallable) is arguably the correct result there rather than a hazard. Every name still routes through `isFeatureFlagName`, so resolution just falls through to off.

The one place it actually bites is the manager's own unit tests: they filter inputs through `isFeatureFlagName`, so exercising any enabled path needs a name that's genuinely in the registry — with an empty registry you can't fake one (`as any` compiles but the guard rejects it, so the test fails). That's the real version of the "some tests will always call it" concern.

I'd rather solve that with a test-only fixture registry if/when we actually reach zero flags than ship a permanent sentinel flag now — the sentinel props up a test need with production surface that stays queryable via `isFeatureEnabled` and shows up in every enumeration. And the moment is self-announcing: whoever removes the last flag gets red manager tests and a natural decision point. Happy to revisit if you'd rather have the guarantee in the type today.

PR #2669, opened 2026-07-24
CODAP-1466: optimize map (and graph) marquee-selection performance

kswenson:
## Summary

Marquee selection on the map was slow on large datasets (profiled on the Tornado-Tracks dataset — thousands of points, with a plugin attached). Profiling with the internal profiler plus Chrome CPU profiles surfaced several independent per-move costs. This PR fixes four of them, taking a marquee move from **~45 ms to ~9 ms (~5×)**. Several fixes live in shared data-display code and also benefit the **graph** marquee.

Fixes CODAP-1466.

## Changes (one per commit)

1. **Delta point restyle** — restyle only the points whose selection toggled instead of all N per selection change (~125 ms → ~2 ms/restyle). Falls back to a full restyle for non-`selectCases` actions or ids that aren't all plotted points. Mirrors the scatter-plot marquee work (#2625).
2. **Don't re-render `MapPointLayer` on every selection change** — it observed the O(N) `dataConfiguration.selection` array via a `useEffect` dependency; now observes selection inside the connecting-lines autorun (as `scatter-plot.tsx` does).
3. **Disable point hit-testing during a marquee drag** — point sprites were hit-tested on every pointermove though the marquee selects via its R-tree (`setPointsInteractive`). ~460 ms → ~125 ms of hit-testing across a drag.
4. **Cache case-by-legend-category grouping** — `allCasesForCategoryAreSelected` re-grouped every case by legend value (O(cases) `getStrValue`) on every selection change; the grouping now lives in a `caseIdsForCategory` cache invalidated only on data/legend changes. **~3.5 s → ~0.004 s** across a drag (the single biggest win).

## Testing

- New unit tests for the category-grouping cache: grouping correctness, stability across selection changes, invalidation on data changes, and selection-check correctness.
- New unit test for `setPointsInteractive`.
- The map-layer delta-extraction and connecting-lines wiring are integration-level (no map/leaflet/pixi component harness); covered by the already-tested shared `setPointSelection` delta path plus the Cypress regression suite.

### DIFF
--- a/v3/src/components/data-display/components/background.tsx
+++ b/v3/src/components/data-display/components/background.tsx
@@ -105,6 +105,9 @@ export const Background = forwardRef<SVGGElement | HTMLDivElement, IProps>((prop
 
   const onDragStart = useCallback((event: PointerEvent) => {
     appState.beginPerformance()
+    // Skip point hit-testing for the duration of the drag: the marquee selects via the R-tree below,
+    // so the pointer event system hit-testing every point sprite on each pointermove is pure overhead.
+    rendererArray.forEach(renderer => renderer?.setPointsInteractive(false))
     selectionTree.current = prepareTree(rendererArray)
     // Event coordinates are window coordinates. To convert them to SVG coordinates, we need to subtract the
     // bounding rect of the SVG element.
@@ -168,9 +171,11 @@ export const Background = forwardRef<SVGGElement | HTMLDivElement, IProps>((prop
     }
     marqueeState.setMarqueeRect({x: 0, y: 0, width: 0, height: 0})
     selectionTree.current = null
+    // Restore point hit-testing now that the drag is over (hover/click behavior).
+    rendererArray.forEach(renderer => renderer?.setPointsInteractive(true))
     dataDisplayModel.setMarqueeMode("unclicked")
     appState.endPerformance()
-  }, [dataDisplayModel, datasetsArray, marqueeState])
+  }, [dataDisplayModel, datasetsArray, marqueeState, rendererArray])
 
   useRendererPointerDownDeselect(rendererArray, dataDisplayModel)
 
--- a/v3/src/components/data-display/components/background.tsx
+++ b/v3/src/components/data-display/components/background.tsx
@@ -179,6 +179,16 @@ export const Background = forwardRef<SVGGElement | HTMLDivElement, IProps>((prop
 
   useRendererPointerDownDeselect(rendererArray, dataDisplayModel)
 
+  // If this component unmounts mid-drag (e.g. the tile is closed), onDragEnd may not run to restore
+  // point hit-testing. Restore it on unmount so points aren't left non-interactive on a renderer that
+  // outlives this marquee overlay. The ref keeps the cleanup on the current renderers while running
+  // only on unmount.
+  const rendererArrayRef = useRef(rendererArray)
+  rendererArrayRef.current = rendererArray
+  useEffect(() => {
+    return () => rendererArrayRef.current.forEach(renderer => renderer?.setPointsInteractive(true))
+  }, [])
+
   // Check if graph has at least one numeric axis that can be zoomed
   const hasNumericAxis = useCallback(() => {
     const xAxisModel = dataDisplayModel.getAxis('bottom')
--- a/v3/src/components/data-display/models/data-configuration-model.ts
+++ b/v3/src/components/data-display/models/data-configuration-model.ts
@@ -527,6 +527,22 @@ export const DataConfigurationModel = types
     }
   }))
   .views(self => ({
+    // The case ids whose legend value is `cat`. Grouping cases by legend value is O(cases) (a
+    // getStrValue per case), so it is cached here and invalidated via clearCasesCache (data/legend/
+    // case changes) — but NOT on selection changes, since category membership doesn't depend on the
+    // selection. allCasesForCategoryAreSelected then only pays the cheap per-case isCaseSelected check
+    // on each selection change (e.g. every marquee move) instead of re-grouping every case.
+    caseIdsForCategory: cachedFnWithArgsFactory<(cat: string) => string[]>({
+      key: (cat: string) => cat,
+      calculate: (cat: string) => {
+        const dataset = self.dataset
+        const legendID = self.attributeID('legend')
+        return (legendID && self.getCaseDataArray(0).filter((aCaseData: CaseData) =>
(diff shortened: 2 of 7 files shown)


bfinzer:
## Review

### Overview

Four independent per-move performance fixes for the map marquee (three of which also benefit the graph marquee), moving a marquee move from ~45 ms to ~9 ms:

1. **Delta point restyle** in `map-point-layer.tsx` — restyles only the toggled points, with a safe full-restyle fallback for actions that aren't `selectCases`-on-plotted-points.
2. **Stop re-rendering `MapPointLayer` on every selection change** — moves the `selection` subscription inside the connecting-lines autorun.
3. **Disable point hit-testing during marquee drag** in `background.tsx` via a new `setPointsInteractive` hook on `PointRendererBase`/`PixiPointRenderer`.
4. **Cache case-by-legend-category grouping** — splits `caseIdsForCategory` out of `allCasesForCategoryAreSelected` so grouping is only recomputed on data/legend changes, not on every selection change. This is the ~3.5s → ~4ms win.

Well-scoped, well-tested, correctness reasoning solid. Approving with a couple of minor suggestions.

### Correctness

- **Delta path fallback is right**: `caseIds.every(id => renderer?.getPointForCaseData({ plotNum: 0, caseID: id }))` catches the case where `selectCases` args include a non-plotted id (hidden case, parent/collection id whose selection ripples to child items). Full restyle in that case avoids missing visible children — necessary because a parent selection expands to items that aren't in `caseIds` themselves.
- **Cache invalidation for `caseIdsForCategory`**: threaded through `clearCasesCache` → invoked by `invalidateCases`, which fires on setAttribute, hidden-case changes, `caseDataHash` changes (case value updates), and dataset changes. Legend changes go through `setAttribute("legend", …)` → `invalidateCases` → clears the cache. Not invalidated on plain selection changes, which is precisely the intended behavior. The tests directly assert stability across selection changes and update on case addition.
- **`setPointsInteractive` unmount cleanup**: the ref pattern (mutating `rendererArrayRef.current` outside a `useEffect`) is a little unusual but correctly keeps the cleanup closure targeting the *current* renderers when unmount happens on a later render. Good defensive coverage for tile-closed-mid-drag.
- **Cache in `allCasesForCategoryAreSelected`**: still invalidated on selection changes via the existing reaction. The split cache preserves the caller's contract exactly.

### Suggestions (all minor)

- **`void dataConfiguration.selection` in `updateConnectingLines`** — this getter is O(visible cases) because it filters `visibleCaseIds` by `isCaseSelected`, so every selection change re-filters. The scatter-plot's equivalent (`scatter-plot.tsx:629`) subscribes to `dataConfiguration?.dataset?.selectionChanges`, a scalar counter — O(1) subscription. Since the PR calls out the scatter pattern as the model, worth mirroring the exact subscription too. (The previous cost was much worse — full component re-render — so this is already a big win either way.)
- **The `caseIds.every(...)` guard** does N `getPointForCaseData` lookups just to *decide* whether to take the delta path, then N more inside the delta loop. In the common marquee case that's fine (small N per move), but for a large single-batch `selectCases` (e.g. legend click on a category with thousands of items) the guard cost is real. Alternative: iterate once and skip missing points inside the loop — but that would silently swallow the parent-selection-cascade case you're guarding against. Current design is safe; a short comment about why the two-pass is preferable might help future readers.
- **`renderer` added to `setupResponsesToDatasetActions` deps** is correct now that the effect body reads `renderer?.getPointForCaseData`. Note that this will tear down and reinstall the `onAnyAction` disposer whenever the renderer swaps (e.g. WebGL context reacquisition) — small window for missed actions during reinstall, unlikely to matter in practice.

### Tests

- `caseIdsForCategory` / `allCasesForCategoryAreSelected` — covers grouping correctness, stability under selection change (the whole point), invalidation on case add, and the reflective read of the current selection. The core caching contract, well-covered.
- `setPointsInteractive` — direct toggle on `PixiPointRenderer` and a no-op verification on the base renderer.
- Delta-restyle path isn't directly unit-tested from the map layer, but the shared `setPointSelection` delta code already has coverage plus the Cypress regression suite. Reasonable.

### Verdict

Approve. Clean, targeted, correct, and the numbers speak for themselves (5×; 3.5 s → 4 ms on the category cache alone). The two subscription-cost points above are polish, not blockers.

bfinzer (approved):

kswenson:
Thanks for the careful read — the correctness section tracks exactly what I intended, especially the note on why the parent-selection cascade forces the full-restyle fallback.

I dug into the three suggestions and I'd like to leave all three as-is. Reasoning:

**1. `void dataConfiguration.selection` → `dataset.selectionChanges`**

The O(visible cases) cost is real, but I don't think swapping is a win here:

- The read is gated on `showConnectingLines`. With lines off — the default, and the marquee hot path this PR targets — the autorun never touches `selection` at all.
- When lines *are* on, the same autorun body runs `connectingLinesForCases()`, removes every `path`, and re-renders all of them. The filter is noise next to that.
- `selection` is actually the more robust dependency. `selectionChanges` is reset to `0` in `data-set.ts:1585` (`initializeVolatileState`, after a snapshot restore replaces `selection`), so a `0 → 0` reset produces no dependency change while the `selection` computed changes regardless. The comment at `use-plot.ts:322-324` flags this same hazard for the counter-based subscription.

One note on the citation: `scatter-plot.tsx:629` doesn't exist — the file is 400 lines and doesn't reference `selectionChanges`. The actual precedent is `use-plot.ts:377`, which is solving a different problem (coalescing selection refreshes into one frame), not connecting-lines invalidation.

**2. Comment on the two-pass `caseIds.every(...)` guard**

I think this one's already covered — `map-point-layer.tsx:361-366` explains that `selectCases` on ids that aren't all plotted points falls back to a full restyle "so no affected point is missed." Happy to expand it if it's not landing.

On the cost: `getPointForCaseData` is a `caseDataKey` string build plus one `Map.get` (`point-renderer-base.ts:366-369`), so even a 10k-id legend click keeps the guard sub-millisecond and well under the restyle loop it gates. `points-state.ts:110` has a `hasPointForCaseData` that would avoid two object allocations per call, but it isn't exposed on `PointRendererBase`, and adding a method for that seems like more churn than the win justifies.

**3. `renderer` in the `setupResponsesToDatasetActions` deps**

No new teardown window here — `refreshPointSelection`'s `useCallback` deps already include `renderer` (`map-point-layer.tsx:300`, untouched by this PR), and `refreshPointSelection` was already in that effect's dep array before this change. So the disposer was already reinstalled on every renderer swap; adding `renderer` explicitly just satisfies exhaustive-deps now that the body reads it directly.

PR #2683, opened 2026-07-29
CODAP-1478: animate residual plot points

bfinzer:
## Summary
- Residual-plot points now transition smoothly to their new positions when the x or y attribute changes, matching the main plot's animation.
- Newly-appearing residual points (residual plot toggle-on, case add) fade in from radius 0.
- Fixes a misleading "defunct axis model" warning that `useSubAxis.renderSubAxis` printed when no axis was registered at the given place; it now silent-skips (as `installDomainSync` already does) and only warns for truly defunct axis references.

## Test plan
- [ ] Scatterplot with movable line / LSRL / plotted function + residual plot: change the X attribute → residual points slide smoothly
- [ ] Same but change the Y attribute → residual points slide smoothly
- [ ] Toggle residual plot on → points fade in from radius 0
- [ ] Add a new case → residual point fades in from radius 0 (no full-radius flash)
- [ ] Change attribute, then within 2s drag the movable line → residuals snap to follow the line in real time
- [ ] Drag a data point → residuals update in real time (no lag)
- [ ] Toggle residual plot off, then undo the toggle → no console warnings

Fixes CODAP-1478

### DIFF
--- a/v3/src/components/axis/hooks/use-sub-axis.ts
+++ b/v3/src/components/axis/hooks/use-sub-axis.ts
@@ -79,6 +79,14 @@ export const useSubAxis = ({
 
     renderSubAxis = useCallback(() => {
       const axisModel = axisProvider.getAxis?.(axisPlace)
+      // Silent skip when no axis is registered at this place — expected during the render
+      // window that follows removing an axis (e.g., a residual-plot toggle-off, or an undo of
+      // a residual-plot-show action): observable reads in reactions above still fire before
+      // React unmounts the SubAxis, and they call renderSubAxis with no axis to render.
+      // Warn only when we do have an axis reference that has gone defunct — that's a genuinely
+      // suspect condition worth surfacing. Matches the undefined/defunct split in
+      // installDomainSync below.
+      if (!axisModel) return
       if (!isAliveSafe(axisModel)) {
         console.warn("useSubAxis.renderSubAxis skipping rendering of defunct axis model:", axisPlace)
         return
@@ -86,7 +94,7 @@ export const useSubAxis = ({
       const multiScale = layout.getAxisMultiScale(axisPlace)
       if (!multiScale) return // no scale, no axis (But this shouldn't happen)
 
-      axisModel && axisProvider.getAxisHelper(axisModel.place, subAxisIndex)?.render()
+      axisProvider.getAxisHelper(axisModel.place, subAxisIndex)?.render()
     }, [axisPlace, axisProvider, layout, subAxisIndex]),
 
     onDragStart = useCallback((event: any) => {
--- a/v3/src/components/graph/adornments/movable-line/movable-line-adornment-component.tsx
+++ b/v3/src/components/graph/adornments/movable-line/movable-line-adornment-component.tsx
@@ -13,6 +13,7 @@ import { useAdornmentCells } from "../../hooks/use-adornment-cells"
 import { useGraphContentModelContext } from "../../hooks/use-graph-content-model-context"
 import {useGraphDataConfigurationContext} from "../../hooks/use-graph-data-configuration-context"
 import { useGraphLayoutContext } from "../../hooks/use-graph-layout-context"
+import { useDataDisplayAnimation } from "../../../data-display/hooks/use-data-display-animation"
 import { isDateAxisModel } from "../../../axis/models/numeric-axis-models"
 import {
   calculateSumOfSquares, computeSlopeAndIntercept, equationString, IAxisIntercepts, lineToAxisIntercepts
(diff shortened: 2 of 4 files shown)


kswenson (approved):
Approving. The design here is solid — the enter/update rewrite is the right shape, and the comments anticipating the d3 transition hazards (interrupt scoping, the `active()` reentrancy guard, the "too late; already running" retarget limitation) made this much easier to review than it would otherwise have been.

I found two defects and fixed them on the branch rather than handing them back, in `47fcb0bb7`. **Please review that commit in return** — it changes a mechanism you designed deliberately, and it shouldn't land on my say-so alone. I'm the only engineer who's looked at this, and you can't approve your own PR, so a second pass from you on my commit is the closest we get to real review coverage on that part.

---

### 1. The attribute-change animation never played (high)

The feature didn't work for the case it exists to serve. Confirmed in the running app before and after the fix.

One attribute change fires the `syncResidualPlot` autorun **three** times. It writes the axis and MultiScale domains that it also reads, so MobX re-schedules it; then the graph layout reflows, because a wider y range widens the left axis tick labels, which shifts the bottom axis and therefore every point's `cx`. Only the first fire knew an attribute had changed. `prevRenderedAttrIDsRef` was advanced by that fire, so every fire after it computed `attrsChanged === false`, took the snap path, and called `selection.interrupt("cxcy")` — cancelling the slide the first fire had just scheduled, while it was still in d3's `CREATED` state and had not painted a frame.

Instrumented trace of a Y-attribute change, before the fix:

```
autorun  attrsChanged=true    → render animate=true,  positionsChanged=true
                                 cy 223.65→216.40, cx 139.63      (slide scheduled)
autorun  attrsChanged=false   → render animate=false, positionsChanged=true
                                 cy0 216.40 (same), cx0 139.63→137.94   ← snap fires
                                 DOM cy still 223.65 — the slide never painted
```

`cy` is identical between the two fires; only `cx` moved, by the ~1.7px of the reflow. That's what makes this unfixable by comparing positions alone — the later repaints of a single logical change carry *genuinely different* targets, so "did anything move" can't separate "same change, reflowed" from "new change, snap to it." (I tried that first; it fixed X-attribute changes and left Y broken, which is how the reflow turned up.)

**Fix:** decide on `isAnimating()` — the signal the main plot's own points already use — instead of the attribute-ID comparison, and skip the cx/cy work entirely when nothing has moved, so a repaint that lands every point where it's already headed leaves an in-flight slide alone. `prevRenderedAttrIDsRef` and the attribute-ID comparison are gone; the result is less machinery than the branch had.

To keep the behavior your design was protecting — a movable-line drag started inside the 2s window must track the cursor, not lag a second behind it — the movable line's drag now calls `stopAnimation()` at drag start. That's the established convention: point drag (`scatter-plot.tsx:137`), dot-plot drag (`use-dot-plot-drag-drop.ts:42`), and axis drag (`use-sub-axis.ts:155`) all do exactly this. Line drag never needed it before, because dragging the line didn't move any points until residual plots existed. **This is the change most worth your scrutiny** — it touches the movable-line adornment, outside this PR's original blast radius.

### 2. Hover and click handlers went stale on retained points (medium)

`selection.join(name)` returns `enter.append(name).merge(update)`, so the pre-PR code re-bound `.style("cursor")` and all three `.on(...)` handlers on every circle on every render, refreshing their closure. Moving them into `selection.enter().append("circle")` meant retained circles kept the handlers bound at creation time.

`tipTextFor` closes over `xAttrID` and `xAttrName`, computed per render. The join is keyed by `caseID`, so circles survive an x-attribute change — precisely the event that stales those values. And because the old attribute still exists in the dataset, `getValue` returned a plausible-looking number rather than blank, so the tooltip was silently wrong rather than obviously broken.

**Fix:** bind cursor and the three handlers on `enterSelection.merge(selection)`, leaving the `animateCxCy` block on `selection` alone so enter circles still skip the slide.

---

### Verified after the fixes

In-app with `?features=residualPlot`, scatter plot + line + residual plot: X-attribute change slides · Y-attribute change slides · tooltip follows the new attribute · toggle-off/on still fades in from dots · movable-line drag inside the animation window still tracks the cursor in real time. `build:tsc` clean, `lint` clean, full Jest suite green (352 suites / 3712 tests).

### Checked and cleared

The `use-sub-axis.ts` change is correct and I'd have flagged it as scope creep if it weren't. `isAliveSafe` returns `false` for `undefined`, so the old code printed "defunct axis model" when there was no model at all. Git archaeology: `61261d8f6` (Sep 2023) originally had the undefined/defunct split, and `6877f38c7` collapsed it by accident while swapping in `isAliveSafe` — while *preserving* the split in `installDomainSync` in the same diff. Your change restores the original intent and matches `use-axis.ts:156` and `use-plot.ts:101-103`. It also can't mask a real missing-axis bug, since `<SubAxis>` only mounts when the axis already exists.

### Follow-up, not blocking

The animation logic has no automated coverage, and issue 1 is the argument for adding some — it produced a *missing* animation rather than a wrong one, which is the hardest kind of defect to catch by eye, and the test-plan box for exactly that case was unchecked. A Cypress check sampling a residual point's `cy` across the transition window (the same check that confirmed the fix by hand) would have caught it. Not blocking here — animation tests are tricky and often flaky, and this repo already has precedent in `bivariate-adornments.spec.ts:319-341`, where dot-radius assertions sit commented out with a TODO for that reason. The failing codecov checks are expected and I don't think they should gate this.

Minor, not addressed: `teardown()` and `exit().remove()` detach nodes without `.interrupt()`ing first, leaving d3 timers ticking against detached nodes for up to 1s. Harmless, but `ConnectingLines.destroy()` does `.interrupt().remove()` — worth matching sometime.

PR #2691, opened 2026-09-03
CODAP-1506: restructure the graph Format palette into labelled sections

kswenson:
Part of CODAP-1506. This is the structural half of the story — the parts that
don't depend on point shapes existing yet. The category rows with their shape
dropdowns need the geometry from CODAP-1504 and the per-category storage from
CODAP-1505, and will land on this branch once those exist.

Targets `leads-point-shapes`, the integration branch for the point-shapes
feature, not `main`.

## What's here

- The graph Format palette's controls are grouped under **Data Points** and
  **Graph** headings. Each section is a region labelled by its heading.
- Palette width 220px → 256px, so a category name has room for roughly 16
  characters instead of 11.
- The legend category list shows 2.5 rows, derived from the row height, so half
  of a third row is visible when there is one.

## Notes for review

**The map palette is deliberately unchanged.** These controls live in
`DisplayItemFormatControl`, which the map layers palette also renders.
Sectioning is opt-in through a `showSectionHeaders` prop that only the graph
passes: the map already heads each layer with the layer's own name and repeats
the controls per layer, so a second heading inside each block would be noise,
and it has no background controls to put in a second section. There's a test
asserting no headings or regions render without the prop.

**The width override is scoped.** The `min-width: 220px` in
`inspector-panel.scss` is shared by every inspector palette in the app,
including map and case table, so changing it there would widen all of them.
This adds an optional `className` prop to `InspectorPalette` and overrides the
width on the graph Format palette alone.

**The row height is 42px, not 38px.** The story assumed 38px.
`.cat-color-picker` sets a 6px margin that overrides the shared 4px from
`.color-picker-row` — equal specificity, so it wins on emit order alone. The
real row is 30 + 2×6 = 42px, today's fixed 75px is 1.79 rows rather than 1.97,
and 2.5 rows is 105px. That margin is now pinned to the same variable the list
height derives from, so emit order no longer decides it.

**One acceptance criterion was already satisfied.** The story asks that "Graph
Background Color" revert to "Background Color". `DG.Inspector.backgroundColor`
is already "Background Color" on main, so that AC means "don't adopt the
prototype's rename" — no change was needed.

Two small changes beyond the letter of the story: `height` → `max-height`, so a
one- or two-category list doesn't reserve empty space; and `overflow-y: scroll`
→ `auto`, so no dead scrollbar appears when there's nothing to scroll. The
original intent — signal that the list scrolls — is now carried by the visible
half-row.

## Testing

5 new tests covering section rendering, the region/heading labelling, which
controls land in which section, and the map's flat layout. Mutation-tested: four
mutations (ignoring `showSectionHeaders`, removing `aria-labelledby`, removing
the heading `id`, swapping the two section titles) each fail at least one test.

`npm run build:tsc`, `npm run lint`, and the inspector test suites all pass.

### DIFF
--- a/v3/src/components/data-display/inspector/display-item-format-control.scss
+++ b/v3/src/components/data-display/inspector/display-item-format-control.scss
@@ -70,6 +70,20 @@
 
 .codap-inspector-palette {
   .palette-form {
+    .palette-section {
+      & + .palette-section {
+        margin-top: 10px;
+      }
+
+      .palette-section-title {
+        font-size: 14px;
+        font-weight: bold;
+        margin: 0 0 4px;
+        padding-bottom: 2px;
+        border-bottom: 1px solid vars.$charcoal-light-5;
+      }
+    }
+
     .slider-row {
       gap: 10px;
 
@@ -166,8 +180,12 @@
     }
 
     .cat-color-setting {
-      height: 75px;
-      overflow-y: scroll;
+      // Derived from the row height so the list always stops mid-row and reads as scrollable.
+      // A fixed height here stops short of the third row entirely, which looks like the end of
+      // the list. max-height rather than height so a one- or two-category list does not reserve
+      // empty space below itself.
+      max-height: vars.$palette-category-list-height;
+      overflow-y: auto;
 
       // Force vertical scrollbar visibility for Chrome/Safari so users know
       // the list is scrollable when there are more than 3 categories
@@ -188,8 +206,11 @@
       scrollbar-width: thin;
       scrollbar-color: gray lightgray;
 
+      // Overrides the shared .color-picker-row margin, which has equal specificity and so wins or
+      // loses on emit order alone. The 2.5-row height above is derived from the same variable, so
+      // the list height tracks whatever this is set to.
       .cat-color-picker {
-        margin: 6px 0;
+        margin: vars.$palette-category-row-margin 0;
       }
     }
 
--- a/v3/src/components/data-display/inspector/display-item-format-control.scss
+++ b/v3/src/components/data-display/inspector/display-item-format-control.scss
@@ -180,10 +180,13 @@
     }
 
     .cat-color-setting {
-      // Derived from the row height so the list always stops mid-row and reads as scrollable.
-      // A fixed height here stops short of the third row entirely, which looks like the end of
-      // the list. max-height rather than height so a one- or two-category list does not reserve
-      // empty space below itself.
+      // A flex column, so the rows are flex items whose vertical margins do not collapse and each
+      // row occupies the full height the 2.5-row limit below is derived from. The list ends
+      // halfway through the third row, and that half row is what signals there is more to scroll
+      // to. max-height rather than height so a one- or two-category list takes only the space it
+      // needs.
+      display: flex;
+      flex-direction: column;
       max-height: vars.$palette-category-list-height;
       overflow-y: auto;
 
@@ -210,6 +213,7 @@
       // loses on emit order alone. The 2.5-row height above is derived from the same variable, so
       // the list height tracks whatever this is set to.
       .cat-color-picker {
+        flex-shrink: 0; // rows keep their height, so the list scrolls rather than compressing
         margin: vars.$palette-category-row-margin 0;
       }
     }
--- a/v3/src/components/data-display/inspector/display-item-format-control.tsx
+++ b/v3/src/components/data-display/inspector/display-item-format-control.tsx
@@ -28,9 +28,7 @@ interface IPaletteSectionProps {
   children: ReactNode
 }
 
-// Groups controls under a heading when given a title, and renders them bare otherwise. The map
-// layers palette heads each layer with the layer's own name and repeats these controls per layer,
-// so a second heading inside each one would be noise; it opts out by passing no title.
+// Groups controls under a heading when given a title, and renders them bare otherwise.
 function PaletteSection({ title, children }: IPaletteSectionProps) {
   const titleId = useId()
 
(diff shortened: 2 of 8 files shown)


v3/src/components/data-display/inspector/display-item-format-control.scss
```diff
@@ -166,8 +180,12 @@
     }
 
     .cat-color-setting {
-      height: 75px;
-      overflow-y: scroll;
+      // Derived from the row height so the list always stops mid-row and reads as scrollable.
+      // A fixed height here stops short of the third row entirely, which looks like the end of
+      // the list. max-height rather than height so a one- or two-category list does not reserve
+      // empty space below itself.
+      max-height: vars.$palette-category-list-height;
```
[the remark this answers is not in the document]
kswenson: Good catch — this was a real bug. The rows are block-level siblings, so their
vertical margins collapse and each row advanced 36px rather than the 42px the
limit was derived from, exposing about 90% of the third row instead of half.

Fixed by making the list a flex column so the rows are flex items and their
margins no longer collapse, with flex-shrink: 0 so they keep their height and
the list scrolls instead of compressing them.

One thing your suggestion didn't cover: preventing the collapse also widens the
visible gap between rows, from a shared 6px to 6px + 6px. The row margin is
halved to 3px so the gap stays at 6px and the spacing matches the design. Row
pitch is 36px and the list is 90px, so the third row is now cut halfway as
intended.

kswenson:

| PR | Story | What it is |
|---|---|---|

**Superseded later:** nothing behavioral. `display-item-format-control.scss` and `en-US.json5` both keep growing in the PRs above — ordinary stacking rather than churn worth flagging here.

**Out of scope by design**, each with a ticket: the PIXI/WebGL renderer (CODAP-1509), legend keys carrying the shape (CODAP-1507), and the residual plot (CODAP-1526).

v3/src/components/data-display/inspector/display-item-format-control.tsx
```diff
@@ -22,6 +23,27 @@ import { PointSizeSlider } from "./point-size-slider"
...
 import "./display-item-format-control.scss"
 
+interface IPaletteSectionProps {
+  title?: string
+  children: ReactNode
+}
+
+// Groups controls under a heading when given a title, and renders them bare otherwise. The map
+// layers palette heads each layer with the layer's own name and repeats these controls per layer,
+// so a second heading inside each one would be noise; it opts out by passing no title.
```
tealefristoe:
```suggestion
// Groups controls under a heading when given a title, and renders them bare otherwise.
```

v3/src/components/data-display/inspector/display-item-format-control.tsx
```diff
@@ -31,18 +53,25 @@ interface IDisplayItemFormatControlProps {
...
-    isTransparent, onBackgroundTransparencyChange, plotBackgroundColor, onBackgroundColorChange
+    isTransparent, onBackgroundTransparencyChange, plotBackgroundColor, onBackgroundColorChange,
+    showSectionHeaders
   } = props
   const { tile } = useTileModelContext()
   const legendAttrID = dataConfiguration.attributeID("legend")
   const attrType = dataConfiguration.attributeType("legend")
+  // Only the graph Format palette is sectioned. The map layers palette renders these controls once
+  // per layer under the layer's own name, and has no background controls to put in a second
+  // section, so it opts out and keeps the flat layout.
```
tealefristoe:
```suggestion
```

v3/src/components/inspector-panel.scss
```diff
@@ -134,10 +134,16 @@ $inspector-panel-border-width: 2px;
   overflow: hidden;
   font-family: Roboto, "museo-sans", sans-serif;
   font-size: 14px;
-  min-width: 220px;
+  min-width: vars.$palette-min-width;
   min-height: 50px;
   font-weight: normal;
 
+  // Scoped to the graph Format palette. The shared min-width above applies to every inspector
+  // palette in the app, so widening it there would widen the map and case table palettes too.
```
tealefristoe:
```suggestion
```

v3/src/components/inspector-panel.tsx
```diff
@@ -269,12 +269,14 @@ interface IInspectorPalette {
   Icon?: ReactNode
   id?: string
   title: string
+  // Applied to the palette element, so an individual palette can override shared palette styling.
+  className?: string
```
tealefristoe:
```suggestion
  paletteClassName?: string
```

tealefristoe (approved):
Looks good 👍 

Most of my suggestions are about trimming comments, which I know is controversial, so take them or leave them.

v3/src/components/data-display/inspector/display-item-format-control.tsx
kswenson: Taken. This and the one below were the same fact written twice, which I hadn't noticed.

I kept the "why" once, on the `showSectionHeaders` prop, since that's where the choice is actually made — the map palette opts out by omitting it, so there's no call site to hang it on.

v3/src/components/data-display/inspector/display-item-format-control.tsx
kswenson: Taken — this was the duplicate. It now lives once, on the `showSectionHeaders` prop declaration.

v3/src/components/inspector-panel.scss
kswenson: Trimmed rather than removed, if that's an acceptable middle ground.

The first sentence was restating what `&.point-format-palette` plainly says, so that's gone. I kept one line for the part that isn't visible from here — that `$palette-min-width` is shared, so widening it rather than overriding it would also widen the map and case table palettes. Happy to drop it entirely if you'd still rather.

v3/src/components/inspector-panel.tsx
kswenson: Taken, and this is the one I should have arrived at myself — a better name instead of a comment explaining the name.

Worth noting for anyone reading later why the comment existed: `InspectorPalette` renders a wrapper, a pointer and the palette, and this class goes to the last of those rather than the outermost element, which isn't what a bare `className` suggests. `paletteClassName` says that without the comment. Only one caller passes it, so the rename was cheap.

PR #2692, opened 2026-09-04
CODAP-1505: store point shape per legend category

kswenson:
Part of the LEADS point-shapes feature. This is the model layer only — nothing
reads these values yet. The controls come with CODAP-1506 and the rendering with
CODAP-1504.

**Stacked on #2691.** Base is `CODAP-1506-format-panel-restructure`, so review
that one first. Once it merges I'll retarget this to `leads-point-shapes` and
the diff will shrink to just this story.

## Where shape lives

Per-category shape goes on the attribute's `CategorySet`, next to per-category
color, so two displays using the same legend attribute agree by construction.
The single no-legend shape goes on `DisplayItemDescriptionModel`, next to point
color. That mirrors how color already works.

The default is stored as absence — assigning `circle` removes the entry rather
than recording it. A document carries only the shapes a user chose, and one that
never used the feature is byte-for-byte unaffected.

## The v2 decision worth reviewing

V2 has no field for point shape, so both halves ride v3 extension namespaces:
per-category on the attribute, the single shape on the graph component. Both
survive a v3 → v2 → v3 round trip; a v2 re-save drops them, which is lossy but
harmless.

**They deliberately do not ride in `_categoryMap`**, which looks like the
obvious home next to colors. I checked V2 on `master` before ruling it out —
`attribute_model.js:486`:

```js
var ignorable = ['__order', 'stroke-color', 'stroke-transparency'];
```

`updateCategoryMap` treats every other key as a category: it appends unknown
keys to `__order`, then `toArchive` writes the result back. A shapes key there
would surface as a phantom category in the user's legend and be **persisted by
v2 itself**. By contrast an unknown key in the attribute's `v3` namespace is
merely mixed onto the SproutCore model by `DG.Attribute.create` and dropped by
`toArchive`'s allowlist, so v2 is unaffected.

There's a test pinning this, because the harmful change is the one that looks
safe.

## Imported shapes are not filtered, unlike colors

Colors are filtered to categories present in the data; shapes are not. The
difference is that every shape entry is a deliberate user assignment, whereas
some paths assign colors automatically by category position — which is why the
color import keeps only a color differing from the one that position would have
produced. Shapes have no generated noise to age out, so a category whose cases
are deleted and later restored keeps its shape. V2 does the same, for the reason
its own comment gives.

## Provisional promotion

A category set stays provisional until the user changes something that must
persist. The predicate watched moves and colors; shapes join it, since assigning
a shape can be the only thing a user does.

This also removes `CategorySet.userActionNames`, which named the promoting
changes but was read by nothing. Its test asserted against the list rather than
the behavior; equivalent coverage now runs against the promotion predicate
itself, where moves, colors and shapes are checked together.

Promotion rebuilds the set from a snapshot and so replaces the instance — a
reference held across a modification goes stale and later writes are silently
lost. That's now documented at both `getCategorySet` and
`promoteProvisionalCategorySet`; the per-category setters re-resolve for this
reason.

## Two deviations from the ACs

- **Undo strings use `V3.` not `DG.`** The AC specifies
  `DG.Undo.graph.changePointShape`, but `DG.*` is for strings inherited from v2
  and shape is new in v3, so these are `V3.Undo/Redo.graph.changePointShape`.
- **`getLegendShapeForCase` is not here.** The per-case resolver mirroring
  `getLegendColorForCase` is a rendering concern; CODAP-1504 adds it where it's
  used.

## Testing

29 new tests. The full-document round-trip tests are the ones that earned their
keep — they caught the promotion gap that every unit test missed, because unit
tests read back through the same in-memory object.

Mutation-tested; each of these fails at least one test: writing shapes into
`_categoryMap`, filtering imported shapes to present categories, accepting
unrecognized shape values, storing the default instead of omitting it, dropping
shapes from the promotion predicate, dropping `moves` from it, and having the
graph importer ignore `v3.pointShape`.

`npm run build:tsc`, `npm run lint`, and the full suite (3805 tests / 357
suites) all pass.

### DIFF
--- a/v3/src/components/data-display/data-display-notifications.ts
+++ b/v3/src/components/data-display/data-display-notifications.ts
@@ -1,6 +1,7 @@
 import { AttributeBinningType } from "../../models/shared/data-set-metadata"
 import { ITileModel } from "../../models/tiles/tile-model"
 import { updateTileNotification } from "../../models/tiles/tile-notifications"
+import { PointShape } from "../../utilities/point-shape-utils"
 import { GraphPlace } from "../axis-graph-shared"
 
 // Shared notification helpers for the graph and map tiles (the V3 "data-display" tiles).
@@ -54,6 +55,14 @@ export function changePointColorNotification(
   tile: ITileModel | undefined, color: string, category: string
 ) {
   return updateTileNotification("change point color", { color, category }, tile)
+}
+
+// Shape has no V2 counterpart, so there is no legacy op string to match. `category` is undefined
+// when no legend attribute is assigned and the shape applies to every point.
+export function changePointShapeNotification(
+  tile: ITileModel | undefined, shape: PointShape, category?: string
+) {
+  return updateTileNotification("change point shape", { shape, category }, tile)
 }
 
 // V2 emits the COMPOUND op string `"change " + <internalName>` from the factory
--- a/v3/src/components/data-display/models/display-item-description-model.ts
+++ b/v3/src/components/data-display/models/display-item-description-model.ts
@@ -1,10 +1,14 @@
 import {Instance, types} from "mobx-state-tree"
 import {applyModelChange} from "../../../models/history/apply-model-change"
 import {defaultPointColor, defaultStrokeColor, kellyColors} from "../../../utilities/color-utils"
+import {kDefaultPointShape, PointShape, pointShapeOrDefault} from "../../../utilities/point-shape-utils"
 
 export const DisplayItemDescriptionModel = types
   .model("DisplayItemDescriptionModel", {
     _itemColors: types.optional(types.array(types.string), [defaultPointColor]),
+    // The shape used when no legend attribute assigns one per category. Scalar rather than an
+    // array like _itemColors: color varies per plot index for multi-y plots, shape does not.
+    _itemShape: types.optional(types.string, kDefaultPointShape),
     _itemStrokeColor: defaultStrokeColor,
     _itemStrokeSameAsFill: false,
     _pointSizeMultiplier: 1, // Not used when item is a polygon in which case it is set to -1
@@ -16,6 +20,9 @@ export const DisplayItemDescriptionModel = types
   .actions(self => ({
     setPointColor(color: string, plotIndex = 0) {
       self._itemColors[plotIndex] = color
+    },
+    setPointShape(shape: PointShape) {
+      self._itemShape = shape
     },
     setPointStrokeColor(color: string) {
       self._itemStrokeColor = color
@@ -44,6 +51,9 @@ export const DisplayItemDescriptionModel = types
     get itemColor() {
       return this.itemColorAtIndex(0)
     },
+    get itemShape(): PointShape {
+      return pointShapeOrDefault(self._itemShape)
+    },
     get itemStrokeColor() {
       return self._itemStrokeSameAsFill ? this.itemColor : self._itemStrokeColor
     },
@@ -58,6 +68,9 @@ export const DisplayItemDescriptionModel = types
     },
     get pointColor() {
       return self.itemColor
+    },
+    get pointShape(): PointShape {
+      return self.itemShape
     },
     get pointStrokeColor() {
       return self.itemStrokeColor
--- a/v3/src/components/data-display/models/display-item-description-model.ts
+++ b/v3/src/components/data-display/models/display-item-description-model.ts
@@ -6,9 +6,15 @@ import {kDefaultPointShape, PointShape, pointShapeOrDefault} from "../../../util
 export const DisplayItemDescriptionModel = types
   .model("DisplayItemDescriptionModel", {
     _itemColors: types.optional(types.array(types.string), [defaultPointColor]),
-    // The shape used when no legend attribute assigns one per category. Scalar rather than an
-    // array like _itemColors: color varies per plot index for multi-y plots, shape does not.
-    _itemShape: types.optional(types.string, kDefaultPointShape),
+    /*
+     * The shape used when no legend attribute assigns one per category. Scalar rather than an
+     * array like _itemColors: color varies per plot index for multi-y plots, shape does not.
+     *
+     * `maybe` rather than `optional` with a default, so the default is stored as absence and a
+     * document that never used shapes does not gain the field when it is opened and saved. That
+     * matches how per-category shapes are stored, and the getter below resolves the absence.
+     */
+    _itemShape: types.maybe(types.string),
     _itemStrokeColor: defaultStrokeColor,
     _itemStrokeSameAsFill: false,
     _pointSizeMultiplier: 1, // Not used when item is a polygon in which case it is set to -1
@@ -22,7 +28,8 @@ export const DisplayItemDescriptionModel = types
       self._itemColors[plotIndex] = color
     },
     setPointShape(shape: PointShape) {
-      self._itemShape = shape
+      // Absence means the default, as it does for a category's shape.
+      self._itemShape = shape === kDefaultPointShape ? undefined : shape
     },
     setPointStrokeColor(color: string) {
       self._itemStrokeColor = color
--- a/v3/src/models/data/category-set.test.ts
+++ b/v3/src/models/data/category-set.test.ts
@@ -406,6 +406,21 @@ describe("CategorySet", () => {
     })
 
 
+    it("keeps a category whose value is a reserved object key", () => {
+      // `map["__proto__"] = x` sets the prototype rather than defining an own property, so a
+      // category with that value would silently vanish from the export. Category values come from
+      // the data and can be any string.
+      const categories = makeSet(["__proto__", "b"])
+      runInAction(() => {
+        categories.setShapeForCategory("__proto__", "star")
+        categories.setShapeForCategory("b", "plus")
+      })
+
+      expect(categories.shapeForCategory("__proto__")).toBe("star")
+      expect(Object.keys(categories.shapeMap).sort()).toEqual(["__proto__", "b"])
+      expect(categories.shapeMap.__proto__).toBe("star")
+    })
+
     it("is observable, so the renderer re-reads when a shape changes", () => {
       const categories = makeSet(["a"])
       const fn = jest.fn()
--- a/v3/src/models/data/category-set.test.ts
+++ b/v3/src/models/data/category-set.test.ts
@@ -437,4 +437,39 @@ describe("CategorySet", () => {
     })
   })
 
+
+  describe("category values that collide with object keys", () => {
+    // Category values come from the data and can be any string, including ones that mean something
+    // to a plain JavaScript object.
+    const makeSet = (values: string[]) => {
+      const tree = Tree.create({
+        attribute: Attribute.create({ id: "aId", name: "a" }),
+        categories: { attribute: "aId" }
+      })
+      values.forEach(v => tree.attribute.addValue(v))
+      return tree.categories
+    }
+
+    it("keeps a color assigned to a category named __proto__", () => {
+      const categories = makeSet(["__proto__", "b"])
+      runInAction(() => categories.setColorForCategory("__proto__", "#ff0000"))
+
+      expect(categories.colorForCategory("__proto__")).toBe("#ff0000")
+      expect(Object.keys(categories.colorMap)).toContain("__proto__")
+    })
+
+    it("does not report an inherited member as a color", () => {
+      // on a normal object, colorMap["constructor"] would return a function
+      const categories = makeSet(["a"])
+      expect(typeof categories.colorForCategory("constructor")).not.toBe("function")
+      expect(categories.colorForCategory("constructor")).toBeUndefined()
+    })
+
+    it("assigns a color to a category named constructor like any other", () => {
+      const categories = makeSet(["constructor", "b"])
+      runInAction(() => categories.setColorForCategory("constructor", "#00ff00"))
+      expect(categories.colorForCategory("constructor")).toBe("#00ff00")
+    })
+  })
+
 })
--- a/v3/src/models/data/category-set.test.ts
+++ b/v3/src/models/data/category-set.test.ts
@@ -406,21 +406,6 @@ describe("CategorySet", () => {
     })
 
 
-    it("keeps a category whose value is a reserved object key", () => {
-      // `map["__proto__"] = x` sets the prototype rather than defining an own property, so a
-      // category with that value would silently vanish from the export. Category values come from
-      // the data and can be any string.
-      const categories = makeSet(["__proto__", "b"])
-      runInAction(() => {
-        categories.setShapeForCategory("__proto__", "star")
-        categories.setShapeForCategory("b", "plus")
-      })
-
-      expect(categories.shapeForCategory("__proto__")).toBe("star")
-      expect(Object.keys(categories.shapeMap).sort()).toEqual(["__proto__", "b"])
-      expect(categories.shapeMap.__proto__).toBe("star")
-    })
-
     it("is observable, so the renderer re-reads when a shape changes", () => {
       const categories = makeSet(["a"])
       const fn = jest.fn()
@@ -456,6 +441,18 @@ describe("CategorySet", () => {
 
       expect(categories.colorForCategory("__proto__")).toBe("#ff0000")
       expect(Object.keys(categories.colorMap)).toContain("__proto__")
+    })
+
+    it("keeps a shape assigned to a category named __proto__", () => {
+      const categories = makeSet(["__proto__", "b"])
+      runInAction(() => {
+        categories.setShapeForCategory("__proto__", "star")
+        categories.setShapeForCategory("b", "plus")
+      })
+
+      expect(categories.shapeForCategory("__proto__")).toBe("star")
+      expect(Object.keys(categories.shapeMap).sort()).toEqual(["__proto__", "b"])
+      expect(categories.shapeMap.__proto__).toBe("star")
     })
 
     it("does not report an inherited member as a color", () => {
--- a/v3/src/models/data/category-set.test.ts
+++ b/v3/src/models/data/category-set.test.ts
@@ -5,7 +5,6 @@ import { jestSpyConsole } from "../../test/jest-spy-console"
 import { Attribute, IAttribute } from "./attribute"
 import { CategorySet, ICategorySet, createProvisionalCategorySet, getProvisionalDataSet } from "./category-set"
 import { DataSet } from "./data-set"
-import { onAnyAction } from "../../utilities/mst-utils"
 
 describe("CategorySet", () => {
   const Tree = types.model("Tree", {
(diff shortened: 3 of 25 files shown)


v3/src/components/data-display/models/display-item-description-model.ts
```diff
@@ -1,10 +1,14 @@
...
 import {applyModelChange} from "../../../models/history/apply-model-change"
 import {defaultPointColor, defaultStrokeColor, kellyColors} from "../../../utilities/color-utils"
+import {kDefaultPointShape, PointShape, pointShapeOrDefault} from "../../../utilities/point-shape-utils"
 
 export const DisplayItemDescriptionModel = types
   .model("DisplayItemDescriptionModel", {
     _itemColors: types.optional(types.array(types.string), [defaultPointColor]),
+    // The shape used when no legend attribute assigns one per category. Scalar rather than an
+    // array like _itemColors: color varies per plot index for multi-y plots, shape does not.
+    _itemShape: types.optional(types.string, kDefaultPointShape),
```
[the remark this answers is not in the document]
kswenson: Good catch — fixed. Switched the field to `types.maybe` and made the setter clear it when the default is chosen, so absence means the default, the same rule the per-category shapes already follow. The getter already resolved undefined to circle, so nothing else changed.

Worth noting for the record that what keeps a saved document unchanged is `JSON.stringify` dropping the undefined value — `maybe` leaves the key present in the in-memory snapshot. The test asserts against the serialized JSON for that reason; asserting on the snapshot object would have passed while proving nothing.

v3/src/models/data/category-set.ts
```diff
@@ -139,6 +140,9 @@ export const CategorySet = types.model("CategorySet", {
   }),
   // user color assignments to categories in an attribute
   colors: types.map(types.string),
+  // user point-shape assignments to categories in an attribute. Stored as strings for the same
+  // reason as colors: the map holds whatever a saved document carries, and reads resolve it.
+  shapes: types.map(types.string),
```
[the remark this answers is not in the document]
kswenson: We considered this but decided against it. The claim is accurate — `shapes: {}` does appear in every CategorySet snapshot — but the `colors` map beside it materializes identically and predates this change, so the snapshot was never unaffected by an empty map. Omitting it would need a `snapshotProcessor` and would make `shapes` behave differently from its sibling, which seems a poor trade for twelve bytes.

You did catch a real overstatement in the commit message, though: "byte-for-byte unaffected" was about the v2 export, where `categoryShapes` is omitted entirely when empty, and it does not hold for v3 snapshots.

v3/src/models/data/category-set.ts
```diff
@@ -213,6 +212,20 @@ export const CategorySet = types.model("CategorySet", {
...
+  },
+  // Only the categories carrying an explicit shape. Categories at the default are omitted, so
+  // exports stay empty until a user actually assigns one.
+  get shapeMap(): Record<string, PointShape> {
+    const map: Record<string, PointShape> = {}
+    self.shapes.forEach((shape, category) => {
+      map[category] = pointShapeOrDefault(shape)
+    })
+    return map
   }
```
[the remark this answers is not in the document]
kswenson: Good catch — fixed with `Object.fromEntries`. Confirmed the failure first: assigning `map["__proto__"]` leaves `Object.keys` empty, so that category's shape really did vanish from the export.

There's now a test for it, which itself had to be corrected: building the input as an object literal `{ __proto__: "star" }` also sets the prototype, so the first version passed without exercising anything.

`colorMap` just above has the same pattern and the same latent problem. Left alone as pre-existing rather than widening this story, but worth its own fix.

v3/src/models/data/v2-category-set-importer.ts
```diff
@@ -9,11 +10,28 @@ import { MinimalMovesFinder } from "./minimal-moves-finder"
...
+   * Every entry here is a deliberate user assignment. Colors differ: some paths assign them
+   * automatically by category position, which is why the color loop below only keeps a color
+   * that differs from the one that position would have produced. Shapes have no such generated
+   * noise to age out, so a category whose cases are deleted and later restored keeps the shape
+   * the user chose for it — which is also what v2 does, deliberately, for the same reason.
+   */
+  const shapes: Record<string, string> = {}
+  Object.entries(categoryShapes ?? {}).forEach(([category, shape]) => {
+    if (isPointShape(shape) && shape !== kDefaultPointShape) shapes[category] = shape
+  })
```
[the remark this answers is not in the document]
kswenson: Fixed the same way, with `Object.fromEntries` over the filtered entries. The test builds its input with `JSON.parse`, which is how the data actually arrives from a v2 document and does create `__proto__` as a real own property.

v3/src/components/data-display/data-display-notifications.ts
```diff
@@ -56,6 +57,14 @@ export function changePointColorNotification(
   return updateTileNotification("change point color", { color, category }, tile)
 }
 
+// Shape has no V2 counterpart, so there is no legacy op string to match. `category` is undefined
+// when no legend attribute is assigned and the shape applies to every point.
+export function changePointShapeNotification(
+  tile: ITileModel | undefined, shape: PointShape, category?: string
+) {
+  return updateTileNotification("change point shape", { shape, category }, tile)
```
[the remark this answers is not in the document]
kswenson: Good catch — added. Four cases alongside the existing helpers: the operation string and shape payload, the category being present when one is assigned and absent when the shape applies to every point, that it emits on map as well as graph, and the missing-tile behavior.

kswenson:

| PR | Story | What it is |
|---|---|---|

**Superseded in #2696 — please review these as they stand here and don't file them against this PR:**

- `shapeForCategory` takes no `shapeIfUnset`, so a category with no shape of its own resolves to circle rather than inheriting the display's shape. #2696 adds inheritance, which is what stops a shape chosen before a legend exists from being discarded the moment one is added.
- `setShapeForCategory` deletes the entry when the shape is circle, because at this point in the stack absence means "the default". Once inheritance lands, absence means "inherit", so #2696 stores every choice including circle — and drops the matching filter in `v2-category-set-importer.ts`, which would otherwise discard an explicit circle on reload.
- `getLegendShapeForCase` has no childmost-collection guard. #2696 adds the one `getLegendColorForCase` already had, and makes the two share it.

**Worth your attention here:** the v2 round-trip strategy. Shapes travel in the attribute's `v3` namespace rather than `_categoryMap`, because v2's `updateCategoryMap` treats every unrecognized key as a category and writes it back — a shape stored there would surface as a phantom legend category in v2.

v3/src/utilities/point-shape-utils.ts
```diff
@@ -0,0 +1,26 @@
...
+ * components, the way color-utils already serves both.
+ */
+export const PointShapes = ["circle", "square", "triangle", "diamond", "star", "plus", "x"] as const
+
+export type PointShape = typeof PointShapes[number]
+
+export const kDefaultPointShape: PointShape = "circle"
+
+export function isPointShape(value?: string): value is PointShape {
+  return PointShapes.includes(value as PointShape)
```
tealefristoe: Maybe it's small enough that it doesn't matter, but this is O(n). Could be O(1) by making a `set` out of `PointShapes`.

v3/src/v2/codap-v2-data-context-types.ts
```diff
@@ -55,6 +55,20 @@ export interface ICodapV2Attribute {
...
+   * `__order`, `stroke-color` and `stroke-transparency` as a category: it appends unknown keys to
+   * __order and writes the result back out, so an extra key there becomes a phantom category in
+   * the user's legend and is persisted. An unknown key here is merely mixed onto the SproutCore
+   * model and dropped by the attribute's toArchive allowlist, so v2 is unaffected and a v2
+   * re-save loses the value rather than corrupting the document.
+   */
+  v3?: {
+    // per-legend-category point shapes, keyed by category value; unset categories use the default
+    categoryShapes?: Record<string, string>
+  }
```
tealefristoe, line 71: Is there a plan for dropping support for v2 at some point, so we don't have to maintain backwards compatibility?

v3/src/models/data/category-set.test.ts
```diff
@@ -353,25 +352,124 @@ describe("CategorySet", () => {
...
+        categories.setShapeForCategory("b", "plus")
+      })
 
-    const fn = jest.fn()
-    const disposer = onAnyAction(tree.categories, action => {
-      if (tree.categories.userActionNames.includes(action.name)) {
+      expect(categories.shapeForCategory("__proto__")).toBe("star")
+      expect(Object.keys(categories.shapeMap).sort()).toEqual(["__proto__", "b"])
+      expect(categories.shapeMap.__proto__).toBe("star")
+    })
```
tealefristoe: Should go under "category values that collide with object keys", and the first test there is basically this same test.

v3/src/models/data/category-set.ts
```diff
@@ -139,6 +140,9 @@ export const CategorySet = types.model("CategorySet", {
   }),
   // user color assignments to categories in an attribute
   colors: types.map(types.string),
+  // user point-shape assignments to categories in an attribute. Stored as strings for the same
+  // reason as colors: the map holds whatever a saved document carries, and reads resolve it.
```
tealefristoe:
```suggestion
  // user point-shape assignments to categories in an attribute
```
Colors doesn't need a comment like this, why does shapes?

v3/src/models/data/category-set.ts
```diff
@@ -259,6 +286,15 @@ export const CategorySet = types.model("CategorySet", {
       self.colors.delete(value)
     }
   },
+  // Storing the default is stored as absence, so a document only carries the shapes a user chose
+  // and a category reverted to circle round-trips as an unset entry.
```
tealefristoe, line 286:
```suggestion
  // and a category reverted to default round-trips as an unset entry.
```

v3/src/models/data/v2-category-set-importer.test.ts
```diff
@@ -60,4 +60,74 @@ describe("importV2CategorySet", () => {
...
+      expect(Object.keys(result?.colors ?? {})).not.toContain("constructor")
+    })
+
+    it("imports assigned shapes", () => {
+      const result = importV2CategorySet(makeAttribute(), undefined, { land: "star", water: "diamond" })
+      expect(result?.shapes).toEqual({ land: "star", water: "diamond" })
+    })
+
+    it("creates a category set when shapes are the only thing to restore", () => {
+      // no colors and no order, so nothing else would justify a category set
```
tealefristoe: Does this test cover anything the previous test doesn't cover?

v3/src/models/data/v2-category-set-importer.ts
```diff
@@ -9,10 +10,31 @@ import { MinimalMovesFinder } from "./minimal-moves-finder"
...
+
+  /*
+   * Shapes come from the attribute's v3 namespace, and are taken as given rather than filtered
+   * against the categories currently in the data.
+   *
+   * Every entry here is a deliberate user assignment. Colors differ: some paths assign them
+   * automatically by category position, which is why the color loop below only keeps a color
+   * that differs from the one that position would have produced. Shapes have no such generated
+   * noise to age out, so a category whose cases are deleted and later restored keeps the shape
+   * the user chose for it — which is also what v2 does, deliberately, for the same reason.
```
tealefristoe: 1. The sentence that starts with "Colors differ" would be more appropriate in a comment where colors is defined, if it's worth keeping at all.
2. The last sentence talks about how v2 handles shapes, but much of this PR goes out of its way to make sure v2 never sees shapes. Either I'm missing something, or there's a problem with the sentence. I also feel like it's not worth including in a comment because it's how I would expect things to work.

v3/src/models/data/v2-category-set-importer.ts
```diff
@@ -9,10 +10,31 @@ import { MinimalMovesFinder } from "./minimal-moves-finder"
...
+   */
+  // fromEntries rather than assignment: `shapes[category] =` with a category value of `__proto__`
+  // sets the prototype instead of defining an own property, losing that category's shape.
```
tealefristoe: How many times have I read a variant of this same comment? 😵

v3/src/models/shared/data-set-metadata.test.ts
```diff
@@ -531,4 +531,43 @@ describe("DataSetMetadata", () => {
...
+
+      categorySet.setShapeForCategory("1", "star")
+
+      expect(tree.metadata.attributes.size).toBe(1)
+      expect(tree.metadata.provisionalCategories.size).toBe(0)
+      // the promoted set carries the shape, so it survives into the saved document
+      expect(tree.metadata.getCategorySet("aId")?.shapeForCategory("1")).toBe("star")
+    })
+
+    it("promotes a provisional category set when categories are reordered", () => {
```
tealefristoe: Indentation, also should add a blank line at the top.

v3/src/models/shared/data-set-metadata.test.ts
```diff
@@ -531,4 +531,43 @@ describe("DataSetMetadata", () => {
...
+    const categorySet = tree.metadata.getCategorySet("aId")!
+    expect(tree.metadata.provisionalCategories.size).toBe(1)
+
+    categorySet.move("3", "1")
+
+    expect(tree.metadata.attributes.size).toBe(1)
+    expect(tree.metadata.provisionalCategories.size).toBe(0)
+  })
+
+  it("does not promote for a shape set back to the default", () => {
```
tealefristoe, line 564: What should happen if you go default -> star -> circle?

v3/src/v2/codap-v2-data-set-importer.ts
```diff
@@ -208,14 +208,16 @@ export class CodapV2DataSetImporter {
...
     attributes.forEach(v2Attr => {
       const {
-        guid, colormap, _categoryMap
+        guid, colormap, _categoryMap, v3
       } = v2Attr
       const attribute = data.getAttribute(toV3AttrId(guid))
       if (attribute) {
         const categorySetInput: Maybe<V2CategorySetInput> = _categoryMap || colormap
-        if (categorySetInput) {
+        // Shapes can be the only thing to restore, so they warrant a category set on their own.
```
tealefristoe:
```suggestion
```

v3/src/v2/codap-v2-types.ts
```diff
@@ -459,6 +459,9 @@ export interface ICodapV2GraphStorage extends ICodapV2BaseComponentStorage {
   // v3 extensions
   v3?: {
     filterFormula?: string
+    // The point shape used when no legend attribute assigns one per category. Per-category
+    // shapes are attribute state and live on the attribute's own v3 namespace.
```
tealefristoe: Almost the exact same comment as `codap-v2-type-utils.ts` lines 106-107, with some subtle changes. Why the changes? Why duplicate the comment in the first place?

v3/src/v2/v2-document-round-trip.test.ts
```diff
@@ -51,3 +54,92 @@ describe("v2 document round-trip (CODAP-1348)", () => {
...
+    shapeFor("water", "diamond")
+
+    const { restoredData } = await roundTrip(document)
+    const restoredAttr = restoredData.attrFromName("habitat")!
+    const restoredSet = getMetadataFromDataSet(restoredData)!.getCategorySet(restoredAttr.id)!
+
+    expect(restoredSet.shapeForCategory("land")).toBe("star")
+    expect(restoredSet.shapeForCategory("water")).toBe("diamond")
+    // a category left at the default comes back at the default
+    expect(restoredSet.shapeForCategory("both")).toBe("circle")
```
tealefristoe:
```suggestion
    expect(restoredSet.shapeForCategory("both")).toBe(kDefaultPointShape)
```

tealefristoe (approved):
Looks good 👍 

Main feedback:
1. Worth looking through the comments and pruning duplicates and unnecessary/distracting comments. I flagged a few, but not all.
2. Probably outside of your control, but when will we stop supporting v2? Something like a quarter of the complexity of this PR comes from supporting v2.
3. It's probably too late to change it, but I'm not sure about explicitly removing the default shape from the category map.
  - If I go default -> star -> circle, I think there's a good chance I want the shape to be circle, not the default, which might change in the future, and which as a user I probably have no concept of.
  - It would be simpler to just save the default shape when the user sets it.
Maybe there's some subtle reasons to do this that I'm missing?

kswenson:
Thanks — a genuinely useful pass. Everything actionable is in `1014d6b49` and `85d56cb07`; replies on each thread.

**On comments (1).** Taken — and since you said you'd flagged a few but not all, I made a pass for the rest rather than just the ones you named. Three more duplications you hadn't flagged:

- Why shapes go in the `v3` namespace rather than `_categoryMap` was explained **four times**, twice at length. It's stated once now, on the `v3` field itself, since that field exists for no other reason.
- That promoting a category set replaces the instance was said both at the promotion and where callers resolve one.
- The `PointShapes` header described what the list is before saying where it lives; only the second isn't visible from the code.

178 → 155 comment lines across the PR. The cuts are repetition, not explanation.

Worth noting two of your flags turned out to be *wrong* comments rather than merely wordy ones — one asserted something untrue about how v2 handles shapes. That's the better argument for pruning: length was hiding errors.

**On v2 (2).** Kirk's answer, which reframes it — we're not supporting the v2 *application*, we're supporting the v2 *file format*, which is part of the plugin API and effectively permanent. Concretely: StoryBuilder stores V2 document states, so without the round-trip, point shapes would silently disappear from any StoryBuilder moment. The complexity buys plugin compatibility rather than deferring a sunset.

**On removing the default from the category map (3).** You're right, and it's already fixed one PR up — Copilot reached the same conclusion on #2696 from the other direction.

Your `default -> star -> circle` case is exactly what broke. Once a category with no shape *inherits* the display's shape (also #2696), absence can't also mean "circle" — so choosing Circle deleted the entry and the getter handed back star. Circle was unselectable for any category whenever the display was set to anything else. #2696 stores every choice including circle, and drops the matching filter in the v2 importer so it survives a reload.

So this PR does show the behavior you're questioning, and it doesn't survive the stack.

v3/src/utilities/point-shape-utils.ts
kswenson: Fixed — it's a Set now. Cheap either way at seven entries, but it's called per value on every import and there's no reason not to.

v3/src/v2/codap-v2-data-context-types.ts
kswenson: Kirk's answer, which reframes it: we're not supporting the v2 *application*, we're supporting the v2 *file format*, which is part of the plugin API and effectively permanent.

Concretely for this PR — StoryBuilder stores V2 document states. Without the round-trip, point shapes would silently vanish from any StoryBuilder moment. So this isn't legacy debt waiting on a sunset date; it's what makes the feature work with an existing plugin.

v3/src/models/data/category-set.test.ts
kswenson: Moved. Kept rather than deleted — the neighboring test covers colors and this one covers shapes, and `__proto__` has bitten this feature more than once, so I'd rather both paths stay pinned.

v3/src/models/data/category-set.ts
kswenson: Right — no reason. Taken.

v3/src/models/data/category-set.ts
kswenson: Taken in spirit, but note this comment is replaced entirely in #2696: the setter stops deleting the default, so it becomes "every choice is stored, circle included". See your other comment about `default -> star -> circle` — you were right, and that's where it's fixed.

Also, my line reads "Storing the default is stored as absence", which isn't a sentence. Thanks for making me look at it.

v3/src/models/data/v2-category-set-importer.test.ts
kswenson: No — you're right, it doesn't. The test above passes only shapes and reads `result?.shapes`, which can't match unless the set was created, so the assertion was already covered.

Deleted, with the intent folded into the remaining test's name and comment.

v3/src/models/data/v2-category-set-importer.ts
kswenson: Both taken, and the second is a real error rather than a style point.

That last sentence claims v2 does the same thing with shapes for the same reason. v2 has no point shapes at all — which is precisely why they travel in the `v3` namespace instead of `_categoryMap`. I was describing v2's *color* behavior and wrote it as though it covered shapes. It's gone.

The "Colors differ" aside is gone too; what's left says only why shapes aren't filtered, which is the part that isn't visible from the code.

v3/src/models/data/v2-category-set-importer.ts
kswenson: Too many. 😄 It was the same explanation at length in two files.

One line at each site now — enough to stop someone "simplifying" `fromEntries` back into an assignment, without retelling the story.

v3/src/models/shared/data-set-metadata.test.ts
kswenson: Fixed. Three tests had drifted from the file's 2-space `it`, including their closing braces.

v3/src/models/shared/data-set-metadata.test.ts
kswenson: It stores circle — and you found this by reasoning about it, which is the same conclusion Copilot reached from the other direction.

At this point in the stack absence means "the default", so choosing Circle deleted the entry. #2696 makes absence mean "inherit the display's shape", at which point deleting is actively wrong: with a display shape of star, choosing Circle for a category gave star straight back, so circle was unselectable. Every choice is stored there, and the v2 importer's matching filter is dropped too.

v3/src/v2/codap-v2-data-set-importer.ts
kswenson: Gone — the condition on the next line says it.

v3/src/v2/codap-v2-types.ts
kswenson: No good reason for either. The two had drifted apart purely from being written at different times.

The type declaration keeps the comment, since that's where a reader meets the field; the props interface that feeds it no longer repeats it.

v3/src/v2/v2-document-round-trip.test.ts
kswenson: Taken — the assertion is about the round trip, not about which shape is the default.

PR #2693, opened 2026-09-04
CODAP-1506: add per-category point shape controls

kswenson:
Completes CODAP-1506, alongside #2691 which did the structural half. Stacked on
#2692 (the model), which is stacked on #2691 — review those first. I'll retarget
each as its parent merges.

Adds the shape dropdown to every legend category row, and to the single *Points*
row shown when no legend attribute is assigned. Behind a new `pointShapes`
feature flag owned by LEADS, so `features=LEADS` enables the whole set.

**Nothing renders a shape yet.** Every point still draws as a circle until
CODAP-1504 lands the geometry. The control writes through to the model from
#2692, so the UI is exercisable now and shapes appear once the renderer can draw
them — which is why the UI is landing before the geometry.

## The controls

The trigger is icon only: with an icon *and* a text label the category name is
squeezed to ~55px and a value like "water" truncates. The open menu carries icon
plus label and marks the current shape in bold. Both take their fill from the
category's own colour, so a row's two controls describe the same category.

Trigger geometry mirrors `.color-picker-thumb` beside it — same padding, same
gap, same 24px inner element — so the two controls in a row are the same size by
construction. Popover values follow the prototype's own `.pt-popover` /
`.pt-listbox`.

## Assets

Seven SVGs exported from the design prototype, adapted to the repo's icon
convention: `fill="currentColor"` so rows can tint them, and the decorative
`title`/`role` attributes dropped so the control supplies the accessible name.

They use the `.nosvgo.svg` extension, which **keeps SVGR** — still React
components, which the `currentColor` tinting requires — while skipping svgo.
That matters because svgo's `removeViewBox` drops the viewBox once `width`/
`height` are present, and those attributes are what give the glyph intrinsic
dimensions. Without them an svg contributes nothing to its parent's intrinsic
width, and the trigger sized itself to padding + arrow alone.

Note the sibling `.nosvgr.svg` is a different rule — that one yields a URL
asset, which can't inherit `color`.

## Two things reviewers should know

**Glyphs are pinned, not laid out.** Each sits in a plain sized box with
`position: absolute; inset: 0`. Laid out as flex items, the svgs rendered
*outside* their parent's box — measured 3px above and left of it in the trigger,
and all seven stacked at one point in the menu. I could not identify the cause
from the stylesheet, the DOM, or the served bundle; taking them out of the flex
layout path removes the failure rather than explaining it. Flagging it as
containment rather than a root-cause fix.

**`shouldFlip={false}` is a workaround.** A menu react-aria decides to flip
above its trigger doesn't render at all, which makes the control unusable rather
than awkward: with the palette near the bottom of the window there's no room
below, so the menu could never be opened. Pinned below, it always appears and
scrolls within its `max-height`. The colour picker in this same palette disables
flipping for the same reason, so the fault is in how popovers position within
the palette rather than in either control. Filing separately; the code comment
says to remove `shouldFlip` once it's fixed.

## Row alignment

The category row distributes with `space-between`, so a control following the
label tracked the label's width and the controls staggered between rows. The
label now absorbs the free space and truncates, pinning both controls to the
right edge and giving the category name the width the panel was widened for.

## Testing

11 new tests. Mutation-verified — each of these fails at least one: ignoring the
feature flag, forcing it off, writing a shape to the wrong category, letting
`SelectValue` render the selected item's children (which duplicated the glyph),
and dropping the hidden value that gives the trigger its accessible name.

Manually verified in the browser across the palette: trigger and menu geometry,
per-category tinting, keyboard operation, and that the menu escapes the category
scroll container rather than being clipped — the AC that called this out as
"the first menu CODAP opens from inside that scrolling list".

`npm run build:tsc`, `npm run lint`, and the full suite (3814 tests / 357 suites)
all pass.


v3/src/components/data-display/inspector/legend-color-controls.tsx
```diff
@@ -137,9 +166,19 @@ export const LegendColorControls = observer(function LegendColorControls(
 
   if (attrType === "color") return null
 
+  // With no legend attribute there are no categories to list, so the same two controls apply to
+  // every point and sit in a single row.
+  const showShape = isFeatureEnabled("pointShapes")
```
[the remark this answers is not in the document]
kswenson: Good catch — accepted. `map-layers-palette.tsx` excludes only pin layers, so polygon layers do get these controls, and with no legend the row was relabeled "Points" and offered a shape that can't reach polygon rendering.

Both gates now include the same `pointSizeMultiplier >= 0` check that already hides the Point Size slider. Since the per-category gate is in a child component that doesn't receive the display description, the decision is made once in `LegendColorControls` and passed down. Added tests covering both the no-legend and categorical polygon cases.

Fixed in 7efc0b0a7.

kswenson:

| PR | Story | What it is |
|---|---|---|

**Superseded in #2696 — no need to flag these here:**

- Three British spellings in comments: "colour swatch" and "behaviour" in `display-item-format-control.scss`, "colour picker" in `point-shape-setting.tsx`. All US-spelled in #2696. (The `colour-coding.pdf` URL in `color-utils.ts` is a real external filename and stays as is.)
- The shape control's sizing and alignment in `display-item-format-control.scss` is iterated again in #2696; the final geometry lives there.

The model code visible as context here is #2692's, so the three items noted on that PR apply to it.

**Worth your attention here:** the react-aria `Select` wiring in `point-shape-setting.tsx`. Two non-obvious things are load-bearing and commented — `shouldFlip={false}` works around CODAP-1510 (palette popovers don't render when they need to flip above the trigger), and the trigger carries `point-shape-thumb` rather than `react-aria-Button` because a plain `className` replaces react-aria's default class.

v3/src/components/data-display/inspector/display-item-format-control.scss
```diff
@@ -68,6 +63,145 @@
...
+    align-items: center;
+    // Less left padding than right: the glyph already carries the artboard's own margin on its
+    // left (a circle spans 16 of the 24 units, the room the star needs), while the disclosure
+    // arrow has no such inset and needs the space.
+    gap: 4px;
+    // contains the absolutely-positioned visually-hidden value, so it cannot escape to the palette
+    position: relative;
+    // Geometry mirrors .color-picker-thumb beside it -- same padding, same gap, same 24px inner
+    // element -- so the two controls in a row are the same size. Height is left to the content, as
+    // it is there, rather than fixed, so they cannot drift apart.
```
tealefristoe: Sounds like this should be a mixin rather than duplicated fields. Also, why does `.color-picker-thumb padding` have `!important` but this one doesn't?

v3/src/components/data-display/inspector/display-item-format-control.scss
```diff
@@ -68,6 +63,145 @@
...
+      position: absolute;
+      inset: 0;
+      display: block;
+      width: 100%;
+      height: 100%;
+    }
+
+    // Matches the color picker's disclosure arrow, so the two controls in a row read as a pair.
+    // margin-left auto pins it right and leaves the glyph left-packed, the way the prototype
+    // flexes its .select-value and .select-arrow. Sized by border rather than width/height, so
```
tealefristoe: What prototype?

v3/src/components/data-display/inspector/display-item-format-control.scss
```diff
@@ -68,6 +63,145 @@
...
+      transition: transform 0.15s ease;
+    }
+
+    &.open .point-shape-arrow {
+      transform: rotate(180deg);
+    }
+  }
+}
+
+// Values follow the prototype's .pt-popover / .pt-listbox so the open menu matches the design.
```
tealefristoe:
```suggestion
```

v3/src/components/data-display/inspector/display-item-format-control.scss
```diff
@@ -68,6 +63,145 @@
...
+    &[data-selected] {
+      font-weight: bold;
+    }
+
+    &[data-focus-visible] {
+      outline: 2px solid vars.$focus-outline-color;
+      outline-offset: -2px;
+    }
+
+    // A plain 24px box, as in the prototype's `.pt-listbox .shape-glyph { flex: 0 0 24px }`.
```
tealefristoe:
```suggestion
```

v3/src/components/data-display/inspector/display-item-format-control.scss
```diff
@@ -68,6 +63,145 @@
...
+    }
+
+    // Pinned rather than laid out, for the same reason as the trigger's glyph.
+    .point-shape-item-svg {
+      position: absolute;
+      inset: 0;
+      display: block;
+      width: 100%;
+      height: 100%;
+    }
```
tealefristoe, line 185: These are almost the same as `point-shape-thumb-value` and `point-shape-thumb-glyph` (except oddly `point-shape-item-glyph` is almost equivalent to `point-shape-thumb-value` instead of `-glyph`). Good opportunity to combine css that's likely intended to stay in sync.

v3/src/components/data-display/inspector/display-item-format-control.scss
tealefristoe, line 1: I think it's worth going through this file to prune comments and consolidate css that's supposed to stay in sync with variables or mixins. In particular, there are comments that describe other css that should be in sync--it would be better to just make the code keep the different fields in sync, rather than relying on comments.

v3/src/components/data-display/inspector/legend-color-controls.tsx
```diff
@@ -34,6 +38,13 @@ export const LegendColorControls = observer(function LegendColorControls(
   const { tile } = useTileModelContext()
   const legendAttrID = dataConfiguration.attributeID("legend")
   const attrType = dataConfiguration.attributeType("legend")
+  /*
+   * Polygons have no point to shape. The map mounts these controls for its polygon layers too, and
+   * a polygon layer marks itself with a negative point size -- the same sentinel that already hides
+   * the Point Size slider. Without this the controls would offer a shape that cannot affect
+   * anything, and relabel a polygon fill row as "Points".
+   */
+  const showShape = isFeatureEnabled("pointShapes") && displayItemDescription.pointSizeMultiplier >= 0
```
tealefristoe: Instead of a comment that describes what `displayItemDescription.pointSizeMultiplier >= 0` means here, it would be better to make a function (a view on `DisplayItemDescriptionModel`?) with a descriptive name that can be used here as well as when hiding the Point Size slider.

v3/src/components/data-display/inspector/legend-color-controls.tsx
```diff
@@ -96,19 +131,74 @@ export const LegendColorControls = observer(function LegendColorControls(
...
+                    <stop key={i} offset={`${stop.offset * 100}%`} stopColor={stop.color} />
+                  ))}
+                </linearGradient>
+              </defs>
+            </svg>
+          </If>
+          <PointShapeSetting propertyLabel={t("V3.Inspector.pointShape")}
+                            shape={displayItemDescription.pointShape}
+                            color={displayItemDescription.pointColor}
+                            fillGradientId={legendBandStops.length > 0 ? gradientId : undefined}
```
tealefristoe: I'd define `hasFillGradient = legendBandStops.length > 0` below the definition of `legendBandStops`, then use it here and in the condition on line 166 so they stay in sync.

v3/src/components/data-display/inspector/legend-color-controls.tsx
```diff
@@ -150,11 +249,15 @@ export const LegendColorControls = observer(function LegendColorControls(
...
+  showShape: boolean
   onCatPointColorChange: (color: string, cat: string) => void
+  onCatPointShapeChange: (shape: PointShape, cat: string) => void
 }
 
 const CategoricalColorControls = observer(function CategoricalColorControls(
-  { categories, dataConfiguration, onCatPointColorChange }: ICategoricalColorControlsProps
+  {
+    categories, dataConfiguration, showShape, onCatPointColorChange, onCatPointShapeChange
+  }: ICategoricalColorControlsProps
```
tealefristoe: Minor style thing, but I'd combine the top and bottom lines here with the lines above/below them.

v3/src/components/data-display/inspector/point-shape-setting.tsx
```diff
@@ -0,0 +1,154 @@
...
+      isOpen={isOpen}
+      onOpenChange={setIsOpen}
+      onSelectionChange={handleChange}
+      selectedKey={shape}
+      data-testid="point-shape-select"
+    >
+      {/*
+        * Icon only. With an icon and a text label the category name is squeezed to about 55px and
+        * a value like "water" truncates; the open menu carries the labels instead.
+        */}
```
tealefristoe:
```suggestion
```

v3/src/components/data-display/inspector/point-shape-setting.tsx
```diff
@@ -0,0 +1,154 @@
...
+      <Button className={clsx("point-shape-thumb", { open: isOpen })} excludeFromTabOrder={disabled}>
+        {/* The span is a plain 24x24 flex item; the glyph is pinned inside it with inset: 0
+            rather than laid out, so nothing about how an svg participates in flex layout can
+            displace it. */}
```
tealefristoe:
```suggestion
```

v3/src/components/data-display/inspector/point-shape-setting.tsx
```diff
@@ -0,0 +1,154 @@
...
+            data-testid="point-shape-glyph" style={glyphStyle} />
+        </span>
+        {/*
+          * Names the current shape for assistive technology, and gives the aria-labelledby that
+          * react-aria puts on the trigger a real element to point at. Without it that reference
+          * dangles, and the accessible name survives only by falling back to aria-label.
+          *
+          * Text only. Rendering the selected item's own children would put a second copy of the
+          * glyph in here, and this span is positioned absolutely, so those copies would escape
+          * the button and pile up over the palette.
```
tealefristoe:
```suggestion
```

v3/src/components/data-display/inspector/point-shape-setting.tsx
```diff
@@ -0,0 +1,154 @@
...
+        {/* The span is a plain 24x24 flex item; the glyph is pinned inside it with inset: 0
+            rather than laid out, so nothing about how an svg participates in flex layout can
+            displace it. */}
+        <span className="point-shape-thumb-value">
+          <ShapeIcon shape={shape} className="point-shape-thumb-glyph"
+            data-testid="point-shape-glyph" style={glyphStyle} />
+        </span>
+        {/*
+          * Names the current shape for assistive technology, and gives the aria-labelledby that
+          * react-aria puts on the trigger a real element to point at. Without it that reference
```
tealefristoe: This seems like a genuinely useful comment, but I don't understand the end of the first sentence.

v3/src/components/data-display/inspector/point-shape-setting.tsx
```diff
@@ -0,0 +1,154 @@
...
+          *
+          * Text only. Rendering the selected item's own children would put a second copy of the
+          * glyph in here, and this span is positioned absolutely, so those copies would escape
+          * the button and pile up over the palette.
+          */}
+        <SelectValue className="codap-visually-hidden">
+          {({ selectedText }) => selectedText}
+        </SelectValue>
+        {/* A real element, as in the prototype, rather than a pseudo-element: a zero-size bordered
+            ::after collapses wherever box-sizing is border-box. */}
```
tealefristoe:
```suggestion
```

v3/src/components/data-display/inspector/point-shape-setting.tsx
```diff
@@ -0,0 +1,154 @@
...
+        * the control unusable rather than merely awkward: with the palette near the bottom of the
+        * window there is no room below, so the menu could never be opened. Pinned below the
+        * trigger it always appears, clipping and scrolling within its own max-height, which is
+        * the lesser failure.
+        *
+        * The colour picker in this same palette disables flipping for the same reason, so the
+        * fault lies in how popovers position within the palette rather than in this control.
+        */}
+      <Popover shouldFlip={false}
+        className={({ defaultClassName }) => `${defaultClassName} point-shape-popover`}>
```
tealefristoe: Minor, but I always try to use `clsx`/`classNames`.

v3/src/components/data-display/inspector/point-shape-setting.tsx
```diff
@@ -0,0 +1,154 @@
...
+        <ListBox>
+          {PointShapes.map(_shape => (
+            <ListBoxItem key={_shape} id={_shape} textValue={shapeLabel(_shape)}>
+              {/* Same containment as the trigger: the span is the flex item, the glyph is
+                  pinned inside it. */}
```
tealefristoe:
```suggestion
```

v3/src/components/data-display/inspector/legend-color-controls.test.tsx
```diff
@@ -528,4 +534,358 @@ describe("LegendBinCountInput", () => {
...
+        />
+      )
+
+      expect(screen.queryByTestId("point-shape-select")).not.toBeInTheDocument()
+      expect(screen.getByTestId("color-swatch-cat-a")).toBeInTheDocument()
+    })
+  })
+
+  describe("with no legend attribute", () => {
+    it("keeps today's single color row when the flag is off", () => {
```
tealefristoe:
```suggestion
    it("keeps a single color row when the flag is off", () => {
```
Maybe there's better wording for this, but I think "today's" will age poorly.

tealefristoe (approved):
Looks good 👍

v3/src/components/data-display/inspector/display-item-format-control.scss
kswenson: Both taken, and the `!important` question has a real answer that's now in the code rather than a reply.

`.color-picker-thumb` *is* a react-aria Button, so the shared `.react-aria-Select .react-aria-Button` rules in `inspector-panel.scss` set its height and padding and have to be overridden. `.point-shape-thumb` passes a plain `className`, which **replaces** react-aria's default class rather than adding to it, so those rules never reach it. That asymmetry cost me a debugging round when I first built this, so it's worth the two lines.

A `palette-thumb` mixin now carries the dozen shared declarations.

v3/src/components/data-display/inspector/display-item-format-control.scss
kswenson: The design prototype the feature was built from — but you're right that naming it here is useless, since nothing in this file tells you where it is or lets you look. All three references are gone; the provenance lives in the geometry module and the commit history.

v3/src/components/data-display/inspector/display-item-format-control.scss
kswenson: Gone.

v3/src/components/data-display/inspector/display-item-format-control.scss
kswenson: Gone.

v3/src/components/data-display/inspector/display-item-format-control.scss
kswenson: Good catch on both counts, including the naming oddity — `point-shape-item-glyph` matches `point-shape-thumb-value`, not `-glyph`, which is confusing.

Two mixins now: `glyph-box` for the 24px box and `pinned-glyph` for the svg inside it, both used by the trigger and the menu items. I left the class names alone rather than renaming, since they're referenced from the tsx and the fill rule — happy to rename in a follow-up if the asymmetry still bothers you.

v3/src/components/data-display/inspector/display-item-format-control.scss
kswenson: Did the pass. Three mixins — `palette-thumb`, `glyph-box`, `pinned-glyph` — replace the declarations that were duplicated with comments saying they had to match.

Doing it turned up exactly the drift you were predicting: a comment claiming "Less left padding than right" sat over a uniform `padding: 3px`. It described a rule that no longer existed, which is the argument for the mixin rather than the comment.

423 → 407 lines, and the remaining comments explain things the CSS can't say — why the arrow is sized by border, why the trigger is `position: relative`.

v3/src/components/data-display/inspector/legend-color-controls.tsx
kswenson: Taken, and this is the better instinct — same as your `paletteClassName` suggestion on #2691: if a comment is explaining what an expression means, the expression should say it.

`DisplayItemDescriptionModel` now has `isPolygon`, used here and by the Point Size slider. That matters beyond tidiness: the two were *supposed* to ask the same question and didn't, which is how this PR shipped a shape control on polygon layers in the first place.

One thing worth reporting — adding the view broke three tests, because the mocks are plain objects carrying `pointSizeMultiplier` but no `isPolygon`. I derived it in the mocks so they can't disagree. Then mutation testing showed the real view had **no** coverage at all: forcing `isPolygon` to `false` broke nothing, because every test ran against a mock that reimplements it. There's a test on the actual model now.

v3/src/components/data-display/inspector/legend-color-controls.tsx
kswenson: Taken — `hasFillGradient` is defined below `legendBandStops` and used in both places.

v3/src/components/data-display/inspector/legend-color-controls.tsx
kswenson: Taken.

v3/src/components/data-display/inspector/point-shape-setting.tsx
kswenson: Gone.

v3/src/components/data-display/inspector/point-shape-setting.tsx
kswenson: Gone.

v3/src/components/data-display/inspector/point-shape-setting.tsx
kswenson: Gone.

v3/src/components/data-display/inspector/point-shape-setting.tsx
kswenson: Fair — that sentence put the mechanism before the point, and buried what the element is actually for.

Rewritten: it's the element react-aria's `aria-labelledby` points at, and without it the trigger is named only by its `aria-label` fallback. The second paragraph explains why it renders text rather than the selected item's children.

v3/src/components/data-display/inspector/point-shape-setting.tsx
kswenson: Gone.

v3/src/components/data-display/inspector/point-shape-setting.tsx
kswenson: Taken — `clsx` now.

v3/src/components/data-display/inspector/point-shape-setting.tsx
kswenson: Gone.

v3/src/components/data-display/inspector/legend-color-controls.test.tsx
kswenson: Taken — "today's" would have read oddly the moment it stopped being today.

PR #2694, opened 2026-09-08
feat: use OIDC authentication for V3 S3 deployments (DEV-185)

emcelroy:
[DEV-185](https://concord-consortium.atlassian.net/browse/DEV-185)

Migrates the v3 S3 deploys from long-lived AWS access keys to GitHub Actions OIDC. Companion to the [v2 migration on `main`](https://github.com/concord-consortium/codap/pull/2695).

Five workflows write to S3. All of them now assume an IAM role instead of using `V3_AWS_ACCESS_KEY_ID` / `V3_AWS_SECRET_ACCESS_KEY`:

| Workflow | Job | Deploys |
|---|---|---|
| `v3.yml` | `s3-deploy` | every push, to `models-resources/codap3/` |
| `release-v3-staging.yml` | `release` | promotes a version to `index-staging.html` and `staging` |
| `release-v3-production.yml` | `release` | promotes a version to `index.html` |
| `release-v3-beta.yml` | `release` | promotes a version to `beta` |
| `release-v3-ai4vs.yml` | `release` | promotes a version to `ai4vs` |

Each job requests an OIDC token and assumes `arn:aws:iam::612297603577:role/codap` via `aws-actions/configure-aws-credentials@v6`.

## AWS setup

The role was created with the canonical [`create-deploy-role.sh`](https://github.com/concord-consortium/starter-projects/blob/main/scripts/create-deploy-role.sh) script from starter-projects. Its trust policy is scoped to this repository and allows both the legacy and immutable OIDC subject formats.

This repo is named `codap` but deploys to `models-resources/codap3/`, so the shared `S3-deploy-by-role-tag` managed policy alone would not grant the right prefix. The role carries an extra inline `codap3-deploy` policy granting object access on `models-resources/codap3/*`.

## Also in this PR

Deletes `s3_website.yml` and `bin/s3_deploy.sh`, the Travis-era v2 deploy. Nothing calls them, there is no `.travis.yml` on any branch, and `master` already removed them in b9a1b1de.

## Follow-up

The v2 CI workflow on `master` still uses `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` to deploy to `models-resources/codap-dev/`. The separate companion PR against `master` will migrate it. The account-level AWS key secrets should not be removed until both PRs have merged and deployed successfully.

[DEV-185]: https://concord-consortium.atlassian.net/browse/DEV-185?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kswenson (approved):
👍 Nice, clean migration — approving. The following review, developed in conjunction with Claude Code 🤖, identified a few optional nits at the bottom, nothing blocking.

## What I checked

The conversion is mechanically uniform across all five workflows, and the one that differed
(`release-v3-staging.yml`, which held the credentials in a *job*-level `env:` block rather
than a step-level one) was converted correctly. Region plumbing survives the change: moving
from `AWS_DEFAULT_REGION` to the action's `aws-region` input still exports both `AWS_REGION`
and `AWS_DEFAULT_REGION` for the later steps in the job.

Dropping the key inputs to `s3-deploy-action@v1` in `v3.yml` is properly supported — its
`action.yml` marks both `required: false` ("Not needed if AWS credentials are already
configured in the environment") and `src/aws-credentials.ts` falls through to the environment,
picking up `AWS_SESSION_TOKEN` automatically. The `v1` tag currently points at the commit that
added that support. The added `deployments: write` is genuinely required rather than
belt-and-braces: the action calls `createDeployment`/`createDeploymentStatus`, and introducing
any `permissions:` block would otherwise have dropped that scope to none.

Best of all, this doesn't have to be taken on faith — the `s3-deploy` job ran green on the
branch and validates the role ARN, trust policy, action version, region, and token scopes all
at once:

```
Assuming role with OIDC
Authenticated as assumedRoleId AROAY5D6H7H4SJIF7SQYN:GitHubActions
Running: aws s3 sync ./v3/dist s3://models-resources/codap3/branch/oidc-auth-for-s3-deployment --delete ...
```

The run log also shows the token scope actually tightening (`Contents: read / Deployments:
write / Metadata: read`) versus the unrestricted write token that job had on `main`.

I verified the housekeeping claims independently rather than taking them on trust, and they
all hold up: no surviving reference anywhere to `V3_AWS_ACCESS_KEY_ID`/`V3_AWS_SECRET_ACCESS_KEY`,
`s3_website`, `s3_deploy`, `DEPLOY_DIR`, or `dist/travis`; no other workflow in
`.github/workflows/` touches AWS or S3, so nothing was missed; `b9a1b1de` is a genuine ancestor
of `origin/master` and its diff is exactly the deletion of those two files; no `s3_website` gem
in either `Gemfile` and no `.travis.yml` on any branch; every relative link in the new
`deploy.md` and the `README.md` addition resolves. And the doc's claim that the role also
covers v2 checks out — `origin/master:.github/workflows/ci.yml` does deploy to
`models-resources/codap-dev/` via `s3-deploy-action`, still on static keys.

The red `codecov/project` mark is a baseline artifact — this PR changes nothing under `v3/src`.

## A note on security, for the record

Three concerns came up while reviewing that I want to record as *checked and dismissed*, since
each sounds alarming and is likely to be raised again by someone reading these workflows later:

- **The trust policy is branch-unrestricted and the role grants the whole `codap3/*` prefix.**
  Accurate as a description of the AWS-side design, but identical before this PR and strictly
  improved by it. `V3_AWS_*` were plain repository secrets with no environment scoping (all five
  environments have empty protection rules and null branch policies), `main` has no enforced
  protection, and forks get neither secrets nor an id-token. The reachable blast radius is
  unchanged; what changes is that there's no longer a long-lived credential to leak, sessions are
  ~1h, and CloudTrail now shows a distinct session identity. Tightening the trust policy with a
  `sub` condition would be a reasonable follow-up in starter-projects.
- **`${{ github.event.inputs.version }}` is interpolated into a `run:` step next to freshly
  configured credentials.** The `run:` blocks are byte-identical context lines in this diff, and
  the credentials were already on those same steps beforehand. Dispatching requires write access,
  and anyone with write access could edit the workflow directly anyway. Post-PR the exfiltrable
  payload is a 1-hour STS session rather than a permanent key pair, so this change makes it *less*
  dangerous, not more.
- **The repo's `default_workflow_permissions` is `write`.** True, but `v3.yml` on `main` has no
  `permissions:` key at all, so `lint`/`build`/`jest`/`cypress` are unchanged here while the one
  job this PR touches got narrower. Adding a workflow-level default would alter four jobs this PR
  doesn't otherwise touch — `cypress` in particular uses `GITHUB_TOKEN` to create commit statuses —
  so it carries its own regression risk and belongs in a separate PR.

## Nits — all optional

- **`deploy.md` slightly understates the staging workflow.** It reads as one entry point per
  workflow, but `release-v3-staging.yml` writes to *two* destinations in two separate steps:
  `index-staging.html` **and** `staging`. Worth a word, mostly so that someone later
  "normalizing" it to match the single-`DEST_FILE` pattern of the other three doesn't silently
  drop the second copy.
- **`contents: read` on the `s3-deploy` job looks redundant** — the job never checks out code
  (`download-artifact` for a same-run artifact uses the runtime token, not `GITHUB_TOKEN`), and
  the four release jobs correctly declare only `id-token: write`. Both forms work; the asymmetry
  is just a little confusing on a read-through.
- **The account ID `612297603577` is repeated across five files.** Fine as-is and consistent with
  the starter-projects convention, but it's five synchronized edits if the account or role name
  ever changes.
- **Pre-existing typo**, free to fix while you're in the file: line 1 of `release-v3-staging.yml`
  says `codap3.cocnord.org`.

PR #2696, opened 2026-09-09
CODAP-1504: draw and hit test point shapes in the canvas renderer

kswenson:
Draws the seven point shapes in the canvas renderer and makes hit testing shape-aware. Third in the LEADS Point Shapes stack, on top of #2693.

Fixes CODAP-1504.

## Geometry

`data-display/renderer/point-shapes.ts` is the single source of shape geometry for every surface that draws one — the canvas renderer today, the PIXI renderer and the legend keys later. Ported from the design prototype's `js/shapes.js`, whose tuned constants these are.

Circle is the reference and is unchanged from what CODAP has always drawn. Every other shape is normalized to about 90% of the circle's area rather than to equal area: straight edges and points read heavier than a circle of identical ink, so the small negative correction is what makes a triangle look like the same size point as a circle. Bounding boxes therefore differ between shapes, which is intentional.

The triangle is centered on its center of area rather than its bounding box. The prototype centers it on the box, which makes it sit `h/6` low — visible as a downward jump when a category is switched to it. In a scatterplot position is the data, so every shape sits on its centroid.

## Hit testing

A point's hit area is the drawn ink **unioned with** the circle of radius `r` that CODAP has always used, not the ink alone. A plus is narrower than a circle across its notches and a star between its arms, so testing the outline by itself would shrink the target for anyone who chose one — and shape is meant to be a free second encoding channel, not something you pay for in click accuracy. The union keeps every shape at least as easy to hit as a circle while adding the ink that extends past it: a star's tips, a square's corners, a triangle's apex.

This surfaced a latent defect in `pointShapeBoundingRadius`, which took half the larger side of the drawn box. That holds only while the box is centered on the point, and the triangle's is not: its apex sits at `1.47r` while half its width is `1.27r`, so a hit area sized from it stopped short of the apex. It now measures from the vertices. Its test had asserted that the radius covered half the extent, which the old implementation returned by construction — a tautology that no asymmetric shape could fail. That test now measures against the vertices too.

Marquee selection is unaffected: it is forwarded to the background element and never goes through the hit tester, and testing a marquee against point positions rather than outlines is what it should do.

## Scope

Behind the `pointShapes` feature flag, which gates the controls: shapes can only be *assigned* by a
flag-enabled user. Rendering is deliberately not gated, so a document that already carries shapes
draws them for everyone rather than silently rendering as circles for a reader without the flag.
That follows the convention from CODAP-1448 of gating affordances rather than registrations or
rendering.

The canvas renderer is the fallback used when Graphics Acceleration is turned off in Options; the
PIXI/WebGL path is CODAP-1509.

One thing worth a reviewer's eye: `getLegendShape` is called per case inside the coordinate
refresh, which runs for every point on every reposition. With no legend it early-returns after
two lookups; with one it does work comparable to the `getLegendColorForCase` already in that
loop. Not measured.

Reviewing commit by commit is likely easier than the combined diff.

### DIFF
--- a/v3/src/components/data-display/data-display-utils.ts
+++ b/v3/src/components/data-display/data-display-utils.ts
@@ -15,6 +15,7 @@ import {
   pointRadiusSelectionAddend, Rect, rTreeRect
 } from "./data-display-types"
 import {IDataConfigurationModel } from "./models/data-configuration-model"
+import { IDisplayItemDescriptionModel } from "./models/display-item-description-model"
 import {CaseDataWithSubPlot} from "./d3-types"
 import { getRendererForEvent, IPoint, IPointStyle, PointRendererBase } from "./renderer"
 
@@ -77,15 +78,25 @@ export const handleClickOnBar = ({ event, dataConfig, barCover }: IHandleClickOn
   setOrExtendSelection(barCover.caseIDs, dataConfig.dataset, extendSelection)
 }
 
+/*
+ * The shape for each case: the one its legend category carries, and the display's own wherever the
+ * legend assigns none. Every path that draws points needs this, so it is built here rather than
+ * rebuilt identically at each of them.
+ */
+export function legendShapeGetter(
+  dataConfig: IDataConfigurationModel | undefined, displayItemDescription: IDisplayItemDescriptionModel
+): (caseID: string) => PointShape {
+  const shapeIfNoCategory = displayItemDescription.pointShape
+  return (caseID: string) =>
+    dataConfig?.getLegendShapeForCase(caseID, shapeIfNoCategory) ?? shapeIfNoCategory
+}
+
 export interface IMatchCirclesProps {
   dataConfiguration: IDataConfigurationModel
   pointRadius: number
   pointColor: string
-  /*
-   * The shape a point is created with, as pointColor is the color it is created with. The refresh
-   * that follows resolves the per-case shape; this is what the point is drawn as until it does, and
-   * several callers create points without refreshing in the same breath.
-   */
+  // The shape a point is created with, as pointColor is the color it is created with. Several
+  // callers create points without refreshing in the same breath.
   pointShape?: PointShape
   pointDisplayType?: PointDisplayType
   pointStrokeColor: string
@@ -157,8 +168,6 @@ export function setPointSelection(
     // When there's no legend, use blue fill for selection instead of a colored stroke
     const useSelectionFill = isSelected && !legendID
     const style: Partial<IPointStyle> = {
-      // Mirrors the fill above: the legend assigns it per case where it can, and the display's own
-      // shape applies everywhere else.
       shape: dataConfiguration.getLegendShapeForCase(caseID, pointShape),
       fill: useSelectionFill ? defaultSelectedColor : fill,
       radius: isSelected ? selectedPointRadius : pointRadius,
--- a/v3/src/components/data-display/data-display-utils.ts
+++ b/v3/src/components/data-display/data-display-utils.ts
@@ -6,6 +6,7 @@ import {
   defaultStrokeOpacity, defaultStrokeWidth
 } from "../../utilities/color-utils"
 import {between} from "../../utilities/math-utils"
+import { kDefaultPointShape, PointShape } from "../../utilities/point-shape-utils"
 import { IBarCover } from "../graph/graphing-types"
 import {isGraphDataConfigurationModel} from "../graph/models/graph-data-configuration-model"
 import {ISetPointSelection} from "../graph/utilities/graph-utils"
@@ -80,6 +81,12 @@ export interface IMatchCirclesProps {
   dataConfiguration: IDataConfigurationModel
   pointRadius: number
   pointColor: string
+  /*
+   * The shape a point is created with, as pointColor is the color it is created with. The refresh
+   * that follows resolves the per-case shape; this is what the point is drawn as until it does, and
+   * several callers create points without refreshing in the same breath.
+   */
+  pointShape?: PointShape
   pointDisplayType?: PointDisplayType
   pointStrokeColor: string
   startAnimation: () => void
@@ -90,7 +97,7 @@ export interface IMatchCirclesProps {
 
 export function matchCirclesToData(props: IMatchCirclesProps) {
   const { dataConfiguration, renderer, startAnimation, stopAnimation, pointRadius, pointColor, pointStrokeColor,
-          pointDisplayType = "points" } = props
+          pointShape = kDefaultPointShape, pointDisplayType = "points" } = props
   // TODO: eliminate dependence on GraphDataConfigurationModel
   const allCaseData: CaseDataWithSubPlot[] = isGraphDataConfigurationModel(dataConfiguration)
     ? dataConfiguration.caseDataWithSubPlot
@@ -111,6 +118,7 @@ export function matchCirclesToData(props: IMatchCirclesProps) {
   renderer?.matchPointsToData(dataConfiguration.dataset?.id ?? '', allCaseData, pointDisplayType, {
     radius: pointRadius,
     fill: pointColor,
+    shape: pointShape,
     stroke: pointStrokeColor,
     strokeWidth: defaultStrokeWidth
   })
--- a/v3/src/components/data-display/data-display-utils.ts
+++ b/v3/src/components/data-display/data-display-utils.ts
@@ -130,7 +130,7 @@ export function setPointSelection(
   props: ISetPointSelection, caseIdsToUpdate?: Iterable<string>, numberOfPlots = 1
 ) {
   const { renderer, dataConfiguration, pointRadius, selectedPointRadius,
-    pointColor, pointStrokeColor, getPointColorAtIndex } = props
+    pointColor, pointStrokeColor, pointShape, getPointColorAtIndex } = props
   const dataset = dataConfiguration.dataset
   const legendID = dataConfiguration.attributeID('legend')
   if (!renderer) {
@@ -149,6 +149,9 @@ export function setPointSelection(
     // When there's no legend, use blue fill for selection instead of a colored stroke
     const useSelectionFill = isSelected && !legendID
     const style: Partial<IPointStyle> = {
+      // Mirrors the fill above: the legend assigns it per case where it can, and the display's own
+      // shape applies everywhere else.
+      shape: dataConfiguration.getLegendShapeForCase(caseID, pointShape),
       fill: useSelectionFill ? defaultSelectedColor : fill,
       radius: isSelected ? selectedPointRadius : pointRadius,
       stroke: isSelected && !useSelectionFill ? defaultSelectedStroke : pointStrokeColor,
--- a/v3/src/components/data-display/inspector/display-item-format-control.scss
+++ b/v3/src/components/data-display/inspector/display-item-format-control.scss
@@ -297,6 +297,30 @@
       }
     }
 
+    // A row carrying both a shape control and a color control -- every category row, and the
+    // single Points row shown when there is no legend attribute. Without this the row distributes
+    // with space-between and each control starts wherever its label happens to end, so the rows
+    // do not line up with each other.
+    .cat-color-picker,
+    .color-picker-row.shape-row {
+      gap: 4px;
+
+      // Only the label may shrink; the controls keep their size.
+      .react-aria-Select {
+        flex: 0 0 auto;
+      }
+
+      // Absorbs the free space, which pins both controls to the right edge. Truncates rather than
+      // pushing them out of view.
+      .form-label.color-picker {
+        flex: 1;
+        min-width: 0;
+        overflow: hidden;
+        text-overflow: ellipsis;
+        white-space: nowrap;
+      }
+    }
+
     .cat-color-setting {
       // A flex column, so the rows are flex items whose vertical margins do not collapse and each
       // row occupies the full height the 2.5-row limit below is derived from. The list ends
@@ -333,22 +357,6 @@
       .cat-color-picker {
         flex-shrink: 0; // rows keep their height, so the list scrolls rather than compressing
         margin: vars.$palette-category-row-margin 0;
-        gap: 4px;
-
-        // Only the label may shrink; the controls keep their size.
-        .react-aria-Select {
-          flex: 0 0 auto;
-        }
-
-        // Absorbs the free space so the controls align across rows rather than each starting
-        // wherever its category name happens to end. Truncates rather than pushing them out.
-        .form-label.color-picker {
-          flex: 1;
-          min-width: 0;
-          overflow: hidden;
-          text-overflow: ellipsis;
-          white-space: nowrap;
-        }
       }
     }
 
--- a/v3/src/components/data-display/models/data-configuration-model.ts
+++ b/v3/src/components/data-display/models/data-configuration-model.ts
@@ -890,6 +890,16 @@ export const DataConfigurationModel = types
        * For categorical it is a map of categories to colors
        * The color type is not handled yet.
        */
+      /*
+       * Changes identity when any category's shape changes, so a display can react to it the way
+       * it reacts to legendColorDomain. Only categorical legends carry shapes; the rest have no
+       * categories to assign one to, so there is nothing to observe.
+       */
+      get legendShapeDomain() {
+        const legendType = self.attributeType('legend')
+        if (legendType !== 'categorical' && legendType !== 'checkbox') return undefined
+        return self.categorySetForAttrRole('legend')?.shapeMap
+      },
       get legendColorDomain() {
         const legendType = self.attributeType('legend')
         switch (legendType) {
--- a/v3/src/components/data-display/models/data-configuration-model.ts
+++ b/v3/src/components/data-display/models/data-configuration-model.ts
@@ -866,10 +866,9 @@ export const DataConfigurationModel = types
        * A point then stands for several children at once, and resolving the legend through any one
        * of them would attribute that child's value to the whole group, so callers fall back.
        *
-       * Lives a block above the two views that use it so they can reach it through `self`. They are
-       * routinely passed around as detached function references -- a plot hands
-       * `dataConfig.getLegendColorForCase` to setPointCoordinates, which calls it bare -- so `this`
-       * inside them is undefined and cannot be used to reach a sibling view.
+       * Must stay a block above the views that use it, so they reach it through `self`. They are
+       * passed around as detached function references -- a plot hands `getLegendColorForCase` to
+       * setPointCoordinates, which calls it bare -- so `this` inside them is undefined.
        */
       get legendCollectionIsMoreChildmost(): boolean {
         const legendID = self.attributeID('legend')
@@ -910,11 +909,8 @@ export const DataConfigurationModel = types
        * For categorical it is a map of categories to colors
        * The color type is not handled yet.
        */
-      /*
-       * Changes identity when any category's shape changes, so a display can react to it the way
-       * it reacts to legendColorDomain. Only categorical legends carry shapes; the rest have no
-       * categories to assign one to, so there is nothing to observe.
-       */
+      // Changes identity when any category's shape changes, so a display can react to it the way it
+      // reacts to legendColorDomain.
       get legendShapeDomain() {
         const legendType = self.attributeType('legend')
         if (legendType !== 'categorical' && legendType !== 'checkbox') return undefined
@@ -970,10 +966,9 @@ export const DataConfigurationModel = types
       /*
        * The shape for a case, mirroring getLegendColorForCase.
        *
-       * Only categorical legends carry shapes. A numeric or date legend has no categories to assign
-       * one to, and a color legend supplies its own colors; those all resolve to the default so
-       * every point in such a plot shares one shape. Callers pass the display's own shape as
-       * `shapeIfNoCategory`, which is what a plot with no legend attribute uses throughout.
+       * Only a categorical legend has categories to carry shapes, so every other kind resolves to
+       * `shapeIfNoCategory` -- the display's own shape, which is also what a plot with no legend
+       * attribute draws throughout.
        */
       getLegendShapeForCase(id: string, shapeIfNoCategory: PointShape = kDefaultPointShape): PointShape {
         const legendID = self.attributeID('legend')
--- a/v3/src/components/data-display/models/data-configuration-model.ts
+++ b/v3/src/components/data-display/models/data-configuration-model.ts
@@ -960,7 +960,7 @@ export const DataConfigurationModel = types
        * The shape for a case, mirroring getLegendColorForCase.
        *
        * Only categorical legends carry shapes. A numeric or date legend has no categories to assign
-       * one to, and a colour legend supplies its own colours; those all resolve to the default so
+       * one to, and a color legend supplies its own colors; those all resolve to the default so
        * every point in such a plot shares one shape. Callers pass the display's own shape as
        * `shapeIfNoCategory`, which is what a plot with no legend attribute uses throughout.
        */
--- a/v3/src/components/data-display/models/data-configuration-model.ts
+++ b/v3/src/components/data-display/models/data-configuration-model.ts
@@ -919,16 +919,20 @@ export const DataConfigurationModel = types
             return 0
         }
       },
+      /*
+       * Whether the legend attribute lives in a collection more childmost than the plotted cases.
+       * A point then stands for several children at once, and resolving the legend through any one
+       * of them would attribute that child's value to the whole group, so callers fall back.
+       */
+      get legendCollectionIsMoreChildmost(): boolean {
+        const legendID = self.attributeID('legend')
+        const legendCollectionID = self.dataset?.getCollectionForAttribute(legendID)?.id
+        const legendCollectionIndex = self.dataset?.getCollectionIndex(legendCollectionID) ?? 0
+        const childmostCollectionID = idOfChildmostCollectionForAttributes(self.axisAttributeIDs, self.dataset)
+        const childmostCollectionIndex = self.dataset?.getCollectionIndex(childmostCollectionID) ?? 0
+        return legendCollectionIndex > childmostCollectionIndex
+      },
       getLegendColorForCase(id: string, colorIfMissing = missingColor): string {
-
-        const collectionOfLegendIsMoreChildmost = () => {
-          const legendCollectionID = self.dataset?.getCollectionForAttribute(legendID)?.id,
-            legendCollectionIndex = self.dataset?.getCollectionIndex(legendCollectionID) ?? 0,
-            childmostCollectionID = idOfChildmostCollectionForAttributes(self.axisAttributeIDs, self.dataset),
-            childmostCollectionIndex = self.dataset?.getCollectionIndex(childmostCollectionID) ?? 0
-          return legendCollectionIndex > childmostCollectionIndex
-        }
-
         const legendID = self.attributeID('legend')
         // todo: When user deletes we are not currently deleting the legend attribute ID. But we should.
         const legendAttribute = self.dataset?.getAttribute(legendID)
@@ -936,7 +940,7 @@ export const DataConfigurationModel = types
           return ''
         }
         const legendType = self.attributeType('legend')
-        if (collectionOfLegendIsMoreChildmost()) {
+        if (this.legendCollectionIsMoreChildmost) {
           return colorIfMissing
         }
         const legendValue = self.dataset?.getStrValue(id, legendID)
@@ -973,6 +977,9 @@ export const DataConfigurationModel = types
 
         const legendType = self.attributeType('legend')
         if (legendType !== 'categorical' && legendType !== 'checkbox') return shapeIfNoCategory
+
+        // as for color: a legend below the plotted cases cannot speak for a parent-level point
+        if (this.legendCollectionIsMoreChildmost) return shapeIfNoCategory
 
         const legendValue = self.dataset?.getStrValue(id, legendID)
         if (!legendValue) return shapeIfNoCategory
--- a/v3/src/components/data-display/models/data-configuration-model.ts
+++ b/v3/src/components/data-display/models/data-configuration-model.ts
@@ -945,6 +945,27 @@ export const DataConfigurationModel = types
           default:
             return ''
         }
+      },
+      /*
+       * The shape for a case, mirroring getLegendColorForCase.
+       *
+       * Only categorical legends carry shapes. A numeric or date legend has no categories to assign
+       * one to, and a colour legend supplies its own colours; those all resolve to the default so
+       * every point in such a plot shares one shape. Callers pass the display's own shape as
+       * `shapeIfNoCategory`, which is what a plot with no legend attribute uses throughout.
+       */
+      getLegendShapeForCase(id: string, shapeIfNoCategory: PointShape = kDefaultPointShape): PointShape {
+        const legendID = self.attributeID('legend')
+        const legendAttribute = self.dataset?.getAttribute(legendID)
+        if (!id || !legendID || !legendAttribute) return shapeIfNoCategory
+
+        const legendType = self.attributeType('legend')
+        if (legendType !== 'categorical' && legendType !== 'checkbox') return shapeIfNoCategory
+
+        const legendValue = self.dataset?.getStrValue(id, legendID)
+        if (!legendValue) return shapeIfNoCategory
+
+        return self.getLegendShapeForCategory(legendValue)
       }
     }))
   .actions(self => ({
--- a/v3/src/components/data-display/models/data-configuration-model.ts
+++ b/v3/src/components/data-display/models/data-configuration-model.ts
@@ -860,6 +860,24 @@ export const DataConfigurationModel = types
       casesInBinAreSelected(quantile: number): boolean {
         const selection = self.getCasesForLegendBin(quantile)
         return !!(selection.length > 0 && selection?.every((anID: string) => self.dataset?.isCaseSelected(anID)))
+      },
+      /*
+       * Whether the legend attribute lives in a collection more childmost than the plotted cases.
+       * A point then stands for several children at once, and resolving the legend through any one
+       * of them would attribute that child's value to the whole group, so callers fall back.
+       *
+       * Lives a block above the two views that use it so they can reach it through `self`. They are
+       * routinely passed around as detached function references -- a plot hands
+       * `dataConfig.getLegendColorForCase` to setPointCoordinates, which calls it bare -- so `this`
+       * inside them is undefined and cannot be used to reach a sibling view.
+       */
+      get legendCollectionIsMoreChildmost(): boolean {
+        const legendID = self.attributeID('legend')
+        const legendCollectionID = self.dataset?.getCollectionForAttribute(legendID)?.id
+        const legendCollectionIndex = self.dataset?.getCollectionIndex(legendCollectionID) ?? 0
+        const childmostCollectionID = idOfChildmostCollectionForAttributes(self.axisAttributeIDs, self.dataset)
+        const childmostCollectionIndex = self.dataset?.getCollectionIndex(childmostCollectionID) ?? 0
+        return legendCollectionIndex > childmostCollectionIndex
       }
     }))
   .views(self => (
@@ -919,19 +937,6 @@ export const DataConfigurationModel = types
             return 0
         }
       },
-      /*
-       * Whether the legend attribute lives in a collection more childmost than the plotted cases.
-       * A point then stands for several children at once, and resolving the legend through any one
-       * of them would attribute that child's value to the whole group, so callers fall back.
-       */
-      get legendCollectionIsMoreChildmost(): boolean {
-        const legendID = self.attributeID('legend')
-        const legendCollectionID = self.dataset?.getCollectionForAttribute(legendID)?.id
-        const legendCollectionIndex = self.dataset?.getCollectionIndex(legendCollectionID) ?? 0
-        const childmostCollectionID = idOfChildmostCollectionForAttributes(self.axisAttributeIDs, self.dataset)
-        const childmostCollectionIndex = self.dataset?.getCollectionIndex(childmostCollectionID) ?? 0
-        return legendCollectionIndex > childmostCollectionIndex
-      },
       getLegendColorForCase(id: string, colorIfMissing = missingColor): string {
         const legendID = self.attributeID('legend')
         // todo: When user deletes we are not currently deleting the legend attribute ID. But we should.
@@ -940,7 +945,7 @@ export const DataConfigurationModel = types
           return ''
         }
         const legendType = self.attributeType('legend')
-        if (this.legendCollectionIsMoreChildmost) {
+        if (self.legendCollectionIsMoreChildmost) {
           return colorIfMissing
         }
         const legendValue = self.dataset?.getStrValue(id, legendID)
@@ -979,7 +984,7 @@ export const DataConfigurationModel = types
         if (legendType !== 'categorical' && legendType !== 'checkbox') return shapeIfNoCategory
 
         // as for color: a legend below the plotted cases cannot speak for a parent-level point
-        if (this.legendCollectionIsMoreChildmost) return shapeIfNoCategory
+        if (self.legendCollectionIsMoreChildmost) return shapeIfNoCategory
 
         const legendValue = self.dataset?.getStrValue(id, legendID)
         if (!legendValue) return shapeIfNoCategory
--- a/v3/src/components/data-display/models/data-configuration-model.ts
+++ b/v3/src/components/data-display/models/data-configuration-model.ts
@@ -862,6 +862,15 @@ export const DataConfigurationModel = types
         return !!(selection.length > 0 && selection?.every((anID: string) => self.dataset?.isCaseSelected(anID)))
       },
       /*
+       * Whether the legend assigns per-category state. Note this is narrower than
+       * isCategoricalAttributeType, which counts a color legend too: that one gives each case a
+       * color of its own, so there are no categories to hang a color or a shape on.
+       */
+      get legendHasCategories(): boolean {
+        const legendType = self.attributeType('legend')
+        return legendType === 'categorical' || legendType === 'checkbox'
+      },
+      /*
        * Whether the legend attribute lives in a collection more childmost than the plotted cases.
        * A point then stands for several children at once, and resolving the legend through any one
        * of them would attribute that child's value to the whole group, so callers fall back.
@@ -912,8 +921,7 @@ export const DataConfigurationModel = types
       // Changes identity when any category's shape changes, so a display can react to it the way it
       // reacts to legendColorDomain.
       get legendShapeDomain() {
-        const legendType = self.attributeType('legend')
-        if (legendType !== 'categorical' && legendType !== 'checkbox') return undefined
+        if (!self.legendHasCategories) return undefined
         return self.categorySetForAttrRole('legend')?.shapeMap
       },
       get legendColorDomain() {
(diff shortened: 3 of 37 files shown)


v3/src/components/data-display/models/data-configuration-model.ts
```diff
@@ -945,6 +957,27 @@ export const DataConfigurationModel = types
...
+        if (!id || !legendID || !legendAttribute) return shapeIfNoCategory
+
+        const legendType = self.attributeType('legend')
+        if (legendType !== 'categorical' && legendType !== 'checkbox') return shapeIfNoCategory
+
+        const legendValue = self.dataset?.getStrValue(id, legendID)
+        if (!legendValue) return shapeIfNoCategory
+
+        return self.getLegendShapeForCategory(legendValue, shapeIfNoCategory)
       }
```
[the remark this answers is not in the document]
kswenson: Good catch, accepted. The doc comment claimed it mirrored `getLegendColorForCase` and it didn't.

The guard was a closure private to the color path; it's now a view both share, so the two can't drift, and the shape path falls back the same way. Test covers a legend attribute in a child collection with the plotted cases in the parent.

Fixed in ef325ef89.

v3/src/components/data-display/renderer/canvas-point-renderer.ts
```diff
@@ -568,16 +585,18 @@ export class CanvasPointRenderer extends PointRendererBase {
         this.ctx.strokeRect(rectX, rectY, width, height)
       }
     } else {
-      // Draw circle (point)
-      this.ctx.beginPath()
-      this.ctx.arc(0, 0, radius, 0, Math.PI * 2)
+      // Draw the point's shape, which is a circle unless the legend assigns another
+      tracePointShape(this.ctx, style.shape ?? kDefaultPointShape, radius)
```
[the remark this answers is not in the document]
kswenson: We're not going to gate rendering here, though you're right that the description overstated it.

The convention in this codebase (CODAP-1448) is to gate affordances rather than registrations or rendering. Gating the renderer would make a document authored with shapes draw as circles for every reader without the flag, which is a worse outcome than the leak it prevents — and since only a flag-enabled user can assign a shape in the first place, exposure is limited to documents they deliberately share.

The PR description claimed nothing was visible by default, which was wrong for a document that already carries shapes. I've corrected it rather than the code.

v3/src/components/graph/hooks/use-plot.ts
```diff
@@ -453,6 +453,25 @@ export const usePlotResponders = (props: IPlotResponderProps) => {
       }, {name: "usePlot [legendColorChange]"}, graphModel)
   }, [graphModel, callRefreshPointPositions])
 
+  /*
+   * A shape change alters only how each point is drawn, so a restyle suffices; positions and masks
+   * are untouched, unlike a legend color change which can also change which points are plotted.
+   *
+   * It has to be a restyle rather than a reposition: only setPointSelection writes the shape into
+   * a point's style. setPointCoordinates, which the reposition path uses, sets the fill but not the
+   * shape, and style updates merge -- so repositioning after a shape change leaves the old shape in
```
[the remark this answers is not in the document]
kswenson: Accepted, and this was the most consequential of the four.

`setPointCoordinates` now takes a `getLegendShape` resolver beside the `getLegendColor` it already had, and the seven graph plots and the map layer supply one — so shape is carried exactly the way color is, which is what it should have been from the start given it's meant as a second encoding channel beside color. Adding a case to a plot with shapes assigned previously drew the new point as a circle.

The resolver is built inside the refresh callback rather than at render scope, so it reads current state whenever the callback runs.

`setPointCoordinates` had no tests at all; these are the first.

Fixed in 481b1c2d8.

v3/src/models/data/category-set.ts
```diff
@@ -220,10 +220,19 @@ export const CategorySet = types.model("CategorySet", {
...
+   * to compute.
+   *
+   * A category with no shape of its own inherits `shapeIfUnset`, which callers set to the
+   * display's own shape. That is what keeps a shape chosen before a legend existed from being
+   * discarded the moment one is added: colors are replaced by something meaningful when a legend
+   * takes over, but shapes would be replaced by nothing.
+   */
+  shapeForCategory(category: string, shapeIfUnset: PointShape = kDefaultPointShape): PointShape {
+    const stored = self.shapes.get(category)
+    return isPointShape(stored) ? stored : shapeIfUnset
```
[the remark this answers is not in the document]
kswenson: Accepted — a real bug, and a good catch on the interaction.

Absence used to mean "circle" and the setter still wrote that meaning by deleting the entry, but inheritance redefined absence as "inherit". With a display shape of star, choosing Circle for a category deleted its entry and the getter returned star straight back.

Every choice is stored now, circle included; a category that was never assigned is still absent, so exports stay empty until a user picks something. Two existing tests asserted the old meaning and had encoded the bug with it — including one expecting a provisional category set *not* to be promoted for an explicit circle, which it should be, since that choice has to survive a reload.

Fixed in 127e25a81.

v3/src/components/graph/plots/scatter-plot/scatter-plot.tsx
```diff
@@ -348,10 +349,16 @@ export const ScatterPlot = observer(function ScatterPlot({ renderer }: IPlotProp
...
+    // rather than as a circle until something later restyles it.
+    const shapeIfNoCategory = graphModel.pointDescription.pointShape
+    const getLegendShape = (anID: string) =>
+      dataConfiguration?.getLegendShapeForCase(anID, shapeIfNoCategory) ?? shapeIfNoCategory
+
     setPointCoordinates({
       dataset, renderer, pointRadius: graphModel.getPointRadius(),
       selectedPointRadius: selectedPointRadiusRef.current,
-      selectedOnly, getScreenX, getScreenY, getLegendColor,
+      selectedOnly, getScreenX, getScreenY, getLegendColor, getLegendShape,
```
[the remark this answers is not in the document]
kswenson: Declining this one, for a reason that's narrower than it first looks.

The "created with no shape" half is already gone: ee23a7610 (after the commit you reviewed) gives `matchCirclesToData` the display shape, so a newly created point is no longer a circle by default.

What's left is that a point created during performance mode carries the display shape rather than its per-category shape until a restyle. But performance mode is a transient state entered for the duration of a user drag, and a shape or color cannot change during a drag — the user is dragging points, not editing a legend. So the only way to reach this is a plugin adding cases mid-drag, and such a point gets the display shape exactly as it gets the display color.

That symmetry is the other half of the reason: `refreshPointPositionsPerfMode` sets no style at all, fill included. Carrying per-category shape there while color stays approximate would make shape diverge from color, and carrying both would defeat the path's purpose, which is to skip per-point style work during a drag.

v3/src/components/graph/hooks/use-plot.ts
```diff
@@ -453,6 +453,25 @@ export const usePlotResponders = (props: IPlotResponderProps) => {
...
+   * place. That is why both shape sources are observed here rather than alongside the other point
+   * properties: the per-category shapes, and the display's own, which is what a plot with no legend
+   * attribute draws throughout.
+   */
```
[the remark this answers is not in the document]
kswenson: Correct, accepted — the comment outlived the thing it described.

It was the reason the reaction was written that way, and it stopped being true the moment `setPointCoordinates` began writing a shape. The restyle path is still the right choice, but because repositioning every point merely to change how they're drawn is wasted work, not because it's the only path that can carry a shape. Reworded to say that.

Fixed in 770ef5bc0.

v3/src/components/map/components/map-point-layer.tsx
```diff
@@ -414,6 +419,25 @@ export const MapPointLayer = observer(function MapPointLayer({mapLayerModel, lay
...
+  /*
+   * A shape change alters only how each point is drawn, so the points are restyled without touching
+   * the heatmap, which has no notion of shape.
+   *
+   * It has to go through refreshPointSelection: that is the only path that writes the shape into a
+   * point's style. refreshPoints sets the radius, fill and stroke but not the shape, and style
+   * updates merge, so refreshing after a shape change leaves the old shape in place. Both shape
+   * sources are observed here -- the per-category shapes, and the layer's own, which is what a
+   * layer with no legend attribute draws throughout.
+   */
```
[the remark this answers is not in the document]
kswenson: Same as the graph-side comment — accepted.

`refreshPoints` writes a shape now, so the claim that `refreshPointSelection` is the only path carrying one is stale. The valid reason survives and is all the comment now gives: it restyles without doing the heatmap work.

Fixed in 770ef5bc0.

v3/src/models/data/category-set.ts
```diff
@@ -288,12 +297,13 @@ export const CategorySet = types.model("CategorySet", {
   },
   // Storing the default is stored as absence, so a document only carries the shapes a user chose
   // and a category reverted to circle round-trips as an unset entry.
+  /*
+   * Every choice is stored, circle included. Absence means the category has no shape of its own and
+   * inherits the display's, which is not the same as having been set to a circle: with a display
+   * shape of star, dropping a circle here would read back as star and make circle unselectable.
+   */
```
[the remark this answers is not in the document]
kswenson: Accepted, and worse than described — the setter carried both comments at once. My edit inserted the new explanation directly beneath the old one instead of replacing it, so it read as "circles are dropped" immediately followed by "every choice is stored, circle included".

The obsolete lines are gone; one contract, stated once.

Fixed in 770ef5bc0.

v3/src/models/data/category-set.ts
```diff
@@ -286,14 +295,13 @@ export const CategorySet = types.model("CategorySet", {
...
   setShapeForCategory(value: string, shape: PointShape) {
-    if (shape && shape !== kDefaultPointShape) {
-      self.shapes.set(value, shape)
-    } else {
-      self.shapes.delete(value)
-    }
+    self.shapes.set(value, shape)
```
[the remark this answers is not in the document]
kswenson: Accepted — good catch, and the important part is that it was a second copy of a contract I only changed in one place.

Storing every choice in `setShapeForCategory` fixed the model, but `v2-category-set-importer.ts` independently filtered circles out, so the exporter wrote the entry and the importer discarded it. A document with a display shape of star and a category deliberately set to circle came back showing star — exactly the bug the setter change was meant to remove, still alive across a reload.

The importer now keeps any recognized shape. Its own comment already said "every entry here is a deliberate user assignment", which the filter contradicted; it now explains why a circle counts as one.

The "drops the default" test is rewritten, and I added a full round-trip test rather than relying on the importer test alone — export and import each decide separately what's worth keeping, so agreeing at one end proves nothing about the other. Both fail if the filter comes back.

Fixed in 5f81368d1.

kswenson:

| PR | Story | What it is |
|---|---|---|

**Nothing here is superseded** — this is the top of the stack, and the spelling, inheritance and persistence revisions noted on #2692 and #2693 all land in this PR.

**Worth your attention:**

- The tuned constants in `point-shapes.ts`. Shapes are normalized to ~90% of the circle's area rather than to equal area, because straight edges and points read heavier at equal ink. Please don't re-derive them; the relative visual weight of the whole set depends on them.
- The hit area is the drawn ink **unioned with** the old circle of radius r. A plus is narrower than a circle across its notches and a star between its arms, so testing the ink alone would make a point harder to click purely because someone chose a shape.
- Shape is threaded through both the paths that *style* a point and the ones that *create* it. That split is where four separate bugs lived during review, so it is the part most worth a skeptical read.
- Rendering is deliberately not gated on the `pointShapes` flag, only the controls are — otherwise a document authored with shapes would draw as circles for every reader without the flag.

Commit by commit is likely easier than the combined diff.

v3/src/components/data-display/inspector/display-item-format-control.scss
```diff
@@ -297,6 +297,30 @@
       }
     }
 
+    // A row carrying both a shape control and a color control -- every category row, and the
+    // single Points row shown when there is no legend attribute. Without this the row distributes
+    // with space-between and each control starts wherever its label happens to end, so the rows
+    // do not line up with each other.
```
tealefristoe, line 303:
```suggestion
    // A row carrying both a shape control and a color control. Without this the row distributes
    // with space-between and each control starts wherever its label happens to end, so the rows
    // do not line up with each other.
```

v3/src/components/data-display/models/data-configuration-model.ts
```diff
@@ -858,6 +860,23 @@ export const DataConfigurationModel = types
...
         return !!(selection.length > 0 && selection?.every((anID: string) => self.dataset?.isCaseSelected(anID)))
+      },
+      /*
+       * Whether the legend attribute lives in a collection more childmost than the plotted cases.
+       * A point then stands for several children at once, and resolving the legend through any one
+       * of them would attribute that child's value to the whole group, so callers fall back.
+       *
+       * Must stay a block above the views that use it, so they reach it through `self`. They are
+       * passed around as detached function references -- a plot hands `getLegendColorForCase` to
+       * setPointCoordinates, which calls it bare -- so `this` inside them is undefined.
```
tealefristoe, line 871:
```suggestion
```

v3/src/components/data-display/models/data-configuration-model.ts
```diff
@@ -890,6 +909,13 @@ export const DataConfigurationModel = types
        * For categorical it is a map of categories to colors
        * The color type is not handled yet.
        */
+      // Changes identity when any category's shape changes, so a display can react to it the way it
+      // reacts to legendColorDomain.
```
tealefristoe, line 913:
```suggestion
```

v3/src/components/data-display/models/data-configuration-model.ts
```diff
@@ -945,6 +962,29 @@ export const DataConfigurationModel = types
...
+       * `shapeIfNoCategory` -- the display's own shape, which is also what a plot with no legend
+       * attribute draws throughout.
+       */
+      getLegendShapeForCase(id: string, shapeIfNoCategory: PointShape = kDefaultPointShape): PointShape {
+        const legendID = self.attributeID('legend')
+        const legendAttribute = self.dataset?.getAttribute(legendID)
+        if (!id || !legendID || !legendAttribute) return shapeIfNoCategory
+
+        const legendType = self.attributeType('legend')
+        if (legendType !== 'categorical' && legendType !== 'checkbox') return shapeIfNoCategory
```
tealefristoe, line 979: This is repeated on line 916. It should be a named function so it's clear what it means and it's not sprinkled throughout.

v3/src/components/data-display/models/data-configuration-model.ts
```diff
@@ -945,6 +962,29 @@ export const DataConfigurationModel = types
...
+
+        // as for color: a legend below the plotted cases cannot speak for a parent-level point
```
tealefristoe, line 981:
```suggestion
```

v3/src/components/data-display/models/data-configuration-model.ts
```diff
@@ -945,6 +962,29 @@ export const DataConfigurationModel = types
...
             return ''
         }
+      },
+      /*
+       * The shape for a case, mirroring getLegendColorForCase.
+       *
+       * Only a categorical legend has categories to carry shapes, so every other kind resolves to
+       * `shapeIfNoCategory` -- the display's own shape, which is also what a plot with no legend
+       * attribute draws throughout.
+       */
```
tealefristoe, line 972:
```suggestion
```
First part of the comment is a useless distraction.

Second part can be removed once `legendType !== 'categorical' && legendType !== 'checkbox'` has a name that explains what it means.

v3/src/components/data-display/renderer/canvas-point-renderer.ts
```diff
@@ -568,16 +585,18 @@ export class CanvasPointRenderer extends PointRendererBase {
         this.ctx.strokeRect(rectX, rectY, width, height)
       }
     } else {
-      // Draw circle (point)
-      this.ctx.beginPath()
-      this.ctx.arc(0, 0, radius, 0, Math.PI * 2)
+      // Draw the point's shape, which is a circle unless the legend assigns another
```
tealefristoe, line 588:
```suggestion
```

v3/src/components/data-display/renderer/canvas-point-renderer.ts
```diff
@@ -568,16 +585,18 @@ export class CanvasPointRenderer extends PointRendererBase {
...
+      tracePointShape(this.ctx, style.shape ?? kDefaultPointShape, radius)
       this.ctx.fillStyle = fill
       this.ctx.fill()
 
       if (strokeWidth > 0) {
         this.ctx.strokeStyle = effectiveStroke
         this.ctx.lineWidth = strokeWidth / effectiveScale // Compensate for scale
         this.ctx.globalAlpha = strokeOpacity ?? 0.4
+        // Rounds the joins, without which the star's 36-degree points and the X's corners grow
+        // spikes at small radii under the default miter.
```
tealefristoe, line 598:
```suggestion
```

v3/src/components/data-display/renderer/point-renderer-types.ts
```diff
@@ -14,6 +15,9 @@ export interface IPoint {
  */
 export interface IPointStyle {
   radius: number
+  // Absent means the default. Optional so every existing caller that builds a style keeps working
+  // and keeps drawing circles.
```
tealefristoe, line 19:
```suggestion
```

v3/src/components/data-display/renderer/point-shapes.test.ts
```diff
@@ -0,0 +1,262 @@
...
+}
+
+describe("point shape geometry", () => {
+  it("covers every shape", () => {
+    PointShapes.forEach(shape => {
+      expect(pointShapeGeometry(shape, 8)).toBeDefined()
+    })
+  })
+
+  it("draws the circle as an arc of exactly r, unchanged from what CODAP already ships", () => {
```
tealefristoe, line 61:
```suggestion
  it("draws the circle as an arc of exactly r", () => {
```

v3/src/components/data-display/renderer/point-shapes.test.ts
```diff
@@ -0,0 +1,262 @@
...
+       * box and cannot see where that box sits, so comparing against half of it is satisfied by any
+       * value big enough for a symmetric shape while still falling short on an asymmetric one.
+       */
+      PointShapes.filter(s => s !== "circle").forEach(shape => {
+        const geometry = pointShapeGeometry(shape, 8)
+        if (geometry.kind !== "polygon") throw new Error(`${shape} should be a polygon`)
+        const furthest = Math.max(...geometry.points.map(p => Math.hypot(p.x, p.y)))
+        expect(pointShapeBoundingRadius(shape, 8)).toBeGreaterThanOrEqual(furthest - 1e-9)
+      })
+    })
```
tealefristoe, line 193: It seems like this test covers everything in the one above and the one below it and more. Am I missing something? If not, I'd get rid of those two tests.

v3/src/components/data-display/renderer/point-shapes.test.ts
```diff
@@ -0,0 +1,262 @@
...
+      const apex = geometry.points.reduce((a, p) => (p.y < a.y ? p : a))
+      const apexDistance = Math.hypot(apex.x, apex.y)
+
+      expect(apexDistance).toBeGreaterThan(pointShapeExtent("triangle", 8).w / 2)
+      expect(pointShapeBoundingRadius("triangle", 8)).toBeGreaterThanOrEqual(apexDistance - 1e-9)
+    })
+  })
+
+  describe("containment", () => {
+    const r = 8
```
tealefristoe, line 209: Why is this a constant here but hardcoded everywhere else in this file?

v3/src/components/data-display/renderer/point-shapes.ts
```diff
@@ -0,0 +1,233 @@
+import { PointShape } from "../../../utilities/point-shape-utils"
+
+/*
+ * The geometry of the seven point shapes, as a function of the point radius.
+ *
+ * One source for every surface that draws a shape. Ported from the design prototype's js/shapes.js,
+ * whose constants these are.
```
tealefristoe, line 7: Seems like a useful comment, but "design prototype's js/shapes.js" will lose meaning quickly.

v3/src/components/data-display/renderer/point-shapes.ts
```diff
@@ -0,0 +1,233 @@
...
+
+export interface IShapePoint {
+  x: number
+  y: number
+}
+
+export interface IShapeExtent {
+  w: number
+  h: number
+}
```
tealefristoe, line 37: These are all generally useful types that should be in a more general types/utils file, if they aren't already.

v3/src/components/data-display/renderer/point-shapes.ts
```diff
@@ -0,0 +1,233 @@
...
+ * A circle is an arc rather than a vertex list, so callers switch on `kind` instead of receiving a
+ * polygonal approximation. Both renderers can draw an arc directly, and approximating one would
+ * change how the shape CODAP already ships is rasterized.
+ */
+export type PointShapeGeometry =
+  | { kind: "circle", radius: number }
+  | { kind: "polygon", points: IShapePoint[] }
+
+// Vertices of a plus: an arm half-width `a` and half-length `L`, optionally rotated. X is this
+// same outline turned 45 degrees -- same ink, same weight, one set of numbers to maintain.
```
tealefristoe, line 49:
```suggestion
// Vertices of a plus and X: an arm half-width `a` and half-length `L`, optionally rotated.
// Plus is the default; X is this same outline turned 45 degrees.
```

v3/src/components/data-display/renderer/point-shapes.ts
```diff
@@ -0,0 +1,233 @@
...
+      const s = K.triangle * r
+      const h = s * Math.sqrt(3) / 2
+      /*
+       * Centered on its center of area, not its bounding box: a centroid sits h/6 below the box
+       * center, so box-centering makes the triangle read as sitting low, and switching a category
+       * to it visibly shifts the points down.
+       *
+       * The prototype centers on the box, buying a predictable hit area and a shared baseline with
+       * the square. On a plot a position is the data, so the bias costs more than the alignment
+       * gains.
```
tealefristoe, line 107:
```suggestion
```

v3/src/components/data-display/renderer/point-shapes.ts
```diff
@@ -0,0 +1,233 @@
...
+}
+
+// The ink area. Used to verify the normalization holds rather than by the renderers.
+export function pointShapeArea(shape: PointShape, r: number): number {
+  return kShapeDefs[shape].area(r)
+}
+
+/*
+ * The drawn bounding box, which is NOT 2r for anything but the circle: a star is about 35% wider
+ * than the circle it replaces. Used to size a shape against a box, and to check the normalization.
```
tealefristoe, line 184:
```suggestion
 * The drawn bounding box. Used to size a shape against a box, and to check the normalization.
```

v3/src/components/data-display/renderer/point-shapes.ts
```diff
@@ -0,0 +1,233 @@
...
+/*
+ * Whether (dx, dy), relative to the point's center, is on the point.
+ *
+ * The drawn ink, unioned with the circle of radius r that CODAP has always used. Shape is a second
+ * encoding channel, so choosing one must not make a point harder to click than it was as a circle:
+ * testing the ink alone would open dead zones between a star's arms and in the notches of a plus,
+ * where the shape is narrower than the circle it replaced. The union keeps every shape at least as
+ * easy to hit as a circle while adding the ink that extends past it -- a star's tips, a square's
+ * corners, a triangle's apex -- so the target matches what is drawn wherever that is generous, and
+ * matches the old circle wherever it is not.
```
tealefristoe, line 226:
```suggestion
 * Whether (dx, dy), relative to the point's center, hits the point.
 *
 * The drawn ink, unioned with the circle of radius r that CODAP has always used. Shape is a second
 * encoding channel, so choosing one must not make a point harder to click than it was as a circle:
 * testing the ink alone would open dead zones between a star's arms and in the notches of a plus,
 * where the shape is narrower than the circle it replaced. The union keeps every shape at least as
 * easy to hit as a circle while adding the ink that extends past it so the target matches what is drawn
 * wherever that is generous, and matches the old circle wherever it is not.
```
I found the first sentence hard to parse, since we're using points in two different senses here: points in pixel space and points in data space (which are represented by shapes that take up many points in pixel space). I almost suggested replacing point with shape in the sentence, but I'm not sure if that's right.

I also think the second paragraph should be reduced more, but I don't want to spend the time doing it.

v3/src/components/data-display/data-display-utils.ts
```diff
@@ -80,6 +81,12 @@ export interface IMatchCirclesProps {
   dataConfiguration: IDataConfigurationModel
   pointRadius: number
   pointColor: string
+  /*
+   * The shape a point is created with, as pointColor is the color it is created with. The refresh
+   * that follows resolves the per-case shape; this is what the point is drawn as until it does, and
+   * several callers create points without refreshing in the same breath.
+   */
```
tealefristoe, line 88:
```suggestion
```
At least get rid of the first sentence.

v3/src/components/data-display/data-display-utils.ts
```diff
@@ -149,6 +157,9 @@ export function setPointSelection(
     // When there's no legend, use blue fill for selection instead of a colored stroke
     const useSelectionFill = isSelected && !legendID
     const style: Partial<IPointStyle> = {
+      // Mirrors the fill above: the legend assigns it per case where it can, and the display's own
+      // shape applies everywhere else.
```
tealefristoe, line 161:
```suggestion
```

v3/src/components/graph/hooks/use-dot-plot.ts
```diff
@@ -76,11 +76,21 @@ export const useDotPlot = (renderer?: PointRendererBase) => {
   const refreshPointSelection = useCallback(() => {
     const pointRadius = graphModel.getPointRadius()
     const selectedPointRadius = graphModel.getPointRadius('select')
+    /*
+     * Read the appearance from the model here rather than closing over the values destructured
+     * above. A restyle can be driven by a reaction, which runs when the model changes and before
```
tealefristoe, line 81:
```suggestion
     * in the hook body above. A restyle can be driven by a reaction, which runs when the model changes and before
```

v3/src/components/graph/hooks/use-dot-plot.ts
```diff
@@ -76,11 +76,21 @@ export const useDotPlot = (renderer?: PointRendererBase) => {
...
+     * the re-render that would rebuild this callback; a captured value restyles the plot to the
+     * appearance it had before that change, so each change only appears when the next one is made.
+     * The other plot types read these at call time for the same reason.
```
tealefristoe, line 84:
```suggestion
```

v3/src/components/graph/utilities/graph-utils.ts
```diff
@@ -503,6 +504,9 @@ export interface ISetPointSelection {
   selectedPointRadius: number,
   pointColor: string,
   pointStrokeColor: string,
+  // The shape used where the legend assigns none -- no legend attribute, or a legend whose type
+  // carries no categories. Optional so callers that predate shapes keep drawing circles.
```
tealefristoe, line 508:
```suggestion
  // carries no categories.
```

v3/src/components/graph/utilities/graph-utils.ts
```diff
@@ -521,6 +525,12 @@ export interface ISetPointCoordinates {
   getScreenX: ((anID: string) => number | null)
   getScreenY: ((anID: string, plotNum?:number) => number | null)
   getLegendColor?: ((anID: string) => string)
+  /*
+   * The shape for a case, resolved at call time like getLegendColor. This path creates points as
+   * well as moving them, so a shape supplied only by the restyle path would leave every newly
+   * created point a circle until something unrelated restyled it.
```
tealefristoe, line 531:
```suggestion
   * This path creates points as
   * well as moving them, so a shape supplied only by the restyle path would leave every newly
   * created point a circle until something unrelated restyled it.
```
I don't really understand the second sentence in this comment, so I can't say whether it should be removed or not. It seems irrelevant for this location, but I'm not sure.

v3/src/components/graph/models/graph-data-configuration-model.test.ts
```diff
@@ -1113,6 +1113,68 @@ describe("DataConfigurationModel legend point shapes", () => {
     expect(categorySet?.shapeForCategory("land")).toBe("diamond")
   })
 
+  it("inherits the display's shape for a category with none of its own", () => {
+    /*
+     * A shape chosen before a legend existed must survive the legend being added. Colors are
+     * replaced by something meaningful when a legend takes over -- distinct palette colors per
+     * category -- but there is no shape palette, so replacing shapes with the bare default would
+     * discard the user's choice and hand back nothing.
+     */
```
tealefristoe, line 1122:
```suggestion
```

v3/src/components/graph/models/graph-data-configuration-model.test.ts
```diff
@@ -1113,6 +1113,68 @@ describe("DataConfigurationModel legend point shapes", () => {
...
+    tree.config.setLegendShapeForCategory("land", "x")
+    expect(tree.config.getLegendShapeForCase(caseId, "plus")).toBe("x")
+  })
+
+  it("works when called as a detached reference, which is how the plots call it", () => {
+    /*
+     * A plot hands these to setPointCoordinates as bare function references --
+     * `getLegendColor = dataConfig.getLegendColorForCase` -- and calls them unbound, so `this`
+     * inside them is undefined. Anything they need from the model has to come through `self`.
+     */
```
tealefristoe, line 1147:
```suggestion
```

v3/src/components/graph/plots/bar-chart/bar-chart.tsx
```diff
@@ -39,6 +39,11 @@ export const BarChart = observer(function BarChart({ abovePointsGroupRef, render
     const pointRadius = graphModel.getPointRadius()
     const legendAttrID = dataConfig?.attributeID('legend')
     const getLegendColor = legendAttrID ? dataConfig?.getLegendColorForCase : undefined
+    // Resolved here, alongside the color, so a point created by this path is drawn with its shape
+    // rather than as a circle until something later restyles it.
+    const shapeIfNoCategory = graphModel.pointDescription.pointShape
+    const getLegendShape = (anID: string) =>
+      dataConfig?.getLegendShapeForCase(anID, shapeIfNoCategory) ?? shapeIfNoCategory
```
tealefristoe, line 46: Duplicated word for word, including the questionable comment, in `binned-dot-plot.tsx`, `case-plot.tsx`, `dot-chart.tsx`, `dot-line-plot.tsx`, `histogram.tsx`, and `scatter-plot.tsx`. Should be abstracted.

v3/src/models/data/v2-category-set-importer.ts
```diff
@@ -21,14 +21,15 @@ export function importV2CategorySet(
 
   /*
    * Taken as given rather than filtered against the categories currently in the data: every entry
-   * is a deliberate assignment, so a category whose cases are deleted and later restored keeps the
-   * shape the user chose for it. Nothing generates a shape, so there is no automatic value to age
-   * out the way the color loop below ages one out.
+   * is a deliberate assignment, circle included -- absence means a category inherits the display's
+   * shape, so an explicit circle is a real choice and dropping it would put that category back on
+   * whatever the display is set to. Nothing generates a shape, so there is no automatic value to
+   * age out the way the color loop below ages one out.
```
tealefristoe, line 27: I'd just drop this whole comment.

v3/src/v2/v2-document-round-trip.test.ts
```diff
@@ -118,6 +118,31 @@ describe("v2 document round-trip of point shapes", () => {
...
   })
 
+  it("brings an explicitly chosen circle back through the round trip", async () => {
+    /*
+     * Absence means the category inherits the display's shape, so a circle has to survive as a
+     * stored value rather than being treated as "nothing to save". Dropped anywhere along the way,
+     * the category comes back inheriting -- showing the display's shape instead of the circle the
+     * user picked. Asserted through the full trip because export and import each decide separately
+     * what is worth keeping.
+     */
```
tealefristoe, line 128:
```suggestion
```

v3/src/models/data/v2-category-set-importer.test.ts
```diff
@@ -88,9 +88,15 @@ describe("importV2CategorySet", () => {
...
     })
 
-    it("drops the default, which is stored as absence", () => {
+    it("keeps an explicitly chosen circle, which absence does not stand for", () => {
+      /*
+       * Absence means the category inherits the display's shape, so a circle in the document is a
+       * choice the user made and not a value to age out. Dropping it here would reinstate the
+       * inherited shape on reload -- a category deliberately set to circle would come back as star
+       * on a display whose shape is star, which is the whole reason the setter stores it.
+       */
```
tealefristoe, line 97:
```suggestion
```

v3/src/models/data/category-set.test.ts
```diff
@@ -378,15 +378,26 @@ describe("CategorySet", () => {
...
+      // absent means the category has none of its own, so it takes what it is given
+      expect(categories.shapeForCategory("a", "star")).toBe("star")
+    })
+
+    it("stores an explicitly chosen circle, which absence does not stand for", () => {
+      /*
+       * Absence means "inherit the display's shape", so it cannot double as "circle". Dropping the
+       * entry here would read back as the display's shape and leave circle unselectable for a
+       * category whenever the display was set to anything else.
+       */
```
tealefristoe, line 393:
```suggestion
```

v3/src/models/data/category-set.ts
```diff
@@ -282,14 +291,13 @@ export const CategorySet = types.model("CategorySet", {
       self.colors.delete(value)
     }
   },
-  // Storing the default is stored as absence, so a document only carries the shapes a user chose
-  // and a category reverted to circle round-trips as an unset entry.
+  /*
+   * Every choice is stored, circle included. Absence means the category has no shape of its own and
+   * inherits the display's, which is not the same as having been set to a circle: with a display
+   * shape of star, dropping a circle here would read back as star and make circle unselectable.
+   */
```
tealefristoe, line 298:
```suggestion
```

tealefristoe (approved):
Looks good 👍

v3/src/components/graph/plots/bar-chart/bar-chart.tsx
kswenson: Taken — and it was eight, not seven: the map layer does the same resolution inline.

`legendShapeGetter` in `data-display-utils` builds the getter from the data configuration and the display description; every call site now just passes it along.

v3/src/components/data-display/models/data-configuration-model.ts
kswenson: Taken. It's `legendHasCategories` now, on the model.

Worth recording why it isn't the existing `isCategoricalAttributeType`: that one counts a **color** legend as categorical, and a color legend is exactly the case this has to exclude — it gives each case a color of its own, so there are no categories to hang a shape on. Reusing it would have been wrong in a way that produced no error.

v3/src/components/data-display/renderer/point-shapes.ts
kswenson: Taken, and better than that — `IShapePoint` was an exact duplicate of `Point`, which `data-display-types` already exports. It's gone and the shared one is used.

`IShapeExtent` had no equivalent, so it moved to the same file as `Extent`.

v3/src/components/data-display/renderer/point-shapes.test.ts
kswenson: Mostly right, but not entirely — two things that test doesn't cover:

- It skips `circle` (`PointShapes.filter(s => s !== "circle")`), so the assertion that a circle's bounding radius is exactly `r` was only in the test above.
- The triangle test's *first* assertion is `apexDistance > extent.w / 2`, which is a different claim: it's the reason the function measures vertices rather than the extent at all.

Folded into two tests that keep both, rather than three. Thanks for pushing — the middle one was doing most of the work.

v3/src/components/data-display/renderer/point-shapes.test.ts
kswenson: No good reason — the file was just inconsistent. There's one radius constant for the whole file now, used everywhere, rather than a constant in one describe and the literal in twenty other places.

v3/src/components/data-display/models/data-configuration-model.ts
kswenson: I'd like to keep this one.

It's the comment that stops someone moving `legendCollectionIsMoreChildmost` back in beside the two views that use it — which is where it was, and which broke legend colors on **every graph**. The plots pass these views around as detached function references, so `this` inside them is undefined; nothing about the code at either end shows that, and the unit tests didn't catch it because they call the views on the instance. The Cypress smoke tests did.

Happy to shorten it, but I think the placement needs a reason attached or it will get tidied back.

v3/src/components/data-display/renderer/canvas-point-renderer.ts
kswenson: Keeping this one too, if that's alright.

`ctx.lineJoin = "round"` looks like a line you could delete with no consequence. Without it the star's 36-degree points and the X's corners grow spikes at small radii under the default miter join — visible, and not something you'd predict from reading the assignment.

v3/src/components/graph/hooks/use-dot-plot.ts
kswenson: I'd keep this. It's the explanation of a bug that already happened once.

The dot plot captured the appearance at render time, and the shape reaction runs before the re-render that would rebuild the callback — so the plot drew the shape it had *before* each change, and the first change appeared to do nothing. Removing the note invites someone to hoist those reads back out of the callback, which is the whole bug.

I did take your reword on the sentence above it.

v3/src/models/data/category-set.ts
kswenson: Keeping a trimmed version of this.

`self.shapes.set(value, shape)` looks like it could obviously skip storing the default. It can't: absence means "inherit the display's shape", so dropping a circle makes it read back as whatever the display is set to, and circle becomes unselectable for a category. That was a real bug on this stack, found on #2692 by you and independently by Copilot.

Two lines now instead of five.

v3/src/models/data/v2-category-set-importer.ts
kswenson: Trimmed rather than dropped.

The part worth keeping is the contrast with the loop immediately below, which *does* filter its entries against the current categories. Without a word, shapes not being filtered reads as an oversight rather than a decision. Two lines now.

v3/src/components/data-display/models/data-configuration-model.ts
kswenson: I'd keep this one, shortened if you like.

`legendShapeDomain` isn't read directly anywhere — it exists as a reaction dependency, and "changes identity when any category's shape changes" is the property that makes it work. A reader of the model has no way to see that from here; the consumers are in `use-plot.ts` and `map-point-layer.tsx`.

v3/src/components/data-display/renderer/point-shapes.ts
kswenson: Taken, and your instinct about the ambiguity is right — "point" was doing two jobs in one sentence.

I've made the first line "Whether (dx, dy), relative to a plotted point's center, hits it", which at least pins which sense is meant. Replacing it with "shape" would be wrong: the thing being hit is a plotted point, and the shape is how it's drawn.

I left the second paragraph as is for now. It's the argument for unioning with the circle, and I'd rather it stayed slightly long than have someone "simplify" it to testing the ink alone and quietly make a plus harder to click than a circle.

v3/src/components/graph/utilities/graph-utils.ts
kswenson: Fair question, and re-reading it the sentence is doing too much.

The point is that `setPointCoordinates` both **creates** points and moves them. The restyle path is what carries the per-case shape, so if this path didn't also write one, a newly created point would be drawn as a circle until some unrelated restyle happened to run. Adding a case to a plot with shapes did exactly that before this was fixed.

Reworded to your suggestion, which keeps that and drops the rest.

v3/src/components/data-display/renderer/point-shapes.ts
kswenson: Agreed — and it was worse than losing meaning: `js/shapes.js` names a file in a prototype repo that no reader of this line can reach.

Reworded to say the constants come from the design prototype the feature was specified from and are tuned rather than derived, which is the part that matters. I removed the other prototype references in this file too, and the ones in the inspector SCSS on #2693 for the same reason.

v3/src/components/data-display/models/data-configuration-model.ts
kswenson: Both taken. The first part was orientation rather than information, and the second is now covered by `legendHasCategories` having a name.

What's left is one line saying what `shapeIfNoCategory` is, which the parameter name alone doesn't give you.

v3/src/components/data-display/models/data-configuration-model.ts
kswenson: Gone.

v3/src/components/data-display/renderer/canvas-point-renderer.ts
kswenson: Gone.

v3/src/components/data-display/renderer/point-renderer-types.ts
kswenson: Gone.

v3/src/components/data-display/renderer/point-shapes.ts
kswenson: Gone.

v3/src/components/data-display/data-display-utils.ts
kswenson: Gone.

v3/src/components/data-display/inspector/display-item-format-control.scss
kswenson: Taken.

v3/src/components/data-display/renderer/point-shapes.test.ts
kswenson: Taken.

v3/src/components/data-display/renderer/point-shapes.ts
kswenson: Taken.

v3/src/components/data-display/renderer/point-shapes.ts
kswenson: Taken.

v3/src/components/graph/utilities/graph-utils.ts
kswenson: Taken.

v3/src/components/graph/hooks/use-dot-plot.ts
kswenson: Taken.

v3/src/components/data-display/data-display-utils.ts
kswenson: Taken.

v3/src/components/graph/models/graph-data-configuration-model.test.ts
kswenson: I'd keep this one — it explains why the test exists rather than what it does, and without it the test looks like a duplicate of one nearby.

v3/src/components/graph/models/graph-data-configuration-model.test.ts
kswenson: I'd keep this one — it explains why the test exists rather than what it does, and without it the test looks like a duplicate of one nearby.

v3/src/v2/v2-document-round-trip.test.ts
kswenson: I'd keep this one — it explains why the test exists rather than what it does, and without it the test looks like a duplicate of one nearby.

v3/src/models/data/v2-category-set-importer.test.ts
kswenson: I'd keep this one — it explains why the test exists rather than what it does, and without it the test looks like a duplicate of one nearby.

v3/src/models/data/category-set.test.ts
kswenson: I'd keep this one — it explains why the test exists rather than what it does, and without it the test looks like a duplicate of one nearby.

PR #2697, opened 2026-09-10
CODAP-1509: draw and hit test point shapes in the PIXI renderer

kswenson:
Draws the seven point shapes in the PIXI/WebGL renderer and hit tests against them. Fifth in the LEADS Point Shapes stack, on top of #2696.

Fixes CODAP-1509.

## What was actually needed

Very little plumbing. Nothing in the style path branches on renderer type, so the work done for the canvas renderer already carries shapes to PIXI through the same `setPointStyle` and `matchPointsToData` calls — and `point-shapes.ts` is consumed unchanged, which is what it was built for. The texture cache needed no change either: its key is the whole style, which now carries the shape, so two shapes cannot collide on one texture.

Two things did need care.

## Centering

A sprite draws its texture around its anchor, which for points is the middle of the texture. Left to size itself, a texture takes the bounds of the ink — and a triangle's ink is not centered on the point, because the shape sits on its center of area. Anchoring the middle of that box would draw a triangle low by about a third of its radius, reinstating in WebGL exactly the positional bias the canvas renderer was fixed for. At r=8 the offset is 2.9px for a triangle and 1.1px for a star; every other shape is symmetric and unaffected.

The texture is given an explicit frame centered on the point instead, sized by a new `pointShapeSymmetricExtent`. Circles keep the path they have always had, including sizing their own texture.

## Hit testing, and a behavior change worth noticing

Sprites now carry a `hitArea` that tests the drawn shape, deferring to the same containment the canvas renderer uses so the two renderers agree on what counts as a hit. That matters because graphics acceleration is a user setting — the same document should not pick differently depending on it.

**This also tightens picking for circles.** Sprites previously had no `hitArea` at all, so they hit tested against the rectangle of their texture: a click up to about 1.4 radii from a point's center would select it. That stops. It is a fix rather than a regression, but it reaches every existing document, so it is worth a reviewer's attention rather than being buried.

Bars keep the rectangular test, which is what a bar is.

## Scope

Behind the `pointShapes` feature flag, which gates the controls. Verified manually with graphics acceleration on, and with it toggled off and on to confirm the two renderers agree.

Not measured: texture churn with many shapes in play. The count stays bounded — shape and color are both functions of the same legend category — but `cleanupUnusedTextures()` still has its single call site in the ticker's idle branch.

### DIFF
--- a/v3/src/components/data-display/renderer/pixi-point-renderer.test.ts
+++ b/v3/src/components/data-display/renderer/pixi-point-renderer.test.ts
@@ -18,6 +18,7 @@ jest.mock("pixi.js", () => {
     destroy() { this.children = [] }
   }
   class MockSprite {
+    hitArea: any = null
     anchor = { x: 0, y: 0, copyFrom(p: any) { this.x = p.x; this.y = p.y } }
     position = { x: 0, y: 0, set(x: number, y: number) { this.x = x; this.y = y } }
     scale = { x: 1, y: 1, set(x: number, y: number) { this.x = x; this.y = y } }
@@ -39,8 +40,11 @@ jest.mock("pixi.js", () => {
   }
   class MockGraphics {
     boundsArea: any = null
-    rect() { return this }
-    circle() { return this }
+    // records what was traced, so a test can tell a polygon from an arc
+    traced: Array<{ op: string, args: any[] }> = []
+    rect(...args: any[]) { this.traced.push({ op: "rect", args }); return this }
+    circle(...args: any[]) { this.traced.push({ op: "circle", args }); return this }
+    poly(...args: any[]) { this.traced.push({ op: "poly", args }); return this }
     fill() { return this }
     stroke() { return this }
     destroy() {}
@@ -68,7 +72,12 @@ jest.mock("pixi.js", () => {
     }
     resize() {}
     render() {}
-    generateTexture() { return new MockTexture() }
+    // keeps every call, so a test can inspect the graphics traced and the frame requested
+    generateTextureCalls: any[] = []
+    generateTexture(options: any) {
+      this.generateTextureCalls.push(options)
+      return new MockTexture()
+    }
     destroy() {}
   }
   return {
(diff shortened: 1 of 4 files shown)


kswenson:
**Reviewing this stack** — five PRs, each based on the one below it, merging bottom-up into `leads-point-shapes`:

| PR | Story | What it is |
|---|---|---|
| #2697 | CODAP-1509 | PIXI/WebGL rendering and hit testing |

**CODAP-1506 is split across #2691 and #2693**, so approving #2691 does not complete that story — it should not leave code review until #2693 is approved too. The other three map one-to-one.

**Nothing here is superseded** — this is the top of the stack.

**Worth your attention:**

- The centering fix. This is the one that would have shipped as a subtle wrong-looking plot rather than an obvious break: a triangle drawn ~2.9px low at r=8, which reads as "the points are slightly off" rather than as a bug in shape rendering.
- The hit-testing change tightens picking for **circles**, in every existing document — sprites previously had no `hitArea` at all and tested against their texture's rectangle. Called out in the description; I'd rather it be argued about than slip through.
- `point-shapes.ts` is consumed unchanged from #2696. If something about the geometry looks wrong here, it belongs on that PR.

Out of scope by design, each with a ticket: legend keys carrying the shape (CODAP-1507), the residual plot (CODAP-1526).

emcelroy (approved):
Looks good 👍 Code makes sense and the feature works well when I test it out in a browser.

There's a conflict with the base branch that needs to be addressed before merging.

Below is a review generated with Claude that raises some issues to consider. In my experience testing the UI, some of the issues are pretty subtle (2, 3 and 4) and may fall into the realm of not worth addressing, but I'll let you be the judge of that.  

# Review: CODAP-1509 point shapes in PIXI

Approach is right. Sharing `isPointInShape` between renderers is the correct call, and the tightened circle picking is worth it. All findings below sit in the two places the PR flags as needing care: the centering and the hit test.

Verified against PIXI 8.18.1 source and by measurement. Tests and lint pass on the branch.

## Fix before merge

**1. Hit testing rebuilds the shape outline on every call.** `pixi-point-renderer.ts:55`

`contains` calls `isPointInShape`, which calls `pointShapeGeometry`, which builds a fresh vertex list — ten `cos`/`sin` pairs for a star. PIXI calls this once per sprite per pointer event; `hitPruneFn` uses the hit area with no cheaper bounds check first. The early return only fires on a hit, and almost every test is a miss.

Measured, 5,000 points, one pointer event: circle 0.02 ms, triangle 0.27 ms, x 0.81 ms,
**star 4.36 ms**. Holding the polygon and rejecting first against `pointShapeBoundingRadius`: 0.015 ms.

The canvas renderer avoids this only because its grid narrows candidates first. PIXI has no such narrowing.

Fix: cache the polygon and bounding radius in `PointShapeHitArea`, recompute only when `syncHitArea` changes shape or radius. Keep containment shared between renderers.

**2. PIXI truncates the frame, so shapes land off centre by a shape-dependent amount.** `pixi-point-renderer.ts:36`

`GenerateTextureSystem` does `region.width = Math.max(region.width, 1/resolution) | 0` but leaves `region.x` alone. A frame of x = -7.8, width 15.6 gives a 15-wide texture with the shape centre at 7.8 and the anchor at 7.5, so the shape draws 0.3 px right of its position and hit area.

Offsets at stroke width 1: square 0.30 px at r=8, x 0.43 px at r=8, triangle 0.38 px at r=5, plus 0.40 px at r=5. Circles are 0.00 at whole radii. Sub-pixel, but it varies by shape, so recoloring a legend category shifts those points.

Fix: whole-number half-extents, so the truncation is a no-op.

**3. Default miter join clips star tips and diverges from canvas.** `pixi-point-renderer.ts:882`

PIXI defaults to `join: "miter"`. `canvas-point-renderer.ts:599` sets `lineJoin = "round"` with a comment saying miter grows spikes on the star's points and the X's corners at small radii. So PIXI both clips and draws differently from canvas.

At stroke width 2 the miter reaches 2.39 px past a star tip against 2 px of padding. Three of the five tips clip: the top by 0.39 px, the two side tips by 0.27 px.

Fix: `join: "round"`. Matches canvas, removes the spikes, and brings the stroke reach inside the existing padding.

**4. Bars-to-points transition ends with a size jump.** `pixi-point-renderer.ts:975-1005`

The sprite animates to `radius * 2` wide, then the texture swaps in at scale 1. A circle texture is `2r + strokeWidth`, so a small mismatch already existed. A star texture is about `2.7r`, so it jumps roughly 40%.

Fix: size the transition from the destination texture's real dimensions.

## Worth hardening

**5. `doMatchPointsToData` updates textures on existing sprites but not hit areas.**
`pixi-point-renderer.ts:493-500`

New sprites get a hit area via `getNewSprite`; existing ones get only a texture. On a display-type change that starts no transition (it needs `sprites.size === caseData.length`), a sprite keeps the wrong test.

Not reachable today: `setPointCoordinates` calls `setPointStyle` on every point straight after, which calls `syncHitArea`. Still worth closing.

The hit area must follow whichever style produced the texture actually assigned, not the point's stored style — in that loop the texture comes from the uniform `style`, and the two can differ.

A points-to-bars test showing an existing shape hit area getting cleared would cover this.

## Notes

- **The map is affected too.** `map-point-layer.tsx:362` passes `getLegendShapeForCase`, so map points take the frame centering and the hit-test cost, not just the tighter circle picking. Include maps in verification.
- **Tighter circle picking, quantified.** Against the old `(2r + strokeWidth)` rectangle the target area drops to about 58% at r=3, and maximum reach from 1.41r to 1.0r. With `pointRadiusMin` = 3 and a size multiplier below 1 that is a small target. Intentional and disclosed; raising it as a product question, not a defect. Any floor belongs in `isPointInShape` so both renderers stay aligned.
- **`pointShapeExtent` is test-only** and already was before this PR, but its comment still says "Used to size a shape against a box." Worth correcting so nobody reaches for it instead of `pointShapeSymmetricExtent`.
- **"keeps every shape at least as easy to hit as a circle"** tests only `plus`. Reaching the circle branch is the contract, so that part is right; parameterising over `PointShapes` would match the name.

kswenson:
Took all five, plus two of the notes. Merged `leads-point-shapes` in; the conflict was the `IShapePoint`/`Point` rename that came out of review on #2696.

**1. Hit-test cost.** Confirmed the mechanism: `EventBoundary.hitPruneFn` calls `hitArea.contains` with no bounds test of its own, and every sprite is `eventMode: "static"`, so one pointer move over empty space is one call per point. `PointShapeHitArea` now holds the outline and a reach cutoff and rejects against the cutoff first. My numbers came out lower than yours — star 0.86 ms rather than 4.36 ms at 5,000 points, different machine — but the ratio is the same, and it was ~100× a circle for nothing.

**2. Frame truncation.** Confirmed at `GenerateTextureSystem.js:79`. The padded extents are rounded up, so the truncation is a no-op.

**3. Miter join.** Now `join: "round"`.

**4. Bars-to-points jump.** The transition sizes from the destination texture's real dimensions, which also closes the stroke-width mismatch circles already had.

**5. `doMatchPointsToData`.** Syncs against the uniform style, since that's the one that drew the texture being assigned — as you noted, not the point's stored style.

On the notes:

- `pointShapeExtent`'s comment now says it checks the tuning, and points at `pointShapeSymmetricExtent` for sizing.
- The `plus`-only test I renamed rather than parameterized. `point-shapes.test.ts` already asserts the circle floor across all seven shapes; running it through PIXI seven times re-tests one delegation. It's now "applies the circle floor to a shape narrower than it".
- **Circle picking floor**: leaving as-is, and agreed it's a product question rather than a code one. If we do want a minimum target at small radii, it belongs in `isPointInShape` so both renderers stay aligned. Happy to open a story if it's worth raising with the team.
- **Maps**: `map-point-layer.tsx` is on the same path, so it takes the centering and hit-test changes too.

Every fix has a test, and each test was checked by breaking the fix and watching it fail.

PR #2699, opened 2026-09-10
CODAP-1526: draw residual points with the category's shape

kswenson:
Draws residual-plot points with the category's shape instead of always a circle. Seventh in the LEADS Point Shapes stack, on top of #2698.

Fixes CODAP-1526.

## Why this exists separately

The residual plot draws its own SVG rather than going through a point renderer, so it is a fourth surface that draws a point — after the canvas renderer (#2696), PIXI (#2697) and the legend key (#2698). It stayed circles whatever the plot above it was drawn with. This is the last one.

## The scope was wider than the code claimed

A comment in `use-residual-plot.ts` said `residualPlotIsApplicable` excludes legends, so the per-case legend lookup was dead. That is not what it does: it turns the plot off for a *categorical* legend only when the line is an **LSRL**. With a movable line or a plotted function a categorical legend is allowed, so the legend path is live — and per-category shapes genuinely reach these points, rather than only the display's own shape. The comment is corrected.

## Rendering

A point is a `<path>` positioned with `translate(cx, cy)`. The radius moves to a `data-r` attribute, because an outline has no radius to read back the way a circle does, and the existing fade-in machinery probes it.

**The grow animation regenerates the outline rather than scaling the point.** A `scale()` transform takes the stroke with it, and the stroke is what marks a point as selected — so a selected point would have faded in with an outline that thickened as it grew. Regenerating via `attrTween` leaves the whole existing fade-in intact: the named `"radius"` transition, the `active()` guard, and the deliberate choice not to redirect a transition already past its starting phase.

The shape comes from the same `getLegendShapeForCase` lookup the main plot uses, so the two plots agree by construction.

**A point is a group, not a bare path.** An unpainted circle of radius r sits behind the outline, because a path takes pointer events on its ink alone — a plus reaches only 58% of its radius toward the notches, so a click there would miss the point and fall through to the background, whose handler deselects everything. Circle plus ink is what `isPointInShape` gives the canvas and PIXI renderers, so all three agree on what counts as a hit.

The stroke joins are rounded, as on the other three surfaces. Both of these requirements are now written down in `point-shapes.ts` rather than rediscovered per surface.

## Testing

Seven Cypress assertions counted `circle` elements and now count paths. The marquee is untouched — it indexes positions it computes itself rather than querying elements.

Eight unit tests cover `drawShape` and `growPointIn`, and a Cypress test covers the hit target by clicking a plus's notch.

The fade-in tests run against real d3 transitions on a real clock. Jest's fake timers cannot drive them: `d3-timer` binds `window.requestAnimationFrame` when it is imported, so it keeps using the real one and its cached clock never advances past the value it first read. The test file says so, since the obvious thing to reach for is fake timers.

*(An earlier version of this description said there was no testable seam. That was true of the hook as written — extracting `growPointIn`, which is "grow this point in and reconcile its shape when it lands", made the fade-in reachable without standing up a graphModel/layout/dataset harness.)*

### DIFF
--- a/v3/cypress/e2e/bivariate-adornments.spec.ts
+++ b/v3/cypress/e2e/bivariate-adornments.spec.ts
@@ -371,7 +371,7 @@ context("Graph adornments", () => {
     cy.get("[data-testid=adornment-checkbox-movable-line]").click()
     cy.get("[data-testid=adornment-checkbox-residual-plot]").find("input").should("not.have.attr", "disabled")
     cy.get("[data-testid=adornment-checkbox-residual-plot]").click()
-    cy.get("*[data-testid^=residual-points-]").find("circle").should("have.length.at.least", 1)
+    cy.get("*[data-testid^=residual-points-]").find("path").should("have.length.at.least", 1)
     cy.get(".axis-wrapper.leftLower").should("exist")
 
     // V2 behavior: removing the line CLEARS (unchecks) the Residual Plot rather than
@@ -392,7 +392,7 @@ context("Graph adornments", () => {
     // Re-checking restores the residual plot. (Legend-triggered clearing follows the same
     // residualPlotIsApplicable path and is covered by the graph-content-model integration tests.)
     cy.get("[data-testid=adornment-checkbox-residual-plot]").click()
-    cy.get("*[data-testid^=residual-points-]").find("circle").should("have.length.at.least", 1)
+    cy.get("*[data-testid^=residual-points-]").find("path").should("have.length.at.least", 1)
     cy.get(".axis-wrapper.leftLower").should("exist")
   })
   it("restyles residual points to reflect selection without tearing down the residual plot", () => {
@@ -402,7 +402,7 @@ context("Graph adornments", () => {
     graph.getDisplayValuesButton().click()
     cy.get("[data-testid=adornment-checkbox-movable-line]").click()
     cy.get("[data-testid=adornment-checkbox-residual-plot]").click()
-    cy.get("*[data-testid^=residual-points-]").find("circle").should("have.length.at.least", 1)
+    cy.get("*[data-testid^=residual-points-]").find("path").should("have.length.at.least", 1)
 
     // A residual point starts unselected — its fill is not the solid selection color.
     cy.get("*[data-testid^=residual-point-]").first().should("not.have.attr", "fill", "#4682b4")
@@ -413,7 +413,7 @@ context("Graph adornments", () => {
     // sibling circle — force the click on the specific first circle.
     cy.get("*[data-testid^=residual-point-]").first().click({ force: true })
     cy.get("*[data-testid^=residual-point-]").first().should("have.attr", "fill", "#4682b4")
-    cy.get("*[data-testid^=residual-points-]").find("circle").should("have.length.at.least", 1)
+    cy.get("*[data-testid^=residual-points-]").find("path").should("have.length.at.least", 1)
     cy.get(".axis-wrapper.leftLower").should("exist")
 
     // Clicking the residual-plot background deselects all cases, reverting the styling.
@@ -428,7 +428,7 @@ context("Graph adornments", () => {
     graph.getDisplayValuesButton().click()
     cy.get("[data-testid=adornment-checkbox-movable-line]").click()
     cy.get("[data-testid=adornment-checkbox-residual-plot]").click()
-    cy.get("*[data-testid^=residual-points-]").find("circle").should("have.length.at.least", 1)
+    cy.get("*[data-testid^=residual-points-]").find("path").should("have.length.at.least", 1)
 
     // No residual point carries the solid selection fill before the drag.
     cy.get('*[data-testid^=residual-point-][fill="#4682b4"]').should("not.exist")
@@ -450,7 +450,7 @@ context("Graph adornments", () => {
     // The marquee covered the strip, so at least one residual point is now selected (selection fill),
     // and the residual plot is not torn down.
     cy.get('*[data-testid^=residual-point-][fill="#4682b4"]').should("have.length.at.least", 1)
-    cy.get("*[data-testid^=residual-points-]").find("circle").should("have.length.at.least", 1)
+    cy.get("*[data-testid^=residual-points-]").find("path").should("have.length.at.least", 1)
   })
 
   it("adds to the existing selection when the residual marquee is shift-dragged", () => {
@@ -460,7 +460,7 @@ context("Graph adornments", () => {
     graph.getDisplayValuesButton().click()
     cy.get("[data-testid=adornment-checkbox-movable-line]").click()
     cy.get("[data-testid=adornment-checkbox-residual-plot]").click()
-    cy.get("*[data-testid^=residual-points-]").find("circle").should("have.length.at.least", 1)
+    cy.get("*[data-testid^=residual-points-]").find("path").should("have.length.at.least", 1)
 
     // Drags a marquee over a horizontal slice of the residual strip, expressed as fractions of its
     // width. pointermove/pointerup go to the document so they reach the marquee's window listeners.
--- a/v3/src/components/graph/plots/scatter-plot/residual-marquee-utils.ts
+++ b/v3/src/components/graph/plots/scatter-plot/residual-marquee-utils.ts
@@ -9,7 +9,7 @@ export interface IResidualPosition {
 
 // Screen position of each residual point, in the graph-SVG coordinate frame: x from the shared
 // getXCoord, y on the lower axis (plotHeight is the top of the residual region). Mirrors the cx/cy
-// used to draw the residual circles, so the hit-test index and the drawn points agree by construction.
+// used to draw the residual points, so the hit-test index and the drawn points agree by construction.
 export function buildResidualPositions(
   residuals: IResidualPoint[], getXCoord: (caseID: string) => number,
   plotHeight: number, lowerScale: (residual: number) => number
--- a/v3/src/components/graph/plots/scatter-plot/use-residual-plot.ts
+++ b/v3/src/components/graph/plots/scatter-plot/use-residual-plot.ts
@@ -1,4 +1,4 @@
-import { active, format, ScaleLinear, select } from "d3"
+import { active, format, interpolateNumber, ScaleLinear, select } from "d3"
 import { tip as d3tip } from "d3-v6-tip"
 import { untracked } from "mobx"
 import { useCallback, useEffect, useRef } from "react"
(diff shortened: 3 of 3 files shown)


kswenson:
**Reviewing this stack** — seven PRs, each based on the one below it, merging bottom-up into `leads-point-shapes`:

| PR | Story | What it is |
|---|---|---|
| #2697 | CODAP-1509 | PIXI/WebGL rendering and hit testing |
| #2698 | CODAP-1507 | Legend keys |
| #2699 | CODAP-1526 | Residual plot — this one |

**CODAP-1506 is split across #2691 and #2693**, so approving #2691 does not complete that story — it should not leave code review until #2693 is approved too. The other five map one-to-one.

**Nothing here is superseded** — this is the top of the stack, and the last of the four surfaces that draw a point.

**Worth your attention:**

- **The fade-in.** A new residual point grows from nothing, and this regenerates its outline at each step rather than scaling it — because a scale would take the stroke with it, and the stroke is how a selected point is marked. That is the part most worth a skeptical look, and the one thing here with no unit coverage.
- **The corrected comment.** The file claimed residual plots never have a legend, which is why the per-case lookup looked dead. It is only excluded for a categorical legend *with an LSRL*, so a movable line or plotted function has one and per-category shapes do reach these points.
- `point-shapes.ts` is unchanged from #2698. Anything about the geometry belongs there.

This completes the set: canvas (#2696), PIXI (#2697), legend keys (#2698), residual plot (#2699). Nothing else in the codebase draws a per-case mark.

emcelroy (approved):
Overall looks good 👍 The review below was generated with Claude and raises a few issues to consider before merging. Issue 1 is difficult to actually trigger using the UI as a human, but you may still want to correct it.

---

The approach is right, and two calls in the description are good ones. Regenerating the outline instead of applying a `scale()` is correct for the reason given — the stroke width is what marks a residual point selected, so a scaled fade-in would have thickened it. The `residualPlotIsApplicable` correction is also right: line 55 excludes a categorical legend **only** when `kind === "lsrl"`, so the legend path is live for a movable line or a plotted function, and the old comment was wrong.

Four things to fix, one inaccurate comment, and a stale base.

## Fix before merge

**1. A shape chosen during a point's fade-in is discarded, and the point stays wrong.**

`use-residual-plot.ts:122`

The `attrTween` closes over `style.shape` at the moment the transition starts. While it runs, `active(this, "radius")` is truthy, so every later `applyResidualStyles` skips the whole radius block and only updates fill and stroke. The tween finishes with the shape it captured.

`d3.active` returns a transition whenever `state > SCHEDULED`, and `transitionDuration` is 1000, so the window is a full second after any point appears — turning the residual plot on, adding cases, re-activating the plot.

I reproduced it with the real logic and real d3 transitions in jsdom: a point fading in as a circle, shape changed to star halfway, ends as a circle. `data-r` is already at its final value, so the else branch would snap it — but nothing schedules another restyle, so it stays a circle beside its neighbors until an unrelated selection change or repaint.

The file already documents this tradeoff for *radius*: "the point-radius-changed-mid-fade case ... is rare enough that finishing at the old target is acceptable." That reasoning was about a slightly wrong size. A wrong shape reports the wrong category, which is what the whole stack exists to encode.

Fix: reconcile the shape when the transition ends, or have the tween read current state rather than a captured value. A regression test for this needs no browser — the reproduction above is a plain jest test.

**2. The click target is the ink only, which is what the main plot deliberately refuses.**

`use-residual-plot.ts:212`

`isPointInShape` unions the drawn ink with the circle of radius r, and says why: "choosing one must not make a point harder to click than it was as a circle: testing the ink alone would open dead zones between a star's arms and in the notches of a plus." Both point renderers hit test that way.

An SVG path takes events on its fill and stroke. The click handler is bound to the path, so the residual plot now does exactly what that comment rules out. For a plus at r=8, the offset (5, 5) is inside the circle (7.07 < 8) and outside the arms (half-width 3.28): a hit in the upper plot, a miss below it.

Worse than a miss — the click falls through to `residual-plot-background`, whose `onClick` deselects every case. So clicking a point's notch clears the selection.

Minimum distance from centre to the fill boundary, as a fraction of r (the stroke pushes this out by half a stroke width):

| circle | square | diamond | triangle | star | plus | x |
|---|---|---|---|---|---|---|
| 100% | 85% | 85% | 74% | 68% | 58% | 58% |

Fix: put the handlers on a group holding an invisible circle of radius r plus the visible path, keeping the tips outside that circle clickable. More invasive than it sounds, because the data join is keyed on `"path"`. #2698 needs the same idea for legend keys, though its backing target is rectangular rather than circular.

This one needs a browser test.

**3. The path string is rebuilt for every point on every restyle.**

`refreshPointSelection` calls `restyleResidualSelection()` with no arguments on every selection change, and its own comment says the `caseIds` delta "doesn't apply", so all residual points are restyled — including on every marquee-drag frame. That O(n) pass is pre-existing. What changed is its cost: the else branch used to set `r` to a number and now builds a path string.

Measured per restyle of 5,000 points: circle 0.26 ms, plus 8.7 ms, star 12.4 ms. Per drag frame.

Fix: skip the rebuild when an element's shape and radius are unchanged — store `data-shape` beside `data-r`. That removes the work rather than caching it. Do **not** add unbounded memoization to `pointShapePathData`: the tween calls it at every intermediate radius, so the key space is not small.

**4. No `stroke-linejoin: round` — the only one of the four surfaces that miters.**

`use-residual-plot.ts:135`

The style block sets fill, stroke, stroke-width and stroke-opacity, and nothing sets the join, so SVG defaults to miter. The canvas renderer sets `lineJoin = "round"`, PIXI passes `join: "round"`, and the legend key got `stroke-linejoin: round` in `c760db798` — all for the reason in the canvas renderer's comment: without it "the star's 36-degree points and the X's corners grow spikes at small radii."

A star's tip miter reaches 1.19 × the stroke width. Residual points use stroke width 2 when selected *and* a legend is present (`useSelectionFill` takes the no-legend case), which is the configuration per-category shapes require — so a selected star grows a 2.4 px spike on each tip at the same radius the upper plot draws cleanly.

Fix: `.attr("stroke-linejoin", "round")` in the same chain. `connecting-lines.ts` already uses that attribute.

## Not accurate

**Residuals are not taken against a per-category line.**

The new comment says "Each point's residual is taken against its own category's line and the adornments store a line per cell," and the description repeats it.

`getAllCellKeys` builds cells from the x, y, top-split and right-split categories only. The legend is not a cell dimension — it appears separately in `getAllCaseSubsetDescriptions`, which is what the LSRL uses to fit a line per legend category, and the LSRL is exactly the case `residualPlotIsApplicable` excludes. That check also rejects `topSplit`, `rightSplit`, and non-numeric x or y. So whenever the legend path runs there is exactly one cell, one line, and every point's residual is measured against it.

Color and shape still correctly say which category a point belongs to. But the justification as written would tell a future reader that residuals are computed per category. Worth rewording, since the point of this commit's other comment change was to stop a comment saying something untrue.

The sibling is still there: `residual-plot-utils.ts:61-63` still says "The Residual Plot's v1 constraints exclude legend / categorical axes." Its conclusion — always the default cell — is right, for a reason that is not.

## Merge the base

The branch is seven commits behind `CODAP-1507-legend-key-shapes`, which has since merged the fixed PIXI branch and added its own mitre fix. A trial merge is clean. Worth doing before the rest, because after it the residual plot is the only surface on the stack still mitering, which makes finding 4 a gap rather than an oversight.

## Notes

- **The reaction itself is wired correctly.** `usePlot` reacts to `[legendShapeDomain, pointDescription.pointShape]` and calls `refreshPointSelection`, which reaches `restyleResidualSelection`. Both display-wide and per-category shape changes do repaint — except during a fade-in, which is finding 1.
- **The marquee claim holds.** `buildResidualPositions` indexes centres and `clampRectToLowerRegion` tests them against a rect. No radius, nothing to change.
- **Stale "circle" comments.** Several survive in `use-residual-plot.ts` (lines 106, 108,
  162, 217, 233), and `scatter-plot.tsx:232` still says the delta "doesn't apply to our SVG circles". Neighboring ones were updated, so these read as missed.
- **Test coverage.** The description says there is no testable seam. That is too strong: finding 1 is reproducible in jest against real d3 transitions, and so is finding 3's caching behavior. Only finding 2 needs a browser.

kswenson:
Took all four, plus the comment and the stale ones. Base merged — nine commits, since #2697 and #2698 both landed since you looked.

**1. Shape discarded mid-fade.** Fixed, and your diagnosis was exact: the tween closes over `style.shape`, and `active()` makes every restyle skip the outline while it runs. The style is consulted once more at the end.

Your point about severity is the one that decided how to fix it rather than whether to. The existing comment accepts a stale *radius* for that second; a stale shape names a wrong category, which is the thing the feature exists to encode.

**2. Ink-only click target.** Fixed with a group holding an unpainted circle of radius r plus the outline. I recomputed your whole table independently and every figure holds — plus and x at 58% is `a√2`, star 68% is the inner vertex, triangle 74% the incircle, square and diamond 85%. The fall-through to the background's deselect handler is the part that made this the one I'd least want to ship.

You were right that it's more invasive than it sounds, and it turned up something: two teardown paths removed points with `selectAll("path").remove()`. Left alone, those would have stripped the outlines and left orphaned target circles behind — invisible, and still catching clicks. Both now remove the group.

**3. Path rebuilt per restyle.** Fixed with a `data-shape` guard beside `data-r`. My measurements came out lower than yours — star 4.89ms, plus 4.69ms, x 5.40ms per 5,000 points against a circle's 0.03ms — and mine exclude the `.attr()` DOM write, so the real figure is above mine either way. The ratio is what matters and it's ~160×. Your warning against memoizing `pointShapePathData` is right: the tween asks for a new radius every frame.

**4. Miter joins.** Fixed.

**The inaccurate comment.** Agreed, and it's worse than you flagged — the *pre-existing* comment said the same thing, so this PR reworded an inaccuracy in a commit whose other comment change existed precisely to stop a comment saying something untrue. Verified your reasoning: `getAllCellKeys` builds from the axis and split categories, both splits are rejected, so there is one cell and one line. `residual-plot-utils` is corrected too.

**On the recurrence.** This was the fourth surface needing rounded joins and the third needing the hit target floored at r. Both requirements are now written down in `point-shapes.ts`. I looked at extracting shared code for them and concluded there isn't any worth writing — canvas sets `lineJoin`, PIXI passes `join`, SVG sets an attribute, and there's no common call site a helper could sit on. So it's a checklist for the next surface, stated as such.

Eight unit tests and one Cypress test; each checked by breaking the fix and confirming the right one fails. The description's claim that there was no testable seam is updated. Jest 3921 passing, lint and tsc clean, `bivariate-adornments.spec.ts` 13/13.

PR #2700, opened 2026-09-12
fix(graph): don't invalidate subplot caches for no-op categories-limit changes

zediiiii:
## Problem

Resizing a graph component with a categorical axis stalls for seconds at a time, on a dataset of only ~1,500 cases. It reproduces on school Chromebooks and on high-end desktops alike, and scales with the number of plotted dimensions (no attributes: fine; one categorical: slow; two: crawls).

## Cause

The categories limit is derived from the axis length:

```js
// axis-utils.ts:73 (and use-sub-axis.ts:184)
const axisLength = layout.getAxisLength(axisPlace),
      numCategoriesLimit = Math.floor(axisLength / kDefaultFontHeight)
dataConfig?.setNumberOfCategoriesLimitForRole(axisPlaceToAttrRole[axisPlace], numCategoriesLimit)
```

`kDefaultFontHeight` is 12, so the limit ticks over roughly every 12 pixels of a resize drag, on each axis. Any change invalidates the subplot caches:

```js
// graph-data-configuration-model.ts:705
if (self.numberOfCategoriesLimitByRole.get(role) !== limit) {
  self.subPlotCases.invalidateAll()
  self.cellMap.invalidateAll()
```

But the limit only affects anything when it is **smaller** than the number of categories — that is the only branch applying the `kOther` clamp (`data-configuration-model.ts:445`):

```js
if (categoryLimitForRole && resultArray.length > categoryLimitForRole) {
  resultArray.length = categoryLimitForRole
  resultArray[categoryLimitForRole - 1] = kOther
}
```

So once the axis is long enough to show every category, going 40 → 41 → 42 changes nothing, yet still discards a cache whose rebuild is O(cells × cases).

## Fix

Normalize limits that cannot change the result to a single value (`undefined`) so they stop counting as changes. Both the base setter and the graph override compare the normalized value.

## Evidence

From a 45-second Chrome performance trace over the slow interaction (main thread, inclusive):

| Inclusive | Frame |
|---|---|
| 25,404 ms | `batchedUpdates$1` (react-dom) |
| 22,815 ms | `runReactions` (MobX) |
| 22,322 ms | `get caseDataWithSubPlot` — `graph-data-configuration-model.ts:717` |
| 22,271 ms | `calculate` (subPlotCases) — `graph-data-configuration-model.ts:556` |

Self time is almost entirely MobX/MST property reads (`get_`, `getObservablePropValue_`, `hasProp`, `reportObserved`).

## Testing

New regression test simulates dragging a resize handle across a 300px range, stepping the limit the way the axis code does, and counts recomputes of `caseDataWithSubPlot`:

| | recomputes |
|---|---|
| before | 50 |
| after | 0 |

A second test asserts a limit *below* the category count still clamps and still introduces the `kOther` bucket, so existing behavior is preserved.

- `npm run build:tsc` — clean
- `npm run lint` — clean
- `npm run test` — 3759 passing (3757 before, plus the 2 new tests)

## Scope / what this does not fix

This addresses the *trigger*, not the cost. `caseDataWithSubPlot` iterates cells and rescans every case per cell, so it is O(cells × cases), and cell count is multiplicative (`getAllCellKeys`, line 484):

```js
const columnCount = topCatCount * xCatCount
const rowCount    = rightCatCount * yCatCount
const totalCount  = rowCount * columnCount
```

Confirmed on a local build: with one categorical attribute, resizing goes from visibly slow to smooth. With two, it is still slow, because every legitimate recompute is then ~C1 × C2 × N case evaluations.

Making that O(cases) looks feasible — `categoricalValueForCaseInRole()` already returns a case's effective category including the `kOther` bucket, and `cellKeyToString()` sorts its entries so a per-case key would match the per-cell key. I did not attempt it here because it decides which subplot every point belongs to (and therefore which mask clips it), and it has to stay correct for numeric axes (`getCategoriesOptions()` returns `xCats: [""]`, not a category list) and for the `kOther` bucket. That seemed worth a separate change by someone with more context on the splitting rules, ideally with a test asserting the new mapping matches the current one case-for-case.

Happy to adjust or split this differently.

### DIFF
--- a/v3/src/components/data-display/models/data-configuration-model.ts
+++ b/v3/src/components/data-display/models/data-configuration-model.ts
@@ -419,6 +419,25 @@ export const DataConfigurationModel = types
   }))
   .views(self => ({
     /**
+     * Normalizes a categories limit so that limits which cannot affect the result all collapse to
+     * the same value (undefined).
+     *
+     * The limit only changes what categoryArrayForAttrRole returns when there are MORE categories
+     * than the limit — that is the only case where the kOther clamp below applies. Axis code
+     * recomputes the limit from the axis length on every layout change
+     * (floor(axisLength / kDefaultFontHeight)), so during a component resize it ticks over roughly
+     * every 12 pixels. Each distinct value invalidates the subPlotCases cache, and recomputing that
+     * is O(cells x cases) — which is why resizing a graph with categorical axes gets slower with
+     * each added dimension. Collapsing every "long enough to show them all" limit to undefined
+     * keeps those no-op changes from invalidating anything.
+     */
+    effectiveCategoriesLimitForRole(role: AttrRole, limit: number | undefined) {
+      if (limit == null || limit <= 0) return undefined
+      const categoryCount = self.categorySetForAttrRole(role)?.values.length
+      if (categoryCount != null && limit >= categoryCount) return undefined
+      return limit
+    },
+    /**
      * @param role
      * @param emptyCategoryArray
      */
@@ -1098,9 +1117,7 @@ export const DataConfigurationModel = types
       categorySet?.setColorForCategory(cat, color)
     },
     setNumberOfCategoriesLimitForRole(role: AttrRole, limit: number | undefined) {
-      if (limit !== undefined && limit <= 0) {
-        limit = undefined
-      }
+      limit = self.effectiveCategoriesLimitForRole(role, limit)
       self.numberOfCategoriesLimitByRole.set(role, limit)
       self.categoryArrayForAttrRole.invalidate(role)
     },
--- a/v3/src/components/graph/models/graph-data-configuration-model.test.ts
+++ b/v3/src/components/graph/models/graph-data-configuration-model.test.ts
@@ -74,6 +74,58 @@ describe("DataConfigurationModel", () => {
     ])
   })
 
+  // Regression: the categories limit is derived from the axis length
+  // (floor(axisLength / kDefaultFontHeight)), so it ticks over every ~12px while a component is
+  // being resized. It only changes any result when it is SMALLER than the number of categories, but
+  // it used to be stored verbatim, and every distinct value invalidated the subPlotCases cache —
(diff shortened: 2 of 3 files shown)


kswenson:
Thank you for this — it's a genuinely valuable contribution, and your work is now part of a larger fix headed for review.

Your diagnosis was the hard part and you got it right: the categories limit is derived from axis length, so it churns every ~12 pixels of a resize drag, and each change was discarding caches whose rebuild is O(cells × cases). Identifying that from a profile, separating the trigger from the cost, and then deliberately *not* attempting the O(cells × cases) rewrite — that's exactly the right read, and it's what made the rest tractable.

We've built on it in #2705, which does the rewrite you scoped out. Your commit is cherry-picked in as the base with your authorship intact. Measured on the document from the original report, a subplot rebuild went from 401,034 dataset reads and 640 ms to 6,008 reads and 11.8 ms — about 54× — and in the browser under CPU throttling the whole pipeline now accounts for around 7% of a resize drag. Your fix is part of that: it suppresses roughly four-fifths of the rebuilds that would otherwise happen.

Two things we adjusted along the way, both quirks of our codebase rather than anything wrong with your reasoning. The axis code sets the categories limit *before* assigning a new attribute, which isn't visible from the code you were editing — so storing the normalized limit could drop the clamp when a higher-cardinality attribute was dropped onto an axis. We changed it to store the raw limit and normalize only for the comparison, which keeps your optimization intact. And `caseDataWithSubPlot` returns the same array instance each time, so a reference-equality reaction never fires on it — which meant the recompute counter in your test reads zero either way. That one caught us out too; we've since filed it as a latent issue in our own code.

Happy to go into more detail on any of it if you're interested. Thanks again — this started as a help-forum report and it's turning into a real fix for it.

zediiiii:
I love this program and I'm glad I was able to contribute. I wish I had
more time to really dig in and come up with a more complete fix! Thanks for
being so responsive and maintaining this wonderful resource.

PR #2702, opened 2026-09-16
CODAP-1511: LEADS Point Shapes — merge the integration branch

kswenson:
Merges `leads-point-shapes`, the integration branch for the LEADS Point Shapes feature, into `main`. The work landed on it as eight separately reviewed PRs; this is the single merge to `main`.

Epic: CODAP-1511.

## The feature

Point shape becomes a second encoding channel alongside color. Seven shapes — circle, square, triangle, diamond, star, plus, x — and every surface that draws a point honors the choice: the canvas renderer, the PIXI/WebGL renderer, the legend keys, and the residual plot.

Where the control appears follows how the legend divides the points:

- **No legend attribute** — one shape for all the points, in the same *Points* row as the point color.
- **Categorical legend** — a shape per category, beside that category's color swatch, plus a display-wide shape that a category with no choice of its own inherits.
- **Numeric or color legend** — one shape for all the points, since there are no categories to attach a shape to. With a numeric legend the glyph is painted in the legend's own gradient, so the control previews the range of points it applies to rather than any one color out of it.

Per-category shape is where the accessibility argument bites: color alone excludes readers who can't separate three hues, and a legend of three identical squares tells them nothing. Shape carries the same distinction redundantly — which is also why the categorical legend keys had to change.

Shape lives next to color in the model — per-category on the attribute's `CategorySet`, the display-wide shape on `DisplayItemDescriptionModel` — so two displays sharing a legend attribute agree by construction. The default is stored as absence: a document that never used the feature is byte-for-byte unaffected. V2 has no field for point shape, so both halves ride v3 extension namespaces and survive a v3 → v2 → v3 round trip.

## The PRs

| PR | Story | |
|---|---|---|
| #2691 | CODAP-1506 | Restructure the graph Format palette into labelled sections |
| #2692 | CODAP-1505 | Store point shape per legend category |
| #2693 | CODAP-1506 | Add the per-category shape controls |
| #2696 | CODAP-1504 | Draw and hit test shapes in the canvas renderer |
| #2697 | CODAP-1509 | Draw and hit test shapes in the PIXI renderer |
| #2698 | CODAP-1507 | Draw legend keys as the category's shape |
| #2699 | CODAP-1526 | Draw residual-plot points with the category's shape |
| #2701 | CODAP-1530 | Explain a legend attribute the display cannot honor |

## What reaches users who never opt in

Assigning a shape is behind the `pointShapes` feature flag, owned by LEADS. Rendering deliberately is not — a document authored with shapes must draw them for every reader rather than silently falling back to circles. That follows the CODAP-1448 convention of gating affordances, not registrations or rendering.

Four changes are visible without the flag, and are the ones worth a second look before merge:

- **Categorical legend keys are shapes, not square swatches** — circles where nothing was assigned, in every document with a categorical legend, on graphs and maps alike. The square existed to show as much color as possible while every point was a circle; that reasoning doesn't survive shape carrying data. The numeric and choropleth legends are untouched.
- **PIXI picking tightens for circles.** Sprites previously carried no `hitArea` and hit tested against their rectangular texture bounds, so a click up to ~1.4 radii off center selected a point. They now test the drawn shape, which is what the canvas renderer has always done — the two renderers have to agree, since graphics acceleration is a user setting.
- **A legend attribute the display can't honor now explains itself — on graphs and maps.** When the attribute lives in a collection more childmost than the plotted cases, each point stands for several of its values at once, so no one color applies and the points draw gray. Everything else used to disagree: the legend rendered nothing at all, taking the attribute's name *and its remove action* with it, and the palette went on offering per-category rows, binning and color controls that changed nothing. The legend now shows the attribute and a sentence saying why it can't be honored, and the palette offers only what still reaches the points — the display-wide shape stays, since shape does reach them.
- **The map follows the graph.** Three further changes were needed to stop it contradicting itself: a map accepts such a legend drop instead of discarding it silently, layers that can honor the attribute rank ahead of those that can't, and a layer set to a heatmap it can't draw shows points rather than nothing.

## Review

All eight PRs were reviewed and approved individually, so this isn't asking for another pass over the code. It's here to run CI over the integrated branch and to give the four behaviors above one look in context.

## Not in this branch

CODAP-1508 (clicking a legend key a second time does not deselect its cases) is the epic's remaining child and is still To Do. It's a pre-existing selection bug, independent of shapes.


kswenson:
Merging without review as the individual PRs have already been reviewed.

PR #2603, opened 2026-05-26
CODAP-1352: emit missing V2 map notifications

kswenson:
## Summary

V2 plugin-event compatibility for the map tile (follow-up to CODAP-1310, analog to CODAP-1351 for graph). 10 commits backfilling the V2 notification ops V3's map was missing — see audit §3.3 (map subsection). V2-source verification ran against `origin/master` (V2's active branch).

### Items wired

New shared file `data-display/data-display-notifications.ts` (graph+map shared helpers):
- `swapCategoriesNotification` — moved from `graph-notifications.ts`, widened to accept map (closes V2 legend-swap gap for both tiles)
- `changePointSizeNotification` — `change point size`
- `toggleStrokeSameAsFillNotification` — `toggle stroke same as fill`
- `changePointColorNotification` — categorical legend, `change point color`
- `changePointColorAndAlphaNotification` — non-categorical, V2's compound op `change changePointColor`
- `changeStrokeColorAndAlphaNotification` — `change changeStrokeColor`
- `changeAttributeColorNotification` — `change attribute color` (low/high spectrum)

New map-only file `map/map-notifications.ts`:
- `changeBaseMapNotification` — `change base map`
- `changeGridSizeNotification` — `change grid size`
- `changeMapCoordinatesNotification` — `change map coordinates` (pan/zoom)
- `hideSelectedCasesNotification` / `hideUnselectedCasesNotification` / `showAllCasesNotification` — map's spaced op strings (distinct from graph's camelCase)

### V3 deviations from V2 (documented per commit)
- Map hide ops carry `numberHidden` for graph-parity (V2 graph already does; V2 map's omission is an inconsistency — additive in V3, plugins can ignore)
- Shared inspector emissions (color/size/stroke) fire for graph too — V3-additive since the V3 controls live in `data-display/`
- `swap categories` now emits from the map legend (V2 oversight — V2 didn't emit from legend for either tile)
- Pan/zoom only emits when the change is undoable (V2 wouldn't emit for the float-precision sync round-trips either)

## Test plan

- [x] Lint + tsc clean
- [x] 3173 Jest tests pass
- [ ] CI build + limited Cypress
- [ ] Copilot review
- [ ] Full regression Cypress (after `run regression` label)

Fixes CODAP-1352.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2704, opened 2026-09-18
CODAP-245: add plan for bringing the Plugin API wiki up to date

bfinzer:
Adds `v3/doc/plugin-api-doc-update-plan.md`. The deliverable for this story is the plan; no documentation is changed here and no code is touched.

The plan is the output of a systematic comparison of the [Data Interactive Plugin API](https://github.com/concord-consortium/codap/wiki/CODAP-Data-Interactive-Plugin-API) wiki page against `src/data-interactive/`. It found five kinds of drift: documented names that are wrong in v3 (so plugins written from the doc fail silently), five registered resources with no documentation at all, missing component types and properties, a stale CODAP-initiated notification catalog, and architecture the page predates. It proposes phases 0–7 to close those, plus a section on making the reference usable by LLMs as well as humans — a generated `plugin-api.json` inventory and tooling to keep the docs from rotting again.

Reviewers: the parts that most need your judgment are **Phase 0**, which is blocked on a human decision — patch the existing 4177-line page in place, split it into per-resource pages (my recommendation), or move the reference into the repo and leave the wiki as a pointer — and the four open questions in §6. The follow-on work should be filed as separate stories.

Fixes CODAP-245


kswenson (changes requested):
👍 Looks great -- the following review, developed in conjunction with Claude Code 🤖, identifies several issues worth addressing before merging.

## PR Review Summary

**Changes:** 1 file, +349 / -0 lines (docs only: `v3/doc/plugin-api-doc-update-plan.md`)

### What it does

Adds a plan for bringing the Data Interactive Plugin API wiki page up to date with v3. The plan comes from comparing the 4177-line wiki page against `src/data-interactive/`. It sorts the drift into five kinds: wrong names, undocumented resources, missing properties and component types, a stale notification catalog, and architecture the page predates. It then proposes phases 0-7, plus a section on making the reference usable by LLMs. No docs or code change here.

### Assessment

A strong, useful plan. I spot-checked most of its load-bearing claims against the code and the wiki. Nearly all of them hold:

- All 38 handlers are registered.
- `pointsAreFusedIntoBars` is at `graph-component-handler.ts:350`, and `pointConfig` doesn't exist.
- `configuration` recognizes only `gaussianFitEnabled`.
- `ActionName` has seven actions.
- The `updateDocumentBegun`/`Ended` notices are at the cited lines.
- `interactiveApi` is gated on the URL parameter, and `dataDisplay` returns `exportDataUri`.
- `toV2Id` returns a number.
- Every cited `doc/*.md` file exists; `plugin-tour-api.md` is 576 lines.
- The plan quotes the notification compatibility audit's examples (`titleChange`, `toggle connecting line`) and §5 accurately, though the audit itself is out of date (see the first issue).
- `dataDisplay`, `interactiveApi`, `tourElements`, `caseCard` and `imageComponentView` really have zero hits on the wiki page.

The main problem is that the notification compatibility audit is out of date (the first issue). The rest are small factual slips, although one of them (the fence count) is a headline number. Fixing them keeps the plan trustworthy as a source document. My thoughts on Phase 0, the Phase 1 callout, and §4.1 follow.

### On Phase 0 (destination)

I'd lean toward **option 3, combined with the per-resource split from option 2**. The plan presents them as alternatives, but they answer two separate questions: *where* the docs live and *how* they are divided up. A split reference can live in the repo just as well as in the wiki.

Reasons to favor the repo:

- **Phase 6 depends on it.** The CI check and the §4.6 docs lint can only verify docs that CI can see. Under options 1 and 2 the docs live in a separate repo (`codap.wiki`). A check in the codap repo would have to fetch the wiki to see whether a matching section exists, and a wiki edit would never re-run the check. So the check can't catch the drift it exists to prevent.
- **Docs change in the same PR as the code.** Adding a handler and documenting it would go through one review. Nothing connects the two today, which is how five resources shipped undocumented.
- **The wiki's open editing isn't being used.** Every edit to the page since 2024 is by a team developer, so moving it gives up nothing in practice.
- **It answers open questions 2 and 4.** `plugin-api.json` and `llms.txt` get a stable, versioned home next to the code. The current wiki text can be frozen as the v2 reference.

On timing: v3 living in `v3/` alongside the stale v2 code on `main` was a choice made early on to make porting v2 code easy. The near-term plan is to remove the v2 code from `main` and promote `v3/` to the top level of the repo. After that, a top-level `doc/` (or even `wiki/`) folder is a natural home for this reference. We don't have to settle the final path now, but it's worth designing the split with that destination in mind.

The cost is findability: existing links point at wiki anchors. Turning the wiki page into a pointer that maps each old section to its new page keeps those links one click from the right place.

### On the Phase 1 "V3 changes" callout

Compatibility with the v2 plugin API was an acceptance criterion for v3. A name that v3 no longer accepts should therefore be a bug or a deliberate deprecation, not a documented "v3 change." Checking the three callout items against v2 (`master`) and the wiki history, they fall into different categories:

- **`pointConfig` and `pointsFusedIntoBars`** were never v2 properties. They were added to the wiki during v3 development, before v3's public release, and the API changed (to `plotType` and `pointsAreFusedIntoBars`) before any release shipped them. No released CODAP ever supported them, so there's nothing to deprecate: the wiki entries are simply errors to correct. That also answers open question 3 (no alias needed).
- **`openCase` / `createCase` / `closeCase`** belong to the legacy Game API (`game_phone_handler.js` in v2), not the Data Interactive API. But v2 did support them and v3 doesn't, so listing them as no longer supported is appropriate. It should name the Game API as their source. Separately, the "Communication Patterns" prose presents them as the way to use the Data Interactive API, so it needs a rewrite around `create item(s)`.

So I'd keep the callout, limit it to the Game API verbs, and correct the two graph properties as plain doc errors. In general, I'd suggest Phase 1 sort each §2.1 item into one of three buckets before labeling it, checking v2 source for each: a v2-compat bug (fix in code), a deliberate v3 change to something v2 or a v3 release shipped (support it as deprecated, or document it as no longer supported), or a doc error (fix the doc). The §4.4 table can list all of these names, since LLMs may have picked them up from the wiki. It should mark the graph properties as doc errors and the Game API verbs as unsupported in v3.

### On §4.1 (the machine-readable companion)

I agree with having a machine-readable companion, but I'd question calling `plugin-api.json` "the single highest-value change." For a model *reading* the docs, markdown generally works as well as JSON or better:

- **Fewer tokens.** JSON spends tokens on quotes, braces and a key repeated for every property. The same inventory as markdown tables is noticeably more compact.
- **Easier to keep track of.** In deeply nested JSON it's harder to tell which object a property belongs to. A heading plus a flat property table keeps that visible.
- **It carries the "why".** The facts that prevent mistakes are prose: when to use `create items`, what triggers each error, how coalescing affects response timing. JSON can only put these in `description` strings.
- **It splits well.** A retrieved markdown section is still a complete answer. A fragment of a large JSON file loses the path that says what it describes.

JSON's advantage is for *tools*: validating generated requests, lookups, CI checks, codegen, and agents that check their own output.

So I'd keep the extracted inventory as the single source of truth and generate from it:

1. **Per-resource markdown** (property tables plus valid `json` examples) as the primary reference for humans and models alike.
2. **JSON Schema** (the standard format, not ad-hoc JSON) for request and response validation. Models know it well, and tools can use it directly.
3. Possibly a **`.d.ts`** of the request and value types. For a model writing plugin code, TypeScript types may be the most compact precise format, and v3's shapes are already TypeScript.

By that ranking, the highest-value LLM change is complete, self-contained markdown pages with valid examples (§4.3 plus the per-resource split), with the schema a strong second.

### Issues

1. **major — §2.4 / Phase 4 treats the notification compatibility audit as current, but some of its findings have since been fixed** (`plugin-api-doc-update-plan.md:84-94`, `:159-164`). The notification compatibility audit (`doc/v2-v3-compatibility-audit.md`) was last updated 2026-05-08. Six stories were filed from it on 2026-05-22, and three of them have since shipped:
   - Done: CODAP-1351 graph adornment/plot notifications (#2600), CODAP-1352 map notifications (#2603), CODAP-1353 case-table/calculator/slider notifications (#2596).
   - Still To Do: CODAP-1354 (V2-compatible default undo/redo notifications on the `document` resource), CODAP-1355 (align user-analytics log events with V2, which affects plugins that match on `formatStr` via `logMessageMonitor`), and CODAP-1356 (V2-compatible item-level `dataContextChangeNotice` operations).

   The plan's own example is already outdated: `titleChange` now puts `type` at the envelope level *and* in `values` (with `diType` added), as in `models/tiles/tile-notifications.ts`. Transcribing the audit as-is would document fixed bugs as v3 behavior and push plugin authors into workarounds they don't need. Phase 4 should start by re-checking the audit against current `main` (and ideally updating the audit itself). The remaining divergences are already covered by 1354-1356, which gives open question 1 an answer: fix them in code rather than document them. Phase 4's dependency then becomes "1354-1356 land or are explicitly deferred."

2. **minor — stale phase cross-references** (`plugin-api-doc-update-plan.md:132`, `:156-157`). The Phase 0 recommendation says the per-resource pages are "generated (Phase 4)". Phase 3 calls itself "the natural first customer of the Phase 4 tooling — consider doing Phase 4 first". The tooling is Phase 6; Phase 4 is the notifications chapter. It looks like the phases were renumbered. A reader following Phase 3's advice would pull the notifications rewrite forward instead of the tooling.

3. **minor — the §4.3 fence statistic is wrong** (`:247`). "246 of the page's 430 code fences are untagged" counts fence *lines*, so each closing fence gets counted as an untagged fence. The page actually has 215 code blocks: 179 `json`, 4 `javascript`, 1 `html`, and **31 untagged**. 26 of those untagged blocks hold the `/* ... */` pseudo-JSON object shapes. The recommendation still stands, since the object shapes really are the untagged blocks, but the scale is overstated about 8x. Suggest: "31 of the page's 215 code blocks are untagged, and 26 of those are the object-shape blocks".

4. **nit — component type count** (`:27`). "10 DI types", but `registerComponentHandler()` is called for 11: caseCard, caseTable, graph, map, slider, calculator, text, webView, guideView, game, imageComponentView. The plan's own §2.3 gets there too (8 documented + 3 missing = 11).

5. **nit — adornment type count** (`:28`, `:57`). "14 types + `Percent` alias", but the §2.2 list names 15 types. 15 registration files call `registerAdornmentHandler` (Count through RegionOfInterest).

6. **nit — attribute `notify` omits `dragMove`** (`:46`). `attribute-handler.ts:101-138` handles `dragStart`, `dragMove` and `dragEnd`; the plan lists only `dragStart | dragEnd`. Worth fixing, since this row is meant to be copied into the docs.

7. **nit — "zero coverage" for `adornment` is slightly overstated** (`:53-57`). The wiki has one example that uses it, "update graph component to include a least squares line" (`component[Diet].adornment`, wiki line ~2860). The adornment resource itself still has no section, so the gap is real. The wording could say "no section; one incidental example".

bfinzer:
Thanks — this was a genuinely useful review. I verified every checkable claim against the code, the wiki repo, v2 on `master`, and Jira, and all seven issues hold. Pushed as 449d507b3.

**Verified and corrected:** the component (11, with `errorTester` gated on `urlParams.errorTester`) and adornment (15) counts; `dragMove` at `attribute-handler.ts:138`; the fence statistic (215 blocks, 31 untagged, 26 of them object shapes — my number counted fence *lines*, so every closing fence became an untagged block); the single incidental `adornment` example at wiki line 2867; the two `Phase 4` → `Phase 6` cross-references.

**Phase 0 — accepted, and it was the most important issue.** You're right that I presented *where* and *how* as one list, and the recommendation was self-undermining: Phase 6's CI check is the plan's entire anti-rot mechanism, and it can't see or be triggered by edits in a separate wiki repo. Recommending option 2 would have prevented the plan's own central deliverable from working. Phase 0 is now two questions, recommending repo + per-resource split, with that argument stated as the deciding one and `v3/doc/` as the starting path.

**Phase 1 — accepted, with one addition.** The bucketing is the best structural suggestion here; treating every divergence as settled v3 behavior was wrong given that v2 compatibility was an acceptance criterion. I added a fourth bucket, **v3 addition**, for the attribute drag path, which has no v2 counterpart and so is neither a bug nor a doc error.

I also confirmed your v2 findings: `pointConfig` and `pointsFusedIntoBars` have zero hits anywhere in v2 source, and both entered the wiki on 2024-12-19. Doc errors, no alias needed — open question 3 answered.

One §2.1 row I'd flag as still untriaged rather than pre-sorted: `configuration`. It *is* a v2 DI resource (`data_interactive_phone_handler.js:255`) and `gaussianFitEnabled` is a v2 global app setting, but I didn't establish whether v2 accepted other configuration names. If it did, "v3 recognizes exactly one" is a compat bug, not v3 behavior. The plan now says so instead of asserting it.

**§2.4 / Phase 4 — accepted.** Confirmed 1351/1352/1353 Done and 1354/1355/1356 To Do, and that my `titleChange` example describes a bug since fixed — v3 now sets `type` at the envelope *and* in `values`, plus `diType` (`tile-notifications.ts:40-49`). Phase 4 now starts by revalidating the audit against `main`, and its dependency is "1354-1356 land or are explicitly deferred."

**§4.1 — conceding the ranking.** "Single highest-value change" was an overclaim; per-resource markdown with valid examples is the highest-value LLM change, JSON Schema second for validation and codegen. One caveat I added to the `.d.ts` idea: v3's DI types are internal and include plenty that isn't public API surface, so it means curating and maintaining a public subset — real work rather than free generation. It's listed as an option to evaluate.

Two follow-ups this surfaced, neither in this PR:

1. `doc/v2-v3-compatibility-audit.md` is stale enough to mislead the next reader the way it nearly misled this plan. Probably its own story.
2. The `configuration` triage above wants a few minutes in v2 source.

kswenson (approved):
👍 Thanks for the thorough revision! I checked 449d507b3 against the first review, with help from Claude Code 🤖. Every issue and discussion point is addressed, and the fixes hold up against the code: 12 `registerComponentHandler` call sites with `errorTester` gated at `error-tester-registration.ts:13`, `type`/`diType` handling in `tile-notifications.ts:40-49`, and the v2 Game API path on `master`. The fourth triage bucket ("V3 addition") is a good addition.

A few observations follow. None of them blocks approval, so decide whether and how to address them.

1. **The `configuration` triage can be closed.** v2's allowlist is `publicConfigurationProperties: ['gaussianFitEnabled']` (`apps/dg/core.js:447` on `master`), and `handleConfiguration` checks both `get` and `update` against it (`data_interactive_phone_handler.js:2407-2431`). So v2 also recognized exactly one name. v3 matches it, which puts this in the **doc error** bucket: the wiki implies an open-ended set that never existed. The "needs triage" bullet (`:213-218`) could become a pre-sorted entry. (Minor, probably not worth documenting: v2's error text is `Not found`, v3's is `unknown configuration "<name>"`.)

2. **The source-of-truth table still presents the audit as current** (`:30`: "already written, 2026-05-08"). This is the one mention the new §2.4 caveat didn't reach. Also, #2706 (CODAP-1545) is now updating the audit. Once it merges, §2.4's "the audit is itself out of date" and Phase 4's "step one is re-validating" will be stale in turn. Pointing at CODAP-1545 in those spots would keep the plan reading correctly whichever lands first.

3. **nit: title.** "Bringing the Plugin API Wiki Up to Date" now recommends moving off the wiki. Perhaps "…Plugin API Reference Up to Date"?

